PR #1, opened 2011-05-07
Doc blocks

elinw:
This is a fork with doc block corrections and formatting. It doesn't do everything that needs to happen but it's a substantial start. 

There are a small number of changes that are outside of doc blocks. 

In cache storage Ian told me to make read, write, destroy and gc  abstract.

JElement is abstract based on the existing tag in the doc block (now removed).

I didn't do this but there is also a question about whether fetch in the toolbar separator should be abstract.

Elin

### DIFF
--- a/libraries/joomla/form/field.php
+++ b/libraries/joomla/form/field.php
@@ -208,7 +208,7 @@ abstract class JFormField
 		// Detect the field type if not set
 		if (!isset($this->type))
 		{
-			$parts = JString::splitCamelCase(get_called_class());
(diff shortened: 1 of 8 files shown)


elinw:
Okay sorry about that; I realized that those abstracts were there because the build here and the hg one missed a sync.

eddieajau:
I've merged a heap of Docblock updates from the svn platform merge branch so I'm assuming that includes a lot of this work.  Can you pull from the master repo and see how it looks.  Thanks.

elinw:
OK will do, yes it does include a lot of the same.

PR #2, opened 2011-05-12
Missing paremeter

jonnsl:
The function getUserState in the JApplication class is missing the $default parameter


LouisLandry:
Great stuff Jonnathan, thanks.

PR #3, opened 2011-05-14
Code from bit bucket

elinw:
As discussed previously.


elinw:
Resubmitted as a branch

PR #4, opened 2011-05-14
AJAX improvements

realityking:
First commit is a small library modification to make it easier to use AJAX or other JS functionality with the Joomla framework.

Second commit is the Mootools upgrade as discusses in the issue tracker.


realityking:
New pull request is here: https://github.com/joomla/joomla-platform/pull/11

Had to slightly change the way I use github.

PR #8, opened 2011-05-15
Docblocks branch

elinw:
This should be just the doc block edits.


elinw:
This should apply cleanly now

elinw:
ok got rid of he commits from realitykind and eddieajau which were introducing non doc block changes to he pull request.

realityking:
I think you broke sth here. Our two commits are still part of the pull request but now there are two additional commits reverting our changes.

elinw:
The diff shows that your stuff is untouched. It's not your commits that are reverted, it's my commit of your commits.

PR #9, opened 2011-05-17
Fixed JDaemonTest

Paladin:
There was a bug in JDaemonTest that crashed the tests if the PCNTL extension wasn't available.
Test code now marks the test as skipped if the extension isn't there.


Paladin:
OK, so I added another fix, this one for the crash caused by lack of JTestConfig in JDatabaseMySQLTest

PR #10, opened 2011-05-18
Let's try this again

Paladin:
Thought I'd submitted this pull request earlier today, but I don't see it anywhere, so maybe I fatfingered it. So we'll try it again. If it's a duplicate, forgive me.

### DIFF
--- a/tests/suite/joomla/database/database/JDatabaseMySQLTest.php
+++ b/tests/suite/joomla/database/database/JDatabaseMySQLTest.php
@@ -59,7 +59,12 @@ class JDatabaseMySQLTest extends JoomlaDatabaseTestCase
 	 */
 	protected function setUp()
 	{
-		$config = new JTestConfig;
+		@include_once JPATH_TESTS . '/config_mysql.php';
+		if (class_exists('JMySQLTestConfig')) {
+			$config = new JTestConfig;
+		} else {
+			$this->markTestSkipped('There is no mysql test config file present.');
+		}
 		$this->object = JDatabase::getInstance(
 			array(
 				'driver' => $config->dbtype,
--- a/libraries/joomla/base/observable.php
+++ b/libraries/joomla/base/observable.php
@@ -74,7 +74,8 @@ class JObservable extends JObject
 		// Iterate through the _observers array
 		foreach ($this->_observers as $observer)
 		{
-			$return[] = $observer->update();
+			$update_args = array('subject' => $this);
+			$return[] = $observer->update($update_args);
 		}
 
 		return $return;
--- a/tests/objects/JObservableTester.php
+++ b/tests/objects/JObservableTester.php
@@ -1,0 +1,20 @@
+<?php
+/**
+ * Abstract observable class to implement the observer design pattern
+ *
+ * @package     Joomla.Platform.UnitTesting
+ * @subpackage  Base
+ * @since       11.1
+ */
+
+require_once JPATH_PLATFORM.'/joomla/base/observable.php';
+
+class JObservableTester extends JObservable
+{
+	public function setState( $setMe )
+	{
+		$this->_state = $setMe;
+		return $setMe;
+	}
+}
+?>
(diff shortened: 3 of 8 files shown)


tests/suite/joomla/database/database/JDatabaseMySQLTest.php
```diff
@@ -59,7 +59,10 @@ class JDatabaseMySQLTest extends JoomlaDatabaseTestCase
 	 */
 	protected function setUp()
 	{
-		$config = new JTestConfig;
+		@include_once JPATH_TESTS . '/config.php';
+		if (class_exists('JTestConfig')) {
```
eddieajau: I'm not sure I understand what is gained. If JTestConfig does not exist we are still going to get an error, no?

Paladin: Yes, but....

The severity of the error changes. The original code crashed my testing system completely at this point with a PHP Fatal error if the config.php file wasn't there; no further tests were executed. This code merely errors out of the tests in this file while letting all the subsequent tests in the suite execute.

Long term it's probably not the best solution. I was putting this code in place as a stopgap, to eliminate the crashes so I could see the "big picture" of the testing. It's the same code someone else used in JLoggerDatabaseTest. 

If you want a different fix for it, you might add this else clause at line 66 to the class_exists test (both in this file and for that matter in JLoggerDatabaseTest):

``` php
else
{
    $this->markTestSkipped("These tests require a test config file be present, and I can't find yours.");
}
```

which will still keep the db tests from running but also kill the "Undefined variable" message for these tests. I decided against this idea because the error message, while not blocking further test execution, calls attention to a problem in the setup of the testing system. An even better solution might be to come up with a "standard" default set of config params and use them when the test config isn't present. I'd planned on addressing that question later, after dealing with the unwritten tests; this was simply a means to get the all the tests to run in the meantime.

tests/suite/joomla/application/cli/JDaemonTest.php
```diff
@@ -33,6 +33,9 @@ class JDaemonTest extends PHPUnit_Framework_TestCase
 	 */
 	public function testWriteProcessIdFile()
 	{
+		if (!defined('SIGHUP')) {
```
LouisLandry: I believe this should be covered now with a separate pull request that was merged.  If you update the platform and look at line 24 of this file you will find a conditional in the setup method that marks the test skipped if the PCNTL extension isn't loaded.  Can you verify that this problem is fixed with that change and if so we can pass on this addition.  I feel it is a more complete solution given that nothing in JDaemon will work without PCNTL.

tests/suite/joomla/database/database/JDatabaseMySQLTest.php
pasamio: First of all some times you want to test without necessarily having the database tests run. I'd err on marking it as skipped. The other issue is what happens when you shift to Oracle/MSSQL/sqlite3 to do testing? The test should at least check that the driver in the test config is MySQL before continuing.

tests/suite/joomla/application/cli/JDaemonTest.php
Paladin: Feel free to reject this commit, then, if you've put a better solution in the tree since I wrote it.

tests/suite/joomla/database/database/JDatabaseMySQLTest.php
Paladin: Shows when I last looked at the drivers. Last time I checked there were only mySQL drivers available. I'm not sure how to parse "I'd err on marking it as skipped" -- you mean you'd rather it was an error than a skip? (I agree, BTW.) As for setting up separate suites with and without db tests, that's beyond the scope of what I was doing. (I'm following the principle of test first, pretty later. You want to create those suites, feel free. I see a few thousand tests I'm going to have to write yet before then.)

PR #11, opened 2011-05-20
AJAX improvements

realityking:
Small library modification to make it easier to use AJAX or other JS functionality with the Joomla framework.


piotr-cz:
There's a problem when rendering messages in language other than English using Joomla.renderMessage().

The DT element classname and content is being set to message type (ie. 'warning'). While it's fine for the class, message header is then forced to be a lowercase English text.

On the orher hand when message type is supplied in document language just like message body, then it won't match relevant class defined in system.css, this is not a solution.

One way how to go around it is to switch DT content from type to _Joomla.JText._(type, type)_. Developer will have to add appriopriate JText strings in JS.

An other would be to change the parameters for renderMessage function, so it would be called

<pre>
Joomla.renderMessages({
    'error': ['Błąd', ['Nieznany błąd']]
});
</pre>


instead of

<pre>
Joomla.renderMessages({
    'error'     ['Nieznany błąd']
});
</pre>

realityking:
I just checked again and I'm pretty sure my JS implementation matches the PHP implementation in JDocumentRendererMessage. (which was the set goal)

piotr-cz:
realityking:
there's one difference I'm talking about:
_libraries/joomla/document/html/renderer/message.php @L 58_ should match _media/system/js/core-uncompressed.js @L 100_ so in in core-uncompressed.js (function Joomla.renderMessages)

_html: type_ 

should be 

_html: Joomla.JText._(type, type)_

PR #12, opened 2011-05-20
New Unit tests

realityking:
A couple of unit tests from my Bitbucket branch.


realityking:
Andrew has now added the JDocumentTest. Easiest is probably if someone just cherry picks the second commit.

PR #13, opened 2011-05-20
Move the rendering of HTML link elements to JDocumentRendererHead.

realityking:
Previously discussed here: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=25251

Note: This has some small backwards compatibility consequences (e.g. if somebody chose to override either of the two classes involved) but it makes JDocument more consistent and opens up future use (head links in other types of documents).

### DIFF
--- a/libraries/joomla/document/html/html.php
+++ b/libraries/joomla/document/html/html.php
@@ -188,9 +188,9 @@ class JDocumentHTML extends JDocument
 	 */
 	public function addHeadLink($href, $relation, $relType = 'rel', $attribs = array())
 	{
-		$attribs = JArrayHelper::toString($attribs);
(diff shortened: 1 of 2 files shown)


LouisLandry:
So the BC stuff is noted, but that is one really good example of the juice being worth the squeeze.  Thanks for this one.

elinw:
Was it discussed somewhere on some list that we would post/document about BC breakage when we know about it?  Maybe start a wiki page if we don't have one already.

realityking:
Started one: http://docs.joomla.org/Potential_backward_compatibility_issues_in_Joomla_1.7_and_Joomla_Platform_11.1

Not sure about the formatting, but at least we won't loose stuff this way.

elinw:
Great! Formatting we can always deal with.

PR #18, opened 2011-05-20
Add JDocumentOpensearch.

realityking:
These are the library changes for this patch: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=25250

This is potentially useful for other apps as well.


LouisLandry:
I think this is a cool idea ... and I like it, but I'd really like to see Unit Tests written for new libraries that we add.  It is really the only way for us to track regressions and know that what is here actually works.  Can you work on getting at least some basic testing done for this and we'll get it merged?

realityking:
I've added a unit test and fixed a small oversight. (Still had the wrong MIME type in from testing)

LouisLandry:
Cool stuff :-)

PR #20, opened 2011-05-20
Use require_once instead of include in JLoader

rvsjoen:
I do not know if there is a good reason for using include over require_once, if there is, please disregard this commit. All unit tests ran fine with this change.

### DIFF
--- a/libraries/loader.php
+++ b/libraries/loader.php
@@ -87,7 +87,7 @@ abstract class JLoader
 
 				// If the file exists attempt to include it.
 				if (is_file($base.DS.$path.'.php')) {
-					$success = (bool) include $base.DS.$path.'.php';
+					$success = (bool) include_once $base.DS.$path.'.php' ;
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


rvsjoen:
Well, the reason I came across this in the first place was looking at using JLoader::import() instead of require_once to load component-specific files inside a component, such as a helper using something like JLoader::import("content", JPATH_COMPONENT_ADMINISTRATOR.DS.'helpers'); which in turn fails with a "class redefinition" error if the helper class is already loaded.

I doubt this is desired behavior as there isn't really a reason as far as I can tell to ever load a file twice.

realityking:
I should have been a little clearer, the issue isn't the _once but the use of required instead of include inside a conditional.

rvsjoen:
Ah, that actually makes more sense, I do not mind rewriting the commit to do include_once instead of require_once.

realityking:
I hate to be a prick but parentheses for statements aren't allowed either. (I didn't notice that you used them before, sorry)

rvsjoen:
Alright, third time is the charm

PR #21, opened 2011-05-21
JForm Validation improvements

jonnsl:
Some improvements to the validation of JForm. see the commits descriptions for more info.


elinw:
Please take a look at my branch which includes validation for url ad tel.

jonnsl:
I forgot to mention that some regex are from mootools

jonnsl:
@elinw according to the php manual "This function (parse_url) is not meant to validate the given URL, it only breaks it up into the above listed parts. Partial URLs are also accepted, parse_url() tries its best to parse them correctly."
so I guess that is not a good idea to use this function to validate urls

elinw:
It's not using parse_url to validate it is using it to break up and then validating each element separately.

PR #22, opened 2011-05-24
Bb bit bucket url stuff updated

elinw:
old request did not have the test

### DIFF
--- a/libraries/joomla/form/rules/url.php
+++ b/libraries/joomla/form/rules/url.php
@@ -1,0 +1,93 @@
+<?php
+/**
+ * @package     Joomla.Platform
+ * @subpackage  Form
+ *
+ * @copyright   Copyright (C) 2005 - 2011 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE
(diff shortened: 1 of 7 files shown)


LouisLandry:
Great work here.  Love the completeness and the thoroughness of the inline commenting.  Thanks for building out those tests too .. will make things far better moving forward.

libraries/joomla/form/rules/url.php
```diff
@@ -0,0 +1,93 @@
...
+		if (!array_key_exists('scheme',$urlParts)){
+			return false;
+		}
+		$urlScheme = (string) $urlParts['scheme'];
+		$urlScheme = strtolower($urlScheme);
+		if (in_array($urlScheme,$scheme) == false){
+			return false;
+		}
+		// For some schemes here must be two slashes. 
+		if (($urlScheme == 'http' || $urlScheme == 'https' || $urlScheme == 'ftp' ||
```
pasamio, line 74: what about using an array here instead and checking for if its in the array?

elinw:
That's a good idea. I'll do that in the next round.

PR #23, opened 2011-05-25
Remove workaround for PHP 4 and PHP 5.2.3 in JFTP and JXMLElement.

realityking:


LouisLandry:
So the destructor in JFTP I am happy to merge, but I don't think we've altered the minimum requirements for the platform yet from 5.2.0 ... and if that is the case I'm not comfortable making the adjustments to JXMLElement quite yet.  Hopefully that makes sense.

realityking:
Ok, I taught the platform would have the same requirement as the CMS (5.2.4). Do you want to do it manually or should I prep a new pull request?

elinw:
http://docs.joomla.org/Joomla_1.6_technical_requirements  says 5.2.4 but other places it says 5.2+

LouisLandry:
Hrm.  I'll go with the docs wiki then.  Happy to do that.  Let's just move forward.  Thanks for the thoughts guys.

PR #25, opened 2011-05-27
Fixes for #22587 and #24285

klas:
Issue 22587
- Change modules cache lifetime to be set as intended (seconds must be converted to minutes)
- Cache handler needs to use all parameters for static variable caching

Also $_handler is a static variable and so can't be referenced as $this->_handler  (this bug was introduced after #22587 report) - change back to self::$_handler

Issue 24285

Fix changes module caching to store only headers that changed after module was executed. Headers before and after callback are compared.

### DIFF
--- a/libraries/joomla/application/module/helper.php
+++ b/libraries/joomla/application/module/helper.php
@@ -389,7 +389,8 @@ abstract class JModuleHelper
 			$cache->setCaching(false);
 		}
 
-		$cache->setLifeTime($moduleparams->get('cache_time', $conf->get('cachetime') * 60));
+		// module cache is set in seconds, global cache in minutes, setLifeTime works in minutes
+		$cache->setLifeTime($moduleparams->get('cache_time', $conf->get('cachetime') * 60) / 60);
 
 		$wrkaroundoptions = array (
 			'nopathway' 	=> 1,
--- a/libraries/joomla/cache/cache.php
+++ b/libraries/joomla/cache/cache.php
@@ -389,11 +389,16 @@ class JCache extends JObject
 	 * @since   11.1
 	 */
 	public function &_getStorage()
-	{
-		if (!isset($this->_handler)) {
-			$this->_handler = JCacheStorage::getInstance($this->_options['storage'], $this->_options);
-		}
-		return $this->_handler;
+	{	
+		$hash = md5(serialize($this->_options));
+	
+		if (isset(self::$_handler[$hash])) {
+			return self::$_handler[$hash];
+ 		}
+		
+		self::$_handler[$hash] = JCacheStorage::getInstance($this->_options['storage'], $this->_options);
+		
+		return self::$_handler[$hash];
 	}
 
 	/**
--- a/libraries/joomla/cache/cache.php
+++ b/libraries/joomla/cache/cache.php
@@ -504,14 +504,30 @@ class JCache extends JObject
 
 		// Document head data
 		if ($loptions['nohead'] != 1) {
-			$cached['head'] = $document->getHeadData();
 
 			if ($loptions['modulemode'] == 1) {
-					unset($cached['head']['title']);
-					unset($cached['head']['description']);
-					unset($cached['head']['link']);
-					unset($cached['head']['metaTags']);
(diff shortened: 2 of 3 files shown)


klas:
e2d0497 contains fixes for #24285

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=24285

AmyStephen:
Louis - 

Would you mind responding to this issue? Is it under consideration - or is a portion of this to be considered? I'm interested in these fixes and would prefer to see them in core than to override the caching environment. Will do so if necessary but just wanted to hear if you were considering this, or not.  Thanks much.

mahagr:
I discussed about similar issue with Sam Moffatt.

Joomla caching in general has an issue where content gets cached and there's no way to prevent caching from happening -- leading to similar situation: breaking functionality or showing wrong information to guests who should not see it.

For example Kunena has discuss plugin, which allows users to comment articles. It works without any major issues until user is a guest. If user is a guest, the first issue you run into is that there's no way to turn off view or module caching once the parent component or module has decided to do the caching. As all the content from plugins and modules get added into the main content from the component, also they will be static or cached. To make forms and any dynamic output to work for guests, there needs to be a simple way to turn off caching in current scope. Same method could be used inside component or module itself to decide that: "no, this output shouldn't be cached even if i first thought it should be". I often find myself being in helper function when I decide not to cache and thus needing some tricks to prevent caching. In my case this works as some pages cannot never be updated by guest and some can always be updated.

Second huge limitation is in System - Cache plugin (page cache). It should be possible to turn the page cache off for example when the page shows notification messages like: "Your session has expired. Please log in again." If cache expires just before guest is shown a message like this, all guests will see that message for the next 5 to 30 minutes.  Also extensions like Kunena Discuss may want to disable page caching to make the the forms to work and to keep some content up to date. Or they may want to disable page caching (maybe limited time to allow cache to be updated) only from the current guest (flag in the session) to show updated content after user has submitted the form.

klas:
@mahagr those are unrelated issues which are both are solvable in 1.6. by component/module developer. See com_content controller for example how to disable cache on POST requests, certain views etc. The same can be done for modules using owncache. Check my JAB11 caching workshop http://vimeo.com/23840973 for some more details

svanschu:
@klas your caching workshop is nice. But can I ask you to bring that into docs.joomla.org? I think a lot of people miss that, because they don't knew your video.

mahagr:
@klas: I missed your workshop in JAB11 and didn't have time to watch it just yet. I'll do it right now, thanks!

AmyStephen:
The very simple reason I am asking that this pull request be considered is that it is now impossible to even activate PHP Strict in development because the $_handler is incorrectly defined as a static variable.  Any reference to the caching causes the application to fail -- and that's with caching turned off. 

I understand core does not support PHP strict at this time. But, it is helpful to use the setting during development in order to keep those errors down for when core is ready. Caching is referenced in so many places that this one single small bug prevents the use of PHP strict for development and testing.

Again, Louis, thanks for considering.

mahagr:
@klas: so you're the one behind Joomla 1.5/1.6 caching, that's good to know. 

The issues I mentioned are not solvable in components or in the modules -- not easily at least. The basic issue is that you cannot prevent caching from happening if you are inside (content) plugin. I mean that there is no way to tell the view that there's dynamic content generated by module or by plugin.

There is also no way to turn off page cache -- JDispatcher needs object to detach plugin from the events and you can only access the System - Cache instance from the plugin itself. It only has two events: onAfterInitialise and onAfterRender. Easiest way to prevent it from caching would be to set a variable (which is then fetched by the plugin) or to add new event to the plugin to detach itself.

klas:
Just one comment about system cache plugin - imagine it as a high speed train. It runs very fast and moves people from one station to next one in a  very efficient way, but it's not easy to stop it or make turns with it. When more flexibility is required we just use a bus or a car. The same thing goes for different cache types.

Anyway - please, let's not discuss any more things not related to issues in pull request. We can continue this on alltogether or via email (in profile), I'll be more than happy to try to help.

AmyStephen:
Thanks Louis

PR #26, opened 2011-05-27
Add batch processing to framework (Feature Request #22590)

mbabker:
Consistent with http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=22590...

This pull request focuses only on the framework changes for this feature request.  Between the 1.6.3 release and the platform merge, batch processing was added to com_content.  Upon my own review and testing, the code used here was applied in a manner that it could be reused throughout components that follow a category->item structure.

Specific to the framework are the following changes:
- JControllerForm: Function batch added
- JModelAdmin: Functions batch, batchAccess, batchCopy, batchMove, and generateNewTitle added
- JHtmlBatch added

If implemented, changes to the CMS can be made as demonstrated via my SVN branch (mbabker).

### DIFF
--- a/libraries/joomla/application/component/controllerform.php
+++ b/libraries/joomla/application/component/controllerform.php
@@ -210,6 +210,33 @@ class JControllerForm extends JController
 	}
 
 	/**
+	 * Method to run batch operations.
+	 *
+	 * Extended classes can override this if necessary.
+	 *
+	 * @param   object   $model  The model of the component being processed.
+	 *
+	 * @return	void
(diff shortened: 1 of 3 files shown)


elinw:
Michael,

These have been added to the libraries on the CMS, is that right? Just asking in case committing both cases potentially causes conflicts.

mbabker:
Yes, this was committed on the SVN already.  I don't believe I've seen a syncing commit yet here though.

PR #29, opened 2011-05-29

Hackwar:

A specific usage example:

One thing that I'm not entirely sure of is the footer rendering. The class so far was pretty much only aimed at the current Joomla backend tables and they all have just one cell for the footer with the content, so I hard coded this behavior in there. Maybe it would be better to change this and have two rows for the footer, one for the footer with each column adressable seperately and one for the content with a columnspan. However I think if we would want to go into that direction, we can still do this in a later stage by simply adding an optional $columnName variable to the addFooter function.


parthlawate:
Excellent Contribution .. More #jpositiv !

Hackwar:
In the meantime I've been thinking about changing the class again a bit. This class would be something like JGridSimple and extend from JGrid, while JGrid would also allow to add more than one header and footer row.

@tkahl
The functionality for sorting, etc. is not included in this on purpose. The idea is to have this as a really low level base class and then either extend it or to use decorators or simply other helper functions on the object. I'm trying to decouple stuff and also keep it as flexible as possible. Nobody should say that they are restricted by this class.

Hackwar:
Since its been a month and you seem to ignore me, I'm closing this request.

eddieajau:
Found it.  Ok, I actually thought this was an interesting idea, but I'm not sure it's ready for a merge just yet.  I like the idea of dynamic columns.  I'd really like to see some wider discussion and use cases.  I think the full object is the right way to go, rather than a static JHtml helper - maybe even break it up into subclasses more JGridRow, JGridCell (hrm, go even as far as JHtmlTag??), and so on because I could see this baby blowing out in scope ultra quickly.  Would love to get some input from someone like Kyle Lebetter on this one because he'd like the developer to have less control over markup :)

Actually, the HTML package needs a rethink anyway so this could be a test case for that maybe?

Also, make sure you allow for heavy use of chaining (may require you to rethink the design a bit).  Could even make it mimic how mootools builds HTML so we dev's only have to learn one API to have a crack at both??

PR #30, opened 2011-05-31
Fix for bug  [#24549] Component install rollback function rollback_menu is unimplemented leaving items in jos_menu (on tracker)

ramdesh:
I've added a patch which:
- Re-uses code from buildAdminMenus() to find out the id of the component 
- Re-uses code from removeAdminMenus() to remove any existing menu items in #_menu which have been created by the component.

### DIFF
--- a/libraries/joomla/installer/adapters/component.php
+++ b/libraries/joomla/installer/adapters/component.php
@@ -1393,10 +1393,73 @@ class JInstallerComponent extends JAdapterInstance
 	 * @return  boolean  True on success
 	 * @since   11.1
 	 */
-	protected function _rollback_menu()
+	protected function _rollback_menu($arg)
 	{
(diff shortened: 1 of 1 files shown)


LouisLandry:
It would seem that this solution is a bit cleaner and leaner than yours Ramindu.  Do you mind confirming that it solves the same problems yours do?

LouisLandry:
Decided we'd go with: https://github.com/joomla/joomla-platform/pull/41  which appears to be a cleaner solution.  Thanks a bunch for the work though.  We appreciate it.

ramdesh:
Sorry I didn't notice the comments on this. I looked at the code from the other pull request and it seems cleaner. :) Anyway, no problem, of course you should go for the better solution.

Thanks!

PR #32, opened 2011-06-02
Cleaning up after JDocument changes

Paladin:
Removing methods under test while ignoring the tests
caused a mess that needed cleaned up.


LouisLandry:
Thanks Arlen.

PR #33, opened 2011-06-02
Fixed JStringTest

Paladin:
Darwin bug was causing tests to fail. Added test for Darwin, and if
tests are running under Darwin, it marks the international string
tests as skipped, and gives the reason why.


LouisLandry:
Thanks for this Arlen.  Here's hoping this one gets fixed upstream.

PR #38, opened 2011-06-05
Add JHtmlBehavior::multiselect() and move the multiselect JavaScript code

realityking:
Add JHtmlBehavior::multiselect() and move the multiselect JavaScript code to core.js.
Add the option to use id's other than "adminForm" with multiselect.

### DIFF
--- a/libraries/joomla/html/html/behavior.php
+++ b/libraries/joomla/html/html/behavior.php
@@ -345,6 +345,23 @@ abstract class JHtmlBehavior
 
 		return;
 	}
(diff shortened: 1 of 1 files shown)


realityking:
Please don't pull this just yet, testing with the CMS revealed an underlying issue with JHtml that we need to handle first.

realityking:
I updated this the same way as the patch on joomlacode. From my perspective this is good to go.

realityking:

PR #43, opened 2011-06-07
Improve JRegistry

chdemko:
See
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=24472

### DIFF
--- a/libraries/joomla/registry/registry.php
+++ b/libraries/joomla/registry/registry.php
@@ -249,10 +249,7 @@ class JRegistry
 	public function loadString($data, $format = 'JSON', $options = array())
 	{
 		// Load a string into the given namespace [or default namespace if not given]
-		$handler = JRegistryFormat::getInstance($format);
-
-		$obj = $handler->stringToObject($data, $options);
-		$this->loadObject($obj);
+		$this->data =  JRegistryFormat::getInstance($format)->stringToObject($data, $options);
 
 		return true;
 	}


Hackwar:
The problem with this is, that you can't overload the loadObject function anymore to change the behavior at a central point, for example to filter the input. Now you need to replace at least two methods.

LouisLandry:
Christophe, in your issue you claim there is a performance improvement.  If it takes up 20% before, what does it take up after?  I tend to agree with Hannes here, unless we can assert that the object coming out of the stringToObject() methods is always acceptable as the internal data structure of a JRegistry object.  If that is the case then we are indeed doing double work.  Perhaps that is something worth exploring?

chdemko:
In the test I have I've done 4-5 months ago, the performance gain was about
20% on the sample data. I will redo the tests and post here the results

2011/6/11 LouisLandry <
reply@reply.github.com>

chdemko:
After doing benchmark (http://code.google.com/p/webgrind/) on sample data using the home page: here are the results
-JRegistry::loadString takes about 24% of the running time before applying patch (total time is 530 ms) on my laptop
-JRegistry::loadString takes about 4% of the running time after applying patch (total time is 424 ms) on my laptop

See
http://joomlacode.org/gf/download/trackeritem/24472/66248/Capture-webgrindof-home-chdemko-public_html-trunk-index.php-MozillaFirefox.png
http://joomlacode.org/gf/download/trackeritem/24472/66249/Capture-webgrindof-home-chdemko-public_html-trunk-index.php-MozillaFirefox-1.png

PR #44, opened 2011-06-09
2 platform changes that will enable Joomfish to interact with database queries more effectively

GeraintEdwards:
see http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26078 and 
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26077

### DIFF
--- a/libraries/joomla/application/component/modellist.php
+++ b/libraries/joomla/application/component/modellist.php
@@ -212,7 +212,7 @@ class JModelList extends JModel
 
 		// Load the total.
 		$query = $this->_getListQuery();
-		$total = (int) $this->_getListCount((string) $query);
+		$total = (int) $this->_getListCount( $query);
 
 		// Check for a database error.
 		if ($this->_db->getErrorNum()) {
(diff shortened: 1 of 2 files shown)


LouisLandry:
Generally speaking I think this is fine.  Are there any other areas where we are casting a JDatabaseQuery object that are problematic for you?  Additionally, does it also make sense to have a __set() method added to JDatabaseQuery?  I'm not sure that I have a well formed opinion on that.

I'd like to get a comment from Andrew on this as well since he authored much of JDatabaseQuery.

GeraintEdwards:
A setter is a good idea (but not essential) but it would allow for unforeseen eventualities.  The only problem with the setter is how it would interact with JDatabaseQueryElement - i.e. which one of the elements would it be setting?

You can always simulate a setter by getting the separate element groups, calling clear and then resetting the relevant element.  e.g. in pseudo code:
$elements = $query->where;
// do something to elements
$query->clear('where');
foreach ($elements as $elem){
   $query->where($elem);
}

Perhaps the last 4 lines would be the code for the __set method?

GeraintEdwards:
Just spotted that I'd put the getter in the JDatabaseQueryElement class instead of the JDatabaseQuery (now corrected).  I'm new to Git and cut and pasted the code into the Git editor and put it in the wrong place :(

PR #45, opened 2011-06-10
Cleared Fatal crash if JTestConfig not found

Paladin:
Added different config file to JDatabaseMySQLTest to allow an independent decision
by the tester on whether these tests should be run. It now requires
config_mysql.php to contain the JMySQLTestConfig class to be present or the tests
will be skipped.


LouisLandry:
Thanks Arlen.

PR #46, opened 2011-06-10
Move JUpdater:arrayUnique() to JArrayHelper::arrayUnique().

realityking:
IMO this sort of utility function is potentially useful for extension as well and should be located in a place where people would actually look for it :)

### DIFF
--- a/libraries/joomla/updater/updater.php
+++ b/libraries/joomla/updater/updater.php
@@ -14,6 +14,8 @@ jimport('joomla.filesystem.folder');
 jimport('joomla.filesystem.archive');
 jimport('joomla.filesystem.path');
 jimport('joomla.base.adapter');
+jimport('joomla.utilities.arrayhelper');
+jimport('joomla.log.log');
 
 /**
  * Updater Class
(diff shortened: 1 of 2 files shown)


libraries/joomla/updater/updater.php
```diff
@@ -139,24 +140,12 @@ class JUpdater extends JAdapter {
...
-		}
-
-		$myArray=array_unique($myArray);
-
-		foreach ($myArray as &$myvalue){
-			$myvalue=unserialize($myvalue);
-		}
-
-		return $myArray;
+		return JArrayHelper::arrayUnique($myArray);
```
LouisLandry: We need to also add a deprecation warning here.  See: https://github.com/joomla/joomla-platform/blob/master/libraries/joomla/log/log.php#L214 for an example in JLog.

realityking: Done.

libraries/joomla/updater/updater.php
```diff
@@ -14,6 +14,8 @@ jimport('joomla.filesystem.folder');
 jimport('joomla.filesystem.archive');
 jimport('joomla.filesystem.path');
 jimport('joomla.base.adapter');
+jimport('joomla.utilities.arrayhelper');
+jimport('joomla.log.logentry');
```
LouisLandry: Shouldn't this just be jimport('joomla.log.log'); ?  I don't know why you'd need to import the JLogEntry class specifically.  Things look good otherwise.

realityking: I was wondering the same but I oriented myself on the file you gave me as an example. Will fix it up.

realityking: Fixed.

LouisLandry:
Awesome.  Thanks a bunch :)

PR #47, opened 2011-06-12
2 platform changes that will enable Joomfish to interact with database queries more effectively (redux)

GeraintEdwards:
Sorry - closed pull request 44 by mistake (I'm new to github).   This new request has the getter added to the correct class (I'd cut and pasted it in the wrong class last time :( ).  To avoid  confusion I'll repeat my comment about setters:

A setter is a good idea (but not essential) but it would allow for unforeseen eventualities. The only problem with the setter is how it would interact with JDatabaseQueryElement - i.e. which one of the elements would it be setting?

You can always simulate a setter by getting the separate element groups, calling clear and then resetting the relevant element. e.g. in pseudo code:
$elements = $query->where;
// do something to elements
$query->clear('where');
foreach ($elements as $elem){
$query->where($elem);
}

Perhaps the last 4 lines would be the code for the __set method?

### DIFF
--- a/libraries/joomla/application/component/modellist.php
+++ b/libraries/joomla/application/component/modellist.php
@@ -212,7 +212,7 @@ class JModelList extends JModel
 
 		// Load the total.
 		$query = $this->_getListQuery();
-		$total = (int) $this->_getListCount((string) $query);
+		$total = (int) $this->_getListCount( $query);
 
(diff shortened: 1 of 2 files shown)


LouisLandry:
Thanks Geraint.

PR #52, opened 2011-06-16
fix for removeUserFromGroup(), Ticket does not exist for it

svanschu:


svanschu:
Bug: removeUserFromGroup did not work, because it tries to unset a array item with the value as key. 

Fix: getting the key, so the unset can work

PR #55, opened 2011-06-18
Don't render meta elements without content

realityking:
CMS Issue: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26184

Just a little markup saver. Also helps a little with some meta elements not valid in HTML5.


realityking:

PR #56, opened 2011-06-18
Use local storage for tabs

realityking:
CMS Issue: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26192

Saves a couple of cookies in the administration without loosing any functionality. Shouldn't have an BC issues.


realityking:

PR #63, opened 2011-06-25
various changes inside database folder

gpongelli:
Hi all, this is a cumulative commit of my contribution about adding PostgreSQL driver inside Joomla-platform.
As discussed in mailinglist, I've changed the "database" folder name to "driver" and created some subfolder, one for each driver present.
Now database.php and databasequery.php contains only abstract functions and some common database task while per-database implementation was moved inside mysql specific file and created a postgresql one.
To check function match from base to specific class I've realized and .ods file, ask me if you need it to check my work ;)

Cheers.


gpongelli:
Close the pull request because I'm going to create a new pull request without the database file movement, concentrating only on file content changes.
It is better that the file movement will be made from an admin of joomla:master when that files will be in a final state for a release, to avoid commit log break because Eclipse Egit doesn't support "--follow" (as explained in mailing list), so actually it's a problem for me to maintain my fork synchronized with joomla:master .

PR #68, opened 2011-07-01
Docblocks

elinw:
More docblocks


elinw:
There's one non doc thing in this, which is that Opensearch was missing an index.html file, so I've added that.

PR #71, opened 2011-07-04
Pass $group in JModelForm

elinw:
Right now when JModelForm preprocesses a form it doesn't send a group argument. As a result all of the content plugins are always loaded since the $group then defaults to content.This includes things like pagebreak.  So this code attempts to find a iikely group name from the context if there is no group. This is particularly noticeable in com_config and com_plugin in the CMS. It might make more sense to do this in the CMS or to let loadForm take group as an argument, but I think it makes sense to try to find the group if there isn't one even if we do that.


elinw:
wrong branch

PR #77, opened 2011-07-06
Fix JForm::load() not replacing form field in same location

rvsjoen:
When loading another form xml definition on a JForm object, and the new xml definition contains field names which overlap, the fields should be replaced in the same location.

In the current implementation, the old field is removed, and then the new field is added to the end of the form, changing the ordering.

This patch fixes the problem with reordering, but I am not sure if it has any other adverse effects, I will leave that up to someone who knows more about the workings of JForm.


LouisLandry:

Cheers.

PR #79, opened 2011-07-07
Fix strict warning in JDatabase

rvsjoen:
Fix strict warning in JDatabase, abstract static functions are not allowed, give test() a body and make it non-abstract


rvsjoen:
Either way is ok by me, the only reasons I can think of to keep it is for documentation purposes, and to enforce a uniform function signature across the database classes.

eddieajau:
I think I would solve it this way.  Make a new abstract protected method in the JDatabase class and convert all the static methods in the child classes to that method.  Then have the static test method in JDatabase do something like this:

public static function test()
{
$class = get_called_class;
$db = new $class;
return $db->enabled();
}

Then the "enabled" method can be declared as an abstract protected function.

However, that only works in PHP 5.3, and we don't want to add a PHP 5.3 requirement on this class, so I think your method, Rune, is the only workaround.  The only change I would suggest is to put a "return false;" in the body, or, you could possible explore whether an interface would work.

rvsjoen:
Something like this ? I would prefer separating the interface into a separate file though, but this is just a proof of concept.

elkuku:
hmm... Does github provide us with the ability to vote for a specific pull request ?
I would give this +1 ;)

eddieajau:
Yes. Something like that looks good. I'll have a better look next week. About to head to the USA.

rvsjoen:
I would not mind putting that into a separate file, but we need a filename/directory/naming convention for interfaces.

LouisLandry:

Cheers.

PR #86, opened 2011-07-09
Catch up  commits to CMS

elinw:
CMS commit 21571


elinw:
These take us up through July 7.

elinw:
Actually there is one change I missed which is in database/table/user/php and at the moment i can't get that into the right place to add here.

elinw:
Forget that, it has been committed by a different root which is why i wasn't getting a change.

PR #90, opened 2011-07-10
Fix use of CRLF.

realityking:


eddieajau:
Thanks.  Wouldn't bind automatically so fixed them on the server (github and SVN).

PR #93, opened 2011-07-11
Make getTable fill in the prefix like JController::getView

rvsjoen:
When using tables in extensions, the default JModel::getTable() method does not automatically include the extension name, instead it uses the prefix 'Table' to for the table to automatically load using getTable it has to be named TableHelloworld and not HelloworldTableHelloworld.

### DIFF
--- a/libraries/joomla/application/component/model.php
+++ b/libraries/joomla/application/component/model.php
@@ -385,10 +385,14 @@ abstract class JModel extends JObject
 	 * @return  JTable  A JTable object
 	 * @since   11.1
 	 */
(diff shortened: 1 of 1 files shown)


chdemko:
This patch could break some compatibility with 3rd party extensions (that use TableMyName for classes name)
You could add
     if ($table = $this->_createTable($name, 'Table', $options)) {
       return $table;
     }
after
     if ($table = $this->_createTable($name, $prefix, $options)) {
       return $table;
     }
to keep compatibility and to give priority to table with namespace

rvsjoen:
Added the proposed code from Christophe to the pull request

PR #94, opened 2011-07-11
Redirect workarounds

rvsjoen:
Remove old workaround for webkit browsers, and make the workaround for msie more targeted. Older webkit browsers will still work, but take a performance hit, however I think it is more important to get rid of workarounds.

https://groups.google.com/forum/#!topic/joomla-dev-framework/8DPWgYnsSQs

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26335

### DIFF
--- a/libraries/joomla/application/application.php
+++ b/libraries/joomla/application/application.php
@@ -378,7 +378,8 @@ class JApplication extends JObject
 			$document = JFactory::getDocument();
 			jimport('joomla.environment.browser');
 			$navigator = JBrowser::getInstance();
-			if ($navigator->isBrowser('msie') && !mb_check_encoding($url, "ascii")) {
+			jimport('phputf8.utils.ascii');
+			if ($navigator->isBrowser('msie') && !utf8_is_ascii($url)) {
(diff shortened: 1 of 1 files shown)


realityking:
Older Webkit browsers is an euphemism. The issue in Webkit was only closed in march this year. That means current Safari and current Safari on iOS are effected. Not to mention all the smaller webkit browsers used in handheld devices. (Chrome 11 is most likely also still affected)

As much as the workaround sucks, I don't think we can get rid of it just yet. Maybe for 1.8.

The IE fix looks good.

GeraintEdwards:
mbstring is not a standard php extension see http://www.php.net/manual/en/mbstring.installation.php - I did a look but could not see if this was a requirement of Joomla 1.7?

There are ideas here http://stackoverflow.com/questions/1473441/php-checks-to-see-if-a-string-is-utf8-encoded-how that don't require mbstring

rvsjoen:
Ok, what about using the phputf8 library for this, see updated diff

PR #96, opened 2011-07-12
JRegistry and JParameter

AmyStephen:
changes for PHP Strict and CMS improvments.

### DIFF
--- a/libraries/joomla/cache/controller/page.php
+++ b/libraries/joomla/cache/controller/page.php
@@ -113,7 +113,7 @@ class JCacheControllerPage extends JCacheController
 	 *
 	 * @since   11.1
 	 */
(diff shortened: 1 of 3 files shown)


eddieajau:
Please resync with the master branch. Thanks.

klas:
a1e0864 is incorrect and partial approach to fixing cache strict errors.  

Caching library uses polymorphic overloading and this by definition means each method's signature will be different so this requires a wider refactoring to get it work with STRICT and keep backwards compatibility at the same time. 

Working solution is available for testing at git://git.forkito.org/cache.git

AmyStephen:
Klas - 

Would you please issue a pull request with your changes? 

Glad to set this aside in favor of your pull request. 

Andrew -

I'll close this for Klas.

PR #102, opened 2011-07-16
Add optional context to JHtmlContent::prepare

mbabker:
See CMS Bug Tracker -> http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26421

This adds an optional context to JHtmlContent::prepare so that in all instances where this method is used to run the onContentPrepare event, a non-generic context can be passed to the event.  This will allow developers to better check the data their plugin's event has received to verify they are not accidentally processing data they do not intend to handle.  There are no backward compatibility issues as this variable is optional and defaults to the already hard coded context.

### DIFF
--- a/libraries/joomla/html/html/content.php
+++ b/libraries/joomla/html/html/content.php
@@ -22,11 +22,11 @@ abstract class JHtmlContent
 	 * Fire onContentPrepare for content that isn't part of an article.
 	 *
 	 * @param   string   The content to be transformed.
-	 * @param   string   The context of the content to be transformed.
 	 * @param   array    The content params.
+	 * @param   string   The context of the content to be transformed.
 	 * @return  string   The content after transformation.
 	 */
-	public static function prepare($text, $context = 'text', $params = null)
+	public static function prepare($text, $params = null, $context = 'text')
 	{
 		if ($params === null) {
 			$params = new JObject;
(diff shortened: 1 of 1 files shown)


ianmacl:
What happens to code that is passing in parameters?  Also, this isn't ideal since this makes the API different than what the plugin expects, which isn't as transparent as one might like.

mbabker:
In the extensions that I work with, nor in the core, could I find a use case where parameters were being passed.  Theoretically, the order of the variables could be changed and those instances which pass no parameters could then just define '' in that spot.

In working on migrating my own site to 1.7, I took a plugin that was built for 1.5 and migrated it myself for use.  With this plugin enabled, all instances of mod_custom were having their content removed and it wasn't until I'd stepped through the code site wide several times before I caught that it was this specific plugin that was the culprit.  Within the plugin, I can't filter on $context currently because the category and featured views of com_content pass this generic context that I am suggesting we allow to be defined, so I have to check that certain properties exist after I retrieve the full $article object from the database.  It was in this step that I was losing mod_custom data, so I had to add an additional check to get around this.

elinw:
You definitely should not break the old api.

Parameters in fact can be passed, and it's quite important as working with loadmodule has been showing me. Many plugins that are display oriented pass parameters.

mbabker:
As I said, I couldn't find an instance in places I looked where $params was passed, and I looked through several 3PD extensions as well as throughout core.  Either way, it is a simple change here to move the context to be defined after $params.

PR #104, opened 2011-07-17
This is a cms catchup commit for issues 25696 and 26284  

elinw:
As part of solving a problem with updating the maximum time was bumped up for
zip. However some shared hosting does not permit this and that can
cause warnings. Therefore @ is added to supress warnings when this
option is not available. Original report and patch for 26284from Brian
Towles.


eddieajau:
Accepted the other pull against staging.

PR #105, opened 2011-07-17
Logging

elinw:
Starting work on making sure logging is happening in all deprecated methods and that nothing is marked for removal before 12.1.


mahagr:
Great to know that deprecated functions will still work in Joomla 1.8 -- it will allow Kunena 1.6 to keep on working in it, as we are only moving to use the new API in Kunena 2.0.

This is really great news to all extension developers as it gives them possibility to re-factor their code in the next major version, thus giving more possibilities and wider version support (J1.5-J1.8).

Kudos!

PR #112, opened 2011-07-22
Make JUpdate also use regular expression for targetplatform

rvsjoen:
When using collections in the updater, the <extension> tag allows you to specify a targetplatformversion using regular expressions, however the <update> tag does not.


LouisLandry:

Cheers.

PR #114, opened 2011-07-22
Allow the infourl to be stored

mbabker:
As part of a change I would like to see in the Extension Manager's Update view, I would like to be able to display the infourl from the update stream on the view.  This would be more relevant to the user as it would most likely point to a release notification, changelog, or other resource the user should review prior to updating versus displaying the URL of the update stream.


mbabker:
Parallel CMS Feature Request - http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26475

mbabker:
Umm... Not quite sure what happened here, but I'll recreate the pull request later on.

PR #120, opened 2011-07-24
Doc blocks

elinw:
work oh  html doc blocks.


elinw:
Merged up, got rid of a conflict.

PR #121, opened 2011-07-24
Jinput

ianmacl:
Added interface to JInput from Japplication.


LouisLandry:
you forgot to jimport JInput.

ianmacl:
Indeed.  I fixed that and added a unit test to ensure I wasn't doing anything else stupid.

PR #125, opened 2011-07-26
Change to case insensitive check

AmyStephen:


LouisLandry:
Hey Amy,

I'm not sure what the specific problem is on this.  Can you give me an example of how this is causing problems?

AmyStephen:
The component path I was sending in is mixed case while the value I have in JPath is stored in lower case.

Naturally, that means this statement will fail even though the edit is intended to check for out of bounds situations and such is not the case. 

if (strpos($path, JPath::clean(JPATH_ROOT)) !== 0) {

By changing the comparison to stripos, the intended testing works, even with strings not treated the same for case.

It just seems a little more flexible for devs without compromising the edit.

AmyStephen:
BTW - it's not a big deal if you prefer not to make that change. I do get it that ThisPath and thispath are different on certain platforms and maybe in some obscure way, changing it could introduce a million and one vulnerability. :-P I've already cleaned up my path handling. It just seemed a little restrictive and I suspect it'll bite others as they build apps on the platform.

Not worth a lot of time.

LouisLandry:
OK, thanks a bunch for the explanation.  I think i'm gonna just close this one for now.  I'd rather spend a little time making sure the impact before I change something like this.  Thanks though!

PR #129, opened 2011-07-26
Fix JForm::load() not replacing form field in same location

rvsjoen:
When loading another form xml definition on a JForm object, and the new xml definition contains field names which overlap, the fields should be replaced in the same location.

In the current implementation, the old field is removed, and then the new field is added to the end of the form, changing the ordering.

This patch fixes the problem with reordering, but I am not sure if it has any other adverse effects, I will leave that up to someone who knows more about the workings of JForm.

### DIFF
--- a/libraries/joomla/form/form.php
+++ b/libraries/joomla/form/form.php
@@ -661,13 +661,7 @@ class JForm
 				if ($current = $this->findField((string) $field['name'], implode('.', $groups))) {
 
 					// If set to replace found fields remove it from the current definition.
-					if ($replace) {
-						$dom = dom_import_simplexml($current);
-						$dom->parentNode->removeChild($dom);
-					}
-
-					// Else remove it from the incoming definition so it isn't replaced.
-					else {
+					if (!$replace) {
 						unset($field);
 					}
 				}


ianmacl:
1) JFormTest::testLoad
Line:1243 There are 2 original ungrouped fields, 1 replaced and 4 new, resulting in 6 total.
Failed asserting that integer:7 matches expected integer:6.

2) JFormFieldHiddenTest::testGetInput
Line:81 The label of a non-hidden element should be some HTML.
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-<label id="hidden-lbl" for="hidden" class="">foo</label>
+

3) JFormFieldSpacerTest::testGetLabel
Line:101 The getLabel method should return something without error.
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-<span class="spacer"><span class="before"></span><span class=""><label id="spacer-lbl" class="">spacer</label></span><span class="after"></span></span>
+<span class="spacer"><span class="before"></span><span class="text"><label id="spacer-lbl" class="">spacer</label></span><span class="after"></span></span>

4) JFormFieldSpacerTest::testGetTitle
Line:101 The getLabel method should return something without error.
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-<span class="spacer"><span class="before"></span><span class=""><label id="spacer-lbl" class="">spacer</label></span><span class="after"></span></span>
+<span class="spacer"><span class="before"></span><span class="text"><label id="spacer-lbl" class="">spacer</label></span><span class="after"></span></span>

Tests need to be fixed before this is committed.

elinw:
These tests are failing because when the replacement happens the results in the two fields replaces, which are hidden and spacer, are different than what happens when they are not replaced and therefore the actual and expected do not match. For example the non hidden element in the basic test is replaced with a hidden element which should not return html.

Hackwar:
Stupid question, but how is the original field replaced with your change? Does simpleXML recognize the fields by the name attribute or something? I thought you would have to find the current element and replace that element itself with your new element...

elinw:
Yes the name attribute.

chdemko:
@rvsjoen Any news on this (the pull request cannot be merged)?

ianmacl:
We'd like to get this in once we have tests and are confident it works correctly.

ianmacl:
(and once it has been rebased, etc)

elinw:
Ian Aaron posted a pull request for this.

https://github.com/joomla/joomla-platform/pull/1069

ianmacl:
Closing in light of [#1069].

PR #147, opened 2011-07-28
Implement a color field for JForm.

realityking:
This uses the native color picker in Opera 11 and falls back to MooRainbow in all other browsers.

Also adds the corresponding JFormRule for validation.


LouisLandry:
Rock and roll :-)

LouisLandry:
So, I pulled this, but please in the future make your pull requests against staging instead of master.

realityking:
Ah sorry, forgot all about that. Should I change the two other pulls?

LouisLandry:
Yes please. Thanks.

On Jul 29, 2011, at 4:48 AM, realityking
reply@reply.github.com
wrote:

PR #155, opened 2011-07-29
Forms--Add an integer validation rule

elinw:
This is a rule we need in the CMS for leading/intro/links/columns and some modules, but I think there are a lot of use cases. Can also be used to validate the integer field type.

Tests included

### DIFF
--- a/libraries/joomla/form/rules/integer.php
+++ b/libraries/joomla/form/rules/integer.php
@@ -47,30 +47,60 @@ class JFormRuleInteger extends JFormRule
 	 *
 	 */
 	public function test(& $element, $value, $group = null, & $input = null, & $form = null)
-	{
-
-		$regexarray = array(
-			'all' => '/^[\+\-]?[0-9]* $/',
-			'positive'=> '/^[\+]?[0-9]*[1-9]*[0-9]* $/',
-			'negative'=> '/^[-?[0-9]*[1-9]*[0-9]* $/',
-			'nonnegative' => '/^[\+]?[0-9]* $/'
-		);
+	{  
 		// If the field is empty and not required, the field is valid.
 		$required = ((string) $element['required'] == 'true' || (string) $element['required'] == 'required');
-		if (!$required && empty($value)) {
+		if (!$required && strlen($value) == 0) {
+
 			return true;
 		}
+
+		// Deal with character strings
+		if (!is_numeric($value)) {
+
+			return false;
+		}
+		// Deal with doubles
+
+		if ( (double) $value - floor($value) > 0) {
+
+			return false;
+		}
+		$value = (int) $value; 
+
 		// If no integertype is specified assume that all are permitted.
 		if (!isset($element['integertype'])) {
 			$element['integertype'] = 'all';
 		}
-		$inttype = $element['integertype'];
-		
-		$regex = $regexarray[$inttype];
-		// Test the value against the regular expression.
-		if (preg_match($regex, trim($value)) == false || $value > $element['max'] || $value < $element['min'] ) {
+		$inttype = (string) $element['integertype'];
+		//Simple elimination of false results
+		if ( isset($element['max'])) {
+			$max = (int) $element['max'];
+			if ( $value > $max) {
+
+				return false;
+			}
+		}
+		if ( isset($element['min'])) {
+			$min = (int) $element['min'];
+			if ( $value < $min) { 
+
+				return false;
+			}
+		}
+		if ($inttype == 'positive' &&  $value < 1) {
+
+			return false;
+		} 
+		elseif ($inttype == 'nonnegative' && $value < 0) {
+
+			return false;
+		} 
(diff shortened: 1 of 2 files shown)


libraries/joomla/access/access.php
```diff
@@ -367,7 +370,9 @@ class JAccess
 	 * @return  array    List of actions available for the given component and section.
 	 *
 	 * @since   11.1
-	 * @todo    Need to decouple this method from the CMS. Maybe check if $component is a valid file (or create a getActionsFromFile method).
+	 *
```
eddieajau, line 396: Can you drop the spacing line after the @since. The last block should be all together.  Thanks.

libraries/joomla/access/rules.php
```diff
@@ -158,9 +166,9 @@ class JRules
 	/**
 	 * Get the allowed actions for an identity.
 	 *
-	 * @param   mixed  $identity  An integer representing the identity, or an array of identities
+	 * @param   mixed   $identity  An integer representing the identity or an array of identities
```
eddieajau: For a @param, code sniffer requires exactly 2 spaces between the type and the var name.

libraries/joomla/application/application.php
```diff
@@ -342,9 +342,9 @@ class JApplication extends JObject
 	 *
 	 * @return  void  Calls exit().
 	 *
-	 * @see     JApplication::enqueueMessage()
-	 *
 	 * @since   11.1
+	 *
```
eddieajau, line 357: Remove this spacing line.  Thanks.

eddieajau:
Elin, could you run code sniffer locally over your forms branch.  The rule at least will need a little reformatting. Thanks.

elinw:
Ok think this is cool now except that if I do exactly 2 then the comment after the return doesn't line up.

ianmacl:
# Test Results

Total Tests: 1339
Assertions: 2276
Failures: 0
Errors: 0
Test Time: 58.009840

eddieajau:
So, I'm thinking that JFormRule could be bolstered a bit because range checks could be applied to text elements as well (for string length).  So, maybe JFormRule should  have a few new methods:  checkMax (is aware of "max" attribute), checkMin (is aware of "min" attribute), checkDataType (which checks the type of data) and checkDataValue (which, by default, does the regex check and invokes checkMin and checkMax).

Then, possibly, this new rule is called JFormRuleNumber which can handle integers or floats.  I'd probably suggest the "integertype" field is changed to:

positive="<boolean"
negative="<boolean>"
zero="<boolean"

That should cover all the combinations that are needed.

We can then add a JFormRuleText that gives us an easy way to check for a string of certainly length, and maybe that includes an arbitrary "regex" field.

Those are rough thoughts anyway.

elinw:
So the range checks would be a length if a string and a value if a number?  Checking datatype as a general issue makes a lot of sense to me because it seems here (and elsewhere) that we are having to do a lot just to do that and it should not really require separate rules, especially since that is just one small part of what you need to check for.  

I like the positive, negative, zero idea which seems elegant even though i don't think it saves and code :).

So you are saying the maker of the form could supply a regex? I like that a lot also. That will save a lot of adding of custom fields.

elinw:
Also, in thinking about this whole issue you'll notice I added logging for misconfiguration e.g. if the max was less than the min. I am wondering if that is something we should think about doing more generally in JForm (e.g. if you forget to put a value attribute for checkbox it will not save, so log when there is a save of checkbox with no value attribute). Also, what should the person filling out the form see in that case? Should the configuration options just be ignored if they contradict each other? Should save not be allowed (in which case you have to tell the user something or they will go crazy trying to save)? Or should it save null?

Hackwar:
is there anything new regarding this?

jlover:
**Checkstyle error details:**
libraries/joomla/form/rules/integer.php:35
Doc comment for var $element does not match actual variable name &$element at position 1
libraries/joomla/form/rules/integer.php:35
Expected 2 spaces after the longest type
libraries/joomla/form/rules/integer.php:42
Doc comment for var $input does not match actual variable name &$input at position 4
libraries/joomla/form/rules/integer.php:44
Doc comment for var $form does not match actual variable name &$form at position 5
libraries/joomla/form/rules/integer.php:73
libraries/joomla/form/rules/integer.php:113
libraries/joomla/form/rules/integer.php:127
libraries/joomla/form/rules/integer.php:132
libraries/joomla/form/rules/integer.php:141
Please end your files with an empty line.

elinw:
Hannes,

Andrew suggested a more general approach so that's what I'm working on.

eddieajau:
Elin, how about we close this pull request for now?

elinw:
Well since you up the ante considerably :P it's not that close unfortunately. You can close it if you want and i'll reopen when i'm ready to push the newer version.  Hopefully I can get  a block of time to finish it up in the next few weeks.

PR #167, opened 2011-08-01
Logging

elinw:
Fix the problem that was making a test fail


elinw:
This needs to get added to the pull request that is in staging already.

PR #171, opened 2011-08-01
Edited libraries/joomla/document/html/html.php via GitHub

AmyStephen:
No longer able to initialize arrays for JS or CSS.

See issue report http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=24598

### DIFF
--- a/libraries/joomla/document/html/html.php
+++ b/libraries/joomla/document/html/html.php
@@ -151,16 +151,16 @@ class JDocumentHTML extends JDocument
 			return;
 		}
 
-		$this->title = (isset($data['title']) && !empty($data['title'])) ? $data['title'] : $this->title;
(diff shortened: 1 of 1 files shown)


eddieajau:
Thanks Amy.  We've changed the system slightly since you last visited.  Can you make a new pull against the "staging" branch.  What happens is that will run the build and, if successful, it will commit those changes to the "master" branch.  We don't push directly against master anymore unless all our testing passes. It's just part of our QA process now (I'll make a note to get the README file updated to that effect).

Also, would you mind adding a list item to the changelog in the following file, describing the fix/issue, as well:

/docs/manual/appendices/changelog.xml

Thanks in advance.

eddieajau:
I've made some notes here about pull requests and a few other things:

https://github.com/joomla/joomla-platform/wiki/Contributing-to-the-Joomla-Platform

If anything is unclear, please let me know.

PR #174, opened 2011-08-02
Graceful return when update manifest cannot be parsed

rvsjoen:
Instead of just dying when the update manifest cannot be parsed, give a warning and move on.

### DIFF
--- a/docs/manual/appendices/changelog.xml
+++ b/docs/manual/appendices/changelog.xml
@@ -29,6 +29,9 @@
 					validation.</para>
 			</listitem>
 			<listitem>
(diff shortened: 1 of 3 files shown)


eddieajau:
Awesome.  Would you mind adding a list item to the changelog in the following file as well:

/docs/manual/appendices/changelog.xml

all I'll get that merged.  Thanks.

rvsjoen:
I will do that, and I will also as per Ian's request modify it to not use the deprecated JError but instead use JLog.
Will update it shortly.

rvsjoen:
Alright, now it should be good. JLog is new to me so you'll have to verify that i've used it as it should.

PR #175, opened 2011-08-02
If given id is ourselves, behave like no id was given

rvsjoen:
JFactory::getUser() and JFactory::getUser($id) does not behave the same even if $id is the same number as the currently logged in user, so that if a plugin has used JUser::bind() to change the user object, JFactory will not give back the correct object if given an id. This patch will make getUser() check if the given id matches the currently logged in user and behave like no id was given if that is the case.

### DIFF
--- a/libraries/joomla/factory.php
+++ b/libraries/joomla/factory.php
@@ -162,20 +162,22 @@ abstract class JFactory
 	public static function getUser($id = null)
 	{
 		jimport('joomla.user.user');
-
-		if (is_null($id))
-		{
+	
+		if (is_null($id)) {
 			$instance = self::getSession()->get('user');
-			if (!($instance instanceof JUser))
-			{
+			if (!($instance instanceof JUser)) {
 				$instance = JUser::getInstance();
 			}
 		}
-		else
-		{
-			$instance = JUser::getInstance($id);
+		else {
+			$current = self::getSession()->get('user');
+			if($current->id != $id){
+				$instance = JUser::getInstance($id);
+			} else {
+				$instance = self::getSession()->get('user');
+			}
 		}
-
+	
 		return $instance;
 	}
 
--- a/libraries/joomla/factory.php
+++ b/libraries/joomla/factory.php
@@ -163,17 +163,23 @@ abstract class JFactory
 	{
 		jimport('joomla.user.user');
 	
-		if (is_null($id)) {
+		if (is_null($id)) 
+		{
 			$instance = self::getSession()->get('user');
-			if (!($instance instanceof JUser)) {
+			if (!($instance instanceof JUser)) 
+			{
 				$instance = JUser::getInstance();
 			}
 		}
-		else {
+		else 
+		{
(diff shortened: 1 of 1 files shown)


eddieajau:
Looks good Rune.  Would you mind just pulling against master to fix up some of the code formatting and also add a line to the changelog.  Thanks in advance.

rvsjoen:
Done, see if its better now

eddieajau:
Almost.  Only thing that needs fixing a bit is that the opening if curly bracket needs to be on a new line.  Otherwise that looks great.  Thanks!

rvsjoen:
This is from the docs/coding-standards/chapters/php.xml file in the platform, and is the reason why I put the opening brace on the same line for if control structures.

"""
Where the control structure is an if, the if and else keywords are put on newlines and the 
control logic is indented. The opening brace is included on the same line as the if or else and 
the closing brace is on a new line.

For all other control structures, such as do, for, foreach, try, catch, switch and while,
both the keyword starts a newline and the opening and closing braces are each put on a new line.
"""

Should I not be looking at that file ? And if no, are there other places with more up to date information on coding standards ?

rvsjoen:
Updated

ianmacl:
# Test Results

Total Tests: 1280
Assertions: 2217
Failures: 0
Errors: 0
Test Time: 57.051470

eddieajau:
I've updated the docs to reflect our current standard.  Basically we changed if to suit the auto-formatters in most IDE's, but particularly Eclipse and seeing as the if block was the only exception, we changed it to be the same as any other block - so we need standard K&R braces.  The pull is still showing the lines in the changelog too, so just pull them out.

Sorry for mucking you around on this one.

rvsjoen:
Ok I am really confused about the braces at this point. I modified a piece of code to have all opening curly brackets on a newline, and now the code style checker is whining about that. Again I turn to the docs/coding-standards/chapters/php.xml which still says that 

"For all all control structures, such as if, else, do, for, foreach, try, catch, switch and while, both the keyword
    starts a newline and the opening and closing braces are each put on a new line."

So....

Option A.

if (true) {
    /\* scratch myself _/
} else {
    /_ keep itching */
}

Option B.

if (true)
{
    /\* scratch myself _/
}
else
{
    /_ keep itching */
}

Which is it, if any ?

eddieajau:
Option B is the right way :)

ianmacl:
# Test Results

Total Tests: 1235
Assertions: 2169
Failures: 0
Errors: 0
Test Time: 57.914007

rvsjoen:
Ok, lets try this again :-)

ianmacl:
# Test Results

Total Tests: 1235
Assertions: 2169
Failures: 0
Errors: 0
Test Time: 52.880001

eddieajau:
Thanks Rune.  phpcs picked up a few other things we missed:

http://build.joomla.org:8080/job/platform/158/checkstyleResult/new/?

Would you mind fixing them up when you have time.  Thanks in advance.

PR #181, opened 2011-08-05
JDocument fluent interface

WebMechanic:
- changed setters and mutators to return `$this` instead of "void".
- cleanup of duplicate `jimport('joomla.document.document')`

### DIFF
--- a/docs/manual/appendices/changelog.xml
+++ b/docs/manual/appendices/changelog.xml
@@ -40,8 +40,21 @@
 				<para>Implemented graceful failure when update manifest is not available</para>
 			</listitem>
 			<listitem>
-				<para>Allow components to remove the admin menu entry in their postflight script, and not 
+				<para>Allow components to remove the admin menu entry in their postflight script, and not
 				throw errors during uninstall if the menu entries are not there.</para>
+			</listitem>
+			<listitem>
+				<para>Implemented fluid interface for setter and mutator methods in document package.
+					JDocument: parse(), addScript(), addScriptDeclaration(), addStyleSheet(),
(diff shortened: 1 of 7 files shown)


eddieajau:
Great patch!  Just a couple of house keeping things though.
1. We need to pull against the staging branch, not master (that sets off our automated builds)
2. I think you might need to update your branch against the original master. I seem to remember already fixing some of those DocBlocks.
3. A few of the DocBlocks will break our sniff rules.
4. Could you add a few lines to the changelog about your patch.  In particular, could you list each method that now supports chaining.

See this document for more information:
https://github.com/joomla/joomla-platform/wiki/Contributing-to-the-joomla-platform

Thanks in advance.  This will be a great improvement.

WebMechanic:
Thanks for the wiki link. It doesn't actuall tell what branch to **pull from** in the first place to apply patches: `master` or `staging`? 
1. so I'll close this one and re-request to the `staging` branch?
2. My fork's master branch appears to be completely off, so I pulled from this repo to create "fluent_doc". At that time the logs said my working copy was based off "Merge pull request #180 from rvsjoen/adminmenus" -- in both master and staging, so there seemed to be no difference.
3. installed phpcs (too) and fixed whitespace
4. that'd be a **long list**, you're sure about that? Should/may I split each file in its own &lt;para>? otherwise this will look very confusing.

ianmacl:
Master and staging are kept in sync unless there are test failures.  Sending requests against staging makes it easier because we can merge right in github.  Once we pull, tests get run in staging and if tests pass they automatically get pushed to master.

PR #187, opened 2011-08-06
Load component's parameters during installation

vietvh:
Related tracker item: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=26360


LouisLandry:
@pasamio can you take a look at this?  It seems like a great idea, but I'd like you to give it a once over.

@vietvh any chance I could get you to run the code sniffer over your pull and make sure you aren't violating any of the existing sniff rules?  Thanks a bunch for the work!

vietvh:
@Louis: I've just made some code formatting according to Joomla code sniffer rules

pasamio:
Does this still work if the developer puts the extension params in the component manifest file? Just looking through it seems to stop using that capability in preference to requiring a config.xml or perhaps I'm missing something.

vietvh:
I just had no intention to support component's params that put in the component manifest file while making this patch, if we need to keep supporting it, i'll redo this patch.

eddieajau:
Cool.  Would you also mind adding to the change log a short note on how a developer would use this feature.  You can use multiple paragraphs if you want. Thanks!

elinw:
Could something similar be done for injecting baseline rules?

pasamio:
I'd suggest if people are using it then we should support it. It is a supported mechanism for setting configuration parameters and sooner or later someone will mark it as a bug if it isn't supported (e.g. "it works if I do this but not this"). It shouldn't be too hard, if you don't find the config xml file just default it to the manifest one and a small amount of logic to handle that.

vietvh:
The patch now supports configuration data stored in component manifest file.

Hackwar:
This looks good. Anything why this couldn't be merged in?

chdemko:
This patch is not update w/r to current trunk

vietvh:
@chdemko: I've updated my repo

libraries/joomla/installer/installer.php
```diff
@@ -89,6 +89,14 @@ class JInstaller extends JAdapter
 	public $manifest = null;
 
 	/**
+	* The configuration manifest XML object
+	*
+	* @var    object
+	* @since
```
eddieajau: Can you add 11.4 here please (I think that's the one check-style error) and change the at-since tag on any new methods to 11.4 please?  Thanks.

vietvh:
Hi Andrew,

I've updated the docs block as well as moving those changes to a new pull request that based on my git branch, not the master like before.

Here is the link: https://github.com/joomla/joomla-platform/pull/638

PR #189, opened 2011-08-06
Sync changes to the media folder from Joomla 1.7.

realityking:
I made this against staging just to be sure, but for the future should we also make the pull requests against staging when no PHP code is involved?


LouisLandry:
Let's just pull everything against staging for consistency.  We may want to create some codesniffer rules against JS and such in the future and it is best to just keep it simple and the same IMO.  Great stuff as always sir.

PR #190, opened 2011-08-06
Introduce JURI::media().

realityking:
This allows accessing media resources trough a different domain which can give performance benefits. Also may make is easier to use a CDN with Joomla.

Note that the change to JBrowser is slight backward compatibility issue because it causes PHP strict warning when used like the old API. I thought it was a good idea to use a strict method here because this saves us from initializing JBrowser when we don't need it.


LouisLandry:
So I get the idea, and I don't disagree with the concept, but lets talk through a better solution here.  I really would rather not add another static method to JURI, and also you don't need to go through JBrowser to get that SSL information.

realityking:
Sure, let's talk trough this.
1. I used the static function to keep the same style as JURI::root() because most of the time this will be used it will replace a call to JURI::root(). I know this is problematic for unit tests, but isn't this better addressed in one go with the other static methods of JURI?
2. I could of course inline the code from JBrowser::isSSLConnection() - but why?

jonnsl:
Why don't you just use a [protocol-relative URL](http://paulirish.com/2010/the-protocol-relative-url/) instead of having two separate variables in the JConfig?

realityking:
It's quite possible that a site owner only has a SSL certificate for the full URL (www.example.org) so it doesn't include subdomains. With two variables in JConfig it is possible to only use a different (sub-)domain when using a non secure connection.

pasamio:
As a process point this should be done against staging not master

realityking:
You're right, didn't pay attention. I'll make a new pull after we've discussed how it should look like.

BTW It would be awesome if you could contact Github and ask if they could add the feature to define the default branch for pull requests on a repro, maybe they'd do it for a project as big as Joomla :)

ianmacl:

# Test Results

Total Tests: 
Assertions: 
Failures: 
Errors: 
Test Time:

ianmacl:

# Test Results

Total Tests: 
Assertions: 
Failures: 
Errors: 
Test Time:

ianmacl:

# Test Results

Total Tests: 
Assertions: 
Failures: 
Errors: 
Test Time:

ianmacl:

# Test Results

Total Tests: 
Assertions: 
Failures: 
Errors: 
Test Time:

ianmacl:

# Test Results

Total Tests: 
Assertions: 
Failures: 
Errors: 
Test Time:

ianmacl:

# Test Results

Total Tests: 
Assertions: 
Failures: 
Errors: 
Test Time:

chdemko:
Pull request is not against staging

realityking:
Closing to start a new pull against staging.

PR #193, opened 2011-08-07
Call to a member function qn() on a non-object

jonnsl:


eddieajau:
Hi Jonnathan. When are you seeing this error?

WebMechanic:
Same here and always failing. Due to the chaining, `$query` won't be initialized until all methods have "returned".
The call of `->leftJoin($query->qn('#__usergroups') ...` fails, because at this point $query is still `null`.

eddieajau:
This actually works fine for me under PHP 5.3. I suspect it's something to do with PHP 5.2.  However, a better solution would be to change `$query->qn('#__usergroups')` to `$db->qn('#__usergroups')` since we already have the database object (JDatabaseQuery::qn is just a proxy back to it anyway).  The auto-formatter also made mince-meat of this block so if each chain could be put on a new line that would be great too.

One last thing.  Could you update the changelog with a short message (see https://github.com/joomla/joomla-platform/wiki/Contributing-to-the-joomla-platform for more information).  It's helps us all keep things up to date if we get in the habit of doing it with the pull request :)

Thanks in advance.

jonnsl:
I'm getting this error under php 5.3.1, anyway... I updated the changelog and did some code formating.
PS: I hate this 'Comment & Close' button.

eddieajau:
Ha! Yes, too close together.  Thanks for that - looks great.  Can you just merge down from joomla-master so I can auto-merge.  Thanks.

jonnsl:
done

PR #199, opened 2011-08-08
Fixed a typo in JHttp (stream_set_timeout vs stream_settimeout)

nikosdion:
Fixed a typo in JHttp (stream_set_timeout vs stream_settimeout) as it caused a fatal error when using the class ;)


LouisLandry:
Hi Nic, thanks for the fix, but can you re-send this pull request against the staging branch instead of master? You can read https://github.com/joomla/joomla-platform/wiki/Contributing-to-the-joomla-platform for more info on that.  Thanks again!

nikosdion:
Oops! I hadn't noticed the branch drop-down. I submitted the patch to the correct branch now :)

PR #202, opened 2011-08-09
Added missing jimport

ianmacl:
Minor bug fix.


eddieajau:
Did you really want this against master?

PR #205, opened 2011-08-10
Added missing jimport

ianmacl:
Let's try this again.  _sigh_


eddieajau:
Is it worth adding anything to the change log for this?

PR #206, opened 2011-08-10
More low hanging fruit

elinw:


elinw:
Not sure why this didn't upddate but the same commit is in the newer pull request.

PR #201, opened 2011-08-08
Fixed a typo in JHttp (stream_set_timeout vs stream_settimeout) as it cau

nikosdion:
Fixed a typo in JHttp (stream_set_timeout vs stream_settimeout) as it caused a fatal error when using the class. Hopefully, this time I got the pull request against the correct branch.

PR #209, opened 2011-08-10
Updated changelog

nikosdion:
Updated change log with the proposed change, as per Andrew's request at https://github.com/joomla/joomla-platform/pull/201#issuecomment-1776347

### DIFF
--- a/docs/manual/appendices/changelog.xml
+++ b/docs/manual/appendices/changelog.xml
@@ -57,6 +57,9 @@
 					JDocumentXml: setName().</para>
 			</listitem>
 			<listitem>
+				<para>Fixed bug in JHtml: stream_set_timeout was misspelled as stream_settimeout, throwing a Fatal Error each time the class was used</para>
+			</listitem>
+			<listitem>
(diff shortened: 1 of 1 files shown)


eddieajau:
Thanks Nic.  Could you just do an update from joomla/joomla-platform - for some reason it doesn't want to automatically merge.  Thanks!

nikosdion:
Hi Andrew! In theory, if I didn't do an Git-n00b mistake, my branch should be updated with everything from the staging branch and you should be able to perform the merge.

eddieajau:
Thanks. Probably wasn't your fault. Branches can get put of sync fairly easily as we pull oth request. Fortunately though, without the pain that we used to experience in svn world ;)

nikosdion:
Yeah, I saw what happened when I pulled the branch back to my working copy. Another change was committed on the same place of the change log and conflicted with my own. Git makes resolving merge conflicts very easy, though. Too easy, in fact! So much better than SVN. If only the CMS would go GitHub, what a joy that would be :)

PR #210, opened 2011-08-10
Formvalidation Improvements

jonnsl:
- Field of JForm can have more than one validation rule (space-separated)
- JForm::validateField don't validate fields that are empty and are not required.(validation rules don't need to check this anymore)
- Personalized error message for each validation rule
- Added more rules to JForm Validator(Alpha, Alphanum, Digits, Integer, maxLength, minLength, noSpace and numeric)
- Validation rules now throw JException(E_WARNING) instead of return false on validation failed, and throw JException(E_ERROR) on error.


jonnsl:
Before someone merge this pull request, I need some help converting the unit tests for the tel and url validation rules and creating new tests for the color and nospace validation rules, for some weird reason the tests that I made/converted freezes the phpunit when there's a failure but works fine if there's no failure.

eddieajau:
This one has fallen off the radar for some reason - sorry.  Can you up update the branch against staging please?  Thanks.

jonnsl:
updated

elkuku:
Small **personal** error analysis: http://elkuku.github.com/pulltester/pulls/210
The warnings do not concern your code ;)

jonnsl:
Fixed.
I think this should be allowed:
    if ($condition) doSomething();

eddieajau:
I would be moderately well off if I had a dollar for each time I found a bug due to inline if statements ;)

libraries/joomla/form/rules/color.php
```diff
@@ -56,12 +54,9 @@ public function test(&$element, $value, $group = null, &$input = null, &$form =
 		// The value must be 6 or 3 characters long
 		if (!((strlen($value) == 6 || strlen($value) == 3) && ctype_xdigit($value)))
 		{
-			return false;
+			throw new JException($this->getErrorMsg($element), -4, E_WARNING);
```
realityking: I don't think we wanna throw JException in new places, it's deprecated.

elinw:
I think it would be great if you would break these out into individual items.

jonnsl:
I will wait until the pull request #636 get commit them I will update my branch, squash some commits and open a new pull request.

PR #230, opened 2011-08-13
Docs and codesniff

elinw:
Some updates to reduce confusion.


elinw:
Oops this was against master.

PR #231, opened 2011-08-13
Joomla CMS [#25836] Strange behaviour in renderModule ($content variable) 

elinw:
From http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=25836 which was in the CMS for a few months waiting for review and now Andrew closed.  It was already tested by people in the CMS. It just eliminates an unnecessary variable which was causing a name clash for some developers.

### DIFF
--- a/libraries/joomla/application/module/helper.php
+++ b/libraries/joomla/application/module/helper.php
@@ -171,10 +171,9 @@ abstract class JModuleHelper
 				$lang->load($module->module, JPATH_BASE, $lang->getDefault(), false, false) ||
 				$lang->load($module->module, dirname($path), $lang->getDefault(), false, false);
 
-			$content = '';
 			ob_start();
 			include $path;
-			$module->content = ob_get_contents() . $content;
+			$module->content = ob_get_contents() . '';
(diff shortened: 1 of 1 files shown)


ianmacl:
# Test Results

Total Tests: 1280
Assertions: 2217
Failures: 0
Errors: 0
Test Time: 56.952886

eddieajau:
I suspect the intention is (or was) that a module can return the content in a $content variable.  On that basis $content should be considered reserved.  This may be a hang over from a previous version and no longer used.  If that is the case, the feature needs to be deprecated, and not removed suddenly.  Not accepting the change.

elinw:
Agreed about the original purpose of $content based on other things related to early versions of the custom module. I am going to copy this comment to the original issue and close it on the CMS tracker.

tompap:
Hello, I wrote the initial issue on the CMS Tracker. I agree that should be set a deprecated. But I would suggest to write somewhere that $content is reserved. Because there is no information about that in the wiki or any other documents. And the word "content" is a common word in joomla word.

PR #234, opened 2011-08-13
Joomla CMS [#26353] Invalid opendir + readdir construct is used in Joomla 

elinw:
Based on http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26353

Addresses an issue when folders are unwritable.

### DIFF
--- a/libraries/joomla/filesystem/path.php
+++ b/libraries/joomla/filesystem/path.php
@@ -78,6 +78,10 @@ class JPath
 
 		if (is_dir($path))
 		{
+			if (!($handle = @opendir($path)))
+			{
+				return $ret;
+			}
 			$dh = opendir($path);
 
 			while ($file = readdir($dh))


ianmacl:
# Test Results

Total Tests: 1280
Assertions: 2217
Failures: 0
Errors: 0
Test Time: 56.808714

ianmacl:

# Test Results

Total Tests: 
Assertions: 
Failures: 
Errors: 
Test Time:

elinw:
It says against staging at the  top???

libraries/joomla/filesystem/path.php
```diff
@@ -78,6 +78,10 @@ class JPath
 
 		if (is_dir($path))
```
eddieajau, line 79: Would it be slightly better to have this?:

if (is_dir($path) && is_readable($path))

Alternatively we could swap to using a directory iterator like we do in JLoader::discover.  That would be much better because we can just catch an exception if there are any problems.

elinw:
Andrew, Christophe raised an issue in the CMS tracker that we don't necessarily know where there could be potential issues with this. What do you think?

eddieajau:
What are the potential issues?

elinw:
Honestly I'm not sure, I think it is more hypothetical.
Would you mind looking at this issue where there is a discussion about the same general topic http://code.joomla.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26298?

elinw:
Should we also do something like you suggest for is_file?

```
        // Note: DirectoryIterator::getExtension only available PHP >= 5.3.6
```

I think it won't solve the problems on the CMS side but it probably makes sense to do when possible.

eddieajau:
This seems to be a bug further upstream.  The problem appears to be the path that is sent to the API, not the API itself.  While a sanity check on the path would still be advisable, it's not the root cause and would only mask the issue raised in the CMS.  I'm going to close this pull request as I think we can do better refactoring that method.

PR #237, opened 2011-08-14
Allow package extension types to use scriptfiles

mbabker:
This pull request makes changes to the package extension installer to allow for a scriptfile to be included on a package.

This will allow for a package to have a single preflight method, for example to check for any dependencies, such as another extension's presence or a minimum Joomla! version.

Additionally, a $results array has been created with each extension's name and $tmpInstaller->install result (which should be true given where it is compiled).  This array is pushed forward to the postflight method, and can be used to display a list of the installed extensions, something many 3PD already do with their custom installation methods (see http://twitpic.com/65xw3i for an example).

Lastly, the TODO concerning the extension_root has been resolved by setting it to JPATH_MANIFESTS/packages/packagename and this is the location the scriptfile is stored in.

### DIFF
--- a/libraries/joomla/installer/adapters/package.php
+++ b/libraries/joomla/installer/adapters/package.php
@@ -90,7 +90,7 @@ class JInstallerPackage extends JAdapterInstance
 		$group = (string) $this->manifest->packagename;
 		if (!empty($group))
 		{
-			$this->parent->setPath('extension_root', JPATH_ADMINISTRATOR . '/manifests/packages/' . implode(DS, explode('/', $group)));
+			$this->parent->setPath('extension_root', JPATH_MANIFESTS . '/packages/' . implode(DS, explode('/', $group)));
 		}
 		else
 		{
--- a/libraries/joomla/installer/adapters/package.php
+++ b/libraries/joomla/installer/adapters/package.php
@@ -105,7 +105,7 @@ class JInstallerPackage extends JAdapterInstance
 		 */
 		// If there is an manifest class file, lets load it; we'll copy it later (don't have dest yet)
 		$this->scriptElement = $this->manifest->scriptfile;
-		$manifestScript = (string)$this->manifest->scriptfile;
+		$manifestScript = (string) $this->manifest->scriptfile;
 
 		if ($manifestScript)
 		{
--- a/libraries/joomla/installer/adapters/package.php
+++ b/libraries/joomla/installer/adapters/package.php
@@ -109,7 +109,7 @@ class JInstallerPackage extends JAdapterInstance
 
 		if ($manifestScript)
 		{
-			$manifestScriptFile = $this->parent->getPath('source').'/'.$manifestScript;
+			$manifestScriptFile = $this->parent->getPath('source') . '/' . $manifestScript;
 
 			if (is_file($manifestScriptFile))
 			{
@@ -118,7 +118,7 @@ class JInstallerPackage extends JAdapterInstance
 			}
 
 			// Set the class name
-			$classname = $element.'InstallerScript';
+			$classname = $element . 'InstallerScript';
 
 			if (class_exists($classname))
 			{
--- a/libraries/joomla/installer/adapters/package.php
+++ b/libraries/joomla/installer/adapters/package.php
@@ -90,7 +90,7 @@ class JInstallerPackage extends JAdapterInstance
 		$group = (string) $this->manifest->packagename;
 		if (!empty($group))
 		{
-			$this->parent->setPath('extension_root', JPATH_ADMINISTRATOR . '/manifests/packages/' . implode(DS,explode('/',$group)));
+			$this->parent->setPath('extension_root', JPATH_ADMINISTRATOR . '/manifests/packages/' . implode(DS, explode('/', $group)));
 		}
 		else
 		{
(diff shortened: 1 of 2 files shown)


pasamio:
Do you have a link to a test package that could be included with the installer sample packages? (they're currently living in the CMS project)

mbabker:
I'll take the example code that is already in the SVN and tweak it to use these changes.  Give me a day or so to get that done.

mbabker:
I've branched the SVN trunk and tweaked the package already there to demonstrate all the features that I'm aware for this extension type.  The scriptfile includes a working example of what I'm proposing with this pull request on postflight.  Since I've got the branch going there, I'll hop through each of the extensions and clean them up then get that merged later.  The branch is at http://joomlacode.org/svn/joomla/development/branches/sample_exts/

eddieajau:
Michael, I've changed the system so we can generate the changelog from the pull request messages.  There's no need to add the changelog anymore.  Your branch will also need updating as git reports it can't be automatically merged.  Thanks.

mbabker:
I've merged to master and in the merge I pulled my changelog entry since it conflicted anyways.  Should be good to go now.

jlover:
mine found only 3 :P
**Checkstyle error details:**
libraries/joomla/installer/adapters/package.php:113
libraries/joomla/installer/adapters/package.php:113
libraries/joomla/installer/adapters/package.php:122

eddieajau:
@pasamio what do you think on this one?

pasamio:
Seems fine to me, I've always wondered about packages having their own scripts but figured that the components, modules or plugins would take it over. Really have no opinion either way.

libraries/joomla/installer/adapters/package.php
```diff
@@ -90,8 +90,7 @@ public function install()
 		$group = (string) $this->manifest->packagename;
 		if (!empty($group))
 		{
-			// TODO: Remark this location
-			$this->parent->setPath('extension_root', JPATH_ROOT . '/packages/' . implode(DS, explode('/', $group)));
+			$this->parent->setPath('extension_root', JPATH_MANIFESTS . '/packages/' . implode(DS, explode('/', $group)));
```
marcel-kanter, line 93: I would not change this, since this changes the path for all files. The scriptfile should they together with the rest anyway.

libraries/joomla/installer/adapters/package.php
```diff
@@ -99,6 +98,56 @@ public function install()
...
+			$manifestScriptFile = $this->parent->getPath('source') . '/' . $manifestScript;
+
+			if (is_file($manifestScriptFile))
+			{
+				// load the file
+				include_once $manifestScriptFile;
+			}
+
+			// Set the class name
+			$classname = $element . 'InstallerScript';
```
marcel-kanter, line 121: Think about sharing the InstallerScript. I would add:
if(!class_exists($classname))
{
  $classname = 'InstallerScript';
}

libraries/joomla/installer/adapters/package.php
```diff
@@ -276,6 +395,46 @@ public function uninstall($id)
...
+			$manifestScriptFile = $this->parent->getPath('extension_root') . '/' . $manifestScript;
+
+			if (is_file($manifestScriptFile))
+			{
+				// Load the file
+				include_once $manifestScriptFile;
+			}
+
+			// Set the class name
+			$classname = $row->element . 'InstallerScript';
```
marcel-kanter, line 413: Think about sharing the InstallerScript. I would add:
if(!class_exists($classname))
{
$classname = 'InstallerScript';
}

libraries/joomla/installer/adapters/package.php
mbabker, line 93: A package extension has no true root. Since it is just a package of extensions, it truly only has a manifest and optional language files.  Therefore, logically, a script file should be placed with the single required file for a package extension, the manifest.  Well, that's my 2 cents anyways.

-Michael

marcel-kanter, line 93: We're thinking the same. I did not came to take a closer look into this code. So JPATH_ROOT and JPATH_MANIFESTS point to equal fodlers?

mbabker, line 93: JPATH_ROOT goes to the site root.  There's no packages folder at the site root.  JPATH_MANIFESTS points to administrator/manifests, which is the path that manifest files for file, library, and package extensions are placed at.  The TODO with the path leads me to think that at one point, there was an intent to use that path, but it's not used, so that was a logical change based on practical use.

-Michael

marcel-kanter, line 93: Hmm. Then the handling of the manifests of component is different to package and my code is copying the script into the base. Should we equalise this for all extension types? Bend component and the rest into JPATH_MANIFESTS?

mbabker, line 93: Negative.  Component, module, and plugin manifests are fine as is and moving them would require numerous other changes.

-Michael

pasamio, line 93: More important, moving them would break any upgraded site which already has these files which would mean adding extra code to check where they are. Essentially packages, and to a lesser extent libraries and files, don't really have a main extension root like components, modules, templates, plugins and languages do have significant files and it makes more sense to co-locate their manifest files with them, as has traditionally been the case.

And yes, it should have been updated.

libraries/joomla/installer/adapters/package.php
pasamio, line 121: No because then you could be lead into a situation where there are duplicated class names in say a package install which would cause a PHP fatal error.

marcel-kanter:
Take a look into https://github.com/joomla/joomla-platform/pull/702

PR #247, opened 2011-08-15
Check if module properly is set

elinw:
In testing the JBs found a regression in that in getModule the module helper is now not setting $module->user to 0 for all modules any more. I did some digging and this seems to be a legacy property from 1.0 and seemed to possible check if you were a super user for menu modules.  As far as I can tell it has no use at all in the current CMS. (Looking at the 1.0 module code explains a lot about some other parts of the module helper.)

### DIFF
--- a/libraries/joomla/application/module/helper.php
+++ b/libraries/joomla/application/module/helper.php
@@ -162,7 +162,7 @@ abstract class JModuleHelper
 		$path = JPATH_BASE . '/modules/' . $module->module . '/' . $module->module . '.php';
 
 		// Load the module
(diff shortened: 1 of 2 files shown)


eddieajau:
I think you can use `empty($module->user)` instead of `(!isset($module->user) || !$module->user )`.  That will give the same effect and maintain the status quo.  I'm not sure what it's there for either so I'm with Rouven. Removing the check blindly is probably not a wise move.

elinw:
Updated to going back to setting that property even though it does nothing at this point since it is referenced in a few other places.

PR #250, opened 2011-08-16
Logging

elinw:
Currently import loads JError and JException both of which are deprecated. What happens is that every time JError or JException are called is that they call JLog to log use of deprecated methods. This causes a number of issues if JLog has not been loaded elsewhere including an infinite loop.  So if we are loading the first two we need to load JLog. Thus far the CMS seems to be managing okay but other applications including installation are not.

### DIFF
--- a/docs/manual/appendices/changelog.xml
+++ b/docs/manual/appendices/changelog.xml
@@ -68,7 +68,7 @@
 				<para>Fixed bug in JHtml: stream_set_timeout was misspelled as stream_settimeout, throwing a Fatal Error each time the class was used</para>
 			</listitem>
 			<listitem>
-				<para></para>
+				<para>Added loading of JLog to import.php.</para>
 			</listitem>
 			<listitem>
 				<para></para>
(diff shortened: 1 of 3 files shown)


libraries/import.php
```diff
@@ -54,9 +54,10 @@ if (!class_exists('JLoader'))
 // Import the factory library.
 JLoader::import('joomla.factory');
 
-// Import the exception and error handling libraries.
+// Import the exception, error handling and logging libraries.
 JLoader::import('joomla.error.error');
 JLoader::import('joomla.error.exception');
+JLoader::import('joomla.log.log');
```
eddieajau, line 60: I think it would be better to move this line into error.php itself.  That keeps import.php as light as possible.

docs/manual/appendices/changelog.xml
```diff
@@ -68,7 +68,7 @@
 				<para>Fixed bug in JHtml: stream_set_timeout was misspelled as stream_settimeout, throwing a Fatal Error each time the class was used</para>
 			</listitem>
 			<listitem>
-				<para></para>
+				<para>Added loading of JLog to import.php.</para>
```
eddieajau, line 71: No need to update the changelog anymore.  We can get this automatically from the merged pull requests.

elinw:
Ok will send that up in a bit.

elinw:
See replacement pull request.

PR #251, opened 2011-08-16
Turn JAuthentication back from static class and updated tests accordingly

pasamio:
This fixes an issue with syncing the CMS where the CMS breaks because the plugins aren't loaded properly. This restores the pre-11.2 behaviour of not using a static class.


pasamio:
Sorry, wrong branch

PR #253, opened 2011-08-17
Logging2

elinw:
This is to replace the other pull request about JLog where the deprecation logging is causing an infinite loop.


elinw:
Replaced

PR #256, opened 2011-08-18
Fixes login and module display issues in the CMS 

eddieajau:
Using database magic calls (qn and q) causes problems in the session database class, so these have been replaced with the fully qualified names.
A small glitch in the module display was fixed.
JAuthentication::authenticate was still being defined as static which caused problems with the plugins.

### DIFF
--- a/libraries/joomla/application/module/helper.php
+++ b/libraries/joomla/application/module/helper.php
@@ -163,7 +163,7 @@ abstract class JModuleHelper
 
 		// Load the module
 		// $module->user is a check for 1.0 custom modules and is deprecated refactoring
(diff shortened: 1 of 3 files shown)


ianmacl:
# Test Results

Total Tests: 1280
Assertions: 2217
Failures: 0
Errors: 0
Test Time: 55.837949

elinw:
Andrew I believe you have that module change wrong. $module->user is only not empty when there is a 1.0 style custom module, which is to say that it is type custom rather than mod_custom. So this in fact should never be  true in a current instance of the CMS.

elinw:
Ok do not write comments on pull requests at 4 in the morning. Sorry about misreading that :P.

PR #257, opened 2011-08-18
Adds JArrayHelper::pivot, improves JArrayHelper::sortObjects and provides full code coverage.

eddieajau:
Alphabetised methods in JArrayHelperTest and adds test for arrayUnique method.

Refactored unreachable code in _fromObject (arrays cannot be processed without the recurse argument being true. Added test case to verify).

Changed testSortObjects to do most of the tests on a mac (ignoring only those that cause problems).

Refactors sortObjects method to use internal static variables for sort properties, instead of some arbitrary [bad practice] global variable.

Added JArrayHelper::pivot which allows you to create pure reverse lookups on an array of scalar, array or object elements, or lookups based on collections of common values or the designated array/object key.


ianmacl:
# Test Results

Total Tests: 1287
Assertions: 2224
Failures: 0
Errors: 0
Test Time: 57.137553

Issue #246, opened 2011-08-14
Suggestion for improved error handling during extension installation (Joomla issue tracker id=25676)


jmcameron:
(From http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=25676)

In line 1235 of libraries/joomla/installer/adapters/component.php, there is the
following line (when building a menu):

```
// If the install fails, exit and roll back changes
if (!$table->setLocation(1, 'last-child') || 
    !$table->bind($data) || 
    !$table->check() || 
    !$table->store()) {
     // Install failed, rollback changes
     return false;
     }
```

I suggest changing this to:

```
// If the install fails, exit and roll back changes
if ( !$table->setLocation(1, 'last-child') ) {
    JError::raiseWarning(1, $table->getError());
    }
if ( !$table->bind($data) ) {
    JError::raiseWarning(1, $table->getError());
    }
if ( !$table->check() )  {
    JError::raiseWarning(1, $table->getError());
    }
if ( !$table->store() ) {
    JError::raiseWarning(1, $table->getError());
    }
```

Each of these functions produces very useful error messages which are being discarded the way the code is current written.   This addition would be very helpful to help extension developers debug their component installation problems. 

-Jonathan

P.S. Although Sam feels that this is not the long-term approach he advocates for error handling, I still feel that something needs to be done on this particular section of code sooner rather than later to make it easier for extension developers to debug their code!

I am willing to do a fork to change this as I suggested above as in interim measure until a more effective long-term solution is implemented.  Just let me know.  -Jonathan
# Follow up comments from the Joomla issue tracker:

Submitted By: Samuel Moffatt
Adddate: 2011-04-24 21:37:31
You don't need to split it out though, just add the JError::raiseWarning before the return as it stands now. The  || will halt execution at the first failure and drop into the false. All you're doing is duplicating what it does now.

In general with 11.1 we're moving towards more logging than the way JError behaves at the moment. I'm not quite sure what that looks like for the installer as it already has a message infrastructure in place of its own that you could set.

---

Submitted By: Andrew Eddie
Adddate: 2011-08-12 21:44:32
Please raise this issue on the Joomla Platform issue tracker
(https://github.com/joomla/joomla-platform/issues), including a link back to
this page, or fork the Platform, make the appropriate changes and issue a pull
request.

Thanks.  Closing issue.

---

Submitted By: Elin Waring
Adddate: 2011-08-14 09:09:46
[...]

PR #264, opened 2011-08-20
Joomla CMS [#25676] Suggestion for improved error handling for libraries/joomla/installer/adapters/component.php 

jmcameron:
Improves error reporting if there is any problem building the admin menus when installing a component.

See:
https://github.com/joomla/joomla-platform/issues/246
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=25676

### DIFF
--- a/libraries/joomla/installer/adapters/component.php
+++ b/libraries/joomla/installer/adapters/component.php
@@ -1441,13 +1441,7 @@ class JInstallerComponent extends JAdapterInstance
 			if (!$table->setLocation(1, 'last-child') || !$table->bind($data) || !$table->check() || !$table->store())
 			{
 				// Install failed, warn user and rollback changes
-				JLog::add($table->getError(), JLog::WARNING, 'installer');
-
-				if (JError::$legacy)
-				{
-					JError::raiseWarning(1, $table->getError());
-				}
-
+				JError::raiseWarning(1, $table->getError());
 				return false;
 			}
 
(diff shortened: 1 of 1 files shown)


ianmacl:
# Test Results

Total Tests: 1382
Assertions: 9860
Failures: 0
Errors: 0
Test Time: 193.078545

eddieajau:
I definitely agree we are missing some error reporting here, but why wouldn't you just use a normal abort and put in:

`$this->parent->abort($this->getError());`

before the `return false;`?

jmcameron:
See the response in https://github.com/joomla/joomla-platform/pull/264

eddieajau:
I think just raise the warning for now.  Don't do the JLog because that will already be handled in JError.

I think with the legacy check either remove it, or 'else' throw an exception.  But, since we don't have good exception support yet, and the Installer package is highly CMS centric at the moment, I think just remove the legacy check.  It'll have to go on the TODO list when we remove JError.

Does that make sense?

jmcameron:
Ok, I'll update the patch and resubmit.  Should this pull request be closed and then I'll create a new one with an updated fork?   If so, you can close it when you respond.

eddieajau:
Hi Jonathan

Just update your master branch.  The pull requests are done via branch so whatever work you push to them before they are merge will be reflected in the request.  Makes it very easy to fix stuff up :)

jmcameron:
Eddie,  I've done the changes you suggested in my fork.  I checked the 'diffs' above and they reflect my changes.   What else do I need to do?  I use git quite a bit, but have not done much forking.  And I'm just starting to figure out github.  Thanks -Jonathan

eddieajau:
That looks fine to me.  You don't need to do anything else.  On to the next bug :)

PR #274, opened 2011-08-21
Issue #246 [Joomla! CMS Issue Tracker Item #25835] JTable:getInstance doesn't manage correctly multiple includePaths

tompap:
Issue #246 [Joomla! CMS Issue Tracker Item #25835] JTable:getInstance doesn't manage correctly multiple includePaths

### DIFF
--- a/libraries/joomla/database/table.php
+++ b/libraries/joomla/database/table.php
@@ -169,10 +169,11 @@ abstract class JTable extends JObject
 			// Search for the class file in the JTable include paths.
 			jimport('joomla.filesystem.path');
 
-			if ($path = JPath::find(JTable::addIncludePath(), strtolower($type) . '.php'))
-			{
-				// Import the class file.
-				include_once $path;
+			if ($paths = JPath::findAll(JTable::addIncludePath(), strtolower($type).'.php')) {
+        			// Import the class files.
+        			foreach($paths as $path) {
+        			   require_once $path;
+      				}
 
 				// If we were unable to load the proper class, raise a warning and return false.
 				if (!class_exists($tableClass))
--- a/libraries/joomla/filesystem/path.php
+++ b/libraries/joomla/filesystem/path.php
@@ -300,4 +300,50 @@ class JPath
 		// Could not find the file in the set of paths
 		return false;
 	}
+	
+	/**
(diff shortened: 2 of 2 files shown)


ianmacl:
So the basic idea is to include every file found with the given name?

I would think that would cause problems if somebody is trying to do overrides of any sort.

ianmacl:
# Test Results

Total Tests: 1389
Assertions: 9867
Failures: 0
Errors: 0
Test Time: 193.526119

LouisLandry:
I'd much rather see this approached simply by deprecating the addIncludePath method and using JLoader::discover() instead.  That will give us a much better ability to override classes and provide a much cleaner loading mechanism as well.  http://api.joomla.org/Joomla-Platform/JLoader.html#discover

tompap:
Hello, I agree that they must be a better solution for the JPath modification. But we need a way to be able to find all the paths that contains a specific file.

Looking to the description of jTable::getInstance:
- Static method to get an instance of a JTable class if it can be found in the table include paths.

This is not the current behaviour of this function, because we just search in the first table include path that match

Another Solution, will be to add an addtionnal parameter to the function JPath::find, $multiplesearch something like that, to tell that we have not to stop on the first occurence but to search all occurence.

My choice was to create a findAll, like we have in php preg_match and preg_match_all

eddieajau:
I agree. We need to head towards using the spl autoload features more.  The addIncludePath is really a hangover from the PHP 4 days.  Thanks for working on a solution, but it's not quite what we are looking for so I'm going to close this issue on that basis.

tompap:
Hello, I really don't understand why you want to close this issue. 
There is a BUG currently in Joomla.

On "Core" Point of view you cannot see the bug, but for third party developer, This issue apperas a lot of time, because getInstance of JTable is not working correctly. If several components are using the same name of Table (common name like User, Content for example), the JTable is not working.

Could you explain why you want to close an existing bug ? That can be reproduced very easily ?

eddieajau:
What you need to add in your code is this:

JLoader::register('MyTableUser', $pathToClass);

or

JLoader::discover('MyTable', $pathToTables);

That will fix your problem and be future proof.

PR #277, opened 2011-08-23
Fixes bad default value for the offset in JString::strrpos; more options for JHtmlTruncate.

eddieajau:
Fixes a platform dependent error (utf8_strrpos expects parameter 3 to be long) caused by the offset of JString::strrpos having a default of false instead of integer zero.

Also adds options to JHtmlString::truncate for allowing a word to be split at the length marker, and also to optionally strip out all HTML tags.

### DIFF
--- a/libraries/joomla/html/html/string.php
+++ b/libraries/joomla/html/html/string.php
@@ -22,68 +22,95 @@ abstract class JHtmlString
 {
 	/**
 	 * Truncates text blocks over the specified character limit and closes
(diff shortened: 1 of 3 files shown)


ianmacl:
# Test Results

Total Tests: 1415
Assertions: 9893
Failures: 0
Errors: 0
Test Time: 199.176902

ianmacl:
# Test Results

Total Tests: 1418
Assertions: 9896
Failures: 0
Errors: 0
Test Time: 184.513546

PR #281, opened 2011-08-23
Added custom names support for "special" columns in JTable

tampe125:
Added custom names support for "special" columns in JTable (ordering, hits etc etc)

In this way you are no longer required to call your columns in a fixed way, but you can add a suffix ie: cu_ordering, cu_hits, etc etc.

### DIFF
--- a/libraries/joomla/database/table.php
+++ b/libraries/joomla/database/table.php
@@ -37,6 +37,55 @@ abstract class JTable extends JObject
 	 * @since  11.1
 	 */
 	protected $_tbl_key = '';
+	
+	/**
+	 * Name of the asset_id column
+	 * 
+	 * @var string
+	 */
+	protected $_asset = 'asset_id';
+	
+	/**
+	 * Name of the access column
+	 * 
+	 * @var string
+	 */
+	protected $_access = 'access';
+	
+	/**
+	 * Name of the hits counter column
+	 * 
+	 * @var string
+	 */
+	protected $_hitsColumn		= 'hits';
+	
+	/**
+	 * Name of the publish column
+	 * 
+	 * @var string
+	 */
+	protected $_pubColumn  		= 'published';
+	
+	/**
+	 * Name of the ordering column
+	 * 
+	 * @var string
+	 */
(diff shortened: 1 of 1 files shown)


ianmacl:
# Test Results

Total Tests: 
Assertions: 
Failures: 
Errors: 
Test Time:

tampe125:
just a question, is everything ok?

ianmacl:
Tests did not complete:
PHP Fatal error:  Maximum function nesting level of '100' reached, aborting! in /home/ian/www/platform/pulltester/checkouts/pull281/libraries/joomla/log/log.php on line 45

Fatal error: Maximum function nesting level of '100' reached, aborting!

This won't be pulled until that is fixed or a reason is determined.  Please try the tests locally.

tampe125:
damn..
ok i'll check it out asap

eddieajau:
I like the general idea, but I'm wondering if it would be better to add a single, protected $columnAlias property which is an array of the custom mappings (note, protected properties don't have an underscore prefix).  The array could be empty by default and this also allows us to expand it at a later date without having to add new properties to the class.

You could also add two other protected methods, getColumnAlias (to use int, for example, the publish method) and setColumnAlias (to use in the constructor).

I don't think you need a configuration option in the constructor.  Rather, force the developer to use that new setColumnAlias I mentioned above when they override the constructor in the child class (or they can just explicitly set the array in the class properties).

What do you think?

tampe125:
thanks eddieajau, it's a good idea.
i still have to check why tests fail :(

tampe125:
sorry but i can't get the test framework up and running  (as stated here https://groups.google.com/forum/#!topic/joomla-dev-cms/M0zeZURSc5Q) ;(

i hope someone else can carry on this pull request

elinw:
Ian, that's the message you get when you run the tests in the libraries folder instead of the joomla folder ...

eddieajau:
I'm going to close this pull for now since there is a bit of work to be done on it one way or the other.  Open another pull when you've had a chance to look at the suggestions for how to improve the feature.  Thanks in advance.

PR #285, opened 2011-08-24
Fixes check style warnings and errors.

eddieajau:


ianmacl:
# Test Results

Total Tests: 1420
Assertions: 9898
Failures: 0
Errors: 0
Test Time: 177.701600

libraries/joomla/base/node.php
```diff
@@ -49,7 +49,7 @@ class JNode extends JObject
 	 *
 	 * If the child already has a parent, the link is unset
 	 *
-	 * @param   JNode  $child  The child to be added
```
ianmacl, line 52: What does our standard say about object/Class Name?  In file above, we have object, here we have JNode?

eddieajau:
The standard doesn't say anything about it (yet), but I'll add that we should aim to include the correct object type, or at least a common ancestor.  This would be preferred to "object", but I was more worried about eliminating check style warnings on this one.

elinw:
What we have kind of been doing is for @return putting the name in some places ... but it is more readable if we put  

@param object $child A JNode object to be added as the child node.
beause some of those object names are very long.. Sooner or later we will have to do a sweep to make this consistent.

eddieajau:
If you do

@param JNode $child Description

IDE's like PDT will give you type hinting and auto-completion on $child because it knows what class it is.

PR #294, opened 2011-08-26
Joomla CMS [#26694] Catchable fatal error: Object of class DateTimeZone could not be converted to string 

eddieajau:
Fixes error in JFactory::getDate where timezone is passed as a DateTimeZone.

Restores backward compatibility of JFactory::getDate to Joomla 1.5 in that is caches the date for a given time.

Adds unit test for JFactory::getDate.

Adds addition test case for JDate constructor where timezone is passed as a DateTimeZone object.

Adds universal JLanguage mock to Joomla test harnesses.

### DIFF
--- a/libraries/joomla/factory.php
+++ b/libraries/joomla/factory.php
@@ -506,18 +506,20 @@ abstract class JFactory
 				$classname = 'JDate';
 			}
 		}
-		$key = $time . '-' . $tzOffset;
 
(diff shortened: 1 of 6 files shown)


ianmacl:
# Test Results

Total Tests: 1280
Assertions: 2217
Failures: 0
Errors: 0
Test Time: 54.984663

PR #296, opened 2011-08-26
Adds headers argument to JHttp get, head and post, and some tests for JHttp.

eddieajau:
This pull adds an addition $headers argument to JHttp's get, head and post methods so you can define additional headers for each request.  It also adds some tests for the JHttp class.


ianmacl:
# Test Results

Total Tests: 1425
Assertions: 9919
Failures: 0
Errors: 0
Test Time: 183.527143

PR #297, opened 2011-08-26
Fix infinite loop when there is no error page

elinw:


ianmacl:
# Test Results

Total Tests: 1422
Assertions: 9913
Failures: 0
Errors: 0
Test Time: 182.538718

PR #301, opened 2011-08-27
Added custom name support for "special" JTable columns (ie ordering, hits etc etc)

tampe125:


tampe125:
i worked on JTableNested too, i hope this helps

libraries/joomla/database/table.php
```diff
@@ -37,6 +37,13 @@ abstract class JTable extends JObject
 	 * @since  11.1
 	 */
 	protected $_tbl_key = '';
+	
+	/**
+	 * Array with alias for "special" columns such as ordering, hits etc etc
+	 * 
+	 * @var array
```
eddieajau: We need an @since 11.3 tag here.  Make sure you line things up properly (look at the docblock above as an example).

libraries/joomla/database/table.php
```diff
@@ -270,6 +278,17 @@ abstract class JTable extends JObject
 		return $this->_getAssetName();
 	}
 
+	protected function getColumnAlias($column)
```
eddieajau, line 293: Docblock is needed for this method and the next.  The @since tag will be 11.3.

libraries/joomla/database/table.php
```diff
@@ -270,6 +278,17 @@ abstract class JTable extends JObject
...
+	{
+		if(isset($this->columnAlias[$column])) return $this->columnAlias[$column];
```
eddieajau: Code formatting is not to standard.  Look at how we do if-blocks in other parts of the code.

libraries/joomla/database/table.php
tampe125: ok i'll add asap.
i didn't know if and when this pull request would be integrated :D

tampe125:
ehm.. did i passed the tests?
i run them locally and everything seems to be ok

elinw:
Oh it looks like the automated tester is broken.

eddieajau:
Looking good so far.  A couple more suggestions:
1. Where we use columnAlias[$column] in getColumnAlias and setColumnAlias it's probably a good idea to strtolower the key of the array.
2. I think we need to sanitise the return value of getColumnAlias, just in case someone puts in a silly typo.
   
   $return = preg_replace('#[^A-Z0-9_]#i', '', $return);

should be enough for now.

Would you also like to have a go at doing a unit test for getColumnAlias and setColumnAlias (with help obviously)?

tampe125:
now the input/output column name check should be ok.

i added test placeholders, if you can help me to set them up i'll do for sure!

eddieajau:
Ok, looking good. The next step is to flesh out the test stubs. The first thing you'll need in an inspector class for JTable. I've written up some notes here about how to build an inspector:

https://github.com/eddieajau/joomla-platform/blob/documentation/docs/manual/chapters/testing.xml

You'll need to create JTableInspector.php that includes the magic __get and __set and also two public overrides for getColumnAlias and setColumnAlias.

Then, you'll include the inspector class in the JTableTest.php file at the head of the file:

```
require_once __DIR__.'/JTableInspector.php';
```

Your tests will probably start something like:

```
$table = new JTable('foo', 'bar', $this->getMockDatabase());
```

setColumnAlias is the easiest to test, something like:

```
$this->setColumnAlias('published', 'state');
$this->assertThat(
    $table->columnAlias['published'],
    $this->equalTo('state'),
    'Tests that the alias was correctly added to the columnAlias array.'
);
```

getColumnAlias is similar, but there are more cases to test:

a) where no alias is available
b) where a good alias is provided
c) where an alias is provided but has bad characters

To make that test, you'll do something like this:

```
$table = new JTable('foo', 'bar', $this->getMockDatabase());
$table->columnAlias = array(
    'published' => 'state'
);

$this->assertThat(
    $table->getColumnAlias('access'),
    $this->equalTo('access'),
    'Tests that when no alias is available, the original column name is returned.'
);
```

You should be able to work out the assertion for the next two cases by yourself with that example :)  But, if you need any help, let me know.  We could also do the last test using a data provider but we can do that another day.

Your next step, and this could be another pull, is to implement this in, for example, JTableContent and check that the alias handling works.  You should be able to remove the JTableContent::publish method from that class now.

Oh, and one last thing, you'll need to merge down from the platform's master branch because this pull won't automatically merge at the moment.  Thanks!

libraries/joomla/database/table.php
```diff
@@ -271,6 +280,46 @@ abstract class JTable extends JObject
...
+		return $return;
+	}	
+	
+	/**
+	 * Method to register a column alias for a "special" column.
+	 * 
+	 * @param string $column      The "special" column (ie ordering)
+	 * @param string $columnAlias The real column name (ie foo_ordering)
+	 * 
+	 * @return void
```
eddieajau: Minor check-style problems in this DocBlock - need 2 spaces between return and void, then line up "string" and "11.3" on that.  Also need 2 spaces after string, and two spaces after $columnAlias.

libraries/joomla/database/table.php
```diff
@@ -271,6 +280,46 @@ abstract class JTable extends JObject
...
+	 * 
+	 * @param   string  $column Name of the "special" column (ie ordering, hits etc etc)
+	 * 
+	 * @return  string  The string that identify the special
+	 * 
+	 * @since   11.3
+	 */
+	protected function getColumnAlias($column)
+	{
+		if(isset($this->columnAlias[$column]))
```
eddieajau: Check-style: Need one space after the if before the bracket.

libraries/joomla/database/table.php
```diff
@@ -271,6 +280,46 @@ abstract class JTable extends JObject
 	}
 
 	/**
+	 * Method to return the real name of a "special" column such as ordering, hits, published
+	 * etc etc. In this way you are free to follow your db naming convention and use the
+	 * built in Joomla functions.
+	 * 
+	 * @param   string  $column Name of the "special" column (ie ordering, hits etc etc)
```
eddieajau: Check-style: need two spaces after $column.

eddieajau:
Fixes looking good!  How are you going with the two unit tests?

tampe125:
hi eddie,
thanks for yuor help, but i'm having a problem: how can i get the mockDatabase?

i see that the method is in JoomlaTestCase, but it's an abastract class, so i need to extend it :/

eddieajau:
You get the mock database by $this->getMockDatabase in your test.  See the example I gave you above.

tampe125:
this is my code

```
public function testSetColumnAlias()
    {
        $table = new JTableInspector('foo', 'bar', $this->getMockDatabase());
        $table->setColumnAlias('published', 'state');
        $this->assertThat(
            $table->columnAlias['published'],
            $this->equalTo('state'),
            'Tests that the alias was correctly added to the columnAlias array.'
        );
    }
```

when i try to run it i get this error:

```
Call to undefined method JTableTest::getMockDatabase()
```

i have to instantiate a new JTableInspector class since JTable is abstract.
any help?

eddieajau:
Looks like I got busy and forgot to follow up on this one.  Could you merge with staging and see if we can get this finished.  Thanks.

tampe125:
hi, sorry but i have changed my pc and i lost everything. atm the moment i'm very busy and i can't re-install & configure everything. can anyone else merge these changes?

tampe125:
is this pull request dead?
should i close it?

realityking:
The pull request doesn't merge anymore. I'll assign this to Andrew but without this being mergeable he's unlikely to do anything.

chdemko:
@tampe125 The pull request cannot be merged

tampe125:
sigh, i'll have to redo everything from scratch :(

ok, i'll close this pull request

PR #305, opened 2011-08-29
Improve Test Coverage

mbabker:
The test coverage should be improved slightly with these additions.  Either methods or test cases have been added to try and cover cases in the code where existing tests missed a couple lines in a function.

### DIFF
--- a/tests/suite/joomla/application/JApplicationTest.php
+++ b/tests/suite/joomla/application/JApplicationTest.php
@@ -210,6 +210,15 @@ class JApplicationTest extends PHPUnit_Framework_TestCase
 	}
 
 	/**
+	 * @todo Implement testStringURLSafe().
+	 */
+	public function testStringURLSafe()
+	{
+		// Remove the following lines when you implement this test.
+		$this->markTestIncomplete('This test has not been implemented yet.');
+	}
+
+	/**
 	 * @todo Implement testGetPathway().
 	 */
 	public function testGetPathway()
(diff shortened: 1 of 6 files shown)


tests/suite/joomla/application/JApplicationTest.php
```diff
@@ -228,6 +237,38 @@ class JApplicationTest extends PHPUnit_Framework_TestCase
...
+	public function testGetHash()
+	{
+		$expResult = MD5("This is a test");
+		/**$mockApplication = $this->getMock('JApplication', array('getHash'), array('site'));
+		$mockApplication->expects($this->once())->method('getHash')->will(
+			$this->returnValue($expResult)
+		);**/
+		require_once(JPATH_PLATFORM . '/joomla/registry/registry.php');
+		$temp = JFactory::$application;
+		JFactory::$application = new JRegistry();
```
eddieajau: Wouldn't it be better to mock JFactory::getConfig here?  For example:

```
// Temporarily override the config cache in JFactory.
$temp = JFactory::$config;
JFactory::$config = new JObject(array('secret' => 'foo'));

$this->assertThat(
    JApplication::getHash('This is a test'),
    $this->equalTo(md5('foo'.'This is a test'),
    'Tests that the secret string is added to the hash.'
);

JFactory::$config = $temp;
```

Just a thought.

eddieajau:
Would you mind adding @since 11.3 to any new tests that don't already have it?

mbabker:
That test is actually a straight up copy of the one for the deprecated JUtility method.  I'm still learning how to build tests, so I'm sure it could be done better.

eddieajau:
Ok, that's cool :)  I've got another pull in the queue to fix the unstable builds so I might wait for that to go through (I made it, so I can't merge it) before I pull this one.  Thanks though - every little bit helps.

mbabker:
And I'll keep throwing commits on here in the mean time for every little bit I'm able to improve on.

PR #308, opened 2011-08-29
CSS optimizations

realityking:


jonnsl:
It would be nice if someone change the default branch to stagging instead of master https://github.com/blog/421-pick-your-default-branch

PR #310, opened 2011-08-29
Fix a regression with controllers for specific formats.

realityking:


eddieajau:
Rouven, would you mind adding a list of the paths that this is supposed to be looking for.  Then I'll have a think about how to craft a unit test for it.  Thanks!

realityking:
Sorry, took me a little bit longer.

Without this patch paths like this are used:

If the task contains a dot, e.g. articles.save (only the part before the dot is used):
BASE_PATH/controllers/articles.format.php

If the task doesn't contain a dot:
BASE_PATH/controllers/controller.format.php

The BASE_PATH is JPATH_COMPONENT if not overwritten.
Format can be XML, JSON and so on, if it is HTML it is omitted.

No with the pull we change the behavior when the task doesn't contain a dot:
If BASE_PATH/controllers/controller.format.php doesn't exist but BASE_PATH/controllers/controller.php does we use the later.

Is that we you need?

jlover:
**Checkstyle error details:**
libraries/joomla/application/component/controller.php:272
libraries/joomla/application/component/controller.php:272
libraries/joomla/application/component/controller.php:282
Expected "if (...)\n...{...}\n...else\n"; found "if (...)\n...{...}\n...else "
libraries/joomla/application/component/controller.php:286
Usage of ELSE IF is not allowed; use ELSEIF instead

PR #312, opened 2011-08-30
Removed hard coded client info

jonnsl:


eddieajau:
Hi Jonathan.  Your branch needs updating.  However, is the CMS getting a patch to manually call addClientInfo?  We just need to synchronise carefully if that's the case.  Though it would raise a small backward compatibility problem so are we better marking that block of code with a JLog deprecated warning and removing the code in 12.1?

PR #313, opened 2011-08-30
Test Coverage Improvements

mbabker:
Another round of slight improvements on the test coverage.  Mainly poking at methods that had one or two lines not covered and a couple methods had no test at all.


eddieajau:
Nice work Michael.

PR #320, opened 2011-08-30
Coding standards: Various little fixes

elkuku:
This adds two new sniffs to prevent the following errors:

```
class Foo
{
    public static function bar()
    {
        /*
         * Usage of "$this" in static methods will cause runtime errors
         * Classes_StaticThisUsageSniff
         */
        echo $this->baz;

        /*
         * Space found before semicolon; expected "'foo';" but found "'foo' ;"
         * WhiteSpace_SemicolonSpacingSniff
         */
        echo 'foo' ;
    }
}
```

### DIFF
--- a/build/phpcs/Joomla/Sniffs/Classes/StaticThisUsageSniff.php
+++ b/build/phpcs/Joomla/Sniffs/Classes/StaticThisUsageSniff.php
@@ -1,0 +1,94 @@
+<?php
+/**
+ * Squiz_Sniffs_Classes_StaticThisUsageSniff.
+ *
+ * PHP version 5
+ *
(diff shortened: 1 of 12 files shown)


eddieajau:
Nikolai, this is pulled against master, not staging. Looks good though.

elkuku:
argh... is there a reason for not setting "staging" as the [default branch](https://github.com/blog/421-pick-your-default-branch) ?

Reopened as [pull/321](https://github.com/joomla/joomla-platform/pull/321)

eddieajau:
No, but I don't have the admin rights to do it :)  I've left a message.

PR #321, opened 2011-08-30
Coding standards: Various little fixes

elkuku:
This adds two new sniffs to prevent the following errors:

```
class Foo
{
    public static function bar()
    {
        /*
         * Usage of "$this" in static methods will cause runtime errors
         * Classes_StaticThisUsageSniff
         */
        echo $this->baz;

        /*
         * Space found before semicolon; expected "'foo';" but found "'foo' ;"
         * WhiteSpace_SemicolonSpacingSniff
         */
        echo 'foo' ;
    }
}
```


eddieajau:
Thanks for cleaning up these edge cases. It's greatly appreciated.

PR #323, opened 2011-08-31
New JDispatcher/JObservable classes to improve speed

rcorral:
I recreated the JDispatcher and JObservable classes to be faster and keep the same functionality. 
I created two unit tests. They can be ran like this: 
phpunit suite/joomla/event/JDispatcherSpeed.php -- runs plugins with current JDispatcher
 phpunit suite/joomla/event/JDispatcher2Speed.php -- uses new JDispatcher class

Each test will create 2000 Joomla plugins, all of them are exactly the same, and will trigger the event that these plugins are attached to, and will add a 1 to a counter. The test will keep track of the counter and how much time it took to execute the event (not the time it took from the initialization of the framework, which is what phpunit does).
Every time that I have tested this, the new class that I created performs over twice as fast. JObserver still has the ability to "discover" events/methods of plugins, but these events need to be registered to the class before they can be discovered on imported plugins. This functionality could be removed, but it certainly aids on the speed

I think it is crucial for Joomla to improve their event api, as larger extensions are developed, shopping carts for example that can have thousands of events, it is important for the developers of these extensions to allow events to be triggered in many places throughout the extension so that 3rd party developers can extend these extensions without having to hack their core code. Therefore we need to think of a better solution than what we currently have, we don't want to compromise speed for extensibility or vice versa.


tests/suite/joomla/event/speedincludes/observable2.php
```diff
@@ -0,0 +1,130 @@
...
+		'onSubmitContact',
+		'onValidateContact',
+		'onContentPrepareData',
+		'onContentPrepareForm'
+		);
+
+	/**
+	 * Constructor
+	 *
+	 * Note: Make Sure it's not directly instansiated
```
pasamio, line 85: This comment doesn't make sense for a public constructor...

PR #324, opened 2011-08-31
New JDispatcher/JObservable classes to improve speed

rcorral:
I recreated the JDispatcher and JObservable classes to be faster and keep the same functionality. 
I created two unit tests. They can be ran like this: 
phpunit suite/joomla/event/JDispatcherSpeed.php -- runs plugins with current JDispatcher
 phpunit suite/joomla/event/JDispatcher2Speed.php -- uses new JDispatcher class

Each test will create 2000 Joomla plugins, all of them are exactly the same, and will trigger the event that these plugins are attached to, and will add a 1 to a counter. The test will keep track of the counter and how much time it took to execute the event (not the time it took from the initialization of the framework, which is what phpunit does).
Every time that I have tested this, the new class that I created performs over twice as fast. JObserver still has the ability to "discover" events/methods of plugins, but these events need to be registered to the class before they can be discovered on imported plugins. This functionality could be removed, but it certainly aids on the speed.

I think it is crucial for Joomla to improve their event api, as larger extensions are developed, shopping carts for example that can have thousands of events, it is important for the developers of these extensions to allow events to be triggered in many places throughout the extension so that 3rd party developers can extend these extensions without having to hack their core code. Therefore we need to think of a better solution than what we currently have, we don't want to compromise speed for extensibility or vice versa.


ianmacl:
# Test Results

Total Tests: 1445
Assertions: 9956
Failures: 0
Errors: 0
Test Time: 195.488348

rcorral:
To note, if approved, before merging to the master branch, the old dispatcher.php and observer.php would need to be replaced with the ones that I created.

rvsjoen:
In that case you should probably do that in the pull request so the unit tests can be executed. If you move them to their correct places in your branch and push, the pull request should update automatically.

rcorral:
The unit tests use the current JDispatcher and JObservable to show the speed difference.
I was thinking it would be easier for someone to give the go-ahead on this pull request, I can then cancel it, replace the two files and then do another request. I am probably overcomplicating it, but it is important for the purpose of the tests, this is what I figured would happen.

eddieajau:
Just a couple of house-keeping things.  

The tests and code should all follow the code-style rules we have set up.  You can find a PHP Code Sniffer standard in the /build folder or a text explanation under /docs.  

It would be better to overwrite the actual classes so that we can see what the differences are.

The unit tests aren't really unit tests.  They are of interests but we'd probably take them out before the final merge.  However, I'd like to see a full set of unit tests done, if possible, on the original classes first.  Then we can apply your changes and ensure nothing breaks backward compatibility or introduces a regression.

On a technical note, I don't think we should be specifying default events in the class itself.  They are very CMS centric and we are trying to reduce the tight coupling of the Platform with the CMS.  I would let the CMS of "other app" add those by hand as it's bootstrapping the framework (so that's a change you would make in, for example. /includes/framework.php in the CMS).

The most important thing in the short term though is take the '2' off the names so we can see the changes :)

Thanks in advance.

PR #325, opened 2011-08-31

Hackwar:

A specific usage example:

After the last pull request, Andrew Eddie told me that I maybe should add support for chaining to the class. This is done in this revision. I also thought about breaking this up into smaller classes, like on a row or even cell base, but that seems to just overcomplicate stuff. So I intentionally stuck with this single monolithic class. I'm not entirely happy with the setting-object-to-a-current-row solution, but it was necessary for the chaining and it requires less code than the previous approach.

Here is a small example how it can be used in the user manager. This is a copy from the header of the table from the backend users view of com_users in Joomla 1.7.0:

``` php
<?php
jimport('joomla.html.grid');

$table = new JGrid(array('class' => 'adminlist'));
$table->setColumns(array('checkbox', 'name', 'username', 'block', 'active', 'groups', 'email', 'lastvisit', 'registerdate', 'id'));

$table->addRow(array(), 1)
->addRowCell('checkbox', '<input type="checkbox" name="checkall-toggle" value="" title="'.JText::_('JGLOBAL_CHECK_ALL').'" onclick="Joomla.checkAll(this)" />', array('width'=> '1%'))
->addRowCell('name', JHtml::_('grid.sort', 'COM_USERS_HEADING_NAME', 'a.name', $listDirn, $listOrder), array('class' => 'left'))
->addRowCell('username', JHtml::_('grid.sort', 'JGLOBAL_USERNAME', 'a.username', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '10%'))
->addRowCell('block', JHtml::_('grid.sort', 'COM_USERS_HEADING_ENABLED', 'a.block', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '5%'))
->addRowCell('active', JHtml::_('grid.sort', 'COM_USERS_HEADING_ACTIVATED', 'a.activation', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '5%'))
->addRowCell('groups', JText::_('COM_USERS_HEADING_GROUPS'), array('class' => 'nowrap', 'width' => '10%'))
->addRowCell('email', JHtml::_('grid.sort', 'JGLOBAL_EMAIL', 'a.email', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '15%'))
->addRowCell('lastvisit', JHtml::_('grid.sort', 'COM_USERS_HEADING_LAST_VISIT_DATE', 'a.lastvisitDate', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '10%'))
->addRowCell('registerdate', JHtml::_('grid.sort', 'COM_USERS_HEADING_REGISTRATION_DATE', 'a.registerDate', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '10%'))
->addRowCell('id', JHtml::_('grid.sort', 'JGRID_HEADING_ID', 'a.id', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '3%'));

echo $table->render();
```

### DIFF
--- a/libraries/joomla/html/grid.php
+++ b/libraries/joomla/html/grid.php
@@ -1,0 +1,350 @@
+<?php
+/**
+ * @package     Joomla.Platform
+ * @subpackage  HTML
+ *
+ * @copyright   Copyright (C) 2005 - 2011 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE
+ */
+
+defined('JPATH_PLATFORM') or die;
+
+/**
+ * JGrid class to dynamically generate HTML tables
+ *
+ * @package     Joomla.Platform
+ * @subpackage  HTML
+ * @since       11.1
+ */
+class JGrid
+{
+	/**
+	 * Array of columns
+	 * @var array
+	 */
+	protected $columns = array();
+	
+	/**
+	 * Current active row
+	 * @var int
+	 */
+	protected $activeRow = 0;
+	
+	/**
+	 * Rows of the table (including header and footer rows)
+	 * @var array
+	 */
+	protected $rows = array();
+	
+	/**
+	 * Header and Footer row-IDs
+	 * @var array
+	 */
+	protected $specialRows = array('header' => array(), 'footer' => array());
+	
+	/**
+	 * Associative array of attributes for the table-tag
+	 * @var array
+	 */
+	protected $options;
+	
+	/**
+	 * Constructor for a JGrid object
+	 * 
+	 * @param array Associative array of attributes for the table-tag
+	 */
+	function __construct($options = array())
+	{
+		$this->setTableOptions($options, true);
+	}
+	
+	/**
+	 * Method to set the attributes for a table-tag
+	 * 
+	 * @param array Associative array of attributes for the table-tag
+	 * @param bool Replace possibly existing attributes
+	 * @return JGrid This object for chaining
+	 */
+	function setTableOptions($options = array(), $replace = false)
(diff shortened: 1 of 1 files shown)


eddieajau:
Thanks Hannes. A few house-keeping things.  You need to pull against staging, not master (that triggers a build on our CI server) - I have to close the issue on that basis, sorry.  You've got a few check-style problems that I can see.  It would be good if you can set up PHP Code Sniffer via PEAR, add our Joomla sniff and get it all formatted and docblocked correctly (if not, the coding standards are under the /docs/ folder and are 'reasonably' up to date).  At this stage the @since tags (all properties and methods need them) will be 11.3.  Last thing is the class needs a full set of unit tests.

Only technical suggestion is to use the magic __toString method instead of the render method.

I would also make a post to the Joomla Platform list (not the Framework list as we are in the process of retiring it) to let people know your work is here to peruse - that's just a suggestion though.

If you can fix up the check-styles and have a full set of passing unit tests, I don't see any other obstacle in adding the code.

Hackwar:
Ok, will do. Regarding the platform mailinglist: Does that mean that I'm not banned from the Joomla mailinglists anymore? Will I be allowed to post to other mailinglists, too?

eddieajau:
Well, technically, no, you are not banned any more because the period is well and truly passed.  However, whatever conditions have been attached for joining back into other lists or groups still stand (beyond my control), but as for the platform list, as long as we have an undertaking that you are prepared to keep things civil and professional (as you would expect to be treated in return), I don't see there will be any problem.

PR #326, opened 2011-08-31

Hackwar:

A specific usage example:

After the last pull request, Andrew Eddie told me that I maybe should add support for chaining to the class. This is done in this revision. I also thought about breaking this up into smaller classes, like on a row or even cell base, but that seems to just overcomplicate stuff. So I intentionally stuck with this single monolithic class. I'm not entirely happy with the setting-object-to-a-current-row solution, but it was necessary for the chaining and it requires less code than the previous approach.

Here is a small example how it can be used in the user manager. This is a copy from the header of the table from the backend users view of com_users in Joomla 1.7.0:

``` php
<?php
jimport('joomla.html.grid');

$table = new JGrid(array('class' => 'adminlist'));
$table->setColumns(array('checkbox', 'name', 'username', 'block', 'active', 'groups', 'email', 'lastvisit', 'registerdate', 'id'));

$table->addRow(array(), 1)
->addRowCell('checkbox', '<input type="checkbox" name="checkall-toggle" value="" title="'.JText::_('JGLOBAL_CHECK_ALL').'" onclick="Joomla.checkAll(this)" />', array('width'=> '1%'))
->addRowCell('name', JHtml::_('grid.sort', 'COM_USERS_HEADING_NAME', 'a.name', $listDirn, $listOrder), array('class' => 'left'))
->addRowCell('username', JHtml::_('grid.sort', 'JGLOBAL_USERNAME', 'a.username', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '10%'))
->addRowCell('block', JHtml::_('grid.sort', 'COM_USERS_HEADING_ENABLED', 'a.block', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '5%'))
->addRowCell('active', JHtml::_('grid.sort', 'COM_USERS_HEADING_ACTIVATED', 'a.activation', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '5%'))
->addRowCell('groups', JText::_('COM_USERS_HEADING_GROUPS'), array('class' => 'nowrap', 'width' => '10%'))
->addRowCell('email', JHtml::_('grid.sort', 'JGLOBAL_EMAIL', 'a.email', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '15%'))
->addRowCell('lastvisit', JHtml::_('grid.sort', 'COM_USERS_HEADING_LAST_VISIT_DATE', 'a.lastvisitDate', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '10%'))
->addRowCell('registerdate', JHtml::_('grid.sort', 'COM_USERS_HEADING_REGISTRATION_DATE', 'a.registerDate', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '10%'))
->addRowCell('id', JHtml::_('grid.sort', 'JGRID_HEADING_ID', 'a.id', $listDirn, $listOrder), array('class' => 'nowrap', 'width' => '3%'));

echo $table->render();
```


ianmacl:
# Test Results

Total Tests: 1296
Assertions: 2248
Failures: 0
Errors: 0
Test Time: 58.024213

eddieajau:
As mentioned before, we need automated check-style tests to pass and a full set of unit tests before we can consider merging.  Let us know when that's done and we can check the work.  Thanks in advance.

Hackwar:
Concerning the style tests: Its telling me that the @version tag is missing, however that tag is nowhere to be found in the whole platform codebase. I don't know what needs to be inserted there. Otherwise I get a warning that the PHP version needs to be specified, but I also don't know how. I used the codesniffer file mentioned in the article on docs.joomla.org If that one is not up-to-date, please tell me where I can get a more recent version.

Concerning the unit tests: They are complete, please otherwise tell me what is missing.

Hackwar:
I also renamed the render function to toString() and added a magic function, too.

eddieajau:
For the code style, are you using the Joomla sniff?

phpcs --standard=Joomla <filename>

Note, you'll need to copy or softlink the standard into the PHP Code Sniffer standards folder, or, add command line arguments to point phpcs to the right sniff.

Test should also have their docblocks done including the @since tag, and while we don't run code-style checks over them, they should still comply.  There are some protected methods not being tested so you'll need an inspector to do those.  I've written some notes about there in this doc:

https://github.com/eddieajau/joomla-platform/blob/documentation/docs/manual/chapters/testing.xml

Hope this helps.

Hackwar:
Yes, I am using the Joomla sniff.

I got issues creating a unit test skeleton with phpunit. It somehow didn't create any output when giving it the class. Don't know why. So I wrote all the stuff by hand, which is the reason for the missing doc blocks in the tests. 

Will fix the rest as much as I can.

eddieajau:
Just another note, in tests you want to isolate the test as much as possible.  For example, when testing the constructor you have this:

```
$this->assertThat(
  $table->getTableOptions(),
  $this->equalTo($options)
);
```

It's better, in my opinion, to test the value of the protected $options variable.  If you use an inspector you'll get access to that and the test can become:

```
 $this->assertThat(
  $table->options,
  $this->equalTo($options),
  'Checks that the options passed to the constructor are correctly set.'
);
```

Adding the note (the third argument) helps self-document and track down problems, particularly if there are multiple assertions in a test.

eddieajau:
Hrm, sounds like it's defaulting to the PEAR sniff for some reason.  At any rate, have a look at any other class (take JCli for example), and you should be able to see the formula.

ianmacl:
# Test Results

Total Tests: 1300
Assertions: 2252
Failures: 0
Errors: 0
Test Time: 56.035857

Hackwar:
After several days of looking for my error, I found out that this document is outdated: http://docs.joomla.org/Joomla_CodeSniffer
It points to a coding standard that is outdated instead of the one in the platform repo... 

Anyway, I fixed up the coding issues and changed the Unittests. Please take a look now.

ianmacl:
# Test Results

Total Tests: 1300
Assertions: 2248
Failures: 0
Errors: 0
Test Time: 55.408828

ianmacl:
# Test Results

Total Tests: 1300
Assertions: 2248
Failures: 0
Errors: 0
Test Time: 56.381765

ianmacl:

# Test Results

Total Tests: 
Assertions: 
Failures: 
Errors: 
Test Time:

Hackwar:
I added notes to the documentation to make it easier for future people trying to set up their systems...

eddieajau:
Thanks for adding the note on the Wiki page.  There appears to be a few bugs in the Sniff so could you fix up the following by hand:

1) } else { should have the brackets and statement on new lines.  The sniff is configured that way but not picking it up.
2) Can you put 2 spaces after each @return
3) Can you put an @since 11.3 line into each of the property and method docblocks (3 spaces between since and 11.3)

As far as tests go, you are missing a bit of coverage on addRow ($special not tested), getRow (if $id !== false), getRows ($special not tested) and deleteRows (unsetting the special header and footer rows not tested).  Fix all that up and we should be right to merge.

Thanks in advance.

Hackwar:
Done.

ianmacl:
# Test Results

Total Tests: 1300
Assertions: 2263
Failures: 0
Errors: 0
Test Time: 56.642137

eddieajau:
Close enough.  Thanks for that.

PR #329, opened 2011-09-01
Make modal able to use percentage for size

rvsjoen:
This will modify _getJSObject() so that if you use for example the following snippet

JHtml::_('behavior.modal', 'a.modal', array( 'handler' => 'iframe', 'size' => array('x' => '960', 'y' => '60%')));

It will auto-scale in the y-direction.


eddieajau:
I don't suppose you'd like to have a go at writing a unit test for _getJSObject would you?

rvsjoen:
I suppose I could give that a shot, but does it make sense to unit test protected functions ?

eddieajau:
Protected methods, while hidden from the public API, can still be used by derived classes by other developers so they need to perform as advertised.  Plus, it ensures that our own usage of the protected API performs as expected.  So, it would be great if you could have a go at it.  If you have any questions about it, I can give you a hand.

rvsjoen:
Ok so I added a commit that shows how this can be done, the only problem is that apparently _getJSObject() does not spit out valid JSON. Not sure how to handle that, I guess I could patch it so that it does, although in that case I have to say I don't understand why something like json_encode() is not used in the first place.

pasamio:
json_encode was added in PHP 5.2 and was a PECL library prior to that. _getJSObject() dates from Joomla! 1.5 which supported PHP4 prior to json_encode being reasonably available. Given we're on PHP 5.2+ now it makes sense to refactor the function to use json_encode if you're willing.

rvsjoen:
After a quick look at this, json_encode() does not allow function calls, which would kill the functionality in for example passing fullScreen = true, to a modal. Since it is based on the value 'window.getSize().x' snippet being passed in the JSObject. I think this is very important behavior in the modal, which is also why I wanted to add auto-scaling in one direction in the first place. 

So this leads me to believe this cannot be done using 'valid' JSON but still valid javascript syntax, and that the unit test will have to be adapted to this scenario instead of reverting to json_encode().

I will look a bit harder.

realityking:
I tried going down the json_encode rule as well (which is significantly faster) but after doing some reading up I too had to realize that JSON only covers a subset of JavaScript objects. Notably JSON can only contain data, not code, which kinda makes sense.

jlover:
**Checkstyle error details:**
libraries/joomla/html/html/behavior.php:892
libraries/joomla/html/html/behavior.php:892
libraries/joomla/html/html/behavior.php:893
Expected "if (...)\n...{...}\n...else\n"; found "if(...)\n...{...}\n...else\n"
libraries/joomla/html/html/behavior.php:895
libraries/joomla/html/html/behavior.php:895
libraries/joomla/html/html/behavior.php:895
libraries/joomla/html/html/behavior.php:895
libraries/joomla/html/html/behavior.php:900
libraries/joomla/html/html/behavior.php:900

eddieajau:
Branch is not merging cleanly.

chdemko:
@rvsjoen The pull request cannot be merged

rvsjoen:
The merge conflict should be fixed, the unit test went down the drain.

realityking:
I tried to use the unit test that used to be in there but it wouldn't pass for me.

rvsjoen:
It didn't pass because the unit test was using json_encode to create the control set. Only problem is that the "json" that is output by this function is not valid. Not sure about the best way to handle it and I don't feel like writing a unit test that actually does custom parsing of some set of invalid json

realityking:
Right, because we're not actually trying output JSON but a JS object. What we could do is just get the serialization of a couple objects and test against those. This way we'll know when something breaks but we won't discover if anything is broken now - still better than nothing thou.

ianmacl:
This is a fairly independent unit that likely doesn't have implications on other areas of the platform.  Perhaps we can get a couple of browser tests with this so we can be reasonably satisfied it works and we can pull it at that point.

ianmacl:
Closing this pull request because it isn't a bug and nobody seems overly interested in it.

PR #333, opened 2011-09-02
Fix inverted attributes in JFormFieldSQL (key_field and value_field)

rvsjoen:
These two attributes are swapped in JFormFieldSQL making the usage of said class a very confusing experience because the 'key_field' is really the value="" of the <option> and the 'value_field' is really the translated string.

This fix will swap them so that key_field is used in inside the option tag as the string and value_field is used in option value="VALUE"


ianmacl:
# Test Results

Total Tests: 1445
Assertions: 9956
Failures: 0
Errors: 0
Test Time: 179.270364

eddieajau:
I see what you are trying to do, and I think I agree, but wouldn't the change introduce a backward compatibility problem?

rvsjoen:
It will, the fix is relatively simple though for people who actually bother to read the changelog. 

I only know that I pulled a hair or 300 trying to figure out what the heck I was doing wrong when the value and key were swapped.

I do not see any other way of doing it, other than _possibly_ rename the 'value' to something else, and keep a backwards mapping into 'value'. Like, 'name' or 'string' or 'text'

eddieajau:
Ok, we don't want to break backward compatibility because I know a) CMS extension devs use the field type and b) they don't read changelogs :)  So I would deprecate the attributes and add new ones that make sense.  Maybe "listvalue" and "listtext"?  It wouldn't hurt to add some comments in the Docblock about what attributes are supported as well.

PR #336, opened 2011-09-04
JView 'get' method now can pass arguments to its model.

tampe125:
Now JView `get` method can pass arguments to model functions.

Usage:

```
//no arguments (no changes from "standard" usage)
$this->get('Data');

//single argument (default model)
$this->get('Data', null, 'myParam');

//more arguments (non-default model)
$this->get('Data', 'myModel', array('myParam1', 'myParam2'));
```

### DIFF
--- a/libraries/joomla/application/component/view.php
+++ b/libraries/joomla/application/component/view.php
@@ -357,12 +357,14 @@ class JView extends JObject
 	 *
 	 * @param   string  $property  The name of the method to call on the model or the property to get
 	 * @param   string  $default   The name of the model to reference or the default value [optional]
+	 * @param   mixed   $args      Arguments passed to model method. It can be a single one, or an
+	 *                             array of arguments [optional]
 	 *
 	 * @return  mixed  The return value of the method
 	 *
 	 * @since   11.1
 	 */
-	public function get($property, $default = null)
+	public function get($property, $default = null, $args = null)
 	{
 
 		// If $model is null we use the default model
@@ -385,7 +387,17 @@ class JView extends JObject
 			if (method_exists($this->_models[$model], $method))
 			{
 				// The method exists, let's call it and return what we get
-				$result = $this->_models[$model]->$method();
+				//get arguments
+				if(!is_array($args))
+				{
+					$args = array($args);
+				}
+				
+				//setup call_user_func_array array
+				$classModel = array($this->_models[$model], $method);
+				
+				$result = call_user_func_array($classModel, $args);
+				
 				return $result;
 			}
 
--- a/libraries/joomla/application/component/view.php
+++ b/libraries/joomla/application/component/view.php
@@ -396,9 +396,7 @@ class JView extends JObject
 				//setup call_user_func_array array
 				$classModel = array($this->_models[$model], $method);
 				
-				$result = call_user_func_array($classModel, $args);
-				
-				return $result;
+				return call_user_func_array($classModel, $args);
 			}
 
 		}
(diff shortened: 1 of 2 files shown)


chdemko:
Thanks for the pull request. Could you add some unit tests on this new feature?

tampe125:
i'd wish but, sadly, i have some problems with test units :/

libraries/joomla/application/component/view.php
```diff
@@ -385,8 +387,16 @@ class JView extends JObject
...
+				//get arguments
+				if(!is_array($args))
+				{
+					$args = array($args);
+				}
+				
+				//setup call_user_func_array array
+				$classModel = array($this->_models[$model], $method);
+				
+				return call_user_func_array($classModel, $args);
```
Hackwar, line 399: Instead of having the check in line 390, you could simply replace that with this changed line for line 399:

``` php
return call_user_func_array($classModel, (array) $args);
```

Hackwar:
I've created a unittest for this and you can find it in this pull request/repo: https://github.com/joomla/joomla-platform/pull/421

Feel free to copy it and add it to this pull.

LouisLandry:
I'm not a big fan of this branch.  IMO, the concept described here violates the separations of concerns.  A view is not supposed to in any way manipulate the state in a model.  Have a look at http://www.phpwact.org/pattern/model_view_controller for some pretty good reading on the concepts.  Specifically a brief discussion about this relationship at http://www.phpwact.org/pattern/model_view_controller#model_view_relationship.

It isn't the job of a view to set "options" as it retrieves information from the model.  The model should simply be a readonly object to the view.  I believe it is the job of the controller to set whatever options are relevant to the model so that things are executed and rendered properly.  If there are two separate data structures that can be returned by a model, then that model should implement getters for those different data structures.

I understand the desire to make things more flexible at every turn, but sometimes things are structured in a certain way for a reason.  It wasn't an oversight that arguments can't be passed into the model getters with the Joomla MVC.  It is a way of enforcing that separation between domain logic and presentation logic.

tampe125:
ok, maybe you're right, but let me explain my situation.
let's say that i have to display published and non-published methods (it's just an example).
the query is exactly the same, i only need to add

```
$query->where('published = 1')
```

now i have to create 2 functions, `getPublishedOnly` , `getEverything` and they differ only for the previous line!
why can't i pass a variable to a function and then save code line?

i repeat, i don't want to pass an ID or something like that, but only a switch

LouisLandry:
I understand your situation very well, and I have come across it in many situations.  There are two approaches that I've taken.  

If it is an either/or case where you only will ever want one or the other, then you can use the controller (which is designed to set state in the model based on the input/request) to set something in the model to make it behave as you want -- either getting everything or just the published methods.

If it is an and case where you want to be able to get both a list of everything as well as a list of published only methods, then I add those methods explicitly as you listed above: `getPublishedOnly`, `getEverything`.  This actually creates a more expressive and explicit public API for your model, and will be easier for people to know exactly what they are getting when they make a call into your API.  Even someone who doesn't understand code well.

I do understand your frustration -- I've been there.  That being said I also believe firmly in the separation of presentation and domain logic, and lots of experience has reinforced that belief.

jlover:
**Checkstyle error details:**
libraries/joomla/application/component/view.php:389
Expected "if (...)\n...{...}\n...else\n"; found "if(...)\n...{...}\n...else\n"
libraries/joomla/application/component/view.php:393
libraries/joomla/application/component/view.php:396

eddieajau:
I'm going to close this and ask that we review the MVC in Q1 of next year.  Thanks.

Issue #339, opened 2011-09-06
loadnextrow and loadnextobject do not iterate


brbrbr:
I would expect loadnextrow and loadnextobject  in database.php to walk tru the result set.

However the cursor is reset  for each call to the function ( like loadrow returns the first row indeed)

eddieajau:
I'd say the methods need to be changed to use $this->cursor.  Are you able to make the required code changes and make a pull request, or do you need help with that?

brbrbr:
Whether it uses the static or $this it will need to check for a valid cursor as well.

I'm not familiar with qit, I have a fork with a patch ( if ( $cursor === null ) {} added. 
Just hit the 'pull request' button or RTFM ?

elinw:
Click pull request, then change the destination from master to staging (just click on the block where it says master and start typing, you'll get a drop down).

brbrbr:
ok,

i hope i didn't mess up the repository to much

eddieajau:
Going to close this issue since we have a pull request in the pipeline.

PR #346, opened 2011-09-07
Joomla CMS [#26744] loadnextrow and loadnextobject do not iterate (#339)

brbrbr:


eddieajau:
Thanks for the pull request.  I think that change might cause problems if you try to use a second (different) query and iterative with loadNext*.  I think you will probably have to check that $this->cursor is not empty and I also think a prerequisite is that you have to fire $db->query first, then loop over $db->loadNextRow (even for the first one).  Does that make sense?  That introduces a slight backward compatible issue but I don't see any other reliable way around the problem.

brbrbr:
using the $this->cursos instead of the static allows to reset the query using setquery/query without the need to iterate thru the result set.

Having a $this->query in loadnextrow/loadnextobject keeps the functions backward compatible, althought I don't think that is a big issue since the current version do not function as intended ( doing the job of loadrow/loadobject)

eddieajau:
Ok, thanks.  I'll take a look when I can.

libraries/joomla/database/database.php
```diff
@@ -933,23 +933,25 @@ abstract class JDatabase implements JDatabaseInterface
 	 */
 	public function loadNextObject($class = 'stdClass')
 	{
-		static $cursor;
-
+	
+		if ( ! $this->cursor ) {
 		// Execute the query and get the result set cursor.
-		if (!($cursor = $this->query()))
```
chdemko, line 939: You could group the test using (!$this->cursor && !($this->cursor = $this->query())

libraries/joomla/database/database.php
```diff
@@ -964,23 +966,27 @@ abstract class JDatabase implements JDatabaseInterface
...
-		static $cursor;
+		
 
+		
+		if ( ! $this->cursor ) {
 		// Execute the query and get the result set cursor.
-		if (!($cursor = $this->query()))
-		{
-			return $this->errorNum ? null : false;
+			if (!($this->cursor = $this->query()))
```
chdemko: idem as above

chdemko:
It remains an issue in your code:
if this code is executed:
$db->setQuery('select \* from jos_content');
var_dump($db->loadNextRow()):
$db->setQuery('select \* from jos_categories');
var_dump($db->loadNextRow()):

We will have the 2 first articles while we are looking for the first article and the first category. I think, it lacks a line in setQuery such that
$this->cursor = false;

brbrbr:
for the first row there is $db->loadRow()

$db->query(...);

resets the cursor.

loadNextRow/Object are odd anyhow, since they are the only load\* functions not resetting the cursor on every call.

chdemko:
$db->query() reset the cursor, sure, but in my example code, it is only called during the first $db->loadNextRow(), the second one does not call it since $this->cursor is not empty

brbrbr:
Hi

I referred to your example code:

$db->setQuery('select \* from jos_content');
var_dump($db->loadNextRow()):
$db->setQuery('select \* from jos_categories');
var_dump($db->loadNextRow()):

should be? :

$db->setQuery('select \* from jos_content');
var_dump($db->loadRow()):
$db->setQuery('select \* from jos_categories');
var_dump($db->loadRow()):

with reset

$db->setQuery('select \* from jos_content');
var_dump($db->loadNextRow());
var_dump($db->loadNextRow());
$db->setQuery('select \* from jos_categories');
$db->query();
var_dump($db->loadNextRow());

Question is what is the supposed usage of loadNextRow?

I found the problem because I needed a function to walk tru a large result set, avoiding overhead.

for smaller set I would use the limit functin to get n rows.

bram

jlover:
**Checkstyle error details:**
libraries/joomla/database/database.php:936
libraries/joomla/database/database.php:941
libraries/joomla/database/database.php:941
Functions and classes must not contain multiple empty lines in a row; found 3 empty lines
libraries/joomla/database/database.php:967
libraries/joomla/database/database.php:967
Functions and classes must not contain multiple empty lines in a row; found 3 empty lines
libraries/joomla/database/database.php:969
libraries/joomla/database/database.php:974
libraries/joomla/database/database.php:974
Functions and classes must not contain multiple empty lines in a row; found 3 empty lines

chdemko:
No more input for this?
- Adding a `$this->cursor = false;` in `setQuery`
- Adding a `protected $cursor;` in `JDatabase`

Also it remains checkstyle errors

gpongelli:
I've already patched this issue on my postgresql branch, check it here
https://github.com/gpongelli/joomla-platform/blob/postgresql/libraries/joomla/database/database.php#L975
and here
https://github.com/gpongelli/joomla-platform/blob/postgresql/libraries/joomla/database/database.php#L1010

Related tests with successful result are here
https://github.com/gpongelli/joomla-platform/blob/postgresql/tests/suite/joomla/database/database/JDatabasePostgreSQLTest.php#L586
and here
https://github.com/gpongelli/joomla-platform/blob/postgresql/tests/suite/joomla/database/database/JDatabasePostgreSQLTest.php#L647

gpongelli:
Improved my code with other tests, adding case where there are "query" or "load" calls before loadNextObject and loadNextRow.
They are six tests, check them [from here](https://github.com/gpongelli/joomla-platform/blob/postgresql/tests/suite/joomla/database/database/JDatabasePostgreSQLTest.php#L705).

chdemko:
See #813

ianmacl:
I'm closing this because it has been superseded by [#903].

PR #351, opened 2011-09-09
Fixed ControlSignatureSniff

elkuku:
This _should_ fix the issue with elses..

**good**

```
if($a)
{
}
else
{
}
```

**bad**

```
if($a)
{
} else {
}
```
## Questions
1. I did a search on the use of `elseif` vs `else if`. Turns out that it is almost 50:50 ;) - Should we care about this ?
2. Would it make sense to add some ''Tests'' with _bad code_ to verify that the sniffs actually work ?

### DIFF
--- a/build/phpcs/Joomla/Sniffs/ControlStructures/ControlSignatureSniff.php
+++ b/build/phpcs/Joomla/Sniffs/ControlStructures/ControlSignatureSniff.php
@@ -15,7 +15,7 @@
  */
 
 if (class_exists('PHP_CodeSniffer_Standards_AbstractPatternSniff', true) === false) {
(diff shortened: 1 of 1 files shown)


ianmacl:
# Test Results

Total Tests: 1445
Assertions: 9956
Failures: 0
Errors: 0
Test Time: 184.116585

eddieajau:
Nice work!

eddieajau:
Oh, I prefer "else if" but "elseif" is valid PHP so I'm personally not fussed which way it goes.  A "bad" sniff file is an interesting idea, we'd just have to have it out of the main /libraries/ folder.

PR #352, opened 2011-09-10
New event called onBeforeModuleLoad

jnilla:
New event called onBeforeModuleLoad that allow users to have fine control over the modules properties before they get loaded and rendered

Thanks to this developer will have more control of the modules before they get rendered, for example
*Change titles, positions name, etc, dynamically 
*Develop plugins to set conditional statements to prevent render a module under certain situations
*Other enhancements related with modules

### DIFF
--- a/libraries/joomla/application/module/helper.php
+++ b/libraries/joomla/application/module/helper.php
@@ -387,6 +387,10 @@ abstract class JModuleHelper
 
 			unset($dupes);
 
+			//Module Plugin
+			$dispatcher	= JDispatcher::getInstance();
+			$dispatcher->trigger('onBeforeModuleLoad', array(&$clean));
+
 			// Return to simple indexing that matches the query order.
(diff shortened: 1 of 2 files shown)


eddieajau:
I have no direct objection to this, but could you discuss it on the CMS developer list first and make sure everyone is happy with this approach.  For instance, I think the name of the event should probably be "onModuleBeforeLoad" but see what people think. There are at least two other extension developers that I know would like to have input into this.  I'll close the pull until you've done that.  Thanks.

jnilla:
hi eddieajau ...  I'm relatively new in this "joomla town" i agree with you with the new name assignation, pleas give me the CMS developer list link to go there and do what you say. thanks

eddieajau:
No problems.  Here you go:
https://groups.google.com/forum/?fromgroups#!forum/joomla-dev-cms

PR #354, opened 2011-09-11
JRouter flexibilization

Hackwar:
This pull request is a small foundation for a larger change to the routing system of Joomla. The goals are to remove some cruft that has accumulated over the years, to make it more flexible and the routing system object oriented and to make writing a component router as simple as possible. Most of the work needs to be done in the CMS, while the code in this pull request is primarily aimed at removing the couplings between the CMS and framework in this area.

The different changes in this pull request and their effects:
1. Every change to the URL is a rule
   Up till now, the changes that were made to a URL were sometimes hardcoded in the router, sometimes they were added rules. Some of those rules were inconvenient (changing the first "-" into ":") and could not be overriden. In this router, every change to a URL is a rule and a rule is a function callback, that is added to the respective list of rules.
2. Changes to a URL are done through a JURI object
3. JRoute::_() accepts both a URL string and an associative array of URL parameters, which are then treated as GET parameters. Up till now, the code normally looks like this: The URL is concatenated from an array of values into a string and then again parsed into an associative array to process it, just to be concatenated into a string later again.
4. Setting a URL to HTTPS now can also be done inside the router by setting the HTTPS parameter of the JURI object to true instead of having to set the parameter in JRoute::_().
5. The specific SEF mode is gone, instead the URL treatment is defined by the rules added to the router.

As is said above, most of the work needs to be done in the CMS by changing JRouterSite, adding the rules, etc. How all that can be done and how the component router later can be simplified to 5 lines of code per component, can be seen in the router plugin here: http://www.joomlager.de/molajorouter.zip
An example how such a router can be extended and how URLs can be improved, can be found in these two plugins:
http://www.joomlager.de/molajowp.zip
http://www.joomlager.de/molajorest.zip
(Please don't get confused by the "Molajo" in the name. That was the working title of the project. This code is not part of the Molajo project.)
The plugins should work in Joomla 1.6


ianmacl:
# Test Results

Total Tests: 1469
Assertions: 10011
Failures: 0
Errors: 0
Test Time: 194.085917

eddieajau:
Thanks Hannes.  This will probably take a while to dig into, and I'm aware other peers and stakeholders have ideas about how the router should work, and I'm not sure who's available for that yet, but here are a few initial comments/questions:
1. Are the changes fully backward compatible with the current class?  If so, can you make sure there are unit tests in place to prove it.  If not, it will have to wait until 12.1 for consideration.
2. Do you have full code coverage on the tests?
3. Can you fix up your code style.  Most of the check points are on this page: https://github.com/joomla/joomla-platform/wiki/Contributing-to-the-joomla-platform  Note that while check style is not performed on the tests, we do expect them to have full docblocks and be well formatted as well.

Hackwar:
Regarding #1: The class is 99% backwards compatible. The only thing missing is the setMode() and getVar() method. The former is not necessary, the later is part of the getVars() function. For setMode() a dummy function could be used instead which does nothing and the getVar() method could be added back in as a proxy to the getVars() method. While the BC is big, this project was not written with backwards compatibility for URL extensions in mind. Component routers are 100% BC, but due to the different workflow, some URL extensions might have problems. Not saying that there have to be problems, just that people tend to use code in strange ways.

Regarding #2: The code should be fully covered by unittests.

Regarding #3: Will do.

Hackwar:
Regarding #1: I added the deprecated methods mentioned above.

Regarding #3: Should be done now.

ianmacl:
# Test Results

Total Tests: 1469
Assertions: 10011
Failures: 0
Errors: 0
Test Time: 186.664040

Hackwar:
A small update to this: I've implemented the routing changes in this branch for the CMS: https://github.com/Hackwar/joomla-cms/tree/jrouter

Is there any news for this pull request?

jgc:
Any update on this pull request?

eddieajau:
I'd really like to see some buy-in and testing from the Joomla CMS or other major downstream stakeholders (like extension developers - the "might have problems" statement concerns me a little) before this gets considered for merging.  On that basis but mainly due to the lack of categorisation features in this issue tracker, I'm going to close this pull until this is done.

Hackwar:
I would have prefered to keep this one open in order to generate more attention, but since I didn't have time to fix the issues in the component router of the CMS so far, this maybe is the better way.

eddieajau:
Yeah - rock and a hard place. I suggest try and keep the topic alive on the platform and CMS/JBS list.  The fact is pull requests don't give you much marketing value.

PR #359, opened 2011-09-13
Converts JDispatcher::getInstance to use a protected static class to store its singleton.

eddieajau:
Currently JDispatcher cannot be mocked due to static storage within the getInstance method.  This method is changed to use a protected static class variable which can be overridden in child classes.  Also adds base tests for JDispatcher and an inspector class that can inject a mock dispatcher if required.

### DIFF
--- a/libraries/joomla/event/dispatcher.php
+++ b/libraries/joomla/event/dispatcher.php
@@ -26,6 +26,14 @@ jimport('joomla.base.observable');
 class JDispatcher extends JObservable
 {
 	/**
+	 * Stores the singleton instance of the dispatcher.
+	 *
+	 * @var    JDispatcher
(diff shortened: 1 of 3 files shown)


ianmacl:
# Test Results

Total Tests: 1427
Assertions: 9925
Failures: 0
Errors: 0
Test Time: 227.497691

WebMechanic:
Nice move :) 
I presume similar changes will be made to other "singleton'ish" classes where $instance[s] are also "hidden" within their respective getInstance() method?

eddieajau:
Yeah, we could do this more consistently.  With the test inspectors we should be able to make JFactory's static properties protected as well, which would make me feel more comfortable :)

PR #367, opened 2011-09-16
Minor improvements to error reporting

eddieajau:
Replaced usage of JException with Exception in the Application package. Added unit test to verify exception is thrown.
Fixed a problem in JError where an infinite loop is detected (defers to the echo handler).
Added a check in the custom error page handler for if the document body is empty, defer to the echo handler otherwise no error will display at all.


ianmacl:
# Test Results

Total Tests: 
Assertions: 
Failures: 
Errors: 
Test Time:

libraries/joomla/error/error.php
```diff
@@ -553,10 +579,18 @@ abstract class JError
 			if (defined('STDERR'))
 			{
 				fwrite(STDERR, "J$level_human: " . $error->get('message') . "\n");
+				if (constant('JDEBUG'))
+				{
+					frwite(STDERR, $trace);
```
chdemko: Should be fwrite instead of 'frwite'

PR #369, opened 2011-09-17
Fix some code style

realityking:
This may not be loved by all but I think the automatic code style correction has made some changes that make the code much less readable by removing the alignment on variable assignments. This is especially bad in large blocks like in JHtmlBehavior.

There are more like these but I wanted to see first if this gets pulled.

### DIFF
--- a/libraries/import.php
+++ b/libraries/import.php
@@ -77,5 +77,5 @@ if (isset($_SERVER['HTTP_HOST']))
 JLoader::import('joomla.base.object');
 
 // Register classes that don't follow one file per class naming conventions.
-JLoader::register('JText', JPATH_PLATFORM . '/joomla/methods.php');
+JLoader::register('JText',  JPATH_PLATFORM . '/joomla/methods.php');
 JLoader::register('JRoute', JPATH_PLATFORM . '/joomla/methods.php');
--- a/libraries/import.php
+++ b/libraries/import.php
@@ -77,5 +77,5 @@ if (isset($_SERVER['HTTP_HOST']))
 JLoader::import('joomla.base.object');
 
 // Register classes that don't follow one file per class naming conventions.
-JLoader::register('JText',  JPATH_PLATFORM . '/joomla/methods.php');
+JLoader::register('JText', JPATH_PLATFORM . '/joomla/methods.php');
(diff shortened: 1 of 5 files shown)


ianmacl:
# Test Results

Total Tests: 1465
Assertions: 10002
Failures: 0
Errors: 0
Test Time: 186.580545

eddieajau:
Responding to WebMechanic, just to be clear, the auto-formatter was used to assist bringing 300,000 lines of code into line in a relatively short period of time (took about a solid equivalent week to do it).  That was a one off operation but one that at least made all the files consistent - you don't have to like the consistency ... I'll still sleep at night :)

The next thing, which is more important, is the Code Sniffer standard which we've tried to keep as close to PEAR as possible, because, like it or not (and I personally don't like some of it), that is an industry standard.

Now, if you don't like the end result, that's your prerogative.  If you want to design a new sniff, rewrite the standards and apply them to our 300K line strong code base, be my guest :)  I've said in a number of places on the web before I think the current standard is a compromise based on the efficiency by which the code could be made consistent in the shortest period of time possible (and that was not a trivial amount of person-hours).  I'm totally open to considering changing the standard, but my only condition is it has to be applied consistently across the board at once as a package deal (more or less, but not a handful of files at a time).  Those wanting to take on that effort are most welcome.

Rouven, I'm happy with logentry, but I think most of the other changes will break sniff rules.  I can understand aligning on equals but if we do that I think we'd need to apply the same rules as docblock and use real spaces, not tabs - and that would get real messy.  However, I'm not happy with alignments within function argument lists at all anymore.

Happy to hear other comment/suggestions/_constructive_ criticism, but, lets respect that some people have already put some thought into the issue, tone down the usage of passionate language and realise that there are as many differences of opinion of formatting as there are lines of code.

LouisLandry:
We should really consider finding/writing a sniff that will handle this sort of alignment.  As @eddieajau mentioned, what was done with respect to all the files was in the name of consistency and having a solid baseline.  We also have much to do to build out a great set of sniffs and would certainly welcome the help if you want to participate @WebMechanic.

Additionally, I've seen _MUCH_ more disgusting code in the last 20 days or weeks than what was produced by the auto-formatter.  I'm absolutely envious if you haven't.

PR #377, opened 2011-09-21
Joomla CMS [#24919] JHtml::_('grid.boolean', $i, $value) should add void(0) if not given a task 

chdemko:
See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=24919


LouisLandry:
Christophe, is this really what you were wanting to do?  Remove the rel tag and associated JSON object with relevant data for the list item?

chdemko:
yes because it's the case where the button is inactive

elinw:
I tried this in a grid in the CMS and it worked fine from what I could tell. It didn't change things when there was a task, and when i removed the last two arguments the icon wasn't clickable.

PR #380, opened 2011-09-21
Joomla CMS [#26844] Improvements to batch processing

mbabker:
There is a bug in the current batch processing method in that when items are copied, if the access level is also changed, then the access change is applied to the original item.  The batchCopy process is modified to return an array of new item IDs to the batch process and the batch process is modified so that the copy/move command is executed first and that the $pks array is reset with the new IDs on a successful copy.

See CMS issue: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26844


eddieajau:
Thanks Michael!

PR #382, opened 2011-09-23
Joomla CMS [#26263] When a default value is defined, cannot save a field with an empty value 

chdemko:
[#26263] When a default value is defined, cannot save a field with an empty value
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26263


elinw:
I'm not sure what the point of a default is if you can save without a value. I think you need to choose one or the other.

elkuku:
**Checkstyle error details:**
libraries/joomla/form/form.php:1661
libraries/joomla/form/form.php:1661

chdemko:
This should be also related to http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27142

elinw:
I think this is mainly a case where people are using default rather than preselect. For example in a list you can define which value is initially selected without making it the default. Then if the user changes it to something else that is null it can save as null. That's html/javascript. You can do it for checkboxes too.  Default is a special case of filter, something that happens later, on save, if no choice is made or no response is typed. It's similar to the integer filter which forces 0 even if the user left it blank.

elinw:
By the way on checkbox there is an attribute checked which will preset the checkbox to checked but still allow save to null.

eddieajau:
This is the intended behavior. If you want the ability to have "nothing", don't include a default.

PR #387, opened 2011-09-24
Moved function static variable to class static variable, will help in ind

ssv445:
Moved function static variable to class static variable, will help in independent unit test cases.


eddieajau:
Thanks. Would you mind fixing up the docblock for the class property before we pull this.  You can see an example of how we've done it in JDispatcher https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/event/dispatcher.php

Thanks in advance.

elkuku:
**Checkstyle error details:**
libraries/joomla/html/toolbar.php:87
Missing function doc comment

PR #391, opened 2011-09-25
Joomla CMS [#25199] Problem with preg_quote in function utf8_ireplace 

chdemko:
See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=25199
and
https://sourceforge.net/tracker/?func=detail&aid=3413766&group_id=142846&atid=753842


eddieajau:
Christophe, can you add a test case to getStr_ireplaceData in JStringTest to make sure we cover this change.

Thanks in advance.

chdemko:
Test case added

PR #394, opened 2011-09-26
Added check to sending header when database connection cannot be established in JFactory

eddieajau:
Fix to JFactory::_createDbo()


elinw:
Andrew is that going to solve the infinite loop on missing database issue in the CMS?

PR #398, opened 2011-09-27
README expanded

gnomeontherun:
The current README is basic, and I have added some additional information about contributing and links. Its not perfect, but a start to making it a more useful document.

Also I ran a compatibility test and PHP 5.2+ was required, not 5.3+.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -9,12 +9,12 @@ You can find out more about Joomla development *(though not platform specific)* 
 
 You can discuss *Joomla Platform* development at: http://groups.google.com/group/joomla-dev-platform
 
-
 Requirements
 ------------
 
-* PHP 5.3+
-
+* PHP 5.2+
+* MySQL 5.04+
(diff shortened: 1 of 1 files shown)


eddieajau:
Hi Jeremy.  I've added the bulk of your comments to the contributing wiki page at https://github.com/joomla/joomla-platform/wiki/Contributing-to-the-joomla-platform and added a link in the README to that page as well.

We have actually deliberately set the minimum spec at PHP 5.3 because some classes (not used in the CMS) do require them, and we don't want to discourage people to write for PHP 5.3.  That said, anything we know is used in the CMS still needs to be compatible, at least until the CMS increases its minimum spec. (expected for 3.0).

I'm going to leave off the MySQL version for now because that's really a function of apps build on top of the platform.

I also left off the bit about extensions because that is very CMS specific and the Platform has a much broader audience.

Thanks for the extra words though - I think they are really helpful for others in the community that are not quite sure how to start.

PR #400, opened 2011-09-27
Joomla CMS [#24973] JHTML select.options with html entities

chdemko:
See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=24973


eddieajau:
I'd really like to see a unit test added to confirm this change works if that's possible.

chdemko:
unit tests added

PR #402, opened 2011-09-27
Adding index.html files

chdemko:
Adding index.html files which are missing in some directories


eddieajau:
Hi Christophe. We talked it over and we'd prefer to leave the index.html files out of the platform repo because the end goal is to have it installable via PEAR or some other mechanism.  You can add them into the CMS and they will stay there when you drop a new version of the platform in.  We could easily add a script to check each folder has one.

PR #403, opened 2011-09-28
Joomla CMS [#26885] JText do not recognize "" no-string

chdemko:
See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26885


eddieajau:
I don't see how this pull request is fixing the problem in JText??

chdemko:
In fact there are 2 errors:
-in the JHtmlJGrid:action method which is calling JText::_ with empty strings
-in JLanguage::_ which should detect an empty string as its first arg

PR #406, opened 2011-09-29
Deprecate JDatabaseMySQL, JDatabaseExporterMySQL and parts of JBrowser

realityking:
For the MySQL stuff I decided not to add the deprecation warnings since PHP only softly deprecated mysql as well. http://news.php.net/php.internals/53799


elinw:
In the system  we use the deprecation tag gives the earliest possible release on which something will be removed. So I would say that JBrowser will be around until probably 12.3 at the earliest and that gives far warning.

realityking:
Why 12.3?

elinw:
Just arbitrary but more than 3 months notice. I don't think we have a standard minimum notice but we've discussed before on list that developers need probably 6-12 months notice.

realityking:
IMO 12.1 is totally fair for these parts(!) of JBrowser. I don't think we should remove stuff in 3.1, this way it won't be in 3.0.

elinw:
I would say minimum 12.2 then. 3 Months is not enough notice.  But I have absolutely no problem with removing deprecated items in 3.1 or even 3.2 with fair warning if it means that users and extension developers will be able to move to the newer version more quickly. It's like a short term legacy layer. We don't want to  make them prioritize removing JBrowser use over moving to UCM or making a Finder plugin.. 
Although that said, it could go into the CMS libraries until 3.0 is released--it really would have to. in fact since not having it available will break 2.5. extensions.

elinw:
Actually in thinking about this some more I would add 
@note   Will be removed when PHP support is removed from the minimum supported version of PHP.

Since we have no real idea when that will be and we are supporting 5.3 for the foreseeable future I think that's adequate even though it breaks the docblocks rules.

.

elkuku:
**Checkstyle error details:**
libraries/joomla/environment/browser.php:26
@deprecated tag comment indented incorrectly; expected 1 spaces but found 3

realityking:
_grr_ Why am I not allowed to align the deprecated tag with the others?

elinw:
I think that is an error in the check, it should be aligned.

realityking:
Closed in favor of #493.

PR #411, opened 2011-10-01
Introduce JURI::media().

realityking:
This allows accessing media resources trough a different domain which can give performance benefits. Also may make is easier to use a CDN with Joomla.

Note that the change to JBrowser is slight backward compatibility issue because it causes PHP strict warning when used like the old API. I thought it was a good idea to use a strict method here because this saves us from initializing JBrowser when we don't need it.

Original pull request: https://github.com/joomla/joomla-platform/pull/190


rvsjoen:
I think this is a nice and simple feature that will unify how to access media for a component. Although you can most certainly argue that static methods are evil, JURI::media() is likely to replace JURI::root() on a 1:1 basis, and it makes it simpler to move to another (non-static) approach later since it will be easy to separate between media access and other file access.

In short, I think the benefits outweigh the drawbacks of adding yet another static function to JURI.

realityking:
If this is ok code wise someone would need to look at the unit tests, one of the mock objects doesn't behave nicely with this.

LouisLandry:
I'd much rather see this feature not depend upon a static method like that.  I'd much rather it be a configuration value that can be interacted with in other ways.  In JWeb I added: https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/application/web.php#L1160 which attempts to solve the same problem you are trying to solve here but by using a different mechanism.

ianmacl:
We're not going to go the route of adding more static methods and properties.  JApplicationWeb::loadSystemUris goes in the direction we want to head in terms of having a store of configurable paths.

PR #415, opened 2011-10-02
Typo in JGrid

Hackwar:
A little bit embarassing, but the JGrid class has a typo in the rendering of the footer rows. It should be <tfoot> instead of <tfooter>. This has been fixed in this pull request.


LouisLandry:
Heh.  Happens to all of us from time to time. :)

PR #416, opened 2011-10-02
Default Layout should not be set in JController->display() as it's already been set in the JView Constructor

Bakual:
Only setting the view layout if a layout is passed in the request.
The default value should not be set here as it already has been set by the JView constructor, allowing for a view to have a different default value than 'default'.

### DIFF
--- a/libraries/joomla/application/component/controller.php
+++ b/libraries/joomla/application/component/controller.php
@@ -660,21 +660,15 @@ class JController extends JObject
 		$document = JFactory::getDocument();
 		$viewType = $document->getType();
 		$viewName = JRequest::getCmd('view', $this->default_view);
-		$viewLayout = JRequest::getCmd('layout', '');
+		$viewLayout = JRequest::getCmd('layout', 'default');
 
-		$view = $this->getView($viewName, $viewType, '', array('base_path' => $this->basePath));
+		$view = $this->getView($viewName, $viewType, '', array('base_path' => $this->basePath, 'layout' => $viewLayout));
 
 		// Get/Create the model
 		if ($model = $this->getModel($viewName))
 		{
 			// Push the model into the view (as default)
 			$view->setModel($model, true);
-		}
-
-		// Set the layout
-		if ($viewLayout)
-		{
-			$view->setLayout($viewLayout);
 		}
 
 		$view->assignRef('document', $document);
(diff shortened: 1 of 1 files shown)


chdemko:
I don't see any objection to accept this pull request

Hackwar:
The best way would be to move this up a bit and put the layout variable into the config array of the JView constructor. So instead of removing the default value in line 663, we could simply replace line 665 with:

$view = $this->getView($viewName, $viewType, '', array('base_path' => $this->basePath, 'layout' => $viewLayout));

and completely remove lines 675 to 678.

Bakual:
I don't think that would be better.
As it is now, the custom view constructor could define a default layout by simply using: 

$config['layout'] = 'new_default';
parent::__construct($config);

If the custom view constructor doesn't set a layout, then the JView class will set it to 'default'.
The controller afterwards would override that default layout only if it's requested.

With your change the JView class would never set a default, since the array key is already present (and it only checks for that). So the controller would need to set the default value (which would be against MVC), but then a custom view class could not set a new default, or it needs much more code to do so (check if a layout is passed in request, basically doing the controller stuff again, which again is against MVC).

Hackwar:
The idea was, that the code looks like this:
$viewLayout = JRequest::getCmd('layout', 'default');

$view = $this->getView($viewName, $viewType, '', array('base_path' => $this->basePath, 'layout' => $viewLayout));

In this case, you always have at least "default" as your layout. If you are overriding this in the constructor of the custom view class like you described, it would still work. There is one downside however. If a constructor in the custom view does the same check:
if(isset($config['layout']))
    $config['layout'] = 'newdefault';
it would fail because the controller sets the layout automatically by default. The code would still be MVC-theory compliant, but you might argue that it is not 10000000% backwards compatible.

Bakual:
Backwards compatibility isn't really an issue since it doesn't work as of today :)

The code in a custom view constructor would then have to look like this:

if (!JRequest::getCmd('layout', ''))
{
    $config['layout'] = 'new_default';
}
parent::__construct($config);

This would work, however it takes calls JRequest twice for exactly the same thing. Alltough only if there is a custom default which probably is rare.
However I'm not sure if it's the best approach. That's up to others than me to decide. I'm happy if it's working in the end :)

Hackwar:
By default, we would have less code executed. A custom view constructor could act like this:

$config['layout'] = JRequest::getCmd('layout', 'customdefault');
parent::__construct($config);

Bakual:
Ah true, that would be easy enough. I agree with you :)

I think we need a new Pull Request now? I'm not that familiar with GitHub yet.

Hackwar:
Not really, simply change the code in your branch and it will automatically show up in this pull request.

Bakual:
Hopefully that worked

Hackwar:
Looks good

PR #417, opened 2011-10-02
Use elseif instead else if.

realityking:
Main reason is that it is much easier to parse when applying other style checks but it's also more consistent and as an added bonus there is a very small speed increase involved (probably < 1%).


build/phpcs/Joomla/Sniffs/ControlStructures/ElseIfDeclarationSniff.php
```diff
@@ -0,0 +1,82 @@
...
+        $phpcsFile->addError($error, $stackPtr, 'NotAllowed');
+      }
+
+    }//end process()
+
+
+}//end class
+
+
+?>
```
chdemko: Why did you add a ?> at the end of the file?

realityking:
Good point, fixed.

LouisLandry:
Fantastic!

elkuku:
I **love** GitHub `:)`

PR #418, opened 2011-10-02
Databasefield and fixes to getConnectors method

elinw:
This addresses some problems in getConnectors which made it not work with the current file structure in the database folder and with the newer naming conventions for the database class. 
It also adds a new field that uses getConnectors to dynamically create a list of available database connection/driver combinations optionally constrained to a list of supported databases for the application.

This will be important for multidatabase support since not all servers will support all databases and not all applications will support all available drivers.


elinw:
Oh that is a messy pull request.

PR #419, opened 2011-10-03
Issues in getConnectors and a databaseconnection field

elinw:
This addresses some problems in getConnectors which made it not work with the current file structure in the database folder and with the
newer naming conventions for the database class. 
It also adds a new field that uses getConnectors to dynamically create a
list of available database connection/driver combinations optionally
constrained to a list of supported databases for the application.

This will be important for multidatabase support since not all servers
will support all databases and not all applications will support all
available drivers.


libraries/joomla/database/database.php
```diff
@@ -169,57 +177,71 @@ abstract class JDatabase implements JDatabaseInterface
...
+		if (isset(self::$connectors))
+		{
+			return self::$connectors;
+		}
 		// Instantiate variables.
 		$connectors = array();
 
 		// Get a list of types.
-		$types = JFolder::folders(dirname(__FILE__));
+		$types = JFolder::files(dirname(__FILE__).'/database');
```
chdemko, line 188: you can add some parameters to JFolder::files to filters the files you need (avoiding the tests done after)

libraries/joomla/database/database.php
```diff
@@ -169,57 +177,71 @@ abstract class JDatabase implements JDatabaseInterface
...
+				// If the class doesn't exist, let's look for it in a folder and register it.
+				if (!class_exists($class)) {
+	
+					// Derive the file path for the driver class.
+					$path = dirname(__FILE__).  '/database/' . $type . '/'. $driverfile;
+	
+					// If the file exists register the class with our class loader.
+					if (file_exists($path))
+					{
+						JLoader::register($class, $path);
```
chdemko, line 221: I think you have to get a third parameter to JLoader::register: true to force the update of the classes list in the loader

libraries/joomla/database/database.php
```diff
@@ -169,57 +177,71 @@ abstract class JDatabase implements JDatabaseInterface
...
-
-		return $connectors;
+				
+				// Sweet!  Our class exists, so now we just need to know if it passes it's test method.
+				if ($class::test())
+				{
+					$connectors[] = $type;
+				}
+			}		
+			self:$connectors = $connectors;
```
chdemko: use self::$connectors (not self:$connectors)

libraries/joomla/database/database.php
```diff
@@ -169,57 +177,71 @@ abstract class JDatabase implements JDatabaseInterface
...
+			
+			return self:$connectors;
 	}
```
chdemko, line 246: it seems to lack a closing bracket

libraries/joomla/database/database.php
elinw, line 188: I'm not sure what parameters you might mean given that we are trying to pick up the databases dynamically from the file names.  The point being we don't know their names in advance.

chdemko, line 188: after that line, you are excluding some files (containing 'exporter, 'importer', ...). Yo can do that while you are calling JFolder::files

chdemko:
Would it be possible to put all files concerning a connector in a folder named

libraries/joomla/database/database/_connector_

i.e. for mysql:

libraries/joomla/database/database/mysql

and name these files

libraries/joomla/database/database/mysql/driver.php: define class JDatabaseMysqlDriver
libraries/joomla/database/database/mysql/exporter.php: define class JDatabaseMysqlExporter
libraries/joomla/database/database/mysql/importer.php: define class JDatabaseMysqlImporter
libraries/joomla/database/database/mysql/query.php: define class JDatabaseMysqlQuery

when a JDatabase::getInstance or JDatabase::getConnectors is called, do a JLoader::discover('JDatabase'.$type, dirname(__FILE__) . '/' . $type);

and replace
$class = 'JDatabaseDriver' . ucfirst(trim($type));
by
$class = 'JDatabase'.ucfirst(trim($type)).'Driver';

elinw:
Christophe, 

That idea was discussed several months ago and will happen eventually. As you can tell, the code I submitted alreay begins the process of looking in folders with the same names as the types.  Please note that this is not a refactoring this is a bug fix (getConnectors is looking in the wrong places for files, for the wrong class names, and so on) and attempt to add a field that is absolutely necessary to be in 11.3 if we are going to have useful multiple database support in any applications.  The bug fixes to getConnector are only consequent to making the field. In the future someone can take up refactoring including reorganizing the files but that is not what this is about.  Let's try to avoid scope creep on bug fixes.

elinw:
Ok fixed the missing : and } .

libraries/joomla/database/database.php
elinw, line 188: It is better to do it that way than writing regex for those strings.

libraries/joomla/database/database.php
elinw, line 221: No, I don't think that is correct. We are just making a list here, not overriding.

LouisLandry:
So I think if you look at the current version of this method in staging: https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/database/database.php#L161

There are really only three lines that need to change here: 176, 188 and 195.  Every other aspect of this method should work very well as expected.  The form field I am fine with, but as for the JDatabase::getConnectors() method I don't think it requires memory caching or that much change.

elinw:
Should we look for driver.php? It seems out of bounds to me at this point. 

In other words are you saying just do $type.php and be done?

elinw:
sigh i's going to be easier to close this.

PR #420, opened 2011-10-08
JGrid: Small refactoring

Hackwar:
After some more thought, the function addRowCell is ill-named and should be setRowCell. I also added getter/setter for the Row Option. The unittests have been changed accordingly.


Hackwar:
Also had to change the behavior of the setRowCell function to completely replace the options of the row cell.

Hackwar:
There was a bug in the rendering of row-attributes, which was fixed in this last commit. I also cleaned up the rendering function a bit. instead of having three functions to render the header, footer and body, we now have only one function that renders table areas and only switches out the changing parts, which is less code and potentially a small performance gain.

chdemko:
Hi Hannes,

Could you please update your branch (patch cannot be applied)

Hackwar:
In the hopes that I did this right: The conflict should be fixed now.

PR #421, opened 2011-10-09
Plugin event and common layouts in JView

Hackwar:
This pull request adds a plugin event "onDisplayView" to display(), which allows to modify the JView object, so that a plugin for example could add columns to a JGrid object or choose a different layout to load.

This example shows a method from a potential plugin and what that plugin could do to the view object

``` php
function onDisplayView($context, $view)
{
    //Return if the view is not the articles view of com_content
    if($context != 'content.articles') {
        return false;
    }

    //Remove the column from the com_content articles list table
    $view->table->deleteColumn('id');

    //Set a different layout than the one that was decided by the controller/URL/router
    $view->setLayout('mycustomlayout');

    //Add a different path for layouts so that you can use your own layouts, for example for a series of templates
    $view->addTemplatePath(JPATH_ROOT.'/plugins/system/contentlayouts/');
}
```

This pull also adds two other features. The first is a parameter $common to display() and loadTemplate(), which loads a common layout from a special folder, so that a generalized form layout for all components can be made or a generalized list layout for backend views. The second is a parameter $data to loadTemplate(), which allows recursive calls, like in the category tree of a category (or categories) view.

This example shows how the $data parameter in a layout might be used:

``` php
if($data) {
    $categories = $data;
} else {
    $categories = $this->categories;
}

echo '<ul>';
foreach($categories as $category) {
    echo '<li>'.$category->name;
    if(count($category->getChildren())) {
        //'childcategories' is this layout
        $this->loadTemplate('childcategories', $common, $category->getChildren());
    }
    echo '</li>';
}
echo '</ul>';
```

I also covered some methods with unittests. Unfortunately I don't know how to cover the two methods that I changed by unittests.


Hackwar:
Anything new on this one?

Hackwar:
The added unittest for get() is for this pull request: https://github.com/joomla/joomla-platform/pull/336 and will not pass completely on this code base.

Hackwar:
Don't know if you are right, but agreeing with you on this one. :-) Done.

libraries/joomla/application/component/view.php
```diff
@@ -49,6 +49,13 @@ class JView extends JObject
 	protected $_defaultModel = null;
 
 	/**
+	 * The plugin group to load for the view
+	 * 
+	 * @var string
+	 */
+	protected $_group = null;
```
eddieajau: Protected class properties don't have an underscore.

Hackwar: I added the underscore in order to keep it in the style of the rest of the class. Should this be changed in this case?

eddieajau: Yes please. Any new work should match the current check-style rules.  Thanks.

Hackwar:
Added a few more unittests for JView

AmyStephen:
Just my opinion, but, a core framework class like JView should not have logic in it specific to a single component/view. Abstracting this concept out so that it is applicable to any component/view would be best.

Hackwar:
Sorry, but which component specific logic do you mean?

AmyStephen:
Hannes - my apologies, I see now that you are sharing an example of usage -- not a hardcoding 'content.article' in JView.

eddieajau:
I'm not overly comfortable with the changes for a few reasons:
1. I'd like to see some buy-in and testing/confirmation from the CMS, that this is the right way to do this.
2. I don't know if onDisplayView is the right way to handle the event. I think that needs more investigation.
3. I'm aware quite a few people want to revisit the MVC triad so I think it would be better to review these kinds of changes with that review.

I'm going to close this pull for the platform for those reasons, for now, but it could be that they could still be applied to the CMS as a view class override which, to be honest, is a better place for these changes in my opinion (but I'm happy to be convinced otherwise in a peer review of the MVC API).

Hackwar:
I kindly ask you to reconsider. This change is extremely usefull for every developer out there and it especially allows existing extensions to use this feature without any change.

Regarding #1: What kind of buy-in do you expect from the CMS? So far there is quite a lot of positive feedback on the mailinglist regarding this, so I would consider this a good addition.

Regarding #2: I looked for a short and concise event name, so this is what came up. I would rather not change it to something like in JModelForm, which I think is pretty long and to specific aimed at "content".

Regarding #3: I don't see the complete refactor of the MVC triad happening in Joomla 2.5 and I'm happy to scrap this feature again when the new implementation is there. However, for the time being, its an unobtrusive and simple way to greatly boost the flexibility. If this does not get into 2.5, it means that most extensions wont be able to use it for the next 18 months.

This feature for me is a pretty crucial part of the JGrid implementation, since it allows me (and several other extensions) to drop our own user-list view and instead augment the existing user manager or extend the details listed for an article in the article list view. 

Please reconsider this request.

eddieajau:
Hannes, which mailing list are you referring to?  On the platform list only Nic replied.  With all due respect "a lot of positive feedback" is hardly accurate.  If there is discussion elsewhere that I'm not aware of, I'm happy to peruse it.

I'm not so much worried about the event name, but how/when/where it's executed.  You also have to realise that the platform is being used for more than the CMS.  A review of the MVC is independent of what is or isn't in Joomla 2.5.

Whatever that case, closing this pull is just that.  It's not ready in my opinion and we don't have enough tools on this tracker to put it in a holding pattern.  It's nothing personal, it doesn't mean it can't go in, it just means it's off the pull request list for now.  That's all.

AmyStephen:
Hannes - Since this approach calls for a plugin, I'm just not understanding how it's any easier than simply overriding the view using a system plugin. I'd be more inclined to support it if it actually made it easier for frontend devs to change the layout, but they are not likely to have these skills. I've looked a few times at this but I end up just sort of scratching my head on how it's better.

+1 on looking at the MVC triad more deeply. Now that we have some experience with it, I think there are good gains to be found.

Hackwar:
I don't want to start making this into a discussion forum, so the short answer is: Try to override a class/view two times. My usecase is a shopping component and a forum component, which both want to display data in the user list view. The first one wants to show open/closed orders and the second one wants to show the number of postings from that user. This currently is not possible, with the combination of JGrid and this pull request, it would be.

eddieajau:
Hannes, I understand what you are saying, but I think you are into very application specific territory with your use-case.  Now, this might turn out to be a great solution, it might not, but I'd like to spend some time on having a plan for where MVC is going, not least of which is working out where new "hooks" should be placed, before we slap more things into it.  I think that's a fair and reasonable approach and I know quite a few people want to join that conversation.

oc666:
Hey all,
sorry to pop-up this issue out of the basement, but really good use case is when plugin needs to override component view layout. It can't be done with the current state of joomla.
See also my post in the list: http://groups.google.com/group/joomla-dev-general/browse_thread/thread/3911061f0e98a734

PR #422, opened 2011-10-09
CodeSnifferTester

elkuku:
This is an **experiment** !
It contains a set of files with potential _coding standards violations_ `bad.php` and their _good_ counterparts `good.php`.
The purpose is
1. Verify that the code sniffer detects the violations and find violations not covered by a sniff.
2. By providing the sniff result from a reference system (ATM mine), other systems could be checked to ensure that the sniffer works the same way on different operating systems (Not sure if that really makes sense, or is applicable on systems like windows...). The files `good.xml` and `bad.xml` contain the code sniffer output.
3. Generate some simple documentation (aka Cheat Sheet ?) by parsing the PHP files and the included comments. A sample HTML output can be seen here: http://elkuku.github.com/joomla-platform/coding-standards/cheat-sheet/
   Parsing into Docbook should also be possible.

Three command line scripts are included:
1. `runtests.php` Runs the tests (phpcs has to be in your PATH) and verifies the results against the reference output. (purpose 2)
2. `maketests.php` Should be executed if anything "changes" (for purpose 2)
3. `makedoc.php` parses the PHP files and generates a HTML output. (Purpose 3)


LouisLandry:
I feel like this may be better suited for it's own repo.  I love seeing tools like this coming to life, and I am incredibly supportive... but I'm not sure that this really belongs in the main repository.  What do you think?

eddieajau:
Nikolai, this is great work but we're going to look at doing this in a new repo so that it can be used elsewhere (eg, for the CMS as well). On that basis I'm going to close this pull and we'll contact you when we have a new repo set up.

elkuku:
That's good news.
As I stated it was only an experiment. It helped me in creating and verifying the sniffs and maybe it could be used to generate some kind of documentation so the (dummy) code would server multiple purposes.
Speaking of which, would you like me to update / add the "new standards" to the docbook documentation ? 

Looking forward to get my hands on the CMS code as well :).

eddieajau:
Yes, please update the standards documents. That would be great. Thanks!

elkuku:
Fine - right now I am at #CPQUITO - but next week I believe I will find some time to work on this one.

elkuku:
Quick question @eddieajau: Do I need to adjust any settings in XMLMind to edit the documentation (like max. line length) ?

eddieajau:
I've got indentation set to 2, and max line length set to 130.

PR #423, opened 2011-10-09
CodeSniffer should run from any directory

elkuku:
Right now it is required to copy or softlink the coding standards to your system standards directory in order to use them.
I find this difficult for testing and it might also raise the barriers for developers.

The reason the sniffs won't run from any directory is the fact that they make one reference to a system standard called "Joomla" - only to set the severity for inline control structures from error to warning...

So, I removed that reference, updated the corresponding sniff file and also modified the build.xml to pick the Joomla! Standard found in the directory instead of the system standard.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -81,7 +81,7 @@
   <exec executable="phpcs">
    <arg value="--report=checkstyle" />
    <arg value="--report-file=${basedir}/build/logs/checkstyle.xml" />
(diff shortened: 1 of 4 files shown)


ianmacl:
This would be great.  I'll test and pull when I get a chance.

LouisLandry:
This rocks.  Thanks a bunch!

PR #424, opened 2011-10-10
Code formatting

elkuku:
This fixes some remaining coding standards violations. Some of them have been introduced recently..

I was not quite sure how to handle multi line ifs
a)

``` ruby
if ($a
    && $b)
{
    true;
}
```

b)

``` ruby
if ($a
    && $b
)
{
    true;
}
```

The coding standard says that version b) is the good one.

I believe multiline ifs, function declarations, and function calls should be formatted in the same way. What do you think ?

Also: The only remaining errors are those "File is being conditionally included; use "include_once" instead" - what to do about that ?

So.. may we teach Mr. Jenkins (or ask Mr. Mc Lennan politely) to include the sniffer when approving pull requests ?

### DIFF
--- a/libraries/joomla/filesystem/folder.php
+++ b/libraries/joomla/filesystem/folder.php
@@ -577,7 +577,8 @@ abstract class JFolder
 		while (($file = readdir($handle)) !== false)
 		{
 			if ($file != '.' && $file != '..' && !in_array($file, $exclude)
-				&& (empty($excludefilter_string) || !preg_match($excludefilter_string, $file)))
+				&& (empty($excludefilter_string) || !preg_match($excludefilter_string, $file))
+			)
 			{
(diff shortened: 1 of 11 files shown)


LouisLandry:
I actually think I like (a) better, and agree fully that all of the multiline stuff should be consistent regardless of which we choose..  Also, I'd like to remove the "file is being conditionally included" sniff altogether.  I don't think there is a cut and dry way to verify what is good/bad there so it doesn't make sense for the code sniffer to look for it.

Thanks a ton for the work on these things.  It makes a huge difference.

eddieajau:
+1 for (a)

eddieajau:
Nikolai, I've pulled this one because it's a good patch, but would love it if you can change the rule to check for the (a) case you show above.  Thanks again!

PR #425, opened 2011-10-11
Sniff fixes

elkuku:
This
- Fixes the else and elseif sniff
- Fixes the standard for multi line function declaration (removes some unnecessary errors)
- Fixes the last line empty not working in classes


LouisLandry:
These changes appear to also have been in a separate pull request that was pulled in.  I'm going to close this one since it seems to be unnecessary.  If there is something I missed please feel free to open a new request.

PR #427, opened 2011-10-15
New JImage class.

LouisLandry:
This is a collaboration between Rune and myself to create a new JImage class for the platform.  There is good unit test coverage and a simple but effective API.  The library supports cropping, resizing, rotating and applying several different filters to an image using GD.


libraries/joomla/media/image.php
```diff
@@ -0,0 +1,768 @@
...
+			JLog::add('The GD extension for PHP is not available.', JLog::ERROR);
+			throw new JMediaException;
+			// @codeCoverageIgnoreEnd
+		}
+
+		// Determine which image types are supported by GD, but only once.
+		if (!isset(self::$formats[IMAGETYPE_JPEG]))
+		{
+			$info = gd_info();
+			self::$formats[IMAGETYPE_JPEG] = ($info['JPEG Support']) ? true : false;
```
pasamio, line 85: Isn't this technically against our coding style/standards?

libraries/joomla/media/image.php
```diff
@@ -0,0 +1,768 @@
...
+			// @codeCoverageIgnoreEnd
+		}
+
+		// Build the response object.
+		$properties = (object) array(
+			'width' => $info[0],
+			'height' => $info[1],
+			'type' => $info[2],
+			'attributes' => $info[3],
+			'bits' => @$info['bits'],
```
pasamio: Why the use of the error suppression operator here on these fields?

libraries/joomla/media/image.php
```diff
@@ -0,0 +1,768 @@
...
+	 *
+	 * @return  integer
+	 *
+	 * @since   11.3
+	 * @throws  JMediaException
+	 */
+	public function getWidth()
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($this->handle) || get_resource_type($this->handle) != 'gd'))
```
pasamio: why the double bracketing?

libraries/joomla/media/image.php
```diff
@@ -0,0 +1,768 @@
...
+	 *
+	 * @return  integer
+	 *
+	 * @since   11.3
+	 * @throws  JMediaException
+	 */
+	public function getHeight()
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($this->handle) || get_resource_type($this->handle) != 'gd'))
```
pasamio: why the double bracketing?

libraries/joomla/media/image.php
```diff
@@ -0,0 +1,768 @@
...
+	 *
+	 * @return  JImage
+	 *
+	 * @since   11.3
+	 * @throws  JMediaException
+	 */
+	function crop($width, $height, $left, $top, $createNew = true)
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($this->handle) || get_resource_type($this->handle) != 'gd'))
```
pasamio: why the double bracketing?

libraries/joomla/media/image.php
```diff
@@ -0,0 +1,768 @@
...
+	 * @return  void
+	 *
+	 * @since   11.3
+	 * @see     JImageFilter
+	 * @throws  JMediaException
+	 */
+	public function filter($type, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($this->handle) || get_resource_type($this->handle) != 'gd'))
```
pasamio: why the double bracketing?

libraries/joomla/media/image.php
```diff
@@ -0,0 +1,768 @@
...
+	 *
+	 * @return  JImage
+	 *
+	 * @since   11.3
+	 * @throws  JMediaException
+	 */
+	function resize($width, $height, $createNew = true, $scaleMethod = JImage::SCALE_INSIDE)
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($this->handle) || get_resource_type($this->handle) != 'gd'))
```
pasamio: copy and paste error it feels like by now, however why the double bracketing?

libraries/joomla/media/image.php
```diff
@@ -0,0 +1,768 @@
...
+	 *
+	 * @return  JImage
+	 *
+	 * @since   11.3
+	 * @throws  JMediaException
+	 */
+	function rotate($angle, $background = -1, $createNew = true)
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($this->handle) || get_resource_type($this->handle) != 'gd'))
```
pasamio: why the double bracketing?

libraries/joomla/media/image.php
```diff
@@ -0,0 +1,768 @@
...
+	 * @return  void
+	 *
+	 * @see     http://www.php.net/manual/image.constants.php
+	 * @since   11.3
+	 * @throws  JMediaException
+	 */
+	function toFile($path, $type = IMAGETYPE_JPEG, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($this->handle) || get_resource_type($this->handle) != 'gd'))
```
pasamio: moar! why the double bracketing?

libraries/joomla/media/image.php
LouisLandry, line 85: Which part?  I don't think we have any rules against using the ternary operator, unless you are seeing something else that is problematic.

libraries/joomla/media/image.php
LouisLandry: Some array offsets are only available for some image formats, so sometimes the value at the offset isn't set therefore it would raise a notice.

libraries/joomla/media/image.php
LouisLandry: Ya, that's a copy paste error.  Technically the second opening brace should be in front of the get_resource_type call to wrap that != expression.

libraries/joomla/media/image.php
pasamio: So why not quickly check it and set it to null or zero? Using the error suppression operator here will cause an avoidable performance hit.

libraries/joomla/media/image.php
pasamio, line 85: no that was it, I thought we had a rule against it but i might be thinking of the short hand if stuff instead

pasamio:
Thanks for fixing up those issues!

libraries/joomla/media/filters/brightness.php
```diff
@@ -0,0 +1,45 @@
...
+	 * @param   resource  $options  An array of options
+	 *
+	 * @return  void
+	 *
+	 * @since   11.3
+	 */
+	function execute($handle, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($handle) || get_resource_type($handle) != 'gd'))
```
elinw: This too?

libraries/joomla/media/filters/contrast.php
```diff
@@ -0,0 +1,45 @@
...
+	 * @param   resource  $options  An array of options
+	 *
+	 * @return  void
+	 *
+	 * @since   11.3
+	 */
+	function execute($handle, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($handle) || get_resource_type($handle) != 'gd'))
```
elinw: and here

libraries/joomla/media/filters/edgedetect.php
```diff
@@ -0,0 +1,44 @@
...
+	 * @param   resource  $options  An array of options
+	 *
+	 * @return  void
+	 *
+	 * @since   11.3
+	 */
+	function execute($handle, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($handle) || get_resource_type($handle) != 'gd'))
```
elinw: isn't search great.

libraries/joomla/media/filters/emboss.php
```diff
@@ -0,0 +1,44 @@
...
+	 * @param   resource  $options  An array of options
+	 *
+	 * @return  void
+	 *
+	 * @since   11.3
+	 */
+	function execute($handle, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($handle) || get_resource_type($handle) != 'gd'))
```
elinw: and here

libraries/joomla/media/filters/grayscale.php
```diff
@@ -0,0 +1,44 @@
...
+	 * @param   resource  $options  An array of options
+	 *
+	 * @return  void
+	 *
+	 * @since   11.3
+	 */
+	function execute($handle, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($handle) || get_resource_type($handle) != 'gd'))
```
elinw: :)

libraries/joomla/media/filters/negate.php
```diff
@@ -0,0 +1,44 @@
...
+	 * @param   resource  $options  An array of options
+	 *
+	 * @return  void
+	 *
+	 * @since   11.3
+	 */
+	function execute($handle, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($handle) || get_resource_type($handle) != 'gd'))
```
elinw: :)

libraries/joomla/media/filters/sketchy.php
```diff
@@ -0,0 +1,44 @@
...
+	 * @param   resource  $options  An array of options
+	 *
+	 * @return  void
+	 *
+	 * @since   11.3
+	 */
+	function execute($handle, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($handle) || get_resource_type($handle) != 'gd'))
```
elinw: here

libraries/joomla/media/filters/smooth.php
```diff
@@ -0,0 +1,45 @@
...
+	 * @param   resource  $options  An array of options
+	 *
+	 * @return  void
+	 *
+	 * @since   11.3
+	 */
+	function execute($handle, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if ((!is_resource($handle) || get_resource_type($handle) != 'gd'))
```
elinw: here

libraries/joomla/media/filters/contrast.php
```diff
@@ -0,0 +1,45 @@
...
+	 */
+	function execute($handle, $options = array())
+	{
+		// Make sure the file handle is valid.
+		if (!is_resource($handle) || (get_resource_type($handle) != 'gd'))
+		{
+			JLog::add('The image is invalid.', JLog::ERROR);
+			throw new JMediaException;
+		}
+
```
Hackwar, line 42: Since that if block is in so many places and we already had a big copy/paste "bug" here, wouldn't it be better to move this code into a method of JImage and JImageFilter? it would make every function 4 lines shorter and if we ever thought that instead of JMediaException we want to have a JImageException or something like that, we would only have to change it in one place...

Hackwar:
You are retrieving the image properties including the attributes bit. Is it expected behavior, that JImage practically clears the metadata of an image? I'm not 100% sure about this, but here is my case: Consider someone is uploading his holiday pictures with a copyright notice and maybe a GPS position and the gallery used for that resizes the image and what not. After the resizing, the metadata would be gone as far as I can see, or am I wrong? Basically, if there is a possibility to wipe the data, but keep it on demand, that would be great...

LouisLandry:
Great point Hannes.  I have some other ideas I'm gonna exercise on this thing too.

vietvh:
I am actually using JImage for my project, but just some basic functionalities like cropping and resizing, it works well so far, looking forward to get it merged into j platform.

LouisLandry:
OK, I've restructured a few things and am now using the SPL exceptions where they seemed most appropriate.

PR #431, opened 2011-10-17
Fixes strict issue in JFTP::get()

AmyStephen:
Strict Standards: Declaration of JFTP::get() should be compatible with that of JObject::get() in /users/amystephen/sites/molajo/libraries/jplatform/joomla/client/ftp.php on line 1741

### DIFF
--- a/libraries/joomla/client/ftp.php
+++ b/libraries/joomla/client/ftp.php
@@ -151,7 +151,7 @@ class JFTP
 	 * @since   11.1
 	 */
 
-	private function __construct($options = array())
+	public function __construct($options = array())
 	{
 
 		// If default transfer type is not set, set it to autoascii detect
@@ -192,7 +192,7 @@ class JFTP
 	 *
 	 * @since   11.1
 	 */
-	private function __destruct()
+	public function __destruct()
 	{
 		if (is_resource($this->_conn))
 		{
@@ -1515,7 +1515,7 @@ class JFTP
 	 *
 	 * @since   11.1
 	 */
-	private function _putCmd($cmd, $expectedResponse)
+	protected function _putCmd($cmd, $expectedResponse)
 	{
 
 		// Make sure we have a connection to the server
@@ -1543,7 +1543,7 @@ class JFTP
 	 *
 	 * @since   11.1
 	 */
-	private function _verifyResponse($expected)
+	protected function _verifyResponse($expected)
 	{
 
 		// Initialise variables.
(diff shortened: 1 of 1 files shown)


eddieajau:
Amy, would you be able to fix up the rest of the methods, adding public and also protected to those with the underscore prefix.  Thanks.

elkuku:
Actually that does not fix the warning (at least on my box)..
JObject::get($property, $default = null)
JFTP::get($local, $remote)

To fix the strict standards warning you would write:
JFTP::get($local, $remote = null)

But that won't make much sense...

IMHO the proper way would be renaming the method to something like JFTP::getFile(), 

BTW: Did you noticed the new error_reporting level in the backend config labeled "development" - that should help in hunting down those **errors** ;)

eddieajau:
Thanks Nikolai. I hadn't got around to doing a hard test yet.  Yes, I agree - the overriding of get here is just plain wrong so I think we need to add change it to getFile and then deprecate get now and remove in 12.1, plus do what we have to to remove the strict error.

Amy, you able to do that?

LouisLandry:
Why even have JFTP extend JObject?  Just remove that dependency and if there is anything it actually uses (I doubt it) ... move them into JFTP itself.  There is absolutely no reason to extend JObject there, we are on php5+ now.

eddieajau:
Good point ...

AmyStephen:
Also, added a zero for the missing microseconds parameter -- showed as an error in phpStorm.

socket_set_timeout($this->_conn, $this->_timeout, 0);

eddieajau:
Thanks Amy.  Couple of things are breaking the unit tests and backward compatibility.  Could you change the constructor and destructor to public just in case someone is using the class directly (ie, $ftp = new JFtp).  Could you also change the other private methods to protected otherwise it becomes difficult to test them.

For the socket_set_timeout method, I'm not sure why phpStorm is complaining about the microseconds argument. It's listed as optional in the PHP manual.  What error is it showing?

AmyStephen:
It certainly does appear to be optional.
http://www.php.net/manual/en/function.stream-set-timeout.php

Here's what phpStorm is saying:
http://img163.imageshack.us/img163/3919/screenshot20111018at104.png

Still in the patch - can't hurt, I suppose. I'll pull it out if you prefer.

eddieajau:
Bizarre.  Yeah, the zero can't hurt.  Unit test (little as they are) pass now.  Thanks.

eddieajau:
Amy, there was one checkstyle failure. We have a rule to have one new line at the end of the file.  I forgot to check that sorry. Could you add a new pull request to fix that minor issue - thanks.

AmyStephen:
I accidentally created a new pull request - not sure what I did. 

https://github.com/joomla/joomla-platform/pull/438

PR #439, opened 2011-10-19
Fix to the warnings call_user_func_array() throws

frt:
Fix to the warnings call_user_func_array() throws when  elements aren't refereces. This warnings happens with php-5.3.x.

### DIFF
--- a/libraries/joomla/event/event.php
+++ b/libraries/joomla/event/event.php
@@ -40,6 +40,12 @@ abstract class JEvent extends JObserver
 		$event = $args['event'];
 		unset($args['event']);
 
+		$temp = $args;
+		$args = array();
+		foreach ($temp as $k => $v) {
+			$args[] = &$temp[$k];
+		}
+
 		/*
 		 * If the method to handle an event exists, call it and return its return
 		 * value.  If it does not exist, return null.


eddieajau:
I run on PHP 5.3 and I don't get any warnings. Is this a strict mode problem?

frt:
The first note in http://br.php.net/manual/en/function.call-user-func-array.php is the best explanation to the problem I can think of.

As I interpreted it, a component expecting a reference will throw a warning. I my case the component was com_jcar http://extensions.joomla.org/extensions/living/education-a-culture/bibliography/10382, and it crashes without my fix.

My fix is similar to what is done in libraries/joomla/html/html.php, in function cal().

I think it is not a strict mode problem (I'm not at it now, but I'm pretty sure that my server isn't configured in strict mode), but a kind of component dependent problem.

HTH

WebMechanic:
@frt if you're on PHP 5.3.6 and have `track_errors = on`, the crash is due to a bug in 5.3.6

It is an `E_STRICT` issue but IMHO the origin of that symtom this patch is trying to fix are the plugins themself. 
They still use the PHP4ish `&$row`, `&$article`, `&$params` etc. to force "pass by reference" in their event handler signatures despite these very arguments are objects (stdClass, JRegistry) and PHP5 can handle object references perfectly fine.
So instead of creating a workaround for a **PHP4-issue** no longer present in PHP5, the function signatures in the plugins should become on par with PHP5.
Read: CMS issue :-)

WebMechanic:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26989

frt:
Agree with the PHP4ish thing. Will see if I can fix that in the component and send to the component developers. Thanks for the link with that patch.

eddieajau:
Ok, I'll close this issue and if it comes back to the platform in any way, let us know.

PR #444, opened 2011-10-19
Implement a getSources() method in JTable

chdemko:
In the JModelAdmin class, onContentBeforeSave and onContentAfterSave are triggered using the table as argument.

The plugins lacks the possibility to access to the source of the data to store additionnal information in external tables.

A simple getSources method implemented in JTable could provide this feature


eddieajau:
Christophe, can you give an example of a use case for this? I don't quite understand what problem is being solved.

Hackwar:
Why can't you use getProperties()?

chdemko:
We can imagine a plugin that injects some fields in a form and these fields have to be stored in an external table (mapping between articles for instance). When the onContentBeforeSave and onContentAfterSave are triggered, the plugin cannot access to these fields since they have been filtered by the bind method (it uses getProperties which returns only the declared fields table).

libraries/joomla/database/table.php
```diff
@@ -70,6 +70,14 @@ abstract class JTable extends JObject
 	protected $_locked = false;
 
 	/**
+	 * Source used when the bind method is called
+	 *
+	 * @var    array|object|null
+	 * @since  11.3
+	 */
+	protected $_source = null;
+
```
elkuku, line 79: Maybe not start a protected var with an underscore to prevent additional code style warnings ;)

PR #445, opened 2011-10-19
Adding the possibility to delay the output of JGrid

chdemko:
Currently, the output of JGrid is constructed using the setRowCell method either by replacing the content of cell, either by concatenating a new content.

It could be interesting to delay the output using this kind of call:

i.e.
$table->setRowCell('mycolumn', array('grid.boolean', $i));

When the
echo $table
will be done, the class will call at this time JHtml::_('grid.boolean', $i). This will allow template designers to override the output of 'grid.boolean'.

### DIFF
--- a/libraries/joomla/html/grid.php
+++ b/libraries/joomla/html/grid.php
@@ -272,10 +272,12 @@ class JGrid
 	 * Set cell content for a specific column for the
 	 * currently active row
 	 *
-	 * @param   string  $name     Name of the column
-	 * @param   string  $content  Content for the cell
-	 * @param   array   $option   Associative array of attributes for the td-element
-	 * @param   bool    $replace  If false, the content is appended to the current content of the cell
+	 * @param   string               $name     Name of the column
+	 * @param   string|object|array  $content  Content for the cell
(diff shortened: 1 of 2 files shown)


Hackwar:
I would vote against this feature. I thought about this feature, too, but decided against it, because I don't think that you could forsee all the ways people will want to use this. One specific issue that I see is, that we could have two plugins trying to modify the same cell after JView filled it with a function callback. The result would be that the first function would have to get the cell content, execute the function, replace whatever it wants to replace in the output, write that back into the cell and the second plugin first has to check if it has to act upon a callback or a string, which will raise confusion.

The last question for me would be, what we are gaining by this. Delaying the function call only does just that. I don't see how it will give us an advantage in the later process.

robschley:
Why not support actual callbacks? What is the purpose of having these pseudo callbacks?

robschley:
Sorry, accidentally closed and reopened.

eddieajau:
I'm going to close this issue and ask that it be discussed on list. If a consensus can be reached, re-pull. It's nothing against the feature, just the lack of suitable workflow tools on GitHub.

PR #446, opened 2011-10-19
JHtmlNumber and tests fail on 32 bit systems

### LINKED ISSUE #443: JHtmlNumber and tests fail on 32 bit systems
Unit tests were recently committed for this class.  The tests demonstrate that on 32 bit systems the method does not work because the byte total overflows.

rvsjoen:
By removing the explicit cast we allow PHP to automatically convert to float when needed in order to avoid overflowing on 32-bit systems.

Fixes #443

### DIFF
--- a/libraries/joomla/html/html/number.php
+++ b/libraries/joomla/html/html/number.php
@@ -35,6 +35,8 @@ abstract class JHtmlNumber
 	 */
 	public static function bytes($bytes, $unit = 'auto', $precision = 2)
 	{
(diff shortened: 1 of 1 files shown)


eddieajau:
Rune, would you mind adding a comment about this bug so that we don't accidentally put the cast back in sometime in the future?  Thanks.

rvsjoen:
Done

eddieajau:
Awesome. Thanks.

PR #447, opened 2011-10-19
Adding test skeletons; Work on JFormField tests

mbabker:
Working on test suite by:
- Adding a handful of skeletons
- Converting JFormField tests to database test cases to allow them to run properly
- Adding test cases for several JFormFields

By my numbers, this adds 17 classes to the test coverage report.


LouisLandry:
It doesn't add to the coverage unless you actually test the class.  And the way it measures that is whether or not the code in the class was executed.  Just adding skeletons is helpful, but it doesn't improve test coverage.  I'll take a look at testing this soon and we can see about pulling it in.  Thanks!

mbabker:
As I look over the test suite, I'm doing what I can with what I know and have looked at elsewhere to actually write tests or implement tests that may not be reporting accurately, such as with these JFormField tests.  In cases where I'm finding no tests or skeletons for the classes, I'm generating skeletons so the class will report test coverage (even at 0%), and for me, that helps to figure out where I should devote some time on the tests.  Who knows, maybe it'll convince others to join in too.

PR #448, opened 2011-10-19
Joomla CMS [#22103] Update /libraries/joomla/form/formfield.php

benharold:
Joomla CMS [#22103] JFormField::getID strips valid characters from id attributes
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=22103

### DIFF
--- a/libraries/joomla/form/formfield.php
+++ b/libraries/joomla/form/formfield.php
@@ -408,7 +408,7 @@ abstract class JFormField
 		}
 
 		// Clean up any invalid characters.
-		$id = preg_replace('#\W#', '_', $id);
+		$id = preg_replace('#[^\w:\.\-_]#', '_', $id);
 
 		return $id;
 	}


eddieajau:
Branch won't merge cleanly.  Needs to be updated.

realityking:
Merging this manually wouldn't be a big issue, but someone else should decide on the merits of this change. I just don't speak regex.

jonnsl:
I think that this is more correct

```
$id = preg_replace('/(^[^a-z]|[^\w:._-])/i', '_', $id);
```

chdemko:
@benharold We cannot authorise the . (dot) character and : (semi-column) in id: it will forbid css styling

if id is 'abc.def-ghi_jkl:hover'

the css rule would be

```
#abc.def-ghi_jkl:hover
```

which does not correspond to an id="abc.def-ghi_jkl:hover" but corresponds to id="abc" class="def-ghi_jkl" and the mouse over the element

jonnsl:
@chdemko
see http://www.w3.org/TR/html401/types.html#type-name
HTML5 is even more permissive http://mathiasbynens.be/notes/html5-id-class

chdemko:
I know that but it forbids css styling

jonnsl:
You just need to escape the special characters http://mathiasbynens.be/notes/css-escapes
http://jsfiddle.net/6262B/
http://mathiasbynens.be/demo/html5-id

chdemko:
Is it fully compatible with all browsers?

benharold:
Since, as Jonathan mentioned, the HTML5 spec is so widely permissive of different characters in the id tag, I would suggest that filtering the id tag has become an antiquated practice, and should be dropped altogether.

Ben

LouisLandry:

PR #450, opened 2011-10-20
Update to remove MD5 from the session id creation funtion

ghost:
Formatting the session id for storage should be the responsibility of the underlying storage engine, not the session manager. Applying MD5[or any other form of hashing] has the possibility of creating the same hash for different id's - thus granting one user access to another's account. Maintain uniqueness of id as long as possible by not hashing it, and let engines format them further if they need to and note the security risk,

### DIFF
--- a/libraries/joomla/session/session.php
+++ b/libraries/joomla/session/session.php
@@ -646,7 +646,7 @@ class JSession extends JObject
 			$id .= mt_rand(0, mt_getrandmax());
 		}
 
-		$id = md5(uniqid($id, true));
+		$id = uniqid($id, true);
 		return $id;
 	}
 
--- a/libraries/joomla/session/storage/database.php
+++ b/libraries/joomla/session/storage/database.php
@@ -113,17 +113,22 @@ class JSessionStorageDatabase extends JSessionStorage
 
 		if ($db->getAffectedRows())
 		{
+
 			return true;
 		}
 		else
 		{
-			// If the session does not exist, we need to insert the session.
-			$db->setQuery(
-				'INSERT INTO ' . $db->quoteName('#__session') .
-				' (' . $db->quoteName('session_id') . ', ' . $db->quoteName('data') . ', ' . $db->quoteName('time') . ')' .
-				' VALUES (' . $db->quote($id) . ', ' . $db->quote($data) . ', ' . (int) time() . ')'
-			);
-			return (boolean) $db->query();
+            // update did not error, so it succeeded but was identical
+            if (JDEBUG)
+            {
+               jimport('joomla.error.log');
+               $log = JLog::getInstance('jcontroller.log.php')->addEntry(
+                array(
+                   	'comment' => sprintf('Session ID %s update is identical to current record', $id)
+                )
+             );
+             }
+            return true;
 		}
 	}
 


Hackwar:
looking at uniqid() on php.net, I can't see any description how long a unique ID generated by this function is. However, I would disagree with your statement, that using an MD5 hash on the result of uniqid() reduces entropy. The few generated IDs that were in the comments of uniqid(), were pretty short, which makes it not very probable, that there would ever be a collision between MD5 hashes of two uniqid() results.

I would grade the benefits of having a fixed-length session ID over the tiny possibility of a hash collision of two calls to uniqid(). We are speaking of maybe even millions of simultaneously open sessions...

eddieajau:
I agree with the premise, but if you are talking the md5 out of JSession, you need to put it back into the storage classes.

ghost:
It depends on the storage class.  Through the use of mt_rand, id is initialized to a string of all numbers which is at least 32 charectors long and at most 31+mt_getrandmax() long.  The default on getrandmax is most likely 2147483647, so thats an upper limit of 41 charectors.  http://php.net/manual/en/function.mt-getrandmax.php and uniqid gives us a string of alphanumeric chars 23 long..so the max length is 64 characters an minimum is 54.

md5 hashing it results in a 32 byte long string.  The benefit of doing this being:
1) Fixed length to fit for being used as a key
2) For storage systems that return data when substrings match, ie when 12345 if it does not exist will retrieve data for key 1234567890

So it is storage system dependent on whether or not it must be md5'd.

I do not believe that is a problem with any of the current storage engines, database, none, xcache, memcache, and apc can all support the larger key size.

While it is traditional to md5 session keys, there is no reason it must be done...[ideally, this should be a user configurable option so users may choose md5, sha1, or whatever.  However, at the moment I'm just concerned with minimizing duplicate key generation possibilities.]

ghost:
Yes, it is rather unlikely that you will end up with duplicate keys.  However, there are a number of sites today that have very real issues with getting load failures caused by failed insert's of session id's - so unlikely or not it is happening.

I honestly don't think MD5 is the cause, but my preference is to avoid it even being an issue.  Short of generating all combinations of 31-41 long numbers combined with all combinations of 23 alphanumeric charectors, and then applying md5 to check - I figured I'd just submit this small pull request.

You can review the discussion here: https://groups.google.com/forum/#!topic/joomlabugsquad/miigEuee-uU on this problem.  I also figured I'd break item's 1 and 2-3 into separate pull requests since 2-3 solves the real most likely culprit while 1 is just for the sake of completeness.

eddieajau:
Mmm, ok, but I think the bigger problem is we have no code coverage to detect a regression. I don't "think" there would be a problem, but there's no way to prove that at present.

elkuku:
**Checkstyle error details:**
libraries/joomla/session/storage/database.php:121 - 130
Tabs must be used to indent lines; spaces are not allowed

libraries/joomla/session/storage/database.php:130
Closing brace indented incorrectly; expected 12 spaces, found 13

libraries/joomla/session/storage/database.php:131
Tabs must be used to indent lines; spaces are not allowed

ghost:
erg...fudge. I'll need to create a Joomla coding standard profile for my IDE.  I prefer spaces, but if we use tabs I'll swap.

PR #453, opened 2011-10-21
Support for overriding command and format in JController::getInstance

pasamio:
Change to JController to permit more flexibility around where it sets some values from the request and instead provide the ability to set it in the options given to the JController::getInstance call. This enables the ability to create a sub-controller with parameters distinct from the request.

### DIFF
--- a/libraries/joomla/application/component/controller.php
+++ b/libraries/joomla/application/component/controller.php
@@ -231,8 +231,8 @@ class JController extends JObject
 
 		// Get the environment configuration.
 		$basePath = array_key_exists('base_path', $config) ? $config['base_path'] : JPATH_COMPONENT;
-		$format = JRequest::getWord('format');
-		$command = JRequest::getVar('task', 'display');
+		$format = isset($config['format']) ? $config['format'] : JRequest::getWord('format');
+		$command = isset($config['command']) ? $config['command'] : JRequest::getVar('task', 'display');
 
 		// Check for array format.
 		$filter = JFilterInput::getInstance();


libraries/joomla/application/component/controller.php
```diff
@@ -231,8 +231,8 @@ class JController extends JObject
 
 		// Get the environment configuration.
 		$basePath = array_key_exists('base_path', $config) ? $config['base_path'] : JPATH_COMPONENT;
-		$format = JRequest::getWord('format');
-		$command = JRequest::getVar('task', 'display');
+		$format = isset($config['format']) ? $config['format'] : JRequest::getWord('format');
+		$command = isset($config['command']) ? $config['command'] : JRequest::getVar('task', 'display');
```
chdemko, line 236: Since JRequest will be deprecated is 12.1, could it be possible to replace this code by

$input = JFactory::getApplication()->input;
$retreive =  isset($config['retreive']) ? $config['retreive'] : 'request'
$format = isset($config['format']) ? $config['format'] : $input->$retreive->get('format', null, 'word');
$command = isset($config['command']) ? $config['command'] : $input->$retreive->get('task', 'display');

And change line 260
JRequest::setVar('task', $task);
to
$input->$retreive->set('task', $task);

Same code has to be changed in display function

ianmacl:
One possibility would be to allow you to provide a JInput object to JController::getInstance which would allow you to inject all your input into the controller.  This would make it easy to make reusable controllers.

eddieajau:
Another possibility is to simply JController::getInstance and move most of its smarts into a controller/MVC factory. That's probably the best way to handle it.

PR #460, opened 2011-10-23
Fixed issue with JFormHelper:loadClass() including only first filename match

### LINKED ISSUE #455: loadElement only includes first matched element file. Problem with multiple elements with same main name.
Sorry for the long explanation here, but it probably makes the problem clear enough:

When having admin views like the module edit page, you have the JForm elements on the right. These are different fields.
The different xml files that are used to create these (some of module itself, some of core, etc) can have an addpath to include extra field files.

So far so good. But, if there are field files with the same file name, things go haywire...

Now there is a possibility to call on these field types with prefixes, to prevent duplicate class names.
Like in my case I have a type="nn_categories"
This will act on the class JFormFieldNN_Categories in the file categories.php in the addpath location. Note, it doesn't look for nn_categories.php but for categories.php

So my thought is that the function that loads the files from the different addpaths shouldn't stop at the first filename that matches (categories.php), but should look for the file that actually has the correct classname (JFormFieldCategories vs JFormFieldNN_Categories vs JFormFieldXXX_Categories)

So there is a problem when 2 extensions (plugins/whatever) are involved in creating the form.
Each with their own addpath and with their own extra field files.
So the first could use type="a_categories" and the second uses type="b_categories".

However, Joomla only loads the first categories.php. So only the first JFormFieldA_Categories is initialized. And not the second  JFormFieldB_Categories - causing all b_categories fields to default to a default text field.

Essentially I think the problem is in file: \libraries\joomla\html\parameter.php in the loadElement method.

Instead of looping through all dirs and including all files that match (maybe double checking for duplicate classnames), it simply does a:
JPath::find($dirs, $file)
Which returns the path of the first file that matches....

It should do a while loop through the dirs, checking if the required class exists in the condition... Or whatever better solution there is...

ghost:
Now it includes all matched files till it finds the required classname. Fixes #455

### DIFF
--- a/libraries/joomla/form/helper.php
+++ b/libraries/joomla/form/helper.php
@@ -198,8 +198,8 @@ class JFormHelper
 			$type = substr($type, $pos + 1);
 		}
 
-        // Try to find the class file.
-        $type = strtolower($type).'.php';
+		// Try to find the class file.
+		$type = strtolower($type).'.php';
 		foreach($paths as $path)
 		{
 			if ($file = JPath::find($path, $type))
--- a/libraries/joomla/form/helper.php
+++ b/libraries/joomla/form/helper.php
@@ -199,8 +199,8 @@ class JFormHelper
 		}
 
 		// Try to find the class file.
-		$type = strtolower($type).'.php';
-		foreach($paths as $path)
+		$type = strtolower($type) . '.php';
+		foreach ($paths as $path)
 		{
 			if ($file = JPath::find($path, $type))
 			{
--- a/libraries/joomla/form/helper.php
+++ b/libraries/joomla/form/helper.php
@@ -198,11 +198,19 @@ class JFormHelper
 			$type = substr($type, $pos + 1);
 		}
 
-		// Try to find the class file.
-		if ($file = JPath::find($paths, strtolower($type) . '.php'))
+        // Try to find the class file.
+        $type = strtolower($type).'.php';
+        foreach($paths as $path)
 		{
-			include_once $file;
-		}
+            if ($file = JPath::find($path, $type))
+			{
+                require_once $file;
+                if (class_exists($class))
+				{
+                    break;
+                }
+            }
+        }
 
 		// Check for all if the class exists.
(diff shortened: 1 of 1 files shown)


eddieajau:
Peter, I think all we need is the unit test to validate the changes.  Thanks in advance.

ghost:
Great. Just to add:
This is just a quickfix till this whole thing is handled in a better lower-level way as mentioned by Louis (using JLoader).
So it might be good to add something along those lines in the comment near this code block...

ghost:
Any info on what the errors are? Anything I can/should do?

elinw:
Peter do you have phpunit installed? You can run the locally. There are instrutions on the wiki and here
https://github.com/joomla/joomla-platform/blob/staging/docs/manual/appendices/analysis.xml.

ghost:
No, I haven't. Pretty new to this. So I'll look into it...

ghost:
So all is ok?

The Checkstyle warnings/errors, does that concern these changes? Can't imagine these few lines generating 453 warnings :S

elkuku:
The warnings come from all over the place - the errors shouldn't ;)

**Checkstyle error details:**
libraries/joomla/form/helper.php:202
libraries/joomla/form/helper.php:203
Expected "foreach (...)\n...{\n"; found "foreach(...)\n...{\n"

ghost:
Sorry, I don't understand. Is everyone using different coding styles?

elinw:
The 453 are the same ones everyone has and have nothing to do with your code. But if the numbers go up it's due to your code.

ghost:
Yes, but what about the 29 errors?
"Checkstyle analysis reported 453 warnings and 29 errors."

elkuku:
I believe, that's a mistake. From my POV your code is OK
Please be patient ;)

ghost:
Ok, great. I am patient. Just worried I am doing stuff wrong. :)

LouisLandry:
Best I can tell this is all good.  Going to merge it now.  Thanks Peter.

PR #461, opened 2011-10-23
Joomla CMS [#26419] Fix access view levels html field to only show levels that the user is authorized for

jmcameron:
Update html access field to only list view levels that the user has access to (except that any super-user can still still them all).

See this original CMN list issue:  http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26419

See https://github.com/joomla/joomla-platform/pull/461


jmcameron:
I discussed this with Elin here at Joomla Day NYC Dev Day and she suggested I submit this patch/pull request.  

Note that super-users will see no change.  But any other user will only see the access levels they are authorized for.  This will probably surprise some authors but it seems like the right thing to do.    Super-user will maintain ability to set any view-level to any content (that uses the 'access' html field).

If accepted and it makes its way to a Joomla CMS release, I'm am willing to write a explanatory paragraph for the release notes.

elkuku:
**Checkstyle error details:**
`libraries/joomla/html/html/access.php:58`
`libraries/joomla/html/html/access.php:58`

jmcameron:
Thanks Nikolai.  I'll updated it tonight.  It was helpful to see the exact reason for the complaints.
-Jonathan

jmcameron:
Strange.  I added spaces around the concat operator and it went from 4 checkstyle warnings to 29?

eddieajau:
Jonathan, can you check the tests on this one please?  Thanks.

libraries/joomla/html/html/access.php
```diff
@@ -46,9 +46,18 @@ public static function level($name, $selected, $attribs = '', $params = true, $i
...
+
 		$query->select('a.id AS value, a.title AS text');
 		$query->from('#__viewlevels AS a');
 		$query->group('a.id');
+		if (!$user->authorise('core.admin'))
+		{
+			// Users that are not super-users can ONLY see the
+			// the view levels that they are authorized for
+			$access_levels = array_unique($user->authorisedLevels());
+			$query->where('a.id in (' . implode(',', $access_levels) . ')');
```
chdemko: You use array_unique here because there is an issue in JAccess::getAuthorisedViewLevels (it lacks a test in the foreach loop:). Here is the right code:

```
    // Find the authorised levels.
    foreach (self::$viewLevels as $level => $rule)
    {
        if ($level != 1)
        {
            foreach ($rule as $id)
            {
                if (($id < 0) && (($id * -1) == $userId))
                {
                    $authorised[] = $level;
                    break;
                }
                // Check to see if the group is mapped to the level.
                elseif (($id >= 0) && in_array($id, $groups))
                {
                    $authorised[] = $level;
                    break;
                }
            }
        }
    }
```

Could you change the code in JAccess and remove your array_unique (this will rerun the test by the way)

chdemko:

chdemko:
@jmcameron The pull request cannot be merged

jmcameron:
I am updating this after catching up to the latest version of the platform.

jmcameron:
I am going to close this pull request.  It needs more testing.  And the platform unit tests need to be updated since the current tests will break until more support is added in the platform unit tests for using JFactory::getUser() to get the current user object.  Once these things are done and the tests pass, I'll start another pull request.

PR #462, opened 2011-10-24
Fix spelling in docblocks and removed @return labels in constructors.

mbabker:
Using the code inspector in PhpStorm, evaluated the following conditions from the results:
- Spelling errors (specifically in code comments and doc blocks)
- A method declared void should not return anything
- Doc block doesn't match method declaration

There was also an instance of DatabaseException that I saw while looking through the results that I switched to JDatabaseException.

### DIFF
--- a/libraries/joomla/database/database.php
+++ b/libraries/joomla/database/database.php
@@ -314,7 +314,7 @@ abstract class JDatabase implements JDatabaseInterface
 			{
 				$instance = new $class($options);
 			}
-			catch (DatabaseException $e)
+			catch (JDatabaseException $e)
 			{
 
 				// Legacy error handling switch based on the JError::$legacy switch.
@@ -432,7 +432,7 @@ abstract class JDatabase implements JDatabaseInterface
 	 *
 	 * @param   array  $options  List of options used to configure the connection
 	 *
-	 * @return  void
+	 * @return  JDatabase
 	 *
 	 * @since   11.1
 	 */
@@ -1022,7 +1022,7 @@ abstract class JDatabase implements JDatabaseInterface
 	 * Method to get an array of the result set rows from the database query where each row is an object.  The array
 	 * of objects can optionally be keyed by a field name, but defaults to a sequential numeric array.
 	 *
-	 * NOTE: Chosing to key the result array by a non-unique field name can result in unwanted
+	 * NOTE: Choosing to key the result array by a non-unique field name can result in unwanted
 	 * behavior and should be avoided.
 	 *
 	 * @param   string  $key    The name of a field on which to key the result array.
@@ -1130,7 +1130,7 @@ abstract class JDatabase implements JDatabaseInterface
 	 * Method to get an array of the result set rows from the database query where each row is an array.  The array
 	 * of objects can optionally be keyed by a field offset, but defaults to a sequential numeric array.
 	 *
-	 * NOTE: Chosing to key the result array by a non-unique field can result in unwanted
+	 * NOTE: Choosing to key the result array by a non-unique field can result in unwanted
 	 * behavior and should be avoided.
 	 *
 	 * @param   string  $key  The name of a field on which to key the result array.
--- a/libraries/joomla/database/database.php
+++ b/libraries/joomla/database/database.php
@@ -432,8 +432,6 @@ abstract class JDatabase implements JDatabaseInterface
 	 *
 	 * @param   array  $options  List of options used to configure the connection
 	 *
-	 * @return  JDatabase
-	 *
 	 * @since   11.1
 	 */
 	protected function __construct($options)
--- a/libraries/joomla/database/database/mysql.php
+++ b/libraries/joomla/database/database/mysql.php
@@ -56,8 +56,6 @@ class JDatabaseMySQL extends JDatabase
 	 *
 	 * @param   array  $options  Array of database options with keys: host, user, password, database, select.
 	 *
-	 * @return  JDatabaseMySQL
-	 *
 	 * @since   11.1
 	 */
 	protected function __construct($options)
--- a/libraries/joomla/database/database/mysql.php
+++ b/libraries/joomla/database/database/mysql.php
@@ -56,7 +56,7 @@ class JDatabaseMySQL extends JDatabase
 	 *
 	 * @param   array  $options  Array of database options with keys: host, user, password, database, select.
 	 *
-	 * @return  void
+	 * @return  JDatabaseMySQL
 	 *
 	 * @since   11.1
 	 */
@@ -389,7 +389,7 @@ class JDatabaseMySQL extends JDatabase
 	 *
 	 * @param   string  $table  The name of the table.
 	 *
-	 * @return  array  An arry of the column specification for the table.
+	 * @return  array  An array of the column specification for the table.
 	 *
 	 * @since   11.1
 	 * @throws  JDatabaseException
--- a/libraries/joomla/database/database/mysqli.php
+++ b/libraries/joomla/database/database/mysqli.php
@@ -56,8 +56,6 @@ class JDatabaseMySQLi extends JDatabase
 	 *
 	 * @param   array  $options  List of options used to configure the connection
 	 *
-	 * @return  JDatabaseMySQLi
-	 *
 	 * @since   11.1
 	 */
 	protected function __construct($options)
--- a/libraries/joomla/database/database/mysqli.php
+++ b/libraries/joomla/database/database/mysqli.php
@@ -56,7 +56,7 @@ class JDatabaseMySQLi extends JDatabase
 	 *
 	 * @param   array  $options  List of options used to configure the connection
 	 *
-	 * @return  void
+	 * @return  JDatabaseMySQLi
 	 *
 	 * @since   11.1
 	 */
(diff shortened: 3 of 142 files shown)


elkuku:
Quite impressive :)
What I love most are the changes of `@return void` to the proper classes - two thumbs up ;)

libraries/joomla/database/database.php
```diff
@@ -432,7 +432,7 @@ abstract class JDatabase implements JDatabaseInterface
 	 *
 	 * @param   array  $options  List of options used to configure the connection
 	 *
-	 * @return  void
+	 * @return  JDatabase
```
eddieajau: Based on previous cases, this should be void (it's the __construct method).

libraries/joomla/database/database/mysql.php
```diff
@@ -56,7 +56,7 @@ class JDatabaseMySQL extends JDatabase
 	 *
 	 * @param   array  $options  Array of database options with keys: host, user, password, database, select.
 	 *
-	 * @return  void
+	 * @return  JDatabaseMySQL
```
eddieajau: Should be return void

libraries/joomla/database/database/mysqli.php
```diff
@@ -56,7 +56,7 @@ class JDatabaseMySQLi extends JDatabase
 	 *
 	 * @param   array  $options  List of options used to configure the connection
 	 *
-	 * @return  void
+	 * @return  JDatabaseMySQLi
```
eddieajau: Should be return void

libraries/joomla/database/database/sqlsrv.php
```diff
@@ -69,7 +69,7 @@ class JDatabaseSQLSrv extends JDatabase
 	 *
 	 * @param   array  $options  List of options used to configure the connection
 	 *
-	 * @return  void
+	 * @return  JDatabaseSQLSrv
```
eddieajau: Should be return void

libraries/joomla/environment/browser.php
```diff
@@ -193,7 +193,7 @@ class JBrowser extends JObject
 	 * @param   string  $userAgent  The browser string to parse.
 	 * @param   string  $accept     The HTTP_ACCEPT settings to use.
 	 *
-	 * @return  void
+	 * @return  JBrowser
```
eddieajau: Should be return void

libraries/joomla/filesystem/archive/bzip2.php
```diff
@@ -31,7 +31,7 @@ class JArchiveBzip2 extends JObject
 	/**
 	 * Constructor tries to load the bz2 extension if not loaded
 	 *
-	 * @return  void
+	 * @return  JArchiveBzip2
```
eddieajau: Should be return void

libraries/joomla/form/form.php
```diff
@@ -78,7 +78,7 @@ class JForm
 	 * @param   string  $name     The name of the form.
 	 * @param   array   $options  An array of form options.
 	 *
-	 * @return  void
+	 * @return  JForm
```
eddieajau: Should be return void

libraries/joomla/html/pagination.php
```diff
@@ -66,7 +66,7 @@ class JPagination extends JObject
 	 * @param   integer  $limit       The number of items to display per page.
 	 * @param   string   $prefix      The prefix used for request variables.
 	 *
-	 * @return  void
+	 * @return  JPagination
```
eddieajau: Should be return void

libraries/joomla/html/pagination.php
```diff
@@ -743,7 +743,7 @@ class JPaginationObject extends JObject
 	 * @param   integer  $base    The number of rows as a base offset.
 	 * @param   string   $link    The link URL.
 	 *
-	 * @return  void
+	 * @return  JPaginationObject
```
eddieajau: Should be return void

libraries/joomla/log/logentry.php
```diff
@@ -77,7 +77,7 @@ class JLogEntry
 	 * @param   string  $category  Type of entry
 	 * @param   string  $date      Date of entry (defaults to now if not specified or blank)
 	 *
-	 * @return  void
+	 * @return  JLogEntry
```
eddieajau: Should be return void

libraries/joomla/log/logger.php
```diff
@@ -35,7 +35,7 @@ abstract class JLogger
 	 *
 	 * @param   array  &$options  Log object options.
 	 *
-	 * @return  void
+	 * @return  JLogger
```
eddieajau: Should be return void

libraries/joomla/log/loggers/database.php
```diff
@@ -73,7 +73,7 @@ class JLoggerDatabase extends JLogger
 	 *
 	 * @param   array  &$options  Log object options.
 	 *
-	 * @return  void
+	 * @return  JLoggerDatabase
```
eddieajau: Should be return void

libraries/joomla/log/loggers/formattedtext.php
```diff
@@ -69,7 +69,7 @@ class JLoggerFormattedText extends JLogger
 	 *
 	 * @param   array  &$options  Log object options.
 	 *
-	 * @return  void
+	 * @return  JLoggerFormattedText
```
eddieajau: Should be return void

libraries/joomla/log/loggers/syslog.php
```diff
@@ -45,7 +45,7 @@ class JLoggerSysLog extends JLogger
 	 *
 	 * @param   array  &$options  Log object options.
 	 *
-	 * @return  void
+	 * @return  JLoggerSysLog
```
eddieajau: Should be return void

mbabker:
So what's "correct"?  There's no consistency right now with the constructor `@return` statements.  In the couple dozen I've just looked at, there's a mix between void, the class name, and no return statement at all, and it seems to lean closer to being a majority class names.

libraries/joomla/log/loggers/w3c.php
```diff
@@ -39,7 +39,7 @@ class JLoggerW3C extends JLoggerFormattedText
 	 *
 	 * @param   array  &$options  Log object options.
 	 *
-	 * @return  void
+	 * @return  JLoggerW3C
```
eddieajau: Should be return void

eddieajau:
Yeah, that's mostly my fault.  The method strictly does not return anything so it should be void.

libraries/joomla/registry/registry.php
```diff
@@ -34,7 +34,7 @@ class JRegistry
 	 *
 	 * @param   mixed  $data  The data to bind to the new JRegistry object.
 	 *
-	 * @return  void
+	 * @return  JRegistry
```
eddieajau: Should be return void

libraries/joomla/utilities/date.php
```diff
@@ -82,7 +82,7 @@ class JDate extends DateTime
 	 * @param   string  $date  String in a format accepted by strtotime(), defaults to "now".
 	 * @param   mixed   $tz    Time zone to be used for the date.
 	 *
-	 * @return  void
+	 * @return  JDate
```
eddieajau: Should be return void

libraries/joomla/utilities/simplecrypt.php
```diff
@@ -31,7 +31,7 @@ class JSimpleCrypt extends JObject
 	 *
 	 * @param   string  $key  Optional encryption key
 	 *
-	 * @return  void
+	 * @return  JSimpleCrypt
```
eddieajau: Should be return void

mbabker:
I think an argument can be made for making the `@return` tags consistent one way or another here.  Personally, I'm in favor of either using the class names or not including a `@return` tag on constructors.

Running phpDocumentor locally on my own component using `@return void`, `@return ClassName`, and no `@return` statement produces different results in the output (this is pretty much the default, so unless the API's implementation has been tweaked in such a way to do something different, it should apply).

`void __construct( [array $config = array()])` - constructor with `@return void`
`PodcastManagerTableFeed __construct( object &$db)` - constructor with `@return PodcastManagerTableFeed` and class name linked to self
`PodcastManagerModelFeeds __construct( [array $config = array()])` - constructor with no `@return` statement

Though the method does not explicitly return something, it does create an instance of a given class, and it seems `@return void` isn't the correct answer for the constructors.

eddieajau:
Ok, scouting around I think you are right.  Seems the consensus is to **not** use `@return` in constructors - I'm fine with that.

elinw:
This was something i was struggling with as I went through because there was so much inconsistency but I believe it is really not quite proper to have a constructor return something. You don't say $something = JSimpleCrypt so it's not exactly a return in the same way that a regular method is.

mbabker:
I'll handle the `@return` statements tonight and tomorrow.  Question - Will this require a change to the code sniffer, and if so, do you think you could handle it @elkuku?

elkuku:
I didn't even noticed that the `@return` statements belong to constructors - bad github diff... ;(
So.. I would assume that the same goes also for destructors ? (just in case)
That would be something like:
- **Standard**: Constructor and destructor comments must not have a `@return` tag.

If this is OK I can modify the corresponding sniff.

mbabker:
Request updated removing the `@return` statements.

eddieajau:
Awesome work Michael. Really appreciate it.

PR #464, opened 2011-10-25
Coding standards: Constructor and destructor comments must not have a @return tag.

elkuku:
This was ment to be added to #462... You are merging too fast ;)
- Removes the `@return` tags from class destructors.
- Fixes the function comment sniff to detect a `@return` tag in class constructors and destructors docblocks.


elinw:
+++++

PR #466, opened 2011-10-26
Adding missing Unittest skeletons for Joomla classes

Hackwar:
This adds 135 unittest skeletons that are currently missing from the unittests.


Hackwar:
Thanks Michael for your work. Merged that into this.

Hackwar:
I'm running the unittests on Windows with Apache and I get a whole range of errors. The last two commits should fix a large range of those.

eddieajau:
I think this is fine but given the volume of changes, I won't be able to get to it until well after the weekend (beside work, have a JDay to prepare for).

elinw:
Might be easier to break this into 3-5 pieces rather than having to have it all 100% perfect.

Hackwar:
This last commit fixes a bug in the JCacheStorage unittests. The setup method is called only after the casesGetinstance method was called, marking all storage engines as not available.

chdemko:
Running tests on my linux laptop:

PHP 5.3.8 (cli) (built: Sep 28 2011 17:14:01) 
Copyright (c) 1997-2011 The PHP Group
Zend Engine v2.3.0, Copyright (c) 1998-2011 Zend Technologies
    with Xdebug v2.1.2, Copyright (c) 2002-2011, by Derick Rethans

mysql  Ver 14.14 Distrib 5.5.14, for Linux (x86_64) using readline 5.1

There were 2 failures:

1) JCacheTest::testStoreAndGet with data set "simple" ('output', array(600, 'file'), 42, '', 'And this is the cache that tries men\'s souls', false)
Should store the data properly
Failed asserting that boolean:false is true.

2) JCliTest::test__construct
Dispatcher property wrong type
Failed asserting that string:foo is an instance of class "JDispatcher".

FAILURES!
Tests: 2350, Assertions: 10741, Failures: 2, Incomplete: 544, Skipped: 75.

Hackwar:
Our unittests are not very stable and failures differ from system to system. You have two failures, I have 20 on my system in this branch. On staging I have more than 30. On the other hand, the unittests run to the end on my system, while joomla-jenkins seems to have issues where unittests are failing completely. I think it would be good to have some coding guidelines for unittests, for example that all included classes are handled by JLoader where possible and that setup() is used for creating an object in $this->object or $this->o.

Regarding the two tests mentioned by you: These two tests are not touched by this pull request. They are already broken in the staging branch.

ianmacl:
There are some instabilities in the unit tests, but generally on a properly configured system they all pass.  The 20 on your system would suggest to me that your database hasn't been properly configured.  That's what I would check.

There are a couple of cache tests that seem to consistently break the first time you run the tests.  This seems to be an initialization issue because of you run the unit tests again they pass.  There are significant issues with the caching system.  There is a lot of logic that has been stuffed in there that probably more properly belongs elsewhere.  When that gets rewritten we'll get that sorted out.

The unit tests consistently run fine for most branches.  I haven't had time to look more closely at what his happening here.  It isn't just on the build server that they fatal - they fatal on my local machine as well.

In general, with unit tests, you want to avoid dependencies on other parts of the library.  I would be against a standard that enforced loading classes using JLoader because of this.  I don't think you can necessarily define a standard here in any event because not all classes are the same and they don't always get loaded the same way.  I haven't thought that deeply on it though, so maybe there is a reasonable standard that would deal with the general case (though we couldn't enforce it).

Hackwar:
The 20 issues on my system are related to JCacheStorage, JDatabaseMySQLExporter, JForm and JDatabaseQuery, mostly toString() stuff and path issues.

I would agree that the dependencies should be reduced as much as possible. When you posted about the fatal errors of these unittests on your system, it looked to me like some files were included twice, so my proposal towards standardizing on JLoader was an attempt to prevent loading a file more than once. It would be one solution to that problem, most likely not the best one.

Hackwar:
Ok, JFormFieldEmailTest.php and JFormFieldEMailTest.php are two different files on Linux. ;-) Deleted the first one, lets see if the unittests work again.

ianmacl:
PHP Fatal error:  Class 'JSessionStorage' not found in /home/ian/www/platform/myplatform/libraries/joomla/session/storage/eaccelerator.php on line 21

Hackwar:
I've installed Ubuntu with a LAMP stack and phpunit and I can't reproduce a fatal error that joomla-jenkins is seeming to have with this branch.

Hackwar:
Closing this pull request. I'm going to split it up into a series of smaller ones to make this more manageable.

PR #468, opened 2011-10-27
Fix JDatabase::getConnectors and add a databaseconnection field

elinw:
This addresses some problems in getConnectors which made it not work with the current file structure in the database folder and with the newer naming conventions for the database class. 
It also adds a new field that uses getConnectors to dynamically create a list of available database connection/driver combinations optionally constrained to a list of supported databases for the application.

This will be important for multidatabase support since not all servers will support all databases and not all applications will support all available drivers.

### DIFF
--- a/libraries/joomla/database/database.php
+++ b/libraries/joomla/database/database.php
@@ -173,26 +173,28 @@ abstract class JDatabase implements JDatabaseInterface
 		$connectors = array();
 
 		// Get a list of types.
-		$types = JFolder::folders(dirname(__FILE__));
+		$types = JFolder::files(dirname(__FILE__).'/database');
 
 		// Loop through the types and find the ones that are available.
 		foreach ($types as $type)
 		{
-			// Ignore some folders.
-			if (($type == 'database') || ($type == 'table') || ($type == '.') || ($type == '..'))
+			// Ignore some files.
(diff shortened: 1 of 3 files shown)


elinw:
Some comments, I had to change 5 lines not the 4 lines that Louis said on the other pull request, but this does now return the format for the connectors described in the docblock. The other issue is that this eliminates support for a file called driver.php in a folder. After some discussion with Ian, we concluded that such a file never existed in core and it is highly unlikely that it exists in an extension. However, when/if this is committed I will add a note about that to the potential backward compatibility issues document. For anyone with an existing driver.php file that is not covered by the newer drivers can easily rename that file and put it in the appropriate location. From my reading of the code the required class name needed to exist anyway so that is not a concern. 

The field does not have a unit test but rather has a skeleton as for the other fields. However, I have tested this in the CMS in the following way.
Add the field and update the database library.
In the installation application fid the database.xml form. Change the database type field to something like
            <field name="db_type" type="databaseconnection" id="db_type" class="inputbox"
                label="INSTL_DATABASE_TYPE_LABEL"
                default="mysqli"
                required="true" supported="mysql,mysqli,postgres"
            />
Including databases you do and do not have on your server.

LouisLandry:
Well done. :-)

PR #470, opened 2011-10-29
Codestyle clean up

elinw:
Some cleanup of database and databaseconnection.


elinw:
A bit more style and other work while I was in that file.

LouisLandry:
Gracias.

PR #472, opened 2011-10-29
Adding a missing jimport which is causing the unit tests to raise an erro

elinw:
Adding a missing jimport which is causing the unit tests to raise an error in menutype.


elinw:
As a note, you'll only get the error if you run the tests in a certain order.

PR #475, opened 2011-10-29
JWeb has several errors using $this->client->engine and $this->client->en

AmyStephen:
JWeb has several errors using $this->client->engine and $this->client->encodings due to the protected property definition.

### DIFF
--- a/libraries/joomla/application/web/webclient.php
+++ b/libraries/joomla/application/web/webclient.php
@@ -56,7 +56,7 @@ class JWebClient
 	 * @var    integer  The detected rendering engine used by the web client.
 	 * @since  11.3
 	 */
-	protected $engine;
+	public $engine;
 
 	/**
 	 * @var    integer  The detected browser used by the web client.
@@ -80,7 +80,7 @@ class JWebClient
 	 * @var    array  The priority order detected accepted encodings for the client.
 	 * @since  11.3
 	 */
-	protected $encodings = array();
+	public $encodings = array();
 
 	/**
 	 * @var    string  The web client's user agent string.


eddieajau:
I haven't seen this error in practice. The magic get should take care of it. See https://github.com/eddieajau/joomla-platform-examples/blob/master/web/detect-client/index.php

AmyStephen:
Yea, not sure, Andrew. 

PHPStorm is pointing it out the following two lines as "Member has Protected Access" 

414 $encodings = array_intersect($this->client->encodings, array_keys($supported));

584 - elseif (!$moved and ($this->client->engine == JWebClient::WEBKIT))

Further, 584 line has a second error "Element Script doesn't have required attribute type."

Was going to share a link to an offer for PHP Storm for free for Joomla - but I see you posted in on that note. In case you want to try it and haven't - the error reporting is incredibly handy.

http://groups.google.com/group/joomla-dev-general/browse_thread/thread/e1396f4fc6f4616%28Message-ID

LouisLandry:
Nope, this is done this way on purpose.  The magic __get() method allows for read-only access to these properties without allowing anyone to write to them.  This is an example of PHPStorm just not understanding the logic it's looking at.

eddieajau:
Amy I think you need to rely less on what php storm says and more on programming principles. I'm really unimpressed about he false positives it's throwing.

AmyStephen:
I will not report these in the future.

eddieajau:
Amy, you can still report them if they make sense. I just ask that you don't put blind faith in what phpStorm says which, in this case, was very obviously wrong.

PR #476, opened 2011-10-30
Work on replacePrefix

elinw:
If we want to replacePrefix and  we don't have $this->tablePrefix  we  need to get it. Thanks Gabriele for pointing out that there was a problem at https://groups.google.com/d/topic/joomla-dev-platform/3dB6LUN890E/discussion.

### DIFF
--- a/libraries/joomla/database/database.php
+++ b/libraries/joomla/database/database.php
@@ -173,7 +173,7 @@ abstract class JDatabase implements JDatabaseInterface
 		$connectors = array();
 
 		// Get a list of types.
-		$types = JFolder::files(dirname(__FILE__) . '/database')
+		$types = JFolder::files(dirname(__FILE__) . '/database');
 
 		// Loop through the types and find the ones that are available.
 		foreach ($types as $type)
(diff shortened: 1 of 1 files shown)


eddieajau:
An empty table prefix is a valid value - not every database will have them. Also, calling the factory method, which I don't think is a good idea, within a loop will potentially lead to recursion, or at best performance degradation.

elinw:
That's exactly the issue, currently the method errors out if you do not have $this->tablePrefix  (either a string or null), but it fails silently on the last iteration of the loop  (unless you echo out the results).

eddieajau:
What's the error?

elinw:
https://gist.github.com/1326876  shows you why I came to that solution, most likely there is something better.

Now you could say just do the factory where you are using it, which is what I did at first, but it seemed more likely to be error prone if you did that instead of putting it in the method. But I do think maybe the loop issue pushes the balance back in that direction.

eddieajau:
Try `JFactory::getDbo()->replacePrefix('#__somename');`

elinw:
:) Works perfectly, thanks.

eddieajau:
If you turn on strict errors if will tell you that you can't call a non-static method statically - because it gives you unpredictable results, as you found out.

PR #482, opened 2011-11-03
Fixed JController::getName()

Hackwar:
JController::getName() calculated the name of the controller upon each function call. This change does it once and then caches the value.


chdemko:
Hannes, could you correct the code style errors?

Hackwar:
Which code style errors do you mean? The ones mentioned by joomla-jenkins are present in the current joomla platform repository and have nothing to do with my pull request. At least the current sniffs don't report any errors on my system for controller.php

elinw:
The warnings are  there, the errors are not. For one thing it looks like @param only has one space after it in a number of locations.

Hackwar:
If you look at the diffs, I didn't change anything coding style related that wasn't in there in the first place.

I also fixed JModel and JView to optimize it like JController

tests/suite/joomla/application/component/JControllerTest.php
```diff
@@ -9,6 +9,57 @@ require_once JPATH_PLATFORM.'/joomla/application/component/controller.php';
...
+		} else {
+			trigger_error('Undefined or private property: ' . __CLASS__.'::'.$name, E_USER_ERROR);
+			return null;
+		}
+	}
+
+	/**
+	* Sets any property from the class.
+	*
+	* @param string $property The name of the class property.
```
elinw: Improper alignmens ..t. add 2 spaces after @param, all of he descriptions with 2 spaces after the longes name.  Put 2 spaces after string.

Same on the next line

That's 6/29.

tests/suite/joomla/application/component/JControllerTest.php
```diff
@@ -9,6 +9,57 @@ require_once JPATH_PLATFORM.'/joomla/application/component/controller.php';
...
+	*/
+	public function __set($property, $value)
+	{
+		$this->$property = $value;
+	}
+	
+	/**
+	 * Calls any inaccessible method from the class.
+	 * 
+	 * @param string 	$name Name of the method to invoke
```
elinw: Please see above, additional 6/29

tests/suite/joomla/application/component/JControllerTest.php
```diff
@@ -64,7 +117,7 @@ class JControllerTest extends PHPUnit_Framework_TestCase
 		// Include JModel as this method is a proxy for JModel::addIncludePath
 		require_once JPATH_PLATFORM.'/joomla/application/component/model.php';
 
-		$path = JPATH_ROOT.'/addmodelpath';
+		$path = JPath::clean(JPATH_ROOT.'/addmodelpath');
```
elinw: concatenate should always have spaces

Hackwar:
All this is in the unittest, not in the platform code. So far, the unittests were excluded from the coding standard... If that is not the case anymore, we have a serious big task at hand. There is not a single file in the /tests folder that adheres to the coding standard.

chdemko:
Mea Culpa, Hannes is right, these code style errors are present in the code of Joomla! and not in its code.

mbabker:
FWIW, if the test suite is going to be included in the code style report, the docs should be updated to reflect that and the task taken on and completed.  Otherwise, I'd update the pull tester to follow the guidance suggested at https://github.com/joomla/joomla-platform/blob/staging/docs/manual/appendices/analysis.xml#L87 and only check J! code.

realityking:
Inspired by this discussion I've ran the autoformatter (with the rules so far pulled to the platform code) on the unit tests too. The result can be found in #486.

elinw:
But it's also true whether or not the rules say to spell things correctly you should spell them correctly, and you should always aim for the highest standards of code. It took more characters to argue about whether correct behavior is required in unitests  (not to mention time and energy of several people who have their own projects) than it would have to fix them.

chdemko:
@Hannes,
There is already a JControllerInspector class in  tests/suite/joomla/application/component/JControllerHelper.php, it's better to add your __set and __call method in this class.

Hackwar:
&lt;Insert a long rant about coding styles here&gt;

I've changed the files as much as I thought was necessary. I did not add all the doc strings for the methods that were auto generated.

elkuku:
Hannes, the code sniffer actually does not run over the unit test folder... But it's good to have this done ;)
On my system I do not have any errors for unit tests or code style for this pull.

Hackwar:
Nikolai, Elin asked these changes to be made. Which is why I'm pretty much confused. Either we want the coding style in the unittests, too, then we have A LOT of work to do, or we don't and the rules would be okay as long as the unittests are good and the doc blocks contain the @since tag.

elkuku:
Don't get me wrong, I am not saying that the unit tests should not comply to the standards - I LOVE coding standards :P
It's just that ATM there is no mechanism to enforce this.
"A lot" of work is actually about 6500 + errors... Maybe we can get a smart auto formatter to do the job..
One way or the other - it's just a decision that has to be made and - someone to DO it ;)

eddieajau:
My practice is to apply coding standards to all code.  It's good practice to have the unit tests conforming to the standards even if we don't enforce it.

PR #485, opened 2011-11-04
Add JForm field type 'Image'

janich:
Images is used for a variety of purposes, so the platform should have at least some basic capability to show an image within a form.
This new form field type is very basic, but yet it should be powerfull enough to support showing an images within any form.

### DIFF
--- a/libraries/joomla/form/fields/image.php
+++ b/libraries/joomla/form/fields/image.php
@@ -34,6 +34,7 @@ class JFormFieldImage extends JFormField
 	 * Use attributes to identify specific fields
 	 *
 	 * @return  string  The field input markup.
+	 *
 	 * @since   11.3
 	 */
 	protected function getInput()
@@ -64,6 +65,5 @@ class JFormFieldImage extends JFormField
 					' id="' . $this->id . '"' . $attr . ' />';
 
 		return implode("\n", $html);
-		
 	}
 }
--- a/libraries/joomla/form/fields/image.php
+++ b/libraries/joomla/form/fields/image.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * @package     Joomla.Platform
+ * @subpackage  Form
+ *
+ * @copyright   Copyright (C) 2005 - 2011 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE
+ */
+
+defined('JPATH_PLATFORM') or die;
+
+jimport('joomla.form.formfield');
+
+/**
+ * Form Field class for the Joomla Platform.
+ * Provides an embedded Image preview
+ *
+ * @package     Joomla.Platform
+ * @subpackage  Form
+ * @since       11.3
+ */
+class JFormFieldImage extends JFormField
+{
+	/**
+	 * The form field type.
+	 *
+	 * @var    string
+	 * @since  11.3
+	 */
+	protected $type = 'Image';
+
+	/**
+	 * Method to get the field input markup for the image.
+	 * Use attributes to identify specific fields
+	 *
+	 * @return  string  The field input markup.
+	 * @since   11.3
+	 */
+	protected function getInput()
+	{
+		// Initialize attributes
+		$attr = '';
+
+		// Initialize some field attributes.
+		$attr .= $this->element['class'] ? ' class="'.(string) $this->element['class'].'"' : '';
+
+		// Initialize JavaScript field attributes.
+		$attr .= $this->element['onclick'] ? ' onclick="'.(string) $this->element['onclick'].'"' : '';
+
+		// Set directory
+		$directory = (string) $this->element['directory'];
+
+		// Compile source
+		$src = JURI::base(true) . '/' . $directory . '/' . htmlspecialchars($this->value, ENT_COMPAT, 'UTF-8');
+
+		$html = array();
+
+		// The image container and tag
+		$html[] = '<div class="fltlft">';
+		$html[] = ' <img src="'. $src .'"'.
+					' id="'. $this->id .'"'.$attr.' />';
+		$html[] = '</div>';
+
+		return implode("\n", $html);
+	}
+}
--- a/libraries/joomla/form/fields/image.php
+++ b/libraries/joomla/form/fields/image.php
@@ -61,7 +61,7 @@ class JFormFieldImage extends JFormField
 		$html = array();
 
 		// The image container and tag
-		$html[] = ' <img src="' . $src . '"'.
+		$html[] = ' <img src="' . $src . '"' .
 					' id="' . $this->id . '"' . $attr . ' />';
 
 		return implode("\n", $html);
--- a/libraries/joomla/form/fields/image.php
+++ b/libraries/joomla/form/fields/image.php
@@ -42,14 +42,14 @@ class JFormFieldImage extends JFormField
 		$attr = '';
 
 		// Initialize some field attributes.
-		$attr .= $this->element['class'] ? ' class="'.(string) $this->element['class'].'"' : '';
-		$attr .= $this->element['alt'] ? ' alt="'.(string) $this->element['alt'].'"' : '';
-		$attr .= $this->element['title'] ? ' title="'.(string) $this->element['title'].'"' : '';
-		$attr .= $this->element['width'] ? ' width="'.(string) $this->element['width'].'"' : '';
-		$attr .= $this->element['height'] ? ' height="'.(string) $this->element['height'].'"' : '';
-		
+		$attr .= $this->element['class'] ? ' class="' . (string) $this->element['class'] . '"' : '';
+		$attr .= $this->element['alt'] ? ' alt="' . (string) $this->element['alt'] . '"' : '';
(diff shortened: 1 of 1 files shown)


libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,68 @@
...
+	 * @var    string
+	 * @since  11.3
+	 */
+	protected $type = 'Image';
+
+	/**
+	 * Method to get the field input markup for the image.
+	 * Use attributes to identify specific fields
+	 *
+	 * @return  string  The field input markup.
```
jlover, line 36: Return comment requires a blank newline after it

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,68 @@
...
+	 * @since   11.3
+	 */
+	protected function getInput()
+	{
+		// Initialize attributes
+		$attr = '';
+
+		// Initialize some field attributes.
+		$attr .= $this->element['class'] ? ' class="'.(string) $this->element['class'].'"' : '';
```
jlover: Concat operator must be surrounded by spaces
(note: this does _not_ apply to the `.=` sign

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,68 @@
...
+		$attr .= $this->element['alt'] ? ' alt="'.(string) $this->element['alt'].'"' : '';
```
jlover: Concat operator must be surrounded by spaces

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,68 @@
...
+		$attr .= $this->element['title'] ? ' title="'.(string) $this->element['title'].'"' : '';
```
jlover: Concat operator must be surrounded by spaces

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,68 @@
...
+		$attr .= $this->element['width'] ? ' width="'.(string) $this->element['width'].'"' : '';
```
jlover: Concat operator must be surrounded by spaces

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,68 @@
...
+		$attr .= $this->element['height'] ? ' height="'.(string) $this->element['height'].'"' : '';
```
jlover: Concat operator must be surrounded by spaces

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,68 @@
...
+
```
jlover: Whitespace found at end of line

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,68 @@
...
+		// Initialize some field attributes.
+		$attr .= $this->element['class'] ? ' class="'.(string) $this->element['class'].'"' : '';
+		$attr .= $this->element['alt'] ? ' alt="'.(string) $this->element['alt'].'"' : '';
+		$attr .= $this->element['title'] ? ' title="'.(string) $this->element['title'].'"' : '';
+		$attr .= $this->element['width'] ? ' width="'.(string) $this->element['width'].'"' : '';
+		$attr .= $this->element['height'] ? ' height="'.(string) $this->element['height'].'"' : '';
+		
+		// Initialize JavaScript field attributes.
+		$attr .= $this->element['onclick'] ? ' onclick="'.(string) $this->element['onclick'].'"' : '';
```
jlover: Concat operator must be surrounded by spaces

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,68 @@
...
+		// Set directory
+		$directory = (string) $this->element['directory'];
+
+		// Compile source
+		$src = JURI::root(true) . '/' . $directory . '/' . htmlspecialchars($this->value, ENT_COMPAT, 'UTF-8');
+
+		$html = array();
+
+		// The image container and tag
+		$html[] = ' <img src="'. $src .'"'.
```
jlover: Concat operator must be surrounded by spaces

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,68 @@
...
+					' id="'. $this->id .'"'.$attr.' />';
```
jlover: Concat operator must be surrounded by spaces

jlover:
The actual error count reported by @joomla-jenkins (including unit test results) may not be accurate - please be patient, all errors will be fixed ASAP ;)

chdemko:
Your field type display only an image but does not give the possibility to get it from the user. And there is already the JFieldMedia class which does this stuff.

elinw:
Christophe, this is designed for a different purpose which is to display an image.  If  you lcook back to the change of mimetypes committed to the CMS you will see the use case for this as a field type. It's not the same thing as the media field. You can read back in the jbs chat on Friday for extensive discussion of this. http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26940

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,69 @@
...
+		// Set directory
+		$directory = (string) $this->element['directory'];
+
+		// Compile source
+		$src = JURI::root(true) . '/' . $directory . '/' . htmlspecialchars($this->value, ENT_COMPAT, 'UTF-8');
+
+		$html = array();
+
+		// The image container and tag
+		$html[] = ' <img src="' . $src . '"'.
```
elkuku: Please move softly to the end of the line, hit the "left arrow key" -- followed by the space bar.
Even more softly.. push your changes here to GitHub, so the code may pass Mr. joomla-jenkins sharp and cruel eagle eyes that wont even miss a single space ;)
You might also be interested in the existing documentation about this (annoying?) procedure: http://docs.joomla.org/Joomla_CodeSniffer

Happy coding ;)

janich: @elkuku, Totally missed that dot. Thanks for the hint (and link) ;)

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,69 @@
...
+		// Compile source
+		$src = JURI::root(true) . '/' . $directory . '/' . htmlspecialchars($this->value, ENT_COMPAT, 'UTF-8');
+
+		$html = array();
+
+		// The image container and tag
+		$html[] = ' <img src="' . $src . '"' .
+					' id="' . $this->id . '"' . $attr . ' />';
+
+		return implode("\n", $html);
```
realityking, line 67: Since this will always be just one line this can be simplified quite a bit.

libraries/joomla/form/fields/image.php
```diff
@@ -0,0 +1,69 @@
...
+		$attr .= $this->element['height'] ? ' height="' . (string) $this->element['height'] . '"' : '';
+
+		// Initialize JavaScript field attributes.
+		$attr .= $this->element['onclick'] ? ' onclick="' . (string) $this->element['onclick'] . '"' : '';
+
+		// Set directory
+		$directory = (string) $this->element['directory'];
+
+		// Compile source
+		$src = JURI::root(true) . '/' . $directory . '/' . htmlspecialchars($this->value, ENT_COMPAT, 'UTF-8');
```
realityking, line 59: This makes this useless for my needs (Gravatar plug-in) since I need to include an image from a remote host. I'd suggest a second method getSrc(). This way I could subclass it at least.

On the other hand, no sure if that's really worth it. Right now my getInput() method only has 3 lines of code.

realityking:
I still like the idea, but I wonder how we can make it more general purpose. One thing would be to be able to override the URL.

The other question is that I don't quite understand the point of the directory attribute. Could you elaborate on that one?

elinw:
I think if you are going to open up to external sources you do need to have a way to limit those to a specific list. There is a lot of potential for abuse in letting users hot link to media external sites that were not designed for the purpose of rendering on other sites. It could easily be or be interpreted as an attack from your site on the site hotlinked. Also I would add an attribute something like INTERNAL so that if you say INTERNAL='True" then you can limit the sources to internal folders only. Still it would be great to incorporate a general solution to allowing use of Gravatar type services.

realityking:
I didn't say I want to be able to directly use external URLs (that certainly wouldn't be good)

What I do want is to put the Url portion into a separate method so I can easily override it in a subclass and still take advantage of the other code. Kinda like JFormFieldList is subclasses dozens of times only overriding the getOptions() method.

elinw:
Yes I know you didn't. I think what you are saying makes sense.

janich:
Im sorry for late reply.

The directory attribute were meant as a limiter for 3rd party devs to use.
In the original proof of concept this field were used to display a user avatar. Since the users should never be able to use any sitewide images, the directory were limited in the plugin's manifest.
This is probably not the best method, but I hope it highlights the idea of it.

And yeah. For this field to work properly, we need to at least separate the url into an separate, overrideable method.
Ill look into that asap.

realityking:
I'm closing this as won't fix.

We don't see this as generic enough to be in the platform.

PR #488, opened 2011-11-05
Code style: adjusting line lengths

elkuku:
This should reduce the amount of warnings reported by the code sniffer and, of course, make the code look much prettier ;)


elkuku:
TBH I would have expected less warnings :P

PR #489, opened 2011-11-05
Fixing control signature rules

chdemko:
Some errors have appeared in https://github.com/joomla/joomla-platform/pull/482. They are due to bad control signature rules

### DIFF
--- a/build/phpcs/Joomla/Sniffs/ControlStructures/ControlSignatureSniff.php
+++ b/build/phpcs/Joomla/Sniffs/ControlStructures/ControlSignatureSniff.php
@@ -50,21 +50,20 @@ class Joomla_Sniffs_ControlStructures_ControlSignatureSniff extends PHP_CodeSnif
 	protected function getPatterns()
 	{
 		return array(
-			'if (...)EOL...{EOL',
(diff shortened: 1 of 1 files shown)


elkuku:
Could you please spell out the errors that appeared in #482 ?
I wrote a set of test files that may help in training the sniffs
https://github.com/elkuku/joomla-platform/tree/sniff-test/tests/code_sniffer/tests
Especially this could be helpful:
https://github.com/elkuku/joomla-platform/blob/sniff-test/tests/code_sniffer/tests/ControlStructures/bad.php

So here is a test case that fails applying your modifications:

```
if (true)
{
    true;
}
else {
    true;
}
```

chdemko:
There were no errors, the rules were buggy

elkuku:
Again. Could you please spell out a 'bug' which is not caught by the former sniff.
Your fix would allow the above mentioned coding standard violation while the actual sniff does not.

Edit: TBH, there seems to be a bug in the CodeSniffer. It's not that your modifications are incorrect, but the code sniffer for some obscure reason won't accept them this way - I spent quite some time trying to fix it, and the actual sniff is the best solution I could find.

PR #490, opened 2011-11-06
Adding the possibility to load (un)compressed css and js file whether debug in on or off

chdemko:
To improve the speed of loading a web page, it may be interesting to automatically detect the debug mode and to load compressed files (css and js) or not.

The logic is as follow:

```
if debug is on
    try to load the uncompressed version of a file named *file*-uncompressed.(js,css)
    else load the classical file *file*.(js,css)
else
    load the classical file *file*.(js,css)
endif
```

The use is as follow:

```
JHtml::_('script', 'com_foo/bar.js');
```

will look for files in media/com_foo/js/

```
if debug is on
    try to load bar-uncompressed.js
    else load bar.js
else
    load bar.js
endif
```

For stylesheets:

```
JHtml::_('stylesheet', 'com_foo/bar.css');
```

will look for files in media/com_foo/css/

```
if debug is on
    try to load bar-uncompressed.css
    else load bar.css
else
    load bar.css
endif
```

Unit tests have been added in modified classes


chdemko:
See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27141 for CMS testing

chdemko:
This has been tested on the CMS and is ready to commit here before committing into the CMS

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&amp;tracker_item_id=27141

PR #492, opened 2011-11-06
Don't run PHPMD on 3rd party libraries.

realityking:


realityking:
This should lower the PMD warnings a bit.

PR #495, opened 2011-11-07
Adding a true autoloader to the Joomla Platform.

LouisLandry:
As of now there is no such thing as a true autoloader for the Joomla Platform.  What we have is a way to explicitly register files with a loader class to be lazy-loaded when necessary.

This pull request adds another method to the SPL autoloader queue that would provide true autoloading functionality for classes that follow a given naming convention for files and class names without having to use `jimport()` or `JLoader` explicitly.

To leverage the autoloader the following conventions must be met:

**Class Names**
- All class names must be prefixed with an upper-case **J** to indicate that the class file will be found within the `libraries/joomla` folder.
-  All class names must be in camel case where the first character after the **J** prefix is upper-case.  For example: `JClass` or `JClassExtended`

**Class File Paths**
- The first _sections_ of a class name indicate the folder path for the class file, and the final _section_ of the class name indicates the class file name.  For the example `JClassExtended` **J** indicates a Joomla library, **Class** indicates the folder path, and **Extended** indicates the file name for the class.  Therefore within `libraries/joomla/class/extended.php` you will find the class `JClassExtended` defined.
- If there is only one section of the class name then that section is used both for a folder path and class file name.  For example `JRegistry` is found defined in the file `libraries/joomla/registry/registry.php`.
## Backward Compatability

The implementation of this new autoloader is supplemental to the current lazy-loading system and should have no impact on backward compatability.  In point of fact, the current lazy-loading system is called first by the SPL autoloading queue, then if the file is not found the new autoloader mechanism is called.  This should make it easy to transition libraries into the appropriate conventions over time without any impact to third party developers or downstream software.
## Further Example

As it happens, the Registry package is a perfect case study for the new autoloader.  With the autoloader enabled, there is no need to ever explicitly call `jimport('joomla.registry.registry');` because the classes already follow this convention perfectly.  You would just simply use JRegistry as if it is already loaded.  The class name to file path mapping is shown below:

`JRegistry` => `libraries/joomla/registry/registry.php`
`JRegistryFormat` => `libraries/joomla/registry/format.php`
`JRegistryFormatINI` => `libraries/joomla/registry/format/ini.php`
`JRegistryFormatJSON` => `libraries/joomla/registry/format/json.php`
`JRegistryFormatPHP` => `libraries/joomla/registry/format/php.php`
`JRegistryFormatXML` => `libraries/joomla/registry/format/xml.php`

_It should be noted that the autoloader gracefully handles when a segment is all caps as in the case above.  The important bit is the transition from lowercase to uppercase._

### DIFF
--- a/libraries/loader.php
+++ b/libraries/loader.php
@@ -8,10 +8,6 @@
 
 defined('JPATH_PLATFORM') or die;
 
-// Register relevant autoload methods.
-spl_autoload_register(array('JLoader', 'load'));
-spl_autoload_register(array('JLoader', 'autoload'));
-
 /**
  * Static class to handle loading of libraries.
  *
@@ -35,74 +31,6 @@ abstract class JLoader
 	 * @since  11.1
 	 */
 	protected static $classes = array();
-
-	/**
-	 * Loads a class from specified directories.
-	 *
-	 * @param   string   $key   The class name to look for (dot notation).
-	 * @param   string   $base  Search this directory for the class.
-	 *
-	 * @return  boolean  True on success.
-	 *
-	 * @since   11.1
-	 */
-	public static function import($key, $base = null)
-	{
-		// Only import the library if not already attempted.
-		if (!isset(self::$imported[$key]))
-		{
-			// Setup some variables.
-			$success = false;
-			$parts = explode('.', $key);
-			$class = array_pop($parts);
-			$base = (!empty($base)) ? $base : dirname(__FILE__);
-			$path = str_replace('.', DS, $key);
-
-			// Handle special case for helper classes.
-			if ($class == 'helper')
-			{
-				$class = ucfirst(array_pop($parts)) . ucfirst($class);
-			}
-			// Standard class.
-			else
-			{
-				$class = ucfirst($class);
-			}
-
-			// If we are importing a library from the Joomla namespace set the class to autoload.
-			if (strpos($path, 'joomla') === 0)
-			{
-
-				// Since we are in the Joomla namespace prepend the classname with J.
-				$class = 'J' . $class;
-
-				// Only register the class for autoloading if the file exists.
-				if (is_file($base . '/' . $path . '.php'))
-				{
-					self::$classes[strtolower($class)] = $base . '/' . $path . '.php';
-					$success = true;
-				}
-			}
-			/*
-			 * If we are not importing a library from the Joomla namespace directly include the
-			 * file since we cannot assert the file/folder naming conventions.
-			 */
-			else
-			{
-
-				// If the file exists attempt to include it.
-				if (is_file($base . '/' . $path . '.php'))
-				{
-					$success = (bool) include_once $base . '/' . $path . '.php';
-				}
-			}
-
-			// Add the import key to the memory cache container.
-			self::$imported[$key] = $success;
-		}
-
-		return self::$imported[$key];
-	}
 
 	/**
 	 * Method to discover classes of a given type in a given path.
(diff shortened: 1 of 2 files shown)


PhilETaylor:
Brings JP in line with pretty much all other PHP frameworks and removes jimport, so all for it!

libraries/loader.php
```diff
@@ -224,6 +225,46 @@ abstract class JLoader
...
+			// If there is only one part we want to duplicate that part for generating the path.
+			$parts = (count($parts) === 1) ? array($parts[0], $parts[0]) : $parts;
+
+			// Generate the path based on the class name parts.
+			$path = JPATH_PLATFORM . '/joomla/' . implode('/', array_map('strtolower', $parts)) . '.php';
+
+			// Load the file.
+			if (file_exists($path))
+			{
+				include_once $path;
```
PhilETaylor: As you are already doing a check to see if this class exists on line 241 can this line not be a simple include instead of an include_once?

LouisLandry: Yep, I can actually.  Additionally I can probably get rid of the if (!class_exists()) call above this block as well.  Will need to test it to make sure obviously.  

The check there was in case the thing was called directly not by the autoloader, but I guess we can make a reasonable assumption that it won't be.  I've even heard rumors that the spl autoloader can call private methods which would mean I can guarantee that it is never called directly.  Thanks for the note.

eddieajau:
Github needs a +1 button :)

jonnsl:
:+1:

elkuku:
:+1:

libraries/loader.php
```diff
@@ -224,6 +225,46 @@ abstract class JLoader
...
+				return true;
```
chdemko: may be the return true could be replaced by a

```
return class_exists($class);
```

(in the case where the included file does not define the class; normally this should not be possible but ...)

chdemko:
:+1:

mahagr:
:+1:

I've added autoloader to Kunena 2.0 a while ago and removing all kimport()'s. There was measurable and surprisingly large performance improvement and at the same time I got rid of a real annoying source of bugs -- people (including me) forgetting to import all the needed classes.

I have one question, though: should every component have their own autoloader or should we make the autoloader to allow mappings like Kunena -> administrator/components/com_kunena/libraries (and using the same rules after that)?

FYI: my autoloader can be found from here: https://github.com/Kunena/Kunena-2.0/blob/master/administrator/components/com_kunena/api.php

ianmacl:
I think it is reasonable to make it flexible to allow use by other packages, but I think at this point we should start simple, make it stable and then improve it from there.

libraries/loader.php
realityking: I think it can call private methods, but only if you use spl_autoload_register() inside the class.

chdemko:
It could be cool if it allows the use by other package: when we will move some classes to the CMS, the autoloader could search also in the joomlacms folder

chdemko:
I've proposed an enhance version for splitting a camel based class name there https://github.com/chdemko/joomla-platform/blob/b7ec1b386f09c7cf5e7f2202247a69bbaa5b3c96/libraries/joomla/string/string.php#L84

mahagr:
@ianmacl: I can fully agree on that, I just had a thought that we should somehow allow auto-loading also from other locations.

How about the class overrides (inside plugins etc)? Do we still use the old registration method in those?

jlleblanc:
:+1:

Looks great. If PHP namespaces are implemented down the road, that would be a good way of handling autoloading from other folders inside `libraries`.

robschley:
:+1:

I dig it. Now we can get away from those stupid file names.

LouisLandry:
OK, I've made some adjustments based on comments here and conversations elsewhere.  Essentially I've made the method private and added a JLoader::setup() method which is called from the Joomla Platform import script that sets up the autoloader callbacks.  I've also removed some unnecessary checks since PHP automatically takes care of those things as it iterates through the autoloaders and the _autoload() method cannot be called from outside of the class anymore.

PR #496, opened 2011-11-08
Adding features to the JForm library + correcting some errors

chdemko:
The purpose of these modifications is to bring two new features to the `JForm` library and two new methods to the library `JString` as a side effect:
- the possibility to prefix field and rule classes by any string (currently, it's only `'J'`). To use the field `FooFormFieldBar` in an xml, simply put `type="foo.bar"`
- the possibility to guess the field type using the class name.
  - A class named `JFormFieldBar` would have his type guessed as `'Bar'`
  - A class named `JFormFieldModal_Bar` would have his type guessed as `'Modal_Bar'`
  - A class named `FooFormFieldBar` would have his type guessed as `'FooBar'`
  - A class named `FooFormFieldModal_Bar` would have his type guessed as `'FooModal_Bar'`
- the method `JString::splitCamelCase` allow to split a string using a camel case notation
  - `"FooBarABCDef"`            becomes  `array("Foo", "Bar", "ABC", "Def")`
  - `"JFooBar"`                 becomes  `array("J", "Foo", "Bar")`
  - `"J001FooBar002"`           becomes  `array("J001", "Foo", "Bar002")`
  - `"abcDef"`                  becomes  `array("abc", "Def")`
  - `"abc_defGhi_Jkl"`          becomes  `array("abc_def", "Ghi_Jkl")`
  - `"ThisIsA_NASAAstronaut"`   becomes  `array("This", "Is", "A_NASA", "Astronaut")`
  - `"JohnFitzgerald_Kennedy"`  becomes  `array("John", "Fitzgerald_Kennedy")`
- the method `JString::ucfirstEach` allow to ucfirst all words of a string
  - `JString::ucfirstEach('dr jekill and mister hyde')` becomes `'Dr Jekill And Mister Hyde'`
  - `JString::ucfirstEach('dr jekill and mister hyde',' ', '_')` becomes `'Dr_Jekill_And_Mister_Hyde'`
  - `JString::ucfirstEach('dr jekill and mister hyde',' ', '')` becomes `'DrJekillAndMisterHyde'`

Last, there was an error in the `JForm` library: the `array_unshift` in the `JFormHelper::loadClass` method only allow to look in the first half of paths.


elinw:
Nice

libraries/joomla/string/string.php
```diff
@@ -65,6 +65,48 @@ abstract class JString
...
+	 *
+	 * @param   string  $string         The source string.
+	 * @param   string  $delimiter      The words delimiter
+	 * @param   string  $new_delimiter  The new words delimiter (null means equal to $delimiter)
+	 *
+	 * @return  string   The string with all words ucfirst.
+	 *
+	 * @since   11.3
+	 */
+	public static function ucfirstEach($string, $delimiter = ' ', $new_delimiter = null)
```
eddieajau: Can we just add this to the "ucfirst" method?  

```
public static function ucfirst($string, $delimiter = null, $newDelimiter = null)
```

If $delimiter is not null, explode on it (and use the utf8 version of the function).  Also, use camelcase for the name of the last arg.

PR #497, opened 2011-11-08
Changing all cases of static instance variables

Hackwar:
Changing all cases of static instance variables in functions to class static properties. This should cover all cases where a class stores itself in an $instance variable and makes all occurences identical. It only changed $instance(s) cases, some methods use different variables for different cases, which have not been touched and still need to be changed.


LouisLandry:
Great!  Thanks Hannes.

PR #498, opened 2011-11-09
Adding JSockets library to manage sockets

fastslack:
JSockets is a library to manage sockets using Joomla! Platform. A simple explanation here: http://matware.com.ar/joomla/creating-a-server-using-joomla-platform.html

### DIFF
--- a/libraries/joomla/sockets/child.php
+++ b/libraries/joomla/sockets/child.php
@@ -1,13 +1,13 @@
 <?php
 /**
- * @version				$Id: 
- * @package				Joomla.Platform
- * @subpackage		JSockets
- * @copyright			Copyright (C) 1996 - 2011 Matware - All rights reserved.
- * @author				Matias Aguirre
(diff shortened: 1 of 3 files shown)


eddieajau:
Matias, you might want to have a look at what I've done here:
https://github.com/eddieajau/joomla-platform-examples/tree/sockets/web/socket

It's based on the work here:
https://github.com/jam1401/PHP-Websockets-Server

Still a bit of cleaning up to do :)

fastslack:
Hi Andrew,

Let me play a little with it :) 

Is any plan to add sockets/websockets support to Joomla! Platform?

eddieajau:
There are no plans, but it would be a desirable feature :)

Just a note, any socket package will require extensive unit tests.  Because of that, and since there are at least 2 solutions in branches I'm going to close this pull request for now.  Can you start a thread on the platform list and we can continue the discussion there (more people can get involved).

Thanks in advance.

PR #501, opened 2011-11-09
Deprecate protected underscore methods and vars

elkuku:
This is a proposal on how to deal with the remaining code style complaints about protected methods and class variables beginning with an underscore, by creating proxies and enclosing the deprecated functions in `@codingStandardsIgnoreStart` and `-End` tags.
A much better solution would be, of course, if the methods could just be renamed.

If this makes any sense you might leave this open for a while and I (or others) might fix the remaing 438 warnings ;)

### DIFF
--- a/libraries/joomla/application/pathway.php
+++ b/libraries/joomla/application/pathway.php
@@ -24,13 +24,29 @@ class JPathway extends JObject
 	 * @var    array  Array to hold the pathway item objects
 	 * @since  11.1
 	 */
-	protected $_pathway = null;
+	protected $pathway = null;
 
 	/**
 	 * @var    integer  Integer number of items in the pathway
 	 * @since  11.1
 	 */
+	protected $count = 0;
+
+	// @codingStandardsIgnoreStart
+	/**
+	 * @var    array  Array to hold the pathway item objects
+	 * @since  11.1
+	 * @deprecated X - use $pathway
+	 */
+	protected $_pathway = null;
+
+	/**
+	 * @var    integer  Integer number of items in the pathway
+	 * @since  11.1
+	 * @deprecated X - use $count
+	 */
 	protected $_count = 0;
+	// @codingStandardsIgnoreEnd
 
 	/**
 	 * @var    array  JPathway instances container.
@@ -48,7 +64,7 @@ class JPathway extends JObject
 	function __construct($options = array())
 	{
 		//Initialise the array
-		$this->_pathway = array();
+		$this->pathway = array();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


libraries/joomla/application/pathway.php
```diff
@@ -212,4 +228,22 @@ class JPathway extends JObject
...
+	 *
+	 * @param   string  $name  Name of the item
+	 * @param   string  $link  Link to the item
+	 *
+	 * @return  JPathway  Pathway item object
+	 *
+	 * @since   11.1
+	 * @deprecated X - use makeItem()
+	 */
+	protected function _makeItem($name, $link)
```
robschley, line 244: This should log that a deprecated method was called.

libraries/joomla/application/pathway.php
```diff
@@ -24,13 +24,29 @@ class JPathway extends JObject
...
+	 * @var    integer  Integer number of items in the pathway
+	 * @since  11.1
+	 */
+	protected $count = 0;
+
+	// @codingStandardsIgnoreStart
+	/**
+	 * @var    array  Array to hold the pathway item objects
+	 * @since  11.1
+	 * @deprecated X - use $pathway
```
robschley, line 39: Do we need to deprecate protected properties? I thought we only had to deprecate public properties/methods?

elkuku, line 39: Not sure. I don't believe that some 3pd out there really extends JPathway, but I believe I have seen someone accessing protected (now) properties in JMenu - Just to be on the safe side maybe..

libraries/joomla/application/pathway.php
elkuku, line 244: Huh, haven’t thought about it.. So that would be 441 log entries more...

libraries/joomla/application/pathway.php
realityking, line 39: In a way it's part of the API too since people may have extended a class.

That's why I'd like to use privates where it makes sense. This way we can change those methods and variables as we see fit without causing API issues.

ianmacl, line 39: I agree with Rouven - protected methods are protected presumably so they can be used by child classes.

Private methods and variables should be used where appropriate.

chdemko:
I've proposed another solution for this fix: ignore var/function beginning with _ if they are marked deprecated

See https://github.com/joomla/joomla-platform/issues/501

LouisLandry:
We could always just change protected $_foo to protected $foo and create a public function __get() / __set() that handles the deprecated usage.

chdemko:
@Louis, it will work for instance vars but PHP does not have (yet) __getStatic and __setStatic method to handle class vars

elkuku:
Christophe, I think you meant #502 `;)` very clever :+1:

I am going to close this one in favor of your solution.

PR #504, opened 2011-11-09
CMS issue [#26439] Editors have problems with quotes in content plugins syntax 

dextercowley:
Fix CMS issue http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26439

JFilterInput is too restrictive with valid HTML content. This change tweaks a regex value to not filter content in well-formed HTML element.


dextercowley:
I should have added that this fix has already been committed to the CMS repo. Thanks.

PR #512, opened 2011-11-12
Convert the old type variable(TYPE->ENGINE)

yama:
Update libraries/joomla/database/database/mysqli.php

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8549&tracker_item_id=27012

https://github.com/yama/joomladb/commit/389246c8bd12fc7984b632b6d69996d5e4cda83b#commitcomment-712962
Thanks

### DIFF
--- a/libraries/joomla/database/database/mysqli.php
+++ b/libraries/joomla/database/database/mysqli.php
@@ -536,6 +536,9 @@ class JDatabaseMySQLi extends JDatabase
 		$this->errorNum = 0;
 		$this->errorMsg = '';
 
+		// Change the old type variable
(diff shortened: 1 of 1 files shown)


eddieajau:
Hi Yama.  I understand the problem you are trying to fix, but, unfortunately, it means we are doing a regex on every query whether it's a create statement or not.  I would prefer to see developers fix their queries.  It's also possible a developer might be trying to connect to an older version of MySQL (doubtful, but possible). On that basis but I'm not going to accept this change.  Sorry.

PR #515, opened 2011-11-12
UT 2/10: Extending bootstrap.php to generate system independent JPATH constants

Hackwar:
Replacing slash in JLoaderTest with DIRECTORY_SEPERATOR to make the test OS independent
Adding unittest to JAccessTest to achieve more code coverage

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -28,19 +28,19 @@ ini_set('display_errors', 1);
  */
 if (!defined('JPATH_TESTS'))
 {
-	define('JPATH_TESTS', realpath(__DIR__));
+	define('JPATH_TESTS', str_replace(DIRECTORY_SEPARATOR, '/', realpath(__DIR__)));
 }
 if (!defined('JPATH_PLATFORM'))
 {
-	define('JPATH_PLATFORM', realpath(dirname(JPATH_TESTS) . '/libraries'));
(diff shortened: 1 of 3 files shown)


realityking:
Is that actually needed? PHP should handle that itself unless we're comparing paths someplace.

Hackwar:
Its exactly for that purpose. It reduces the number of str_replace that we have to use in the unittests to make the path comparisons work on Windows systems. Without this change, I have up to 50 failures in the testsuite. Since PHP is handling this otherwise well, this is just for the unittests and not for the normal applications.

eddieajau:
That sounds more like a problem with the unit test assertion.  Probably just need a test helper to fix paths before any comparison is made.

Hackwar:
Theoretically I agree, that we need a test helper for path comparisons, but why shouldn't we make sure that the constants that we use are made consistent on all platforms for comparisons already?

Hackwar:
Besides that, this pull request also fixes two other issues in unittests, that raise the code coverage by a bit.

eddieajau:
Just separate out the coverage code into a different request.  On the constants, I hear what you are saying but I don't think what you've done is going to be reliable, and it introduces some really wierd code (eg, $ds).  I think you are better off running the tests and giving us the failure report to we can work on a better solution for the windows tests.

Hackwar:
The $d is actually an error on my side. Let me fix that and then lets see again if we can simply re-open this request. :-)

Hackwar:
Hmm, looks like the correct way would be to use realpath(). So lets see that we do this with realpath() then...

PR #523, opened 2011-11-12
UT 10/10: Adding unittests for updater, user and utilities package

Hackwar:
Adding unittests for updater, user and utilities package and JRoute, JText
Removing deprecated methods from code coverage from JUpdater and JUser


Hackwar:
This pulltester shows the tests to pass and no errors: http://elkuku.github.com/pulltester/pulls/523.html

ianmacl:
1) JUpdaterTest::test__construct
JDatabaseException: JLIB_DATABASE_ERROR_CONNECT_DATABASE

All four were database connection problems.  I'll have to look into why that was.  They do run fine locally.

ianmacl:
Looking at your code it appears as though you aren't ensuring preconditions such as database set in JFactory::$database.  You just do your instantiation without worrying about isolation.

Hackwar:
I was assuming that you only have to do that preconditions statement for the current class, so since JUpdater does not use a database connection anywhere in the code, I would have pointed ad JAdapter and its unittests to check these preconditions properly...

ianmacl:
As a general rule, any test that could eventually call JFactory should set the initial conditions of JFactory properly.  This does mean our code needs to be fixed, but until it is fixed, we have to live with this situation and set up these dependencies or else the tests become really brittle and we see issues like this pop up, where the tests might run fine one place and not somewhere else.

A lot of work has been put into making sure that if the environment is setup properly, the tests will pass.

eddieajau:
Any process on getting the tests to work properly?

Hackwar:
I'm closing this one, since its going to be easier to start over with the missing tests than to fix this again.

PR #525, opened 2011-11-13
Allows JWeb to have a configurable themes path.

elinw:
This adjusts JWeb::Render to use a configuration variable "themes.base" if it is available and then if not to use the convention based path.

### DIFF
--- a/libraries/joomla/application/web.php
+++ b/libraries/joomla/application/web.php
@@ -372,6 +372,7 @@ class JWeb
 			$options['directory'] = $this->get('themes.base');
 		}
 		// Fall back to constants.
+		else
 		{
 			$options['directory'] = (defined('JPATH_BASE') ? JPATH_BASE : dirname(__FILE__)) . '/themes';
 		}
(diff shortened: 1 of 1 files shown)


elinw:
Thanks to Angie for finding this issue while working with the Flickr example application.

eddieajau:
Good idea, but I'd prefer to see this done programmatically, something like:

```
if ($this->get('themes.base'))
{
 $options['directory'] = $this->get('themes.base');
}
// Fall back to constants.
{
 $options['directory'] = (defined('JPATH_BASE') ? JPATH_BASE : dirname(__FILE__)) . '/themes';
}
```

This allows for the theme path to change on the basis of input (eg, auto-switching on subdomains, etc).

Unit tests will also need to be updated slightly to check this new feature.

LouisLandry:
I'm with @eddieajau on this one.  Good find though, and good thoughts on the fix.

elinw:
Ok I wasn't really that happy with what I had.

PR #528, opened 2011-11-14
Remove unnecessary HTML files.

LouisLandry:
The platform project does not have a need for for the index.html files found in all Joomla CMS code folders to prevent directory browsing from web servers.  Presumably if someone is using the platform independent of the Joomla CMS then there will be steps taken to prevent directory browsing either by placing the platform outside of the server document root, or by some other web server directive.


elkuku:
:+1: How about adding a .htaccess file to prevent direct access - just in case ?

PR #533, opened 2011-11-15
JArchive work

realityking:
-Add unit tests
-Add isSupported() function to all adapters


libraries/joomla/filesystem/archive/zip.php
```diff
@@ -142,13 +142,25 @@ class JArchiveZip extends JObject
...
+
+	/**
 	 * Method to determine if the server has native zip support for faster handling
 	 *
 	 * @return  boolean  True if php has native ZIP support
 	 *
 	 * @since   11.1
 	 */
-	public function hasNativeSupport()
+	public static function hasNativeSupport()
```
eddieajau, line 163: Is this going to cause a problem with backward compatibility?

realityking, line 163: Yes and no. Calling it dynamically works, but you'll probably get a E_STRICT warning.

realityking, line 163: I read this up. Calling it dynamically doesn't cause any warning, so there's no backwards compatibility issue.

PR #536, opened 2011-11-17
Clone2

elinw:
This adds a  __clone method for handling deep cloning to JDatabaseQuery and JDatabaseQueryElement and tests for those methods.


elinw:
Thanks Ian and Rob.

PR #537, opened 2011-11-18
Add TRUNCATE method to JDatabaseQuery

mbabker:
JDatabaseQuery currently has no TRUNCATE method. In at least MySQL and PostgreSQL, the TRUNCATE TABLE `#__content` query exists, and this is used a bit in Finder.


mbabker:
Well crap...

PR #540, opened 2011-11-19
Added missing references =& instead of assignments = for array of module...

beat:
Proposing to add missing references =& instead of assignments = for array of module objects in the references chain. JModuleHelper::_load returns a reference, but it is used is by assignment, instead of by reference assignment.

### DIFF
--- a/libraries/joomla/application/module/helper.php
+++ b/libraries/joomla/application/module/helper.php
@@ -33,7 +33,7 @@ abstract class JModuleHelper
 	public static function &getModule($name, $title = null)
 	{
 		$result = null;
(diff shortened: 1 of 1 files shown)


eddieajau:
Thanks Beat.

beat:
You are welcome, Andrew. Thank you for merging, developing and maintaining.

I discovered here that contributing small fixes to Joomla Framework on GitHub within the web-browser directly is not only way simpler than with any other patching or svn methods, but also fun and impressively working well!.

eddieajau:
:) Github is awesome (except for the little things that really annoy me, hehe).

PR #542, opened 2011-11-19
Add truncate function to JDatabase

mbabker:


eddieajau:
Michael, can you change the method name to truncateTable (that leaves us room to use a truncate method to nuke the whole db).  Otherwise that's a good add.

eddieajau:
Thanks Michael!

PR #543, opened 2011-11-19
Update PHPMailer to version 5.2

mbabker:
Update PHPMailer library to 5.2.  Added the POP3 class, removed the English language file as it is set within the code and not distributed anymore, and alphabetized the Joomla! language file.


libraries/phpmailer/phpmailer.php
```diff
@@ -56,7 +53,7 @@ class PHPMailer {
    * Sets the CharSet of the message.
    * @var string
    */
-  public $CharSet           = 'utf-8';
+  public $CharSet           = 'iso-8859-1';
```
eddieajau: Does this change in charset affect us at all?

mbabker: I looked at the PHPMailer 5.1 tag, and it has the ISO charset as well.
I'm thinking we might've changed it internally, so I'll go ahead and
commit that change.

On 11/23/11 12:23 PM, "Andrew Eddie"
<reply+p-247525-461c72b0f132fbf05483aed73d562b0c4bb0e472-368545@reply.githu
b.com> wrote:

PR #546, opened 2011-11-20
Allow application to set $options.

elinw:
Sometimes an application would like to use a different folder name for templates or a second template, but currently this is not always possible, at least not simply when using JWeb.  For example it is easy to create theme2 but to actually access it when it is in a folder that is not called themes is very complex. One way that should work is to be able to do something like this in the application:

```
    $this->options = array(
        'template' => $this->get('theme2'),
        'file' => 'index.php',
        'params' => '',
        'directory' => 'templates',
    );
```

However currently this will always be overwritten by the hard coded options in JWeb.  This pull request simply adds checks for existing options before JWeb sets them.

### DIFF
--- a/libraries/joomla/application/web.php
+++ b/libraries/joomla/application/web.php
@@ -374,12 +374,13 @@ class JWeb
 			);
 		}
 
-		if (!isset($options['directory'])){
+		if (!isset($options['directory']))
+		{
 			if ($this->get('themes.base'))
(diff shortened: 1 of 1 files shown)


elinw:
Long run it might be better to set the options  array elements template, file, params individually so that applications don't have to repeat them, but I'll wait for feedback on that.

PR #547, opened 2011-11-22
Add JOIN support to JDatabaseQuery UPDATE query

mbabker:
JDatabaseQuery::__toString() currently does not handle JOIN statements in the UPDATE case.  This causes UPDATE queries that are joined to another table to fail due to the table not being joined.  Added support for this as well as a unit test testing the toString update case.


eddieajau:
Good one.  I didn't know you could do that.  Thanks!

PR #553, opened 2011-11-24
Output only one deprecation warning for JRequest.

realityking:
Don't load JRequest if _JREQUEST_NO_CLEAN is set.

### DIFF
--- a/libraries/joomla/environment/request.php
+++ b/libraries/joomla/environment/request.php
@@ -23,6 +23,8 @@ define('JREQUEST_NOTRIM', 1);
 define('JREQUEST_ALLOWRAW', 2);
 define('JREQUEST_ALLOWHTML', 4);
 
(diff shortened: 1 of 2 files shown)


realityking:
I also added a small deprecation notice for some Joomla 1.0 compatibility stuff.

libraries/joomla/environment/request.php
```diff
@@ -23,6 +23,8 @@ define('JREQUEST_NOTRIM', 1);
 define('JREQUEST_ALLOWRAW', 2);
 define('JREQUEST_ALLOWHTML', 4);
 
+JLog::add('JRequest::getURI() is deprecated.', JLog::WARNING, 'deprecated');
```
LouisLandry: This probably shouldn't just reference JRequest::getURI() ... it should reference the entire class.

realityking: That's what you get for doing a bad copy & paste job. Fixed.

PR #561, opened 2011-11-27
Remove the stub files at the old locations for JString and JLog.

realityking:
The autoloader picks them up at the right location anyways.


eddieajau:
Should we do this now or after 11.4?

realityking:
I'll leave that for you guys to decide. The autoloader should pick up the new location without issues, but anyone who has a require or include for that file would be out of luck.

robschley:
JString has lived in its current home since August. Not sure how long we should wait before removing these stubs.

https://github.com/joomla/joomla-platform/blob/7ac88ec3e4cb352e3099a428c71139ed7cc601ac/libraries/joomla/string/string.php

realityking:
Well it hasn't made it to the CMS until a week ago.

IMO the question is how many people use require instead of jimport()

eddieajau:
Let's leave it till after 11.4 (whoever merges, remove annotation in title please).

PR #563, opened 2011-11-28
Fix ? Relative paths used in unit tests are compared to absolute paths

elkuku:
I had problems running this test on my system.
When the test is executed on its own it runs fine, as only relative paths are used. But when run inside a suite it fails.

I hope this is correct, maybe I missed something else more obvious from the setup instructions ?

### DIFF
--- a/tests/suite/joomla/html/JHtmlTest.php
+++ b/tests/suite/joomla/html/JHtmlTest.php
@@ -600,7 +600,7 @@ class JHtmlTest extends JoomlaTestCase
 		// we do a test for the case that the js is in the templates directory
 		JHtml::script($urlpath . $urlfilename, false, true);
 		$this->assertArrayHasKey(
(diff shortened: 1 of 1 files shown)


elkuku:
Well, at least the changes don't break joomla-jenkins tests ;-)

elinw:
This was failing for me too, due to 
https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/factory.php#L856

Which is I think because there is no JRequest loading, but I thought it might make more sense to update the code not to use JRequest.

elkuku:
Do you mean you also get the fatal error on JRequest when running the whole suite ?
Unit tests are really strange beasts :P

elinw:
No you're right, I get yours when I run the whole suite and the JRequest when I run just the html.

elkuku:
After upgrading PHPUnit to version 3.6.4 these errors disappeared

PR #564, opened 2011-11-28
Code style: Mark protected member vars as deprecated

elkuku:
Mark protected member vars as deprecated if they are prefixed with an underscore. Add proper pendant.
Package: application

I used:
`@deprecated use $XXX or declare as private`
Not sure as of which version to deprecate - please advise.
It should also be verified if the var could be declared as private, including the underscore.

This will fix 45 code style warnings.

### DIFF
--- a/libraries/joomla/application/component/helper.php
+++ b/libraries/joomla/application/component/helper.php
@@ -24,6 +24,15 @@ class JComponentHelper
 	 * @var    array
 	 * @since  11.1
 	 */
+	protected static $components = array();
+
+	/**
+	 * The component list cache
+	 *
+	 * @var    array
+	 * @since  11.1
+	 * @deprecated use $components declare as private
+	 */
 	protected static $_components = array();
 
 	/**
@@ -38,11 +47,11 @@ class JComponentHelper
 	 */
 	public static function getComponent($option, $strict = false)
 	{
-		if (!isset(self::$_components[$option]))
+		if (!isset(self::$components[$option]))
 		{
 			if (self::_load($option))
 			{
-				$result = self::$_components[$option];
+				$result = self::$components[$option];
 			}
 			else
 			{
@@ -53,7 +62,7 @@ class JComponentHelper
 		}
 		else
 		{
-			$result = self::$_components[$option];
+			$result = self::$components[$option];
 		}
 
 		return $result;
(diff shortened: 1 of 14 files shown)


realityking:
Not sure if this is a good idea (and it's kinda cheating too). IMO we should deprecate them when we know whatever we wanna make them private or if we wanna rename them. Just adding the note we'll do this or that really isn't helpful for devs and just hides the problem.

elkuku:
Well, yeah.. I believe it is not more than an "interim solution", I wouldn't call it "cheating" ;) but I believe it should be good enough to make "remember" that there is still some work to do.
When it comes to removing/changing the deprecated stuff some more careful thinking will be required on how to proceed. On the other hand, there are very few private methods/properties right now, so some of these might be good candidates.
It is also an attempt to continue @chdemko's work on #502

eddieajau:
I know it's the PEAR standard, and it probably made sense when using PHP4, but I'm starting to wonder whether it makes sense to have the underscore rule for private variables.

elkuku:
You mean private variables should also go without underscore ? I would totally agree as I personally hate those underscores.
It would be a slight modification of the sniff, and some more warnings.

The intent of this pull is to "mask" those warnings, so they won't appear in the "reports" @joomla-jenkins kindly generates for every pull request, while marking vars and methods as deprecated so they could be changed when there it's "time".

eddieajau:
Yes, remove the underscores.  I came across a case where I was using some smart get methods, but it wouldn't work with private method name because they "had" to have an underscore prefix.  Probably a discussion for the list for another day.

elkuku:
So would it be OK to continue with the "rest" ? Just to be absolutely clear...
1) Mark the **protected** vars and methods as deprecated, as in this pull.
2) Leave the **private** vars and methods as they are - for now ?

eddieajau:
Yes.

libraries/joomla/application/component/helper.php
```diff
@@ -24,6 +24,15 @@ class JComponentHelper
...
+	protected static $components = array();
+
+	/**
+	 * The component list cache
+	 *
+	 * @var    array
+	 * @since  11.1
+	 * @deprecated use $components declare as private
+	 */
 	protected static $_components = array();
```
realityking, line 36: What's the use of this variable now? The code in this class doesn't use it.

eddieajau, line 36: Probably isn't any use for it anymore.

PR #565, opened 2011-11-28
Joomla CMS [#27246] JFormFieldTimezone default value not set

radiant-tech:
Server config value is set to wrong variable name.  See CMS Bug Tracker #27246.

### DIFF
--- a/libraries/joomla/form/fields/timezone.php
+++ b/libraries/joomla/form/fields/timezone.php
@@ -55,7 +55,7 @@ class JFormFieldTimezone extends JFormFieldGroupedList
 		// If the timezone is not set use the server setting.
 		if (strlen($this->value) == 0)
 		{
-			$value = JFactory::getConfig()->get('offset');
+			$this->value = JFactory::getConfig()->get('offset');
 		}
 
 		// Get the list of time zones from the server.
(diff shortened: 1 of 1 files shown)


libraries/joomla/form/fields/timezone.php
```diff
@@ -55,7 +55,7 @@ class JFormFieldTimezone extends JFormFieldGroupedList
 		// If the timezone is not set use the server setting.
 		if (strlen($this->value) == 0)
 		{
-			$value = JFactory::getConfig()->get('offset');
+			$this->value = JFactory::getConfig()->get('offset');
```
chdemko, line 61: I agree that old code was buggy, but the new one introduce a new behavior: in the cms, when you edit a user in the administrator after applying the patch, its timezone is set to the global one (which is different of keeping the value equal to 'Use default'). I suggest to replace the code by:

```
    $keyField = $this->element['key_field'] ? (string) $this->element['key_field'] : 'id';
    $keyValue = $this->form->getValue($keyField);

    // If the timezone is not set and this is a new item use the server setting.
    if (strlen($this->value) == 0 && empty($keyValue))
    {
        $this->value = JFactory::getConfig()->get('offset');
    }
```

PR #569, opened 2011-11-30
Documentation - Variable types

mbabker:
To better help developers, I've started reviewing the "undefined methods" inspection on PhpStorm and trying to update docblocks to more accurately reflect what each parameter type is and what is being returned.  Most of these changes are based on what is commented in the docblock or based on the instance type that is being returned.

### DIFF
--- a/libraries/joomla/html/parameter/element/calendar.php
+++ b/libraries/joomla/html/parameter/element/calendar.php
@@ -31,10 +31,10 @@ class JElementCalendar extends JElement
 	/**
 	 * Fetch a calendar element
 	 *
-	 * @param   string             $name          Field name
-	 * @param   string             $value         The date value
-	 * @param   JSimpleXMLElement  &$node         JSimpleXMLElement node object containing the settings for the element
-	 * @param   string             $control_name  Control name
+	 * @param   string       $name          Element name
+	 * @param   string       $value         Element value
+	 * @param   JXMLElement  &$node         JXMLElement node object containing the settings for the element
+	 * @param   string       $control_name  Control name
 	 *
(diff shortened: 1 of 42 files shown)


libraries/joomla/html/parameter/element/calendar.php
```diff
@@ -31,10 +31,10 @@ class JElementCalendar extends JElement
 	/**
 	 * Fetch a calendar element
 	 *
-	 * @param   string  $name          Field name
-	 * @param   string  $value         The date value
-	 * @param   object  &$node         JSimpleXMLElement node object containing the settings for the element
-	 * @param   string  $control_name  Control name
+	 * @param   string             $name          Field name
+	 * @param   string             $value         The date value
+	 * @param   JSimpleXMLElement  &$node         JSimpleXMLElement node object containing the settings for the element
```
elkuku: I believe this should be JXMLElement instead of JSimpleXMLElement

mbabker: Possibly.  I'd have to get an environment set up where I'm calling JElement instances to accurately figure it out, but I'm hoping that in this instance (and the other few instances that are "properly" commented), the comment stating it's a JSimpleXMLElement is correct.

elkuku: I believe the others are also incorrect - I just don't wanted to spam ;)
JSimpleXMLElement has been deprecated and the functionality is now covered by JXMLElement which extends SimpleXMLElement (the PHP one).

eddieajau:
Thanks Michael.  That's awesome work.

PR #570, opened 2011-11-30
Fix a problem with the close method in the MySQLi driver.

elinw:
If no actual connection exists (even though the object exists) and you try to close the connection you get a warning. This adds a check to see if the connection is null before trying to close.

### DIFF
--- a/libraries/joomla/database/database/mysqli.php
+++ b/libraries/joomla/database/database/mysqli.php
@@ -155,7 +155,7 @@ class JDatabaseMySQLi extends JDatabase
 	 */
 	public function __destruct()
 	{
-		if ($this->get('connection') != null && is_object($this->connection))
+		if (is_resource($this->connection))
 		{
 			mysqli_close($this->connection);
 		}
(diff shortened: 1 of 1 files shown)


eddieajau:
I think it would be better to add $connection in JDatabaseMySQLi as a class property then it always exists.  Something like:

```
/**
 * The database connection resource.
 *
 * @var object
 * @since 11.4
 */
protected $connection;
```

elinw:
ok

elinw:
Andrew the problem is that the object does exist but it is empty.

elinw:
This is a better solution I think.

elinw:
I looked at the build site and this checkstyle error is not related to the file this change is in.

elinw:
oops closed by accident.

eddieajau:
I've merged, but I think it's still worth declaring that connection property explicitly in the class.

elinw:
Yes I think I agree with that.  Will do a separate request.

eddieajau:
Going to have to revert this as it causes a regression failure in the CMS.

elinw:
Andrew it also causes failure as it is, so I remade the pull request with the original proposal which is just to additionally check that the $connection is not null in addition to checking if it is an object. No one ever responded on list with a better suggestion

PR #571, opened 2011-12-01
Code style: Sniff concatenation

elkuku:
This fixes a small error in the concat operator sniff as well as some violations that had been escaped - until now ;)


eddieajau:
Good stuff.

PR #572, opened 2011-12-01
Code style: Sniff operator spacing

elkuku:
This ensures that operators are surrounded by one space.


libraries/joomla/application/component/controlleradmin.php
```diff
@@ -225,7 +225,7 @@ class JControllerAdmin extends JController
 
 		// Initialise variables.
 		$ids = JRequest::getVar('cid', null, 'post', 'array');
-		$inc = ($this->getTask() == 'orderup') ? -1 : +1;
+		$inc = ($this->getTask() == 'orderup') ? - 1 : + 1;
```
eddieajau: Maybe it's just me, but I don't like the space where it applies to a negative number.

elkuku:
Actually I wasn't quite sure myself when I changed it.
There was a man who drove his car into a lake because his navigator told him to go straight. Don't trust the machines ;)

So I changed the machine, err the sniff, to look if there is a ternary "?" before the operator, and also that people might use a + sign for a positive number ;)

eddieajau:
Just a side note, I'm fixing up all the missing "public" stuff in the functions.

elkuku:
I think this is related to one of the other pulls ;)
But yes, good stuff. When it is done I believe the sniff should give an error instead of a warning on missing scope modifiers.

eddieajau:
Sounds good.

LouisLandry:
Is this going to allow for people to have assignment operations line up?

``` php
$foo    = 'bar';
$foobar = 'snafu';
$bar    = 'high';
```

?

elkuku:
Yes, this is still allowed.

LouisLandry:
Great, thanks.  Is it worth going over things and also fixing all the new issues here?  There appears to be at least one error.

elkuku:
Well, I would be curious about honorable Mr. Jenkins error.. I ran my "personal" jenkins, without errors. Could be a hickup :P
http://elkuku.github.com/pulltester/pulls/572

> all the new issues ?

I merged the master, found 1 (one) new error, fixed it, pushed, let's see what Mr. Jenkins says ;)

PR #573, opened 2011-12-01
Code style: Sniff cast spacing

elkuku:
This ensures that casts do not contain spaces and are followed by one space.


eddieajau:
Nice one.  That's been annoying me.

PR #576, opened 2011-12-01
Updates to docs directory

mbabker:
- Some grammar checks
- Updated CHANGELOG


eddieajau:
Michael, I apologise in advance. Ian's been working on a revised format that we can publish and I needed to pull that first.  Subsequently, I've thrown you pull out of whack.  Also, with the changelog, since we've got good HTML docs now I think what we'll do is just change that to a "revision history" page and link out to the changelog articles on developer.joomla.org for the full version.  How does that sound?  I'm sorry if this means some/a lot of your work has been wasted.

mbabker:
It isn't a waste.  Honestly, I'm sitting in the middle of nowhere right now with nothing better to do, so little things like this are basically just time killers for me at the moment.

mbabker:
OK, I'm merged up to master and got the changes.  I think everything's correct in terms of the change set.  The only thing I can't confirm 100% accuracy on is the changelog.xml file since the diff doesn't show because of the number of changes.

eddieajau:
Thanks - merged. I'll fix up that changelog later.

PR #577, opened 2011-12-01
Refactor JDatabaseMySQLi + Unit Tests

mbabker:
In a bit of a radical move, I've refactored JDatabaseMySQLi to extend JDatabaseMySQL.  Since these drivers build queries in a very close manner, and each other MySQLi class extends the MySQL companion, this helps to eliminate duplicate code and allows the focus to be solely on driver differences (calling mysqli_ versus mysql_ functions primarily).

In addition to this change, I've implemented a set of tests for JDatabaseMySQLi, similar in nature to JDatabaseMySQL's own tests.  In the few tests that have been created, there are no apparent major issues from refactoring the mysqli driver in this way.

Lastly, a bug in JDatabaseMySQL::getTableKeys was fixed in that it was incorrectly referencing `$this->db->quoteName` (db reference not needed here).


LouisLandry:
I really like this approach.  Hopefully I can take a deeper look at it soon and get it merged in, or one of the other maintainers can do the same sooner than I can get to it.  I just have limited time this week.

eddieajau:
Looks like a winner to me.

PR #579, opened 2011-12-02
Fix up missing method scope declarations.

eddieajau:
Fixes up new warnings generated by a new sniff.


elkuku:
Actually Jenkins is failing because ..
_total number of annotations exceeds the threshold 500:_
which is related to #575 
Should be good after merging this one

PR #583, opened 2011-12-02
Change the way core.js is loaded.

realityking:
Since core.js requires MooTools anyways we don't need to be able to load it without MooTools.


eddieajau:
Done, thanks.

PR #584, opened 2011-12-02
JPath::check() throws Error 20 when running in a chroot jail

nikosdion:
When JPATH_ROOT is an empty string –i.e. we run inside a chroot jail– JPath::check() always raises error 20 (snooping out of bounds). By definition of the chroot jail, we can snoop out of bounds in this case. Therefore this check should always succeed. Ergo the proposed change.

### DIFF
--- a/libraries/joomla/filesystem/path.php
+++ b/libraries/joomla/filesystem/path.php
@@ -177,7 +177,7 @@ class JPath
 		}
 
 		$path = JPath::clean($path);
-		if (strpos($path, JPath::clean(JPATH_ROOT)) !== 0)
+		if (!empty(JPATH_ROOT) && strpos($path, JPath::clean(JPATH_ROOT)) !== 0)
 		{
 			// Don't translate
 			JError::raiseError(20, 'JPath::check Snooping out of bounds @ ' . $path);


nikosdion:
For more information on a real-world case, please take a look at this support thread on my forum: https://www.akeebabackup.com/support/forum/akeeba-backup-support/admin-cant-access-to-backend-after-restore-in-other-directory/54376.html#p54376 The user gave me FTP access, so I was able to track the problem on his server. This problem is practical. I have come across numerous shared hosts running in chroot jails over the last five years and, right now, when the users of those hosts try to install Joomla! they will believe that newer versions of Joomla! are broken (AFAIK, Joomla! 1.5 did work on those hosts, albeit very quirky).

eddieajau:
Thanks - merged. I think ultimately we need to make that more configurable because the idea of JPATH_ROOT is only directly applicable to the CMS.

elkuku:
BTW..... I believe the jenkins tests are failing (mine also) because it seems that you should use the language construct [empty()](http://www.php.net/manual/en/function.empty.php) on variables only:

> empty() only checks variables as anything else will result in a parse error.

I get a :
`Parse error: syntax error, unexpected ')', expecting T_PAAMAYIM_NEKUDOTAYIM  in .../libraries/joomla/filesystem/path.php on line 180`

Let me say: WOW `=;)`

eddieajau:
_sigh_ I need to clean my glasses.  Can my logic is correct with this change:

```
if (constant('JPATH_ROOT') && strpos($path, JPath::clean(JPATH_ROOT)) !== 0)
```

nikosdion:
Oops! Writing code after midnight directly on GitHub doesn't help making sure it's correct. I should actually have typed:
if( (JPATH_ROOT != '') && strpos($path, JPath::clean(JPATH_ROOT)) !== 0) 
The problem is not whether JPATH_ROOT is defined or not (it is defined), but if it equals to an empty string.

## 

Nicholas K. Dionysopoulos
Lead Developer, AkeebaBackup.com
Web: http://www.AkeebaBackup.com
Blog: http://www.dionysopoulos.me/blog

eddieajau:
I was just as blind :)  Nic, can you make a new pull to fix this.  Thanks.

nikosdion:
Made pull request #594 to fix this. Thanks!

PR #585, opened 2011-12-02
Add missing type hints to newer classes.

realityking:
Add missing type hints to newer classes.

Does it make sense to check older classes too or should I wait for 11.4 to be released?

### DIFF
--- a/libraries/joomla/application/input/cookie.php
+++ b/libraries/joomla/application/input/cookie.php
@@ -28,7 +28,7 @@ class JInputCookie extends JInput
 	 *
 	 * @since   11.1
 	 */
-	public function __construct($source = null, $options = array())
+	public function __construct(array $source = null, array $options = array())
 	{
 		if (isset($options['filter']))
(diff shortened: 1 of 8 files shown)


libraries/joomla/application/input/cookie.php
```diff
@@ -28,7 +28,7 @@ class JInputCookie extends JInput
 	 *
 	 * @since   11.1
 	 */
-	public function __construct($source = null, $options = array())
+	public function __construct(array $source = null, array $options = array())
```
robschley, line 31: If $source needs to be an array, null should not be the default value.

realityking, line 31: As far as PHP is concerned that's absolutely fine. http://php.net/manual/en/language.oop5.typehinting.php

robschley, line 31: You're right. I'm not thinking today... we do this all the time.

PR #588, opened 2011-12-03
Rename filterinput.php and filteroutput.php files to suit autoloader.

chdemko:
The new autoloader allows to rename these files


realityking:
Did you use git mv to move the two files? It'd be good to keep the history on those files intact.

chdemko:
yes

realityking:
Weird, github isn't registering this as moved files. And the history for the new files have just one entry (https://github.com/chdemko/joomla-platform/commits/filter/libraries/joomla/filter/input.php)

robschley:
It seems to have done the same thing for https://github.com/joomla/joomla-platform/pull/589 and I used git mv.

chdemko:
I think it's because we have written some new lines in the old files

realityking:
Maybe we'll have to do those in two commits, sucks loosing history.

On a related note, do we rally need those stubs? I really wonder how many people use require instead of jimport().

PR #589, opened 2011-12-03
Moved string.stringnormalize to string.normalize.

robschley:


eddieajau:
If we are moving the file, do you want to fix up the "z" as well?

robschley:
Yes, I can do that too.

robschley:
Should be done now.

robschley:
I fixed the pull request not being able to merge correctly. Should be ready to go now.

PR #592, opened 2011-12-03
Database Package Refactoring

orware:
Significant refactoring of the database package. Reorganized the various types of objects into folders (driver, query, exporter, importer). Added in a new base PDO class that can be extended from rather easily to create new drivers. As an example, a new Oracle PDO-based driver has been created. A new Preparable Interface for Queries has been added, allowing you to bind variables to a query. A new Limitable Interface for Queries has been added, allowing the SQL query to have the special (usually Database Vendor specific) limiting method added to the SQL string.


elinw:
We're not using index.html files in the platform.

libraries/joomla/database/database.php
```diff
@@ -1524,12 +1605,18 @@ public function debug($level)
 	/**
 	 * Diagnostic method to return explain information for a query.
 	 *
-	 * @return  string  The explain output.
+	 * @return      string  The explain output.
 	 *
 	 * @deprecated  12.1
-	 * @since   11.1
```
elinw, line 1530: I'm not sure why @since would change.

libraries/joomla/database/database.php
```diff
@@ -1406,27 +1466,11 @@ public function setQuery($query, $offset = 0, $limit = 0)
 	 * @return  void
 	 *
 	 * @since   11.1
-	 * @throws  JDatabaseException
+	 * @throws  RuntimeException
 	 */
 	abstract public function transactionStart();
 
 	/**
-	 * Method to truncate a table.
```
elinw, line 1414: Why is this being removed? This was rather important for postgres support if i understand correctly.

libraries/joomla/utilities/garbagecron.txt
```diff
@@ -0,0 +1,23 @@
+<?php
```
elinw, line 1: I believe this has been moved.

libraries/joomla/database/driver/mysqli.php
```diff
@@ -0,0 +1,754 @@
...
+	}
+
+	/**
+	 * Destructor.
+	 *
+	 * @since   11.1
+	 */
+	public function __destruct()
+	{
+		if (is_object($this->connection))
```
elinw, line 154: I have a pull request in for this already, this should be is_resource I believe.
update yes that change was merged.

elinw:
:)  Really nice

Obviously you need to sync up with the current build of the platform and take out the index files etc but my bigger question is shouldn't each removed method be calling its replacement? Or is that being handled programatically (it's a lot of code, I could easily be missing something)?

libraries/joomla/database/database.php
```diff
@@ -173,7 +173,7 @@ public static function getConnectors()
 		$connectors = array();
 
 		// Get a list of types.
-		$types = JFolder::files(dirname(__FILE__) . '/database');
+		$types = JFolder::files(dirname(__FILE__) . '/driver');
```
elinw, line 176: The databaseconnection field should be updated to use this as part of this commit not only because it is cleaner but because moving the files .will break the field

orware:
Hi Elin, I did try and do a merge from the platform's staging branch prior to adding my database refactoring commit (it was just a single one at the end of all of the craziness above). I didn't actually want this pull request to include all of those other commits, but the pull request is including all of them. I still don't know enough about git to know why it does that.

Thank you for your comments...I haven't tried to resolve any of them aside from removing the index.html files from the folders in the database package, mainly because I don't know what the right answers are for those other questions.

realityking:
I think your branch is based on the repro before it got rebased. Ian may know how to fix it.

elinw:
Yeah, I have had pull requests like this before.
I think you can rebase this branch onto another branch that you create  with a clean pull. You should be able to just use the 
https://github.com/orware/joomla-platform/commit/413e2321a686f62513123eb13a12a2dbcde51b23
commit.
 I'm not positive on all this but I also think it is better to use git mv to move files around.

orware:
Since I didn't put together this pull request cleanly from the platform's master branch I created a new one this morning and cherry picked the two previous commits into it in the following new pull request:
https://github.com/joomla/joomla-platform/pull/610

@elin if any of your previous comments still apply to the new pull request, please let me know and I'll try and make any needed changes.

PR #593, opened 2011-12-03
Fix build.xml

mbabker:
I may or may not have inadvertently broken the phpcs target in build.xml by missing a comma, so pointed out by @elkuku.  Whoops.


robschley:
Merged.

PR #596, opened 2011-12-03
Fix some more PHPMD warnings

realityking:
and deprecate some stuff so they'll be fixed in the near future.

### DIFF
--- a/libraries/joomla/installer/installer.php
+++ b/libraries/joomla/installer/installer.php
@@ -143,8 +143,22 @@ class JInstaller extends JAdapter
 	 * @return  boolean  Allow overwrite switch
 	 *
 	 * @since   11.1
+	 * @deprectaed 12.1 Use JInstaller::isOverwrite()
 	 */
 	public function getOverwrite()
+	{
+		JLog::add('JInstaller::getOverwrite() is deprectaed. Please use JInstaller::isOverwrite() instead', JLog::WARNING, 'deprecated');
+		return $this->isOverwrite();
+	}
+
+	/**
+	 * Get the allow overwrite switch
+	 *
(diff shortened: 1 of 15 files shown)


libraries/joomla/installer/installer.php
```diff
@@ -143,9 +143,23 @@ public static function getInstance()
 	 * @return  boolean  Allow overwrite switch
 	 *
 	 * @since   11.1
+	 * @deprectaed 12.1 Use JInstaller::isOverwrite()
 	 */
 	public function getOverwrite()
 	{
+		JLog::add('JInstaller::getOverwrite() is deprectaed. Please use JInstaller::isOverwrite() instead', JLog::WARNING, 'deprecated');
+		return self::isOverwrite();
```
chdemko: should it be

```
return $this->isOverwrite();
```

?

libraries/joomla/installer/installer.php
```diff
@@ -206,9 +220,23 @@ public function setRedirectURL($newurl)
 	 * @return  boolean
 	 *
 	 * @since   11.1
+	 * @deprectaed 12.1 Use JInstaller::isUpgrade()
 	 */
 	public function getUpgrade()
 	{
+		JLog::add('JInstaller::getUpgrade() is deprectaed. Please use JInstaller::isUpgrade() instead', JLog::WARNING, 'deprecated');
+		return self::isUpgrade();
```
chdemko: should it be

```
return $this->isUpgrade();
```

?

realityking: fixed

libraries/joomla/installer/installer.php
realityking: fixed

PR #597, opened 2011-12-03
Test if a constant exists.

elinw:
Modify the test that checks whether JPATH_ROOT exists.


mbabker:
@nikosdion already has a fix for this (introduced in #584) in #594.

elinw:
hm I'm not sure which  I think is better.

ianmacl:
This has already been fixed in the aforementioned pull request.

PR #599, opened 2011-12-03
Add a connection property to the  MySQLi driver

elinw:


elinw:
As per Andrew's request https://github.com/joomla/joomla-platform/pull/570

ianmacl:
Closing this because the $connection property isn't needed - it is already defined in the parent class.

PR #600, opened 2011-12-03
Code style: Fix a superfluous tab character

elkuku:
This will remove a tab character - not more not less...

This is the "1 error" that joomla-jenkins shows on all pull requests.


eddieajau:
Thanks.

PR #603, opened 2011-12-04
Replace JFormFieldEditors with a more general solution.

realityking:
While working on integrating the captcha patch into the current trunk I found this little jewel.

### DIFF
--- a/libraries/joomla/form/fields/plugins.php
+++ b/libraries/joomla/form/fields/plugins.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ * @version		$Id: editors.php 13698 2009-12-11 18:38:48Z eddieajau $
+ * @copyright	Copyright (C) 2005 - 2009 Open Source Matters, Inc. All rights reserved.
+ * @license		GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+defined('JPATH_BASE') or die;
+
+jimport('joomla.html.html');
+require_once dirname(__FILE__).DS.'list.php';
+
+/**
+ * Form Field class for the Joomla Framework.
+ *
+ * @package		Joomla.Framework
+ * @subpackage	Form
+ * @since		1.6
+ */
+class JFormFieldPlugins extends JFormFieldList
+{
+	/**
+	 * The field type.
+	 *
+	 * @var		string
+	 */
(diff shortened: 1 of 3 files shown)


libraries/joomla/form/fields/plugins.php
```diff
@@ -0,0 +1,87 @@
...
+ * @subpackage Form
+ * @since      11.4
+ */
+class JFormFieldPlugins extends JFormFieldList
+{
+	/**
+	 * The field type.
+	 *
+	 * @var		string
+	 * @since   11.4
```
eddieajau: Should be 2 spaces after @since

eddieajau:
Rouven, would you mind adding the test skeleton file with this pull.  Otherwise a very smart idea.

realityking:
I added the same tests that we had for JFormFieldEditors. I hope they work.

realityking:
I'm a bit lost here where the three checkstyle errors come from

According to http://elkuku.github.com/pulltester/pulls/603 they are:

libraries/joomla/form/fields/plugins.php:18
@package tag comment indented incorrectly; expected 4 spaces but found 4
libraries/joomla/form/fields/plugins.php:19
@subpackage tag comment indented incorrectly; expected 1 spaces but found 1
libraries/joomla/form/fields/plugins.php:20
@since tag comment indented incorrectly; expected 6 spaces but found 6

which just doesn't make any sense.

libraries/joomla/form/fields/plugins.php
```diff
@@ -0,0 +1,87 @@
...
+defined('JPATH_PLATFORM') or die;
+
+jimport('joomla.form.formfield');
+JFormHelper::loadFieldClass('list');
+
+/**
+ * Form Field class for the Joomla Framework.
+ *
+ * @package    Joomla.Platform
+ * @subpackage Form
```
elkuku: I think you should use 2 spaces instead of 1 after @subpackage and align the others accordingly.
But yes, the test is fishy - `@FIXME`

realityking: Allright, let's try that.

realityking: Yay, that worked!

robschley:
Merged.

PR #604, opened 2011-12-04
Use less deprecated functions in the platform

realityking:
This should make a start in getting rid off deprecated stuff. This handles the following three functions:
JError:isError()
JUtility::getHash()
JUtility::getToken()

The first one can be replaced with the instanceof operator, the other two just moved to a different class.


eddieajau:
Jenkins reports 1 test error.

realityking:
Should be fixed now.

PR #605, opened 2011-12-04
strict static

ghost:
strict static should be called static


eddieajau:
I'm pretty sure all the methods are supposed to be used statically.  Would you mind doing that and fixing up all the methods if that is the case.  Thanks in advance.

PR #607, opened 2011-12-05
Batch Processing work for CMS 2.5

mbabker:
To coincide with joomla/joomla-cms#29, this pull request merges in all library work for this feature improvement.

Covered specifically in the libraries:

JControllerForm:
- batch() - Added logic to build an array of contexts to improve ACL permissions checks, converted JRequest calls to JInput

JModelAdmin:
- batch() - Handle receiving and forwarding contexts array to all batch functions
- batchAccess() - Rewritten to check edit permission on the item being edited (no ACL checks previously)
- batchCopy() - Converted JRequest to JInput, create permission check now looks at the category permissions where the item is being copied to
- batchLanguage() - Rewritten to check edit permission on the item being edited (no ACL checks previously)
- batchMove() - Converted JRequest to JInput, create permission check now looks at the category permissions where the item is being copied to, edit permission check uses contexts array to check item permissions

JHtmlBatch:
- user() - Added user widget to display a list of users

JHtmlUser:
- Added this class to display either a list of groups or users

Tests added for JHtmlBatch::user() and JHtmlUser methods.

### DIFF
--- a/libraries/joomla/application/component/controllerform.php
+++ b/libraries/joomla/application/component/controllerform.php
@@ -252,11 +252,28 @@ class JControllerForm extends JController
 	public function batch($model)
 	{
 		// Initialise variables.
-		$vars = JRequest::getVar('batch', array(), 'post', 'array');
-		$cid = JRequest::getVar('cid', array(), 'post', 'array');
+		$input	= JFactory::getApplication()->input;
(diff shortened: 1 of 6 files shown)


eddieajau:
Thanks. Merged.

PR #609, opened 2011-12-05
Deprecation Work - Replace JException

mbabker:
This pull request removes most instances of JException in favor of Exception.  All but the Form package are handled because I couldn't sort out where the unit test suite was failing with this change in that package.

### DIFF
--- a/libraries/joomla/database/table.php
+++ b/libraries/joomla/database/table.php
@@ -133,7 +133,7 @@ abstract class JTable extends JObject
 
 			if (empty($fields))
 			{
-				$e = new JException(JText::_('JLIB_DATABASE_ERROR_COLUMNS_NOT_FOUND'));
+				$e = new Exception(JText::_('JLIB_DATABASE_ERROR_COLUMNS_NOT_FOUND'));
 				$this->setError($e);
 				return false;
 			}
@@ -425,13 +425,14 @@ abstract class JTable extends JObject
 	 *
 	 * @link    http://docs.joomla.org/JTable/bind
 	 * @since   11.1
+	 * @throws  Exception
 	 */
 	public function bind($src, $ignore = array())
 	{
 		// If the source value is not an array or object return false.
 		if (!is_object($src) && !is_array($src))
 		{
-			$e = new JException(JText::sprintf('JLIB_DATABASE_ERROR_BIND_FAILED_INVALID_SOURCE_ARGUMENT', get_class($this)));
+			$e = new Exception(JText::sprintf('JLIB_DATABASE_ERROR_BIND_FAILED_INVALID_SOURCE_ARGUMENT', get_class($this)));
 			$this->setError($e);
 			return false;
 		}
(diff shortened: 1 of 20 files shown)


libraries/joomla/database/table.php
```diff
@@ -118,6 +118,9 @@ public function __construct($table, $key, &$db)
 	 * Get the columns from database table.
 	 *
 	 * @return  mixed  An array of the field names, or false if an error occurs.
+	 *
+	 * @since   11.1
+	 * @throws  Exception
```
realityking: The exception is never actually throw is it? It's just returned.

realityking:
Have you tested this with the CMS? JException seems to offer some methods Exception doesn't, I wonder if they are actually used.

mbabker:
Looked through the CMS.  In instances where JException is used, it only uses the first two params, which are the same as Exception.  There may be an instance or two where the catch statement needs to be changed, otherwise, there should be no issues.

libraries/joomla/database/table.php
eddieajau: Aye. Should remove the @throws.

LouisLandry:
This is fantastic.  Can you give me an idea of what you were running into problems with in the form package?  I may be able to help.

mbabker:
The `$valid->get('level')` switch at https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/form/form.php#L1119

This might work out correctly now that #604 has been merged.

LouisLandry:
Ah.

That probably needs to be rewritten to look more like:

``` <?php
    // Attempt to validate the field.
    try
    {
        $this->validateField($field, $group, $value, $input);
    }
    // There was an invalid field due to unexpected value, etc.
    catch (RuntimeException $e)
    {
        array_push($this->errors, $e);
        $return = false;
    }
    // There was a problem actually validating the field.
    catch (LogicException $e)
    {
        JError::raiseWarning(0, $e->getMessage());
        return false;
    }
```

What this would require is that form fields actually throw exceptions instead of returning them.  I don't know that we can just do that right now, but we can certainly mark some stuff deprecated and start handling that case.

realityking:
@mbabker
It should have helped. The code I substituted is 1:1 what is in JError::isError()

realityking:
How about this for 11.4:

```
// Attempt to validate the field.
try
{
    $valid = $this->validateField($field, $group, $value, $input);

    // @deprecated 12.1
    if ($valid instanceof JException)
    {
        JLog::add('JFormRule subclasses should not return JExceptions. Instead either a RuntimeException or LogicException should be thrown. ', JLog::WARNING, 'deprecated');
        switch ($valid->get('level'))
        {
            case E_ERROR:
                JError::raiseWarning(0, $valid->getMessage());
                return false;
                break;

            default:
                array_push($this->errors, $valid);
                $return = false;
                break;
        }
    }
}
// There was an invalid field due to unexpected value, etc.
catch (RuntimeException $e)
{
    array_push($this->errors, $e);
    $return = false;
}
// There was a problem actually validating the field.
catch (LogicException $e)
{
    JFactory::getApplication()->enqueueMessage($e->getMessage(), 'warning');
    return false;
}
```

PR #610, opened 2011-12-05
Database Package Refactoring (Clean Version)

orware:
Significant refactoring of the database package. Reorganized the various types of objects into folders (driver, query, exporter, importer). Added in a new base PDO class that can be extended from rather easily to create new drivers. As an example, a new Oracle PDO-based driver has been created. A new Preparable Interface for Queries has been added, allowing you to bind variables to a query. A new Limitable Interface for Queries has been added, allowing the SQL query to have the special (usually Database Vendor specific) limiting method added to the SQL string.


elkuku:
In case you feel lost about the 117 errors...
Here is some documentation: http://docs.joomla.org/Joomla_CodeSniffer
Please pay special attention to the note in usage related to the platform that is different to the (nonexistent) code style of the CMS  ;)
I have set up a **personal** page that might also help to resolve the checkstyle errors. About the unit tests - I am not sure as results seems to differ across different operating systems.
http://elkuku.github.com/pulltester/pulls/610 <= for reference only !

orware:
Thanks Nikolai, I've had trouble getting CodeSniffer to work and I haven't gone back to try again. If there's a quick way on somebody else's end to create the report for me I can correct the style errors and as many of the warnings as I can.

chdemko:
I would prefer to have a directory structure like this:

```
libraries/database/engine/mysql/server.php (JDatabaseEngineMysqlServer class)
                               /query.php (JDatabaseEngineMysqlQuery class)
                               /exporter.php (JDatabaseEngineMysqlExporter class)
                               /importer.php (JDatabaseEngineMysqlImporter class)
libraries/database/engine/postgres/server.php (JDatabaseEnginePostgresServer class)
                                  /query.php (JDatabaseEnginePostgresQuery class)
                                  /exporter.php (JDatabaseEnginePostgresExporter class)
                                  /importer.php (JDatabaseEnginePostgresImporter class)
```

This will allow the Joomla! autoloader to work without jimport and this will be easier for maintaining db engines (one folder for one engine).

orware:
The way the folders have been currently organized (and the files renamed),
aside from the ones in the table directory, since I didn't touch those,
should all conform to the new autoloader standard.

I don't agree to changing the word "driver" to "engine" as the driver
keyword more correctly signifies the purpose of those files and is the more
commonly used term.

Additionally, I know we have two different viewpoints on how to organize
the files into folders...the way that I've organized them, and then one
similar to the one you proposed.

I chose to organize them the way I did because it seemed more logical to me
(I didn't really need to change the class names at all, only the filenames
once I placed them in the correct folder), considering the types that I saw
(driver, query, importer, exporter type files) and because I liked the way
the driver type was at the very end of the class name instead of in the
middle, but that's just my preference.

If others feel strongly about one organization over the other it might be
good for them to chime to see if there's a strong argument for one over the
other (I simply wanted to organize things according to the new autoloader
without having to modify class names).

On Tue, Dec 6, 2011 at 6:40 AM, Christophe Demko <
reply@reply.github.com

robschley:
I don't like the "engine" approach either as engine is also used to define types of tables such as in MySQL you have MyISAM and InnoDB which are "engines" and we don't want to mix up that terminology.

chdemko:
'engine' or 'driver' as you want, my main concern was about files structure

ianmacl:
So the question really comes down class names:
JDatabaseDriverMysqlServer vs JDatabaseServerMysql
JDatabaseDriverPostgresQuery vs JDatabaseQueryPostgres

Personally I am in favour of the latter as it means we don't need to have the word driver (or engine) in all of our database classes which could make them more ambiguous as to their actual purpose.

elinw:
I'm with Ian on that one. I think it is more readable also.  I'm wondering though about what was raised by Andrew about grouping all the MySQL derived ones together  https://groups.google.com/forum/#!searchin/joomla-dev-platform/drizzle/joomla-dev-platform/cUMgx-351wQ/E4MYobCicdUJ ... I don't know how much of a real issue/convenience that is for any other database but for MySQL we have MySQLi and Drizzle and probably others many of which just have a small number of specific differences but his example of the possibility of different versions is also potentially important. Also Louis had a somewhat different approach earlier in that thread.

orware:
@ianmacl  With the current approach the word "driver" isn't currently in all of the database classes. Some examples:
JDatabaseQueryMySQL
JDatabaseQueryOracle
JDatabaseImporterMySQL
JDatabaseDriverMySQL
JDatabaseDriverOracle

The latest commit should fix all of the CodeSniffer errors...I'm going to take a look at the unit tests next.

eddieajau:
Do we introduce any significant backward compatibility issue with this?

libraries/joomla/database/driver/oracle.php
```diff
@@ -0,0 +1,404 @@
...
+	 * Note: Use query objects with bound variables instead!
+	 *
+	 * @param   string   $text   The string to be escaped.
+	 * @param   boolean  $extra  Optional parameter to provide extra escaping.
+	 *
+	 * @return  string  The escaped string.
+	 *
+	 * @since   11.4
+	 */
+	public function escape($text, $extra = false)
```
ianmacl: Should this be implemented before we pull to master?

orware:
@eddieajau The main things I could think of that would cause possible backwards compatibility issues would be if somebody directly instantiated JDatabaseMySQL classes in the past or something like that (since the autoloader needs the class to be called JDatabaseMysql to instantiate it correctly)...that's mainly because I removed registering the new file locations from all of the files in the database/database directory (I had them initially, but Louis recommended I remove them and just leave the JLog deprecated calls in). Existing MySQL stuff has basically remained untouched, though I have added some new functionality into JDatabaseQuery along with the new Preparable/Limitable interfaces.

@ianmacl I'd definitely love to implement the escape method properly for Oracle, but I simply just don't know exactly what needs to be done for that at the moment, though I could try researching it some more tonight. The main reason why it isn't as big of a concern with the Oracle driver is because prepared statements are used instead so escaping becomes unnecessary. For example:

```
    $query->setQuery('select * from spriden where spriden_pidm = :pidm');
    $query->bind(':pidm', 999999);
    $this->db->setQuery($query);
```

Or even reuse most of your query in a loop and just rebinding the variables:

```
    $insert = range(1, 100);

    $query->setQuery('INSERT INTO tablename (id, requests) VALUES (:idnumber, :requestnumber)');

    foreach($insert as $id)
    {
        $query->bind(':idnumber', $id);
        $query->bind(':requestnumber', $id);
        $this->db->setQuery($query);

        $this->db->execute();
    }
```

Since I think the Oracle class in particular will only be used in custom components with people wanting to tie into another system to retrieve data (mainly), using prepared statements like the ones above will pretty much be required...I don't see us running Joomla on top of Oracle in the near future, but I will try and find some additional information related to escaping and see if I can implement the method more accurately.

robschley:
The only thing I don't like is that there isn't an actual JDatabaseDriver class. I think if we're going to use the convention of JDatabaseDriverMysql, JDatabaseDriverOracle, JDatabaseDriverSqlsrv, etc. we should rename JDatabase to JDatabaseDriver.

elinw:
It seems like we're adding an extra layer of hierarchy for  no reason that has been explained.  is there something else besides a folder named driver that would go into the actual database folder? I think the way it is now shows the problems with not enough hierarchy, but if there is not something else anticipated it doesn't make sense for me.

The bigger question is whether it makes sense to organize by database or object type. I think tifhere is something to be said for having packages by database if the idea is that eventually people will pick the packages they need and leave the rest.

orware:
@robschley I was thinking along changing the name initially, but thought against renaming JDatabase for two reasons:
1. Mainly, I liked JDatabase over JDatabaseDriver
2. JDatabase feels more like a factory in a way for getting instances of any driver, query object, importer, exporter, etc. so that made me feel like it wasn't necessary to add the Driver bit at the end.

@elinw Here's a comparison between what we have currently:

```
joomla/
    database/
        database/
        table/
```

So in the current scenario we have driver classes, query classes, and the more recent importer/exporter classes all residing in the database/database folder.

In the new organization we have the following:

```
joomla/
    database/
        database/ (for the backwards compatible classes)
        driver/
        exporter/
        importer/
        query/
        table/
```

For me I don't feel that's much of an extra hierarchy, and I enjoy the fact that now all of the files are organized in a way that makes sense, at least for me. And actually, most of this reorganization is really related to making the needed changes so that the autoloader would work properly, and in a way that didn't modify much of the class names (aside from the driver ones, I didn't have to rename the query, exporter or importer classes...aside from the casing change I had to make the other day from "JDatabaseMySQL" to "JDatabaseMysql" since the autoloader uses camel casing as part of its convention for locating files.

elinw:
Yeah I see how you are thinking and that structure doesn't have the extra level. Happy to lose the database/database.

eddieajau:
Note added to title (remove on merge). We won't consider this until after 11.4.

gpongelli:
@orware 
Why don't you use "git mv" instead of moving files directly?
With former command you can keep history of moved files, that I think it's better.
Look at [my commit](https://github.com/joomla/joomla-platform/pull/532/files#diff-10), it is highlighted as
<b>tests/{ddl.sql → ddl_mysql.sql}</b>
instead of simply removing one file and adding the same with another path or name.
Hoping to see this changes in another pull request ;)

Bye,
Eng. Gabriele Pongelli

gpongelli:
Just to be clear, "git mv" allow other developer (maintainer of master branch too) to do a rebase without losing their changes on moved files instead of merging and have conflicts/painful state.
I've tried to merge this branch on a copy of my postgresql branch, the result was a conflict state.
Then I've tried rebasing these commits on another copy of my branch to keep my changes and add these one after mine, the result was conflict state worse than previous.

Using "git mv" instead, with rebasing --onto and then merge (as explained [here](http://learn.github.com/p/rebasing.html) ), the result is a fast-forward commit, without any conflict.

These are the commands I've done:

<pre><code># git branch base_br master
# git branch moving master
# git checkout moving
Switched to branch 'moving'
# cd libraries/joomla/database/database
# git mv mysql.php ../
# cd ../../
# git commit -m "moved file"
[moving 79f0b06] moved file
 1 files changed, 0 insertions(+), 0 deletions(-)
 rename libraries/joomla/database/{ => }/database/mysql.php (100%)
# git rebase --onto base_br base_br moving
Already on 'moving'
Current branch moving is up to date.
# git checkout base_br
Switched to branch 'base_br'
# git merge moving
Updating bdfe082..79f0b06
Fast-forward
 .../joomla/database/{ => }/database/mysql.php      |    0
 1 files changed, 0 insertions(+), 0 deletions(-)
 rename libraries/joomla/database/{ => }/database/mysql.php (100%)</code></pre>


As you can see there's no conflicts, it's a clear solution and it will not create problems to other developer's workspace ;)

Eng. Gabriele Pongelli.

elinw:
Currently this pull request is out of sync with a number of impacted files. 
I would like to make the following proposal.
This should be broken into two  phases or steps.
Step one should be the introduction of the new code in the present location, database/database.
Once those are merged do Step Two which would be to move all of the files to the new locations without making anadditional changes in the same action as moving.
Merge those.
Then we will not have to worry about regressions due to having to hand merge all f the changes from the current libraries not to mention the loss of history. Done carefully I think it can be seamless. Personally, if the code is good I have no way of testing) I have no problem at all with having additional classes and methods available in 11.4, but  I think rearranging thngs probably rightly should be done after 11.4. It just adds a level of complexity that is unnecessary in the short term.

gpongelli:
I agree on Elin thought, hoping that moving action will be done with "git mv" ;) .

chdemko:
Fixed in #906

PR #608, opened 2011-12-05
Add the highlighter behavior

realityking:
This is based on Michaels pull request #371. I made some changes to make it more flexible and most importantly moved it to JHtmlBehavior.

PR #611, opened 2011-12-05
Code style: Modify the FunctionCallSignatureSniff

elkuku:
Modify the FunctionCallSignatureSniff to allow double quotes as the last/first content on a line in multi line function calls.

This will remove two errors on #608 that "doesn't look that bad" ;)

It will allow the following construct:

``` php
<?php
$foo = bar("
    baz
");
```

Seems that those JS-guys like that style..


eddieajau:
Thanks. Merged.

PR #304, opened 2011-08-29
ORM for Joomla-Platform

juliopontes:
ORM for jplatform

This is my ORM for JPlataform, using Activy Records.

Discussion topic: http://groups.google.com/group/joomla-dev-framework/browse_thread/thread/a7fb7dbce6d2e8d2?pli=1

Original project: https://github.com/juliopontes/joomla-orm

Cheers,

Júlio Pontes - Brazil

PR #612, opened 2011-12-06
Added new JStringInflector class

juliopontes:
I´ve implemented JStringInflector and tried to create unittest too need help to test.

https://github.com/joomla/joomla-platform/pull/304#issuecomment-2261744


robschley:
The file should be named 'inflector.php' to follow the new autoloading conventions.

eddieajau:
That's an awesome start Julio.  Here's some initial comments:
- Code-style needs to be fixed up a bit (run it through phpcs with the Joomla standard)
- subpackage needed in file docblock
- @tutorial is not needed (it doesn't exist anyway)
- @link needs to got to JStringInflector (not JInflector - does the page exist?)
- Should be a concrete class, not an abstract class with static methods (much easier to test) or static properties. It's probably worth adding a string factory to create the concrete classes in the string package though.
- We use en-GB for coding so change the "ize" to "ise" please :)
- In the method docblocks, you are using @var instead of @param (should be @param [3 spaces] vartype [2 spaces min] $varname [2 spaces min] Var description
- countable, singularise, pluralise and add (I'd rename add to addRule) should be public methods.
- I'd make matchRegexRule private and rename it to just _match. I think you can make the loop more efficient by checking for if ($regex[0] == '/') ... do the regex. That means you aren't doing a regex on plain words that will be replaced.
- add/addRule should check that the ruletype is valid, or use an alias (eg, singular|plural|countable)
- I'd rename countable to isCountable and add matching isPlural and isSingular
- What do you think about renaming singularise and pluralise to toSingular and toPlural ?

Might also be worth looking at cakePHP's inflector to see if there are ideas to borrow:

https://github.com/cakephp/cakephp/blob/master/lib/Cake/Utility/Inflector.php

At the very least, there are some extra rules we can borrow.

Otherwise, it's a good start :)

juliopontes:
Andrew,

see my changes and give feedback.

juliopontes:
I need help to know how can I run these tests?

mbabker:
http://developer.joomla.org/platform-manual.html - Appendix A Chapter 2 gives you some basic info on running PHP_CodeSniffer, where the errors you're getting are coming from.

juliopontes:
Thanks, working on this. :)

elkuku:
Here is some documentation: http://docs.joomla.org/Joomla_CodeSniffer
Please pay special attention to the note in usage related to the platform that is different to the (nonexistent) code style of the CMS  ;)
I have set up a **personal** page that might also help to resolve the checkstyle errors. About the unit tests - I am not sure as results seems to differ across different operating systems.
http://elkuku.github.com/pulltester/pulls/612 <= for reference only !

realityking:
Could you please rename the file from stringinflector.php to inflector.php. This way it will be picked up the autoloader.

robschley:
You still have 72 style errors. Let us know when you get those sorted out and we'd be happy to merge this.

juliopontes:
Thanks Rob,

I'm working on fix style. :)

juliopontes:
I've tried to run this command: 

phpcs --report=checkstyle --standard=build/phpcs/Joomla /var/www/git/joomla-platform/libraries/joomla/string/inflector.php

and get a fatal error: 

PHP Fatal error:  Uncaught exception 'PHP_CodeSniffer_Exception' with message 'Referenced sniff Joomla.ControlStructures.InlineControlStructure does not exist' in /usr/share/php/PHP/CodeSniffer.php:814

Anyone can help me?

elkuku:
You should merge/rebase the latest changes from the master (into your master..). Your build.xml is quite old, as well as the sniffs..

**Edit**: Oh and you don't need to use absolute paths.. just `cd` to the platform "root" and run

`phpcs --standard=build/phpcs/Joomla libraries/joomla/string/inflector.php`

You also have both inflector.php and stringinflector.php in your repo, so you actually have 72 errors = 36 errors \* 2 ;)

juliopontes:
Waiting feedback;

juliopontes:
I've tried:

phpcs --report=checkstyle --standard=build/phpcs/Joomla libraries/joomla/string/inflector.php

and result:
<?xml version="1.0" encoding="UTF-8"?>
// checkstyle version="1.3.2">
// file name="/var/www/git/joomla-platform/libraries/joomla/string/inflector.php">
// /file>
// /checkstyle>

How can I check warnings?

elkuku:
You may safely omit the warnings as they do not concern your code. I believe they should be fixed/not displayed/masked (#564) - anyway - Those warnings are **expected** `:(`

Another **edit**.. if you omit `--report=checkstyle` the output should be a lot easier to read (to the human eye) ;)

WebMechanic:
Nice. How is this supposed to be used in other languages than english? _l10n_ anyone?
Using `private static` prevents subclassing and overriding and should be changed to `protected static`.
There is no notion of the language these rules actually apply to. A script might want to use the proper rules for a language other than plain english? 
Many languages might well be unable to use such an inflector at all 'cos they are too complex to fall into simple rules like "singular" and "plural", but that's another topic.
I've been toying a while with a local "hack" of the CMS's en_GBLocalise and de_DELocalise to host the rules, which to me makes sens putting them in those files 'cos they provide language specific settings anyway:
de_DELocalise https://gist.github.com/1450988
en_GBLocalise https://gist.github.com/1450989 
Rules were collected from throughout the internets incl. some for German I came up with.

juliopontes:
I guess we can add two methods like:

JStringInflector::getLocalise('en_GB');
JStringInflector::addLocalisePath();

What do you think?

juliopontes:
Anyone can check if codestyle its ok?

realityking:
The last comment from jenkins looks ok (254 warning 0 errors)

juliopontes:
My code has 254 warning? if yes, How can I see these warnings?

realityking:
No, that's the entire platform. Your code seems to be fine.

juliopontes:
Thanks, I´ve guess the solution of have localize class its a good idea, but need feedback.

propose of implementation:

strings/inflector/localize/en_GB.localize.php
strings/inflector/localize/pt_BR.localize.php
strings/inflector.php

JStringInflector::getLocalise('en_GB');

If anyone want to add a custom localize just use:

JStringInflector::addLocalisePath(JPATH_COMPONENT.DS.'localise');
JStringInflector::getLocalise('es_ES);

ps.: By deffault will load en_GB

What do you think?

robschley:
Let's not let the perfect get in the way of the good. This is a good start. If someone wants to build on it and add localisation support, great. But I don't think that we should get in the way of using this now.

juliopontes:
agreed, so waiting for merge this version. and we can think on this point on future.

elkuku:
The unit tests run fine for you ? I think they need an update because of the name change ;)

LouisLandry:
Yep, this is good stuff Julio.  The tests don't appear that they will work because you have things like: 

```
/**
 * Method to test JStringInflector::countable().
 *
 * @return  void
 *
 * @dataProvider  isCountable
 * @since   11.3
 */
```

and the actual data provider method is:

```
public function seedIsCountable()
```

Fix that and we are ready to merge I believe.

WebMechanic:
Since inflection is a matter of language, that's coincidently made up of strings, I think this class ought to belong to the _language_ package just like `latin_transliteration.php`. If someone picks up on it to add support for other languages, it's all in one place.
I still like to see `static private $_rules` to become `static protected $_rules` to at least lower the barriers for subsclassing.
Thanks.

juliopontes:
waiting feedback

juliopontes:
Hi guys, I'm add 2 new methods isSingular and isPlural. Please wait for merge when I've finish these methods.

juliopontes:
some fixes on toSingular regex, and finished 2 methods isSingular and isPlural.

juliopontes:
@eddieajau @robschley I've implement a auto assign in JView.

<pre><code>
    /**
     * Execute and display a template script.
     *
     * @param   string   $tpl         The name of the template file to parse; automatically searches through the template paths.
     * @param   boolean  $autoassign  Auto assign data to view
     *
     * @return  mixed  A string if successful, otherwise a JError object.
     *
     * @see     fetch()
     * @since   11.3
     */
    public function display($tpl = null, $autoassign = true)
    {
        if ($autoaassign)
        {
            if (JStringInflector::isPlural($this->getName()))
            {
                $model = $this->getModel($this->getName());
                $items = $model->get('Items');
                $this->assignRef('items', $items);
            }
            elseif (JStringInflector::isSingular($this->getName()))
            {
                $model = $this->getModel($this->getName());
                $item = $model->get('Item');
                $this->assignRef('item', $item);
            }
        }

        $result = $this->loadTemplate($tpl);
        if ($result instanceof Exception)
        {
            return $result;
        }

        echo $result;
    }
</code></pre>


The idea is do automatic assign for view, like example: 
if you have a view "articles" will check if is Plural, get model with same viewName and call getData() method. after that we just assign "items" to your view.
other else "artcle" is a singular word so it display one a single item. Your tmpl has a "item" var with your item.

What do you think about? If agreed I can send this patch in this same pull request?

Waiting feedback

LouisLandry:
When running the tests for the string package on this branch I get:

```
FFF............................................................  63 / 184 ( 34%)
...................................SSSSSS.....SSSSSSSS......... 126 / 184 ( 68%)
..........................................................

Time: 2 seconds, Memory: 12.25Mb

There were 3 failures:

1) Warning
The data provider specified for JStringInflectorTest::testIsCountable is invalid.
Method isCountable does not exist


2) Warning
The data provider specified for JStringInflectorTest::testToPlural is invalid.
Method toPlural does not exist


3) Warning
The data provider specified for JStringInflectorTest::testToSingular is invalid.
Method toSingular does not exist


FAILURES!
Tests: 184, Assertions: 169, Failures: 3, Skipped: 14.
```

You need to fix that up.  Also, after running the code sniffer on the string package I get:

```
FILE: ...ents/Code/Joomla/Platform/sandbox/libraries/joomla/string/inflector.php
FOUND 2 ERROR(S) AFFECTING 1 LINE(S)
 247 | ERROR | Expected 1 space before "=="; 0 found
 247 | ERROR | Expected 1 space after "=="; 0 found

Time: 1 second, Memory: 17.75Mb
```

That will need to get fixed up as well.  Otherwise I think this is good enough to get merged.

juliopontes:
@LouisLandry and about auto assign?

LouisLandry:
Let's keep that a separate topic from this pull request.  There is really no need to put them together that I can see.

juliopontes:
waiting feedback

eddieajau:
I've been working on some refactoring to this.  Will push up when I can.

eddieajau:
Julio, I sent you a pull request for some refactoring.  See https://github.com/eddieajau/joomla-platform/tree/inflector

juliopontes:
Andrew, what should I do with a pull request? "This pull request cannot be automatically merged."

eddieajau:
Can you update your master branch against staging and I'll see what I can do.

juliopontes:
Andrew, updated. Waiting feedback

juliopontes:
Anyone have news?

eddieajau:
Julio, I can't see where you've merged my branch in.  Would it be easier for me to make a pull request from my branch?

juliopontes:
@eddieajau Sorry, I´ve understand to merge with oficial joomla/staging. So, how we can fix it?
a) merge my branch with your branch?
b) I´ve fork u branch and commit these files?

eddieajau:
Julio, would you merge my inflector branch into this branch (your master branch), and then push your master branch up to the server.

juliopontes:
@eddieajau I hope that works. :) I´ve tried merge here. Take a look and give me a feedback

eddieajau:
Thanks. We'll wait for Jools to test it.

eddieajau:
Passed.  I'll need someone else to merge as I've had a hand in this one.

libraries/joomla/string/inflector.php
```diff
@@ -0,0 +1,462 @@
...
+
+	/**
+	 * Cached inflections.
+	 *
+	 * The array is in the form [singular => plural]
+	 *
+	 * @var    array
+	 * @since  11.4
+	 */
+	private $_cache = array();
```
chdemko, line 84: I think it will be better to use two arrays:

```
/**
 * Cached singular inflections.
 *
 * The array is in the form [singular => plural]
 *
 * @var    array
 * @since  11.4
 */
private $_singular = array();

/**
 * Cached plural inflections.
 *
 * The array is in the form [plural => singular]
 *
 * @var    array
 * @since  11.4
 */
private $_plural = array();
```

Since array_search used in _getCachedSingular performs its operation using linear search.

libraries/joomla/string/inflector.php
```diff
@@ -0,0 +1,462 @@
...
+	 * @return  mixed  The cached inflection or false if none found.
+	 *
+	 * @since   11.4
+	 */
+	private function _getCachedPlural($singular)
+	{
+		$singular = JString::strtolower($singular);
+
+		// Check if the word is in cache.
+		if (isset($this->_cache[$singular]))
```
chdemko, line 164: use (see https://github.com/joomla/joomla-platform/pull/612/files#r418760)

```
if (isset($this->_cache[$singular]))
{
    return $this->_singular[$singular];
}
```

libraries/joomla/string/inflector.php
```diff
@@ -0,0 +1,462 @@
...
+	 *
+	 * @return  mixed  The cached inflection or false if none found.
+	 *
+	 * @since   11.4
+	 */
+	private function _getCachedSingular($plural)
+	{
+		$plural = JString::strtolower($plural);
+
+		return array_search($plural, $this->_cache);
```
chdemko, line 185: use (see https://github.com/joomla/joomla-platform/pull/612/files#r418760)

```
if (isset($this->_plural[$plural]))
{
    return $this->_plural[$plural];
}
```

libraries/joomla/string/inflector.php
```diff
@@ -0,0 +1,462 @@
...
+		if ($plural === null)
+		{
+			$plural = $singular;
+		}
+		else
+		{
+			$plural = JString::strtolower($plural);
+		}
+
+		$this->_cache[$singular] = $plural;
```
chdemko, line 241: use (see https://github.com/joomla/joomla-platform/pull/612/files#r418760)

```
$this->_singular[$singular] = $plural;
$this->_plural[$plural] = $singular;
```

libraries/joomla/string/inflector.php
```diff
@@ -0,0 +1,462 @@
...
+	 * Adds a specific singular-plural pair for a word.
+	 *
+	 * @param   string  $singular  The singular form of the word.
+	 * @param   string  $plural    The plural form of the word. If omitted, it is assumed the singular and plural are identical.
+	 *
+	 * @return  JStringInflector  Returns this object to support chaining.
+	 *
+	 * @since   11.4
+	 */
+	public function addWord($singular, $plural =null)
```
chdemko, line 270: add a blank space after =

libraries/joomla/string/inflector.php
```diff
@@ -0,0 +1,462 @@
...
+	 */
+	public static function getInstance($new = false)
+	{
+		if ($new)
+		{
+			return new JStringInflector;
+		}
+		elseif (!is_object(self::$_instance))
+		{
+			self::$_instance = new JStringInflector;
```
chdemko: if JStringInflector can be extended, it's better to use (php 5.3)

```
self::$_instance = new static;
```

libraries/joomla/string/inflector.php
```diff
@@ -0,0 +1,462 @@
...
+	 *
+	 * @return  JStringInflector
+	 *
+	 * @since   11.4
+	 */
+	public static function getInstance($new = false)
+	{
+		if ($new)
+		{
+			return new JStringInflector;
```
chdemko: if JStringInflector can be extended, it's better to use (php 5.3)

```
return new static;
```

chdemko:
Some questions and remarks:
- why some properties are private (cannot be accessed if class is extended)?
- change comments to be 12.1
- it's better to use 2 arrays for caching (performance issue)
- we can use new static in php 5.3 for late state bindings (if class is extended)

eddieajau:
I have a preference for private properties unless someone gives me a good reason why they want to make it protected.  It's easy to change private to protected, but we have, many times, had the case where developers have been using protected variables inappropriately.  So it's just a conservative approach to take when releasing new API (I don't want people messing with bits that I don't want them messing with).

I'll fix up the little formatting issues (could have sworn I ran phpcs, but maybe I didn't).

I use one array in the form "singular word" => "plural form".  It actually saves a lot of duplicate data storage doing it that way.  You can use two arrays but you have to use a completely different methodology for the lookups (you will end up storing twice as much data).  If you want to do it that way, fine by me, but someone else can do the hard work :)

Yes, we can use LSB now so I'll make that change.

libraries/joomla/string/inflector.php
elkuku, line 270: Seems there is a sniffer rule missing or not working as expected..

realityking:
Also the since tags need to be changed to 12.1

chdemko:
@juliopontes Could you please update your pull request to get it merged?

juliopontes:
@chdemko Update with joomla/joomla-platform?

chdemko:
@juliopontes update with

```
@since  12.1
```

and

```
new static;
```

in getInstance

juliopontes:
Done, please check.

chdemko:
Please, could you change the @ since in the tests also + call new static in getInstance  https://github.com/joomla/joomla-platform/pull/612/files#L0R323

juliopontes:
@chdemko done.

jonnsl:
you could have squashed into a single commit.

PR #613, opened 2011-12-06
Update JHtmlForm::token() test

mbabker:
Update the test to mock the session due to a unit test failure locally, and use getFormToken in the test as well.

### DIFF
--- a/tests/suite/joomla/html/html/JHtmlFormTest.php
+++ b/tests/suite/joomla/html/html/JHtmlFormTest.php
@@ -14,8 +14,41 @@ require_once JPATH_PLATFORM.'/joomla/html/html/form.php';
  *
  * @since  11.1
  */
-class JHtmlFormTest extends PHPUnit_Framework_TestCase
+class JHtmlFormTest extends JoomlaTestCase
 {
+	/**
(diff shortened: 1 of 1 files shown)


elkuku:
This fixes the UT error on my side too.
Not sure why they would only appear "locally", but it works :)

mbabker:
Well, they weren't showing up from Jenkins reports.  So, obviously, he's
got some super secret squirrelly stuff going on to keep from getting that
failure  ;-)

On 12/7/11 1:29 AM, "Nikolai Plath"
<reply+i-2459345-87575b293c399da6f106104af839bb02795b25c5-368545@reply.gith
ub.com> wrote:

realityking:
The form.php stuff got added with #604.

elkuku:
> So, obviously, he's got some super secret squirrelly stuff going on to keep from getting that failure ;-)

This is, honestly, the only reasonable explanation :P
The error also does not appear on the build server but the test gets executed - another JSecret&trade;...

mbabker:
OK, so I've rebased the branch based on the merge of #604, so now all that's being done here is updating the unit test.

LouisLandry:
Hotness.

PR #617, opened 2011-12-06
Don't load JRequest when it isn't used.

realityking:


libraries/import.php
```diff
@@ -67,7 +66,7 @@ class_exists('JLoader') or die;
  */
 if (isset($_SERVER['HTTP_HOST']))
 {
-	JLoader::import('joomla.environment.request');
+	JLoader::register('JRequest', JPATH_PLATFORM . '/joomla/environment/request.php')
```
elkuku: Missing semicolons produce nice parse errors ;)

realityking: Gosh darn. Should be fixed now.

elinw:
Rouven i know that in my environment there are places where tests fail because JRequest is not loaded even though they pass the tests that jenkins runs, so I would be extremely cautious about this.

realityking:
Unit tests don't go trough import.php, so this shouldn't affect unit tests.

The unit test with JRequest issues probably need a require_once to fix them.

PR #618, opened 2011-12-06
Code style: Fix FileCommentsSniff

elkuku:
This sniff produces an erroneous message like:
`@foo tag comment indented incorrectly; expected 4 spaces but found 4'`

Last time this one appeared was in #603 as [commented](https://github.com/joomla/joomla-platform/pull/603#issuecomment-3033879) by Rouven.


robschley:
Merged.

PR #619, opened 2011-12-07
More autoloader + jimport() stuff

realityking:


realityking:
Urgh, don't hit enter when the cursor is still on the title.

The two jimports() are not required anymore thanks to #604.
Moving JStringNormalize should be done now since it's a pretty new class and not many people will have it used yet.

tests/suite/joomla/string/JStringNormalizeTest.php
```diff
@@ -7,7 +7,7 @@
  * @license     GNU General Public License version 2 or later; see LICENSE
  */
 
-require_once JPATH_PLATFORM . '/joomla/string/stringnormalize.php';
```
chdemko: See #589

realityking: Duh. I'll take it off again ;)

libraries/joomla/application/application.php
```diff
@@ -99,7 +99,6 @@ class JApplication extends JObject
 	 */
 	public function __construct($config = array())
 	{
-		jimport('joomla.utilities.utility');
 		jimport('joomla.error.profiler');
```
chdemko, line 102: JUitility is used in this file. Why do you want to remove the jimport ?

realityking, line 102: MAybe I'm just dense - but where? I couldn't find it. (Note that getHash was moved from JUtility to JApplication)

PR #620, opened 2011-12-07
Add git as a valid url protocol.

elinw:
This commit adds git as an accepted protocol when using the url rule for field validation.


robschley:
Merged. Thanks!

PR #621, opened 2011-12-08
constant() generates E_WARNING if given constant isn't *defined*

WebMechanic:
reverts some changes from 3e0bfb52


robschley:
Merged. Thanks!

PR #623, opened 2011-12-09
Add url filtering, improve url field, update  tests to reflect changes.

elinw:
This adds a url filter to the existing filters in Jform. The filter requires that a url have a protocol. If a protocol is not present it uses http://.  It also updates the tests to include tests of the url filtering option. This filter can optionally be used with the url field type or the url rule or both, but this is not required. 

As part of this change, the url field is built out and incorporates a new element, relative. By default (with the exception of some educated guessing based on the host and index.php) the url filter assumes that a url without a protocol is an external url that is only missing a protocol. If the relative element is used, the filter assumes that a url without a protocol is relative to the local root.

### DIFF
--- a/libraries/joomla/form/fields/url.php
+++ b/libraries/joomla/form/fields/url.php
@@ -22,7 +22,7 @@ JFormHelper::loadFieldClass('text');
  * @see         JFormRuleUrl for validation of full urls
  * @since       11.1
  */
-class JFormFieldUrl extends JFormField
+class JFormFieldUrl extends JFormFieldText
 {
 	/**
 	 * The form field type.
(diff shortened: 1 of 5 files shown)


elinw:
I do think in the long run filters should probably be broken out into their own folder.

libraries/joomla/form/fields/url.php
```diff
@@ -31,4 +31,39 @@ class JFormFieldUrl extends JFormFieldText
```
realityking, line 30: Du we still use anything from JFormFieldText? If not this should not be extended from it.

libraries/joomla/form/fields/url.php
```diff
@@ -31,4 +31,39 @@ class JFormFieldUrl extends JFormFieldText
 	 * @since  11.1
 	 */
 	protected $type = 'Url';
+		/**
```
realityking: Missing space.

libraries/joomla/form/fields/url.php
```diff
@@ -31,4 +31,39 @@ class JFormFieldUrl extends JFormFieldText
...
+		* Method to get the field input markup.
+		* When used in conjunction with the url filter in JFormField, ifthe relative element is
+		* false the method assumes most URLS are external.
+		* When relative is true a url that does not include a protocol is assumed to be local.
+		* This method does not validate a url which should be done using JFormRuleUrl.
+		*
+		* @return  string
+		*
+		* @see     JFormRuleUrl, JForm::Filter
+		* @since   11.1
```
realityking: Should be 11.4

elinw:
Good catches

realityking:
You got the wrong one for the since tag, $type should have 11.1, getInput() 11.4.

I think most of those 54 code style errors are du to missing spaces in string concats.

elinw:
Actually I think a lot were due to having an extra tab on all of the lines in the method.

elinw:
I wanted to mention that this is quite important for the CMS since currently there is all sorts of different kinds of solutions to forcing a protocol in the many different places where urls are expected (various components including weblinks, newsfeeds, content, contact, several modules and plugins, the url field in general (since we do have this field). I think it is a useful filter to have for web applications which do often have urls embedded one way or another. But if it's not useful for the platform I will propose to put it into the cms libraries so I can clean up that code.

PR #624, opened 2011-12-09
Joomla CMS [#26582] Recursive permanent deletion of nested categories 

radiant-tech:
Proposed in response to CMS Tracker #26582 (http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26582).  

Currently if a parent cat in trash is removed ALL child cats (regardless of state and whether selected for removal) are removed.  JModelAdmin calls the delete function without the second parameter that JTableNested would look for to decide if child cats should be removed.

### DIFF
--- a/libraries/joomla/database/table/category.php
+++ b/libraries/joomla/database/table/category.php
@@ -191,12 +191,12 @@ class JTableCategory extends JTableNested
 
 	/**
 	 * Override method to delete a node and, optionally, its child nodes from the table
-	 * 
+	 *
 	 * @param   integer  $pk        The primary key of the node to delete
 	 * @param   boolean  $children  True to delete child nodes, false to move them up a level.
-	 * 
+	 *
 	 * @return  boolean  True on success.
-	 * 
+	 *
 	 * @see     JTableNested::delete
 	 * @since   11.4
 	 */
@@ -206,7 +206,7 @@ class JTableCategory extends JTableNested
 		{
 			$children = false;
 		}
-		
+
 		return parent::delete($pk, $children);
 	}
 
--- a/libraries/joomla/database/table/category.php
+++ b/libraries/joomla/database/table/category.php
@@ -190,6 +190,23 @@ class JTableCategory extends JTableNested
 	}
 
 	/**
+	 * Override method to delete a row and, optionally, child rows from the table
+	 * 
+	 * @param	integer	$pk		The primary key of the row to delete
+	 * @param	boolean	$children	True to delete child rows, false to move them up a level.
+	 * 
+	 * @return	boolean
+	 * @see JTableNested::delete
+	 */
+	public function delete($pk, $children = null)
+	{
+		if (empty($children)) {
+			$children = false;
+		}
+		return parent::delete($pk, $children);
+	}
+
+	/**
 	 * Overridden JTable::store to set created/modified and user id.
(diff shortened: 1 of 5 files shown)


libraries/joomla/database/table/category.php
```diff
@@ -190,6 +190,23 @@ public function bind($array, $ignore = '')
 	}
 
 	/**
+	 * Override method to delete a row and, optionally, child rows from the table
+	 * 
+	 * @param	integer	$pk		The primary key of the row to delete
+	 * @param	boolean	$children	True to delete child rows, false to move them up a level.
+	 *
```
eddieajau: Need to fix some code-style in this docblock:
- 3 spaces after @param.
- "The" and "True" in the variable descriptions must line up (padding with spaces but with a minimum of two after the variable name).
- Need a blank comment line after @return
- "JTableNested" needs to line up with "boolean"
- Need to add an "@since 11.4" after @see (line up "11.4" with "JTableNested")

libraries/joomla/database/table/category.php
```diff
@@ -190,6 +190,23 @@ public function bind($array, $ignore = '')
...
+	 * 
+	 * @param	integer	$pk		The primary key of the row to delete
+	 * @param	boolean	$children	True to delete child rows, false to move them up a level.
+	 * 
+	 * @return	boolean
+	 * @see JTableNested::delete
+	 */
+	public function delete($pk, $children = null)
+	{
+		if (empty($children)) {
```
eddieajau: Curly brace must be on a new line.

radiant-tech:
Thanks Andrew.  Revised per your comments.

eddieajau:
There's a test error there.  You have to keep the same method signature:

```
/**
 * Method to delete a node and, optionally, its child nodes from the table.
 *
 * @param   integer  $pk        The primary key of the node to delete.
 * @param   boolean  $children  True to delete child nodes, false to move them up a level.
 *
 * @return  boolean  True on success.
 *
 * @link    http://docs.joomla.org/JTableNested/delete
 * @since   11.1
 */
public function delete($pk = null, $children = true)
```

That should fix the check style violations as well.

elkuku:
About the 1 failure from the unit tests, I don't know about Mr. Jenkins error, but mine is this:
http://elkuku.github.com/pulltester/pulls/624
Maybe this helps..

radiant-tech:
Nikolai ... in the pull request I've now added test functions for the new delete functions in JTableCategory and JTableMenu which I assume is the source of the failure error.  Is there somewhere else where I need to change the assertions?

oc666:
Is this pull going to be in J!2.5?

radiant-tech:
Mark knew I was working on the unit test -- will send him a note now that everything passed.

chdemko:
I'm much in favor to find a solution modifying only CMS files since there could be backward compatibility issues. I've proposed a patch there http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26582

oc666:
Christophe, totally agree with you. 
Will test it later.
Thanks

radiant-tech:
Thank you Christophe for the fix.  Closing.

oc666:
Thanks Christophe

PR #626, opened 2011-12-09
Explicitly set useStorage param to 'true':'false' so that 'null' will no...

thinkeryvin:
...t get defaulted to 'true' by the tabs.js. This should fix issue where setting useCookie=>'false' still resulted in tabs.js using the cookie storage.


realityking:
Looks good to me.

LouisLandry:
Thanks for taking the time to submit this.

PR #631, opened 2011-12-10
Fix JTableUsergroup.

realityking:
Another issue discovered by the CMS system tests.


eddieajau:
Why are we quoting dot notation?

realityking:
What do you mean, Andrew?

eddieajau:
Why is this line:

```
$query->where($db->quoteName('c') . '.' . $db->quoteName('rgt') . ' <= ' . (int) $this->rgt);
```

not just:

```
$query->where('c.rgt <= ' . (int) $this->rgt);
```

It doesn't necessarily affect this pull, but I'm just wondering why all the extra work for a known case. It seems overkill.

realityking:
Ah, gotcha.

PR #632, opened 2011-12-11
Deprecate JDate::toMySQL

chdemko:
In favor of JDate::toSQL

### DIFF
--- a/libraries/joomla/utilities/date.php
+++ b/libraries/joomla/utilities/date.php
@@ -527,7 +527,7 @@ class JDate extends DateTime
 	 */
 	public function toMySQL($local = false)
 	{
-		JLog::add('JDate::toMySQL() is deprecated.', JLog::WARNING, 'deprecated');
+		JLog::add('JDate::toMySQL() is deprecated. Use JDate::toSQL() instead.', JLog::WARNING, 'deprecated');
 		return $this->format('Y-m-d H:i:s', $local, false);
 	}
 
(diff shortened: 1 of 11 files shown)


ianmacl:
Do all databases expect the data to be in the same format?

elinw:
In deprecation notices if possible please state what to use instead.

chdemko:
@Ian: not necessarily: in my code, I call $dbo->getDateFormat() in the toSQL() method

eddieajau:
I think the method name should be toSql, but other than at it looks fine.

eddieajau:
Actually, can we also change all of the instances of toMysql to toSql in the platform code, so it's full deprecated in the API?

libraries/joomla/utilities/date.php
```diff
@@ -523,14 +523,35 @@ public function toISO8601($local = false)
...
+	 *
+	 * @param   boolean    $local  True to return the date string in the local time zone, false to return it in GMT.
+	 * @param   JDatabase  $dbo    The database driver or null to use JFactory::getDbo()
+	 *
+	 * @return  string     The date string in SQL datetime format.
+	 *
+	 * @link http://dev.mysql.com/doc/refman/5.0/en/datetime.html
+	 * @since   11.4
+	 */
+	public function toSql($local = false, $dbo = null)
```
eddieajau: One minor thing:

```
public function toSql($local = false, JDatabase $dbo = null)
```

eddieajau:
Awesome, I like it!

PR #633, opened 2011-12-11
Use the autoloader to rename form/formfield.php and form/formrule.php

chdemko:


realityking:
Yes please! :)

eddieajau:
Does this allow us to get rid of the `JFormHelper::loadFieldClass('list');` code as well?

realityking:
I don't think so. For this we'd have to rename the field**s** folder to field.

chdemko:
no because they are stored in the 'fields' folder instead of the 'field' folder (note the "s").

eddieajau:
Ok, we can attack that another time then.

PR #634, opened 2011-12-11
Fix some typos

elinw:
Fix some conditions in JForm that have = instead of ==.


chdemko:
This code is not buggy. The assignment is done inside the test.

PR #636, opened 2011-12-12
Allow form rules to throw exceptions.

realityking:


jonnsl:
:+1:

eddieajau:
I think we should go a bit further with this.  validateRule should have @throws added to the docblock and allow for logic exceptions, such as changing:
    return new JException(JText::sprintf('JLIB_FORM_VALIDATE_FIELD_RULE_MISSING', $type), -2, E_ERROR);
to
    throw new LogicException(JText::sprintf('JLIB_FORM_VALIDATE_FIELD_RULE_MISSING', $type));

JRule should throw a LogicException on an invalid rule, and be marked to throw RuntimeExceptions otherwise for failed validation and change the return false to throwing a RuntimeException.  Similarly change the derived rule classes to throw and exception rather than returning false.

What do you think?

elinw:
That makes a lot of sense to me. I've written a bunch of rules and mixing the two together has always felt wrong to me.

realityking:
I agree. But wouldn't the example Andrew gave better be a RuntimeException?

eddieajau:
Actually, that example is better as an InvalidArgumentException.  For failing rules, I'd use a RuntimeException.

realityking:
Ok I think I've done it as Andrew suggested. Note I haven't tested the code yet, but I'd be grateful if someone would check if my approach is sound.

realityking:
I realized I handled the legacy JExceptions in the wrong place but this should be good. I haven't tested this in the wild yet so please don't merge quite yet. I'd be glad thou if someone looked over the code.

Backward compatibility should be maintained with this unless one is directly using the (protected method) JForm::validateField() or is directly using JFormRuleEquals (which is hardly possible) so I think we're pretty safe.

chdemko:
@realityking For fixing the unit test: https://gist.github.com/1868665

eddieajau:
Closing. I'm pretty sure we settled on returning exceptions, not throwing them.

PR #637, opened 2011-12-12
(>11.4) Code style: Mark protected member vars as deprecated - part 2

elkuku:
Continuing from #564
This fixes the warnings on public and protected class vars beginning with an underscore.
Remaining warnings are caused by missing scope modifiers.


realityking:
Please don't mess with JParameter. It's legacy code that will probably be dropped rather soon, no need to change anything there.

libraries/joomla/database/table.php
```diff
@@ -28,6 +28,15 @@
 	 * @var    string
 	 * @since  11.1
 	 */
+	protected $tbl = '';
```
jonnsl: This will probably  cause error in the store method because it will try to store this variable into the database https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/database/database.php#L783

elkuku:
@realityking right. I removed the vars again and changed the `@deprecated` message as in JSimpleXMLElement. Do you know of other classes that are deprecated ?
@jonnsl Good catch - that's kind of a show stopper `:(` I believe JDatabase::insertObject() should use another "method" to identify internal fields.. reverted

elkuku:
Look ma, lotsa zeros :)

realityking:
JElement and it's subclasses (part of the JParameter system) are also deprecated.

LouisLandry:
Let's pull this right after 11.4 has been tagged... what do you think?

elkuku:
The CMS might not be "ready" for this change yet.. so - Perfect ;)

chdemko:
I think we can find a global solution for these $_name protected variables:

in JObject

```
/**
 * magic get method
 *
 * @param   $propertyName  Property name
 *
 * @return  mixed  the property value
 *
 * @since   12.1
 * @deprecated in 12.3
 */
public function __get($propertyName)
{
    if ($propertyName[0] == '_' && property_exists($this, $newPropertyName = substr($propertyName, 1)))
    {
        JLog::add(get_called_class() . '::$' . $propertyName . ' will be deprecated in Joomla! 12.3. Use ' . get_called_class() . '::$'. $newPropertyName . ' instead.', JLog::WARNING, 'deprecated');
        return $this->$newPropertyName;
    }
    else
    {
        // Trigger an error
        return $this->$propertyName;
    }
}

/**
 * magic set method
 *
 * @param   $propertyName  Property name
 * @param   $value         Property name
 *
 * @return  void
 *
 * @since   12.1
 * @deprecated in 12.3
 */
public function __set($propertyName, $value)
{
    if ($propertyName[0] == '_' && property_exists($this, $newPropertyName = substr($propertyName, 1)))
    {
        JLog::add(get_called_class() . '::$' . $propertyName . ' will be deprecated in Joomla! 12.3. Use ' . get_called_class() . '::$'. $newPropertyName . ' instead.', JLog::WARNING, 'deprecated');
        $this->$newPropertyName = $value;
    }
    else
    {
        $this->$propertyName = $value;
    }
}
```

An in all descendant classes of JObject, we can rename $_name protected variable to $name (this will ensure backward compatibility)

The same can be applied to other "base" class as JCacheController and so on

chdemko:
@elkuku Any news on this?

elkuku:
No, not really.. If there is a chance to get it merge it might need an update. If not - feel free to close it :(

elkuku:
Closing this as it has been split up into several smaller pull requests that simply rename the protected members.

See: Rename protected members in JXXX..

PR #638, opened 2011-12-12
Joomla CMS [#26360] Fixed component's parameters are not loaded during installation.

vietvh:
Related tracker item: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=26360


LouisLandry:
This looks good to me.  @pasamio would you take a look at this and pull it if you feel comfortable with the implementation.  Thanks a bunch @vietvh.

chdemko:
This looks good to me

pasamio:
I'm kind of worried how this will behave with those users who have nested manifest files throughout their extension. For example a separate plugin or similar. I don't see any test cases which handle that sort of edge case and if we're changing the way manifest detection should work, we should really be looking for a variety of different situations and ensuring it behaves consistently.

vietvh:
Could you be more specific about the nested manifest files case,? Is there any ready package so that I can test?

So far the test source that I've mainly used is /tests/_data/installer_packages in the cms trunk repository.

pasamio:
If you have a component without a config.xml which has inside itself a second component with a config.xml, how does your patch behave? Does it default the params to the component's XML file or does it pick up the config.xml from the child component that is being shipped uncompressed as well?

chdemko:
@vietvh Any news on this?

vietvh:
Closed for now as it need some updates.

PR #641, opened 2011-12-12
Joomla CMS [#26893] add filterText method to JComponentHelper with new customList feature

radiant-tech:
See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26893 for description of problem having Text Filtering confined to Articles.

Incorporates feature requeste started here: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26482

Logic for filtering is explained in the test instructions of #26893.  Custom black list allows SU to override the hard-corded default tagBlackList and attrBlackList of JFilterInput.

### DIFF
--- a/libraries/joomla/application/component/helper.php
+++ b/libraries/joomla/application/component/helper.php
@@ -95,12 +95,13 @@ class JComponentHelper
 	}
 
 	/**
-	* Applies the global text filters to arbitrary text as per settings for current user group
-	* 
-	* @param	text The string to filter
-	* 
-	* @return	string The filtered string
-	* @since	11.3
+	* Applies the global text filters to arbitrary text as per settings for current user groups
+	*
+	* @param   text  $text  The string to filter
+	*
+	* @return  string  The filtered string
+	*
+	* @since   11.4
 	*/
 	public static function filterText($text)
 	{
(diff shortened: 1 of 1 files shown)


LouisLandry:
Why would this go into the component helper class?  Also, if we were going to add this I'd really like to see some unit tests to ensure that the code works as expected into the future --- and to document expected behavior.

eddieajau:
The other way to look at it is does it matter because I don't know that the component helper belongs in the platform anyway. We should probably excise it in 12.1.  Just an alt point of view.

LouisLandry:
That's valid as well.  I am happy to merge this if that is the approach we take.

radiant-tech:
Well I would be VERY happy to see it merged (fingers crossed to get the new text filters into 2.5).  But I do want to make sure it is in a logical class location.  Given that the filterText method treated globally can also be used in module editor fields, do you see a problem with it in JComponentHelper (given the thought of moving that class out of platform)?

LouisLandry:
The only issue I see is that what we are talking about (filtering text based on user/application settings) has nothing to do with a component.  Granted it is a common place to put that logic, but I would expect that to be either tied closer to the user object, or the application object.  Again though, I don't have a problem with it being here because ultimately this is a construct of the CMS and I expect that this class will end up out of the platform and into the CMS in the near future.

eddieajau:
Merged.

PR #646, opened 2011-12-13
Fix an issue caused by pull request #554.

realityking:
We removed some jimports() without renaming the class so it can be picked up by the autoloader.


eddieajau:
Done. Thanks.

PR #648, opened 2011-12-14
Implement Github Forks API, add missing tests.

ianmacl:
This adds forks to the Github APIs handled by the platform and completes unit tests for these classes.


elkuku:
I believe those pesky checkstyle errors should be marked red `;-)`

PR #658, opened 2011-12-19
Add ability for JDatabase::quoteName to auto split and quote dotted strings

eddieajau:
Allows for the quoteName method to auto split and quote each part of a dotted string.  For example:

```
$db->quoteName('a.foo');
```

Will return `'a'.'foo'`.

Adds tests for quoteName and improves code coverage of JDatabase.


LouisLandry:
Some since tags need to be adjusted for 11.4, but this looks great.

eddieajau:
_sigh_ had one of those stuck-brain moments.  Fixed.

PR #661, opened 2011-12-19
Make more classes autoloadable.

realityking:
Remove some unnecessary jimport() calls.


eddieajau:
Are the unit tests running ok your end?

realityking:
Can't really check. My unit test set up is somehow broken, I really need to fix it.

elkuku:
It says something like the autoloader fails to autoload ;) http://elkuku.github.com/pulltester/pulls/661

realityking:
Is that from after my last push?

elkuku:
It features a time stamp ;)
But it seems to be taken before your last push - now the error is different.. http://elkuku.github.com/pulltester/pulls/661

PR #665, opened 2011-12-21
Modified line 50 to support all databases. We needed to escape ' instead...

hooduku:
Modified line 50 to support all databases. We needed to escape ' instead pf ".
Can this be merged into platform?

Thanks
sudhi

### DIFF
--- a/libraries/joomla/form/fields/plugins.php
+++ b/libraries/joomla/form/fields/plugins.php
@@ -47,7 +47,7 @@ class JFormFieldPlugins extends JFormFieldList
 			$query  = $db->getQuery(true);
 			$query->select('element AS value, name AS text');
 			$query->from('#__extensions');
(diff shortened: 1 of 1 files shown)


libraries/joomla/form/fields/plugins.php
```diff
@@ -47,7 +47,7 @@ protected function getOptions()
 			$query  = $db->getQuery(true);
 			$query->select('element AS value, name AS text');
 			$query->from('#__extensions');
-			$query->where('folder = "' . $folder . '"');
+			$query->where('folder = \'' . $folder . '\'');
```
chdemko, line 50: please use

```
$query->where('folder = ' . $db->q($folder));
```

LouisLandry:
Fixed in a separate pull.

PR #672, opened 2011-12-23
setting name via config array

janschoenherr:
Hi,

it would be nice to set the name via config. Otherwise the default template search path might be wrong.

Regards

### DIFF
--- a/libraries/joomla/application/component/view.php
+++ b/libraries/joomla/application/component/view.php
@@ -221,7 +221,7 @@ class JView extends JObject
 		{
 			if (array_key_exists('name', $config))
 			{
-				$this->_name = $config['name'];
+				$this->_name = $this->name = $config['name'];
 			}
 			else
 			{


chdemko:
Could you propose a solution that include the test on empty($this->_name) ? ie:

```
    if (empty($this->_name))
    {
        if (array_key_exists('name', $config))
        {
            $this->_name = $this->name = $config['name'];
        }
        else
        {
            $this->_name = $this->getName();
        }
    }
    else
    {
        $this->name = $this->_name;
    }
```

janschoenherr:
Shouldn't it always be empty? It's in the constructor.

chdemko:
It could have been set in the declaration of the class

realityking:
The actual issue is that $this->name has been introduced without wiring it up.

Is _name supposed to be private or protected? Either way we should use it the way we want to use it in the long term and use __set() and __get() for b/c for 2.5.

elkuku:
Trying to implement a `__get()` method in JView I get a
`Notice: Indirect modification of overloaded property has no effect`
followed by a
`Fatal error: Cannot assign by reference to overloaded object`
both from JView::asignRef() that also does a check if the first character is a `_` :(

Looking at $name and $_name, turns out that both are in use but $name has never been declared:

``` php
<?php
class JView extends JObject
{
    public function __construct($config = array())
    {
        // Set the view name
        if (empty($this->_name))
        {
            if (array_key_exists('name', $config))
            {
                $this->_name = $config['name'];
            }
            else
            {
                $this->_name = $this->getName();
            }
        }
...
    }

    public function getName()
    {
        if (empty($this->name))
        {
            $r = null;
            if (!preg_match('/View((view)*(.*(view)?.*))$/i', get_class($this), $r))
            {
                JError::raiseError(500, JText::_('JLIB_APPLICATION_ERROR_VIEW_GET_NAME'));
            }
            if (strpos($r[3], "view"))
            {
                JError::raiseWarning('SOME_ERROR_CODE', JText::_('JLIB_APPLICATION_ERROR_VIEW_GET_NAME_SUBSTRING'));
            }
            $this->name = strtolower($r[3]);
        }

        return $this->name;
    }

}
```

So there is definitely something more here.

@janschoenherr Could you give an example of

> ...the default template search path might be wrong

?

janschoenherr:
@elkuku sure:

If I create an instance of a subclass of JView

`class FooView extends JView {`

and pass the constructor an array with the name

`new FooView(array('name' => 'bar'))`

the constructor of JView sets 

`$this->_name = 'bar'`

then continues to

`$this->_setPath('template', $this->_basePath . '/views/' . $this->getName() . '/tmpl');`

and here it calls 

`$this->getName()`

this returns empty, because $this->name is not set yet and the name is not being retrieved from the class name.

You'll end up with something like

`$this->_setPath('template', $this->_basePath . '/views//tmpl');`

but expect something like

`$this->_setPath('template', $this->_basePath . '/views/bar/tmpl');`

realityking:
This has been fixed in #1144 in a different way. Thanks for contributing!

PR #676, opened 2011-12-27
Updates to fix updater system

pasamio:
This pull request:
- fixes CLIENT undefined index problem
- remove an undefined index warning
- removes the INFOURL addition until CMS has accepted the change
- and also removes a commented debugging statement that somehow made it in.
# "CLIENT" undefined index warning

In pull request 667 the concept of changing "CLIENT" to "CLIENT_ID" is introduced to remove an undefined index "CLIENT" problem. This is actually a side effect of CMS tracker 24338 [2,3] where the appropriate entry in the update_cols was mercilessly renamed from "CLIENT" To "CLIENT_ID" leading us to the predicament we presently find ourselves in.

I suspect that the initial change is itself a reaction to CMS Issue # 24305 [4,5] which goes to show the law of unintended consequences is truly afoot.
# Undefined index warning

An extra guard is added around the last_check_timestamp check in the update cache code that was added. There are situations when this value may not exist and could trigger a warning.
# Removal of INFOURL addition

The removal of the "INFOURL" tag is two fold. First it's in the wrong place, it should live in the _update_cols of the updateadapter class. Secondly until the CMS accepts the DB schema change[6], it breaks loading extensions which made the changes in this pull request rather hard to test using the current Joomla! 2.5 beta. Given this change was merged into the platform in July and it is now December, it is like that the CMS is presently broken but people just don't realise (the collection type is unaffected).

Refs:
[1] https://github.com/joomla/joomla-platform/pull/667
[2] https://github.com/joomla/joomla-platform/commit/28819069e156bb749371a0d08c30d1d7cd62a366#libraries/joomla/updater/updateadapter.php
[3] http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=24338
[4] https://github.com/joomla/joomla-platform/commit/120f114471ebd2842aa72a8d3958aa09525aeb16
[5] http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=24305
[6] http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26635

### DIFF
--- a/libraries/joomla/updater/adapters/extension.php
+++ b/libraries/joomla/updater/adapters/extension.php
@@ -123,7 +123,7 @@ class JUpdaterExtension extends JUpdateAdapter
 		$tag = $this->_getLastTag();
 		//if(!isset($this->$tag->_data)) $this->$tag->_data = '';
 		//$this->$tag->_data .= $data;
-		if (in_array($tag, $this->_updatecols) || $tag == 'INFOURL')
+		if (in_array($tag, $this->_updatecols))
 		{
 			$tag = strtolower($tag);
 			$this->current_update->$tag .= $data;
@@ -144,7 +144,6 @@ class JUpdaterExtension extends JUpdateAdapter
 		$url = $options['location'];
 		$this->_url = &$url;
 		$this->_update_site_id = $options['update_site_id'];
-		//echo '<p>Find update for extension run on <a href="'. $url .'">'. $url .'</a></p>';
 		if (substr($url, -4) != '.xml')
 		{
 			if (substr($url, -1) != '/')
@@ -189,6 +188,11 @@ class JUpdaterExtension extends JUpdateAdapter
 		xml_parser_free($this->xml_parser);
 		if (isset($this->latest))
 		{
+			if (isset($this->latest->client) && strlen($this->latest->client))
+			{
+				$this->latest->client_id = JApplicationHelper::getClientInfo($this->latest->client, 1)->id;
+				unset($this->latest->client);
+			}
 			$updates = array($this->latest);
 		}
 		else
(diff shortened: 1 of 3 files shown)


elinw:
Given that it was only done for 11.3 which was only merged maybe a month ago, it's not suprising that the cms hasn't necessarily caught up perhaps.Date of commit to the platform is not really relevant if the platform hasn't been merged. :)
How about if we just get the CMS to commit the change?
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26635

mbabker:
It's been a while since I did the INFOURL change, but if I remember correctly, I originally started out by putting it in the _updatecols property, but the column wouldn't get handled correctly.  I can always test again to see if that holds true or if it was something at the time that I was missing, but that's why it is where it's at now.

infograf768:
Please indeed test, Michael and change cms patch to include a new sql update.

gnomeontherun:
This appears to make sense, and the patch for this fixes the update collection notifier in the CMS, since it now returns an empty array instead of the error messages.

pasamio:
@elinw the point is that 2.5 beta is broken and the change isn't the best; if the CMS tests and applies that change then we don't have an issue however at the moment the CMS has a bug

@mbabker not sure why it wouldn't be working if you added it there. I can't see code wise why it would make a difference as the "OR" should be the same as adding it to the array logically speaking.

mbabker:
Well, it's working now with INFOURL in the _updatecols property.  I'm regenerating the CMS patch right now with that change, so I suppose it can be added here as well.

elinw:
Ok so we are making progress?

mbabker:
CMS Pull Request @ joomla/joomla-cms#65

chdemko:
Fix #667

PR #513, opened 2011-11-12
Deprecating JNode and JTree and switching off code coverage

Hackwar:
JNode and JTree are classes that are not usable. If someone ever extended these classes, he had to rewrite all methods. Thus deprecating it directly for 12.1.

PR #677, opened 2011-12-27
Deprecating JNode and JTree and switching off code coverage

Hackwar:
JNode and JTree are classes that are not usable. If someone ever extended these classes, he had to rewrite all methods. Thus deprecating it directly for 12.1.

This is a new pull request, since I messed up the earlier one by merging the wrong stuff into the branch. The earlier request is #513.

### DIFF
--- a/docs/coding-standards/en-US/chapters/php.xml
+++ b/docs/coding-standards/en-US/chapters/php.xml
@@ -1,6 +1,7 @@
-<?xml version='1.0' encoding='utf-8' ?>
-<!DOCTYPE chapter PUBLIC "-//OASIS//DTD DocBook XML V4.5//EN" "http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd" [
-<!ENTITY % BOOK_ENTITIES SYSTEM "Coding_Standards.ent">
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE chapter PUBLIC "-//OASIS//DTD DocBook XML V4.5//EN"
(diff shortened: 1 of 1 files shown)


realityking:
Couple of points:
1. JNode is referenced in docblocks in JCategories. Those should probably be removed.
2. Does it really make sense to ignore code coverage for deprecated classes? Seems a bit like cheating
3. Could you add the deprecation logging?
4. The CMS uses JTree once in JAdminCssMenu. Could you take a look at cleaning that up? That would be great.

Hackwar:
to 1.: I will take a look at JCategories. I'd actually like to write a new JNode/JTree which JCategories and also JMenu can use to build upon, unifying all this a bit more (and making development a bit easier from my POV)
to 2.: Yes, it might be cheating, but I already invested quite some time to write unittests for those two classes and their code is so broken that I was unable to get proper unittests that also passed without rewriting both classes. It is similar to the JObserver/JObservable classes.
to 3.: will do shortly.
to 4.: Can I make a pull request against the CMS repo for that?

realityking:
re No. 4. I think that's fine however you'll have to open an accompaning issue on the tracker too.

Hackwar:
Here is the pull request on the CMS: https://github.com/joomla/joomla-cms/pull/144

eddieajau:
I think JNode and JTree should also move to the legacy tree.

eddieajau:
Let's update the branch, move it to the legacy tree and make them for removal in 12.3 unless we hear otherwise.

PR #678, opened 2011-12-29
Correct a bug in the url form filter.

elinw:
Correct a bug in the url form filter.


elinw:
For some reason Andrea thinks the letter t matters in urls. Not to mention h and p. So fussy. Sigh.

PR #679, opened 2011-12-29
Changed quoteName function.

gpongelli:
Now quoteName can take an array of names to quote and each can have dot notation.
Added alias support "AS" as second parameter, same type of first parameter, so if first parameter is a string second need to be a string, if first parameter is an array second need to be an array.
Special mention to first parameter array: each array element will be quoted and can be aliased with second parameter array that must be same length and contain null value for those element that doesn't need alias, so for example

<pre><code>$db->quoteName( array('a', 'b'), array('alias_a', 'alias_b') );</code></pre>

using " as quote character, this call will return an array with

<pre><code>Array
(
    [0] => '"a" AS "alias_a"'
    [1] => '"b" AS "alias_b"'
)</code></pre>

<p>
while this example

<pre><code>$db->quoteName( array('a', 'b'), array(null, 'alias_b') );</code></pre>

using " as quote character, this call will return an array with

<pre><code>Array
(
    [0] => '"a"'
    [1] => '"b" AS "alias_b"'
)</code></pre>

<p>
More complex example

<pre><code>$db->quoteName( array('a.c', 'b.d'), array(null, 'alias_b_d') );</code></pre>

will return an array with

<pre><code>Array
(
    [0] => '"a"."c"'
    [1] => '"b"."d" AS "alias_b_d"'
)</code></pre>

<p>
This pull request contain also changes to relative quoteName test to try array quote and alias parameter.

### DIFF
--- a/libraries/joomla/database/database.php
+++ b/libraries/joomla/database/database.php
@@ -1240,28 +1240,74 @@ abstract class JDatabase implements JDatabaseInterface
 	 * Wrap an SQL statement identifier name such as column, table or database names in quotes to prevent injection
 	 * risks and reserved word conflicts.
 	 *
-	 * @param   mixed  $name  The identifier name to wrap in quotes, or an array of parts to quote with dot-notation.
+	 * @param   mixed  $name  The identifier name to wrap in quotes, or an array of identifier names to wrap in quotes.
+	 * 							Each type supports dot-notation name.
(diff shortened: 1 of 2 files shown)


robschley:
Shouldn't this logic be moved to JDatabaseQuery?

gpongelli:
Why inside JDatabaseQuery?
During a [discussion](https://groups.google.com/forum/#!topic/joomla-dev-platform/4wCYZ6P4Dog) was told to to this changes.

Issue #386, opened 2011-09-24
JFile::write() uses $buffer by reference


Denitz:
Why ?
public static function write($file, &$buffer...

We can't have this code:
JFile::write('somefile.php', 'text');

Fatal error: Cannot pass parameter 2 by reference...

mahagr:
I've also hit this "feature" when trying to create dummy html file into empty directory. It works perfectly in J!1.5 but not in J!1.7, which makes the function not to be backwards compatible -- and there's no documentation that the API has changed.

eddieajau:
The reference is probably there to save on memory (JFile is notorious for consuming it).

PR #681, opened 2011-12-30
Add ability to use any context with setMetaData & getMetaData

ghost:
JDocument/setMetaData  is currently limited to the name and http-equiv contexts.
http://docs.joomla.org/JDocument/setMetaData

In HTML5 the meta tag also now excepts the 'itemprop' and 'charset' contexts. Other specs such as Open Graph use custom meta tag contexts like 'property'.

To add support for these new and future uses of the meta tag, I'd like to propose making the third argument of JDocument/setMetaData a mixed value (boolean (legacy) ~ string).

_HTML5 Microdata - itemprop_
http://www.whatwg.org/specs/web-apps/current-work/multipage/semantics.html#the-meta-element

_Open Graph - property (Non-Valid)_
http://ogp.me/


realityking:
I don't think this is the right solution.

For charset we already have a dedicated function. We just need to get a HTML5 switch for the renderer to take care of the new ways.

As for itemprop, that would be more interesting. Personally I'd kinda like to take the wait and see approach on the Microdata/RDFa stuff and come up with a more unified approach until we start modifying API.

bigbangireland:
Thanks Rouven,

Is there a discussion going on about Joomla and HTML5 or is there a thread I may have missed that you could point me to? It had been discussed on the ideas forum but nothing seemed to come of it. goo.gl/g1md6

It's not a mater of if microdata will be the standard, but when. The when will be affected by how quickly platforms such as Joomla adapt it. Google, Bing and Yahoo are all pushing microdata with schema.org. http://goo.gl/OfZcr

Just because contexts such as 'property' (Open Graph / Facebook) are not valid, doesn't mean they are not used. In fact OG metatags are widely used in Joomla installations. This pull request would at least add the ability to group all metatags neatly.

realityking:
Right, I didn't get to add that (very simple) code in time for the PBF event. I got sidetracked by RL stuff and had other priorities on my J! work.

Since I had to do a JDocument pull request today anyways I've added the code for it there. It's #386. If we get lucky it gets merged in time for 2.5.

bigbangireland:
Cheers Rouven

bigbangireland:
Sorry about that @joomla-jenkins new to this pull request stuff ;)

Changes reverted.

chdemko:
@bigbangireland The pull request cannot be merged

LouisLandry:

PR #683, opened 2011-12-30
Fix a bug in JDocument.

realityking:
Fix the bug that meta elements with the attribute http-equiv weren't set corretly. This breaks API compatability on JDocument::SetMetadata().


realityking:
Slightly different implementation that keeps me from finding out how to mock that thing on short notice.

realityking:
Instead of this #697 and #699 should be merged. I'll fix this one up with the remaining stuff when those are merged.

realityking:
I split everything except the bug fix off so we can deal with each separately.

realityking:
Any news on this?

PR #684, opened 2011-12-31
Add a message queue to JApplicationWeb

elkuku:
While I was playing with the new toy JApplicationWeb, I found that the system messages would be handy for this type of application.

Also removes some duplicated code from JApplication as well as the deprecated property `$_messageQueue`

### DIFF
--- a/libraries/joomla/application/application.php
+++ b/libraries/joomla/application/application.php
@@ -53,15 +53,6 @@ class JApplication extends JObject
 	protected $messageQueue = array();
 
 	/**
-	 * The application message queue.
-	 *
-	 * @var    array
-	 * @since  11.1
-	 * @deprecated use $messageQueue or declare as private
-	 */
-	protected $_messageQueue = array();
-
-	/**
 	 * The name of the application.
 	 *
 	 * @var    array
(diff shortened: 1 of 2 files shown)


piotr-cz:
+1

I find message queue is extemely useful in applications based on Joomla Platform.

With JP 11.4 when you use `<jdoc:include type="message" />` in theme/ buffer
JDocument calls JWebapplication::getMessageQueue but fails with PHP errors.

eddieajau:
I'm pretty sure we decided we'd just use JLog for this and the application would implement a listener for it (compare what we did with dealing with JError::raiseWarning and JError::raiseNotice).

elkuku:
<del>That doesn't sound bad.
So the functions an members corresponding to the "messagequeue" in JApplication should be deprecated ?

<del>The code I proposed only adds the functionality to JApplication**Web**, where I believe it make some sense..
Actually I am using the message queue to replace the calls to JError::raiseWarning and JError::raiseNotice in CMS extensions;)
But JLog also sounds like a good replacement.</del>

Please scratch that - this is what happens when you are looking at "old" pull requests ;)

PR #686, opened 2011-12-31
for Platform 12.1 new self, you are able to extends the jform with the extended instance 

romacron:
JForm::getInstance(**); works fine.

for extra formhandling like

class myForm extends JForm{
}

$personalForm=myForm::getInstance(**);
it gets the instance of JForm
i am not able to get the instance of myForm


chdemko:
Your solution does not work because you don't use late static bindings (http://www.php.net/manual/en/language.oop5.late-static-bindings.php) which is not available in php5.2

```
<?php
class JForm {
    public static function getInstance() {
        return new self();
    }

    public static function getInstance2() {
        return new static();
    }
}

class JForm2 extends JForm {}

echo get_class(JForm2::getInstance());  // JForm
echo get_class(JForm2::getInstance2()); // JForm2
echo get_class(JForm::getInstance2()); // JForm
```

The right solution would be to use new static which is not avaliable in php5.2

Could you change your code and propose this modification for platform 12.1

romacron:
mist ;-)

Yes of course, i strictly ignored php5.2 

stupid question: where can i find the repo for "Platform 12.1"

realityking:
Since the repro is now PHP 5.3 and up we can use LSB now.

Somehow your pull request got bad, you're original commit is gone but you got a dozen commits in here that are already part of the platform. Could you make a new pull request or fix this one? Thank you.

realityking:
I'm gonna close this since the commits in the pull have nothing to do with the description and the author doesn't respond anymore.

romacron:
the previous messages didn't arrived my mail account. 
Im not able to reopen this ticket.

libraries.joomla.form.form line 2000 in getInstance()

// Instantiate the form.
$forms[$name] = new JForm($name, $options);
to 
// Instantiate the form.
$forms[$name] = new self($name, $options);

so developers can use

$myOwnFormClass = happyForm::getInstance();

realityking:
self won't work as Christophe pointed out, you'll have to use static. Could you open up a new pull request for that?

PR #688, opened 2011-12-31
Correct a typo in variable name

davidhurley:
Fixed error in local variable reference. 

Addressed in Tracker #27561

### DIFF
--- a/libraries/joomla/application/component/view.php
+++ b/libraries/joomla/application/component/view.php
@@ -541,7 +541,7 @@ class JView extends JObject
 	 */
 	public function getName()
 	{
-		if (empty($this->name))
+		if (empty($this->_name))
 		{
 			$r = null;
 			if (!preg_match('/View((view)*(.*(view)?.*))$/i', get_class($this), $r))
@@ -552,10 +552,10 @@ class JView extends JObject
 			{
 				JError::raiseWarning('SOME_ERROR_CODE', JText::_('JLIB_APPLICATION_ERROR_VIEW_GET_NAME_SUBSTRING'));
 			}
-			$this->name = strtolower($r[3]);
+			$this->_name = strtolower($r[3]);
 		}
 
-		return $this->name;
+		return $this->_name;
 	}
 
 	/**


pasamio:
Have you read the docblock comment for JView::$_name ?

davidhurley:
Yes.

davidhurley:
Sorry, are you saying I should have updated that definition? It appears to be set as a string consistently throughout the file.

realityking:
Somehow sometimes _name is used and sometimes name. This will confuse quite a few people. We should come up with a fix that maintains b/c.

elkuku:
The little "detail" is found in the docblock as `@deprecated` ;)
Do you get any errors or whatever that caused you to propose the change ?

davidhurley:
Yes, as reported in the bug tracker (referenced originally) I could not instantiate a new JView manually for inclusion because the function failed to pull the correct model. Regardless of deprecation or not I would certainly think we would want the code to be correct and fix a minor problem instead of attempting to defend an error by claiming deprecation. :)

elkuku:
Sorry I missed the tracker ID ;)
The problem could be a change introduced about two month ago ([ref](https://github.com/joomla/joomla-platform/commit/04cd4c7ae268eceab196a15bbcce0de8d600d6b1#L1L447)) when `name` is used in `getName()` but `_name` is used in the `__construct()` function.

Could you try to change `_name` to `name` in the constructor and see if this fixes the issue ?

It would be really cool if you could post a small snippet showing "how" you are instantiating the view. Just to be sure..

@realityking: I agree ;)

davidhurley:
Yep, I will give that a shot. 

And I'll post a short code block when I get on a computer.

elkuku:
great :)
Another (small) observation: The docblock declaration states `@var    array` which is obviously bad - while you are on it, could you also change this to `string` ?

davidhurley:
Yeah, I saw that and I will update it too ;)

elkuku:
Just a small note while we are on it..

`JView::assign()` and `JView::asignRef()`

``` php
<?php
foreach (get_object_vars($arg0) as $key => $val)
{
    if (substr($key, 0, 1) != '_')
    {
        $this->$key = $val;
    }
}
```

While I never really understood the use of these functions (`$this->assignRef('foo', $foo)` vs `$this->foo = $foo` since you have to access them as `$this->foo` in your view template anyway), they might cause some trouble in the "new" approach of deprecating underscored protected member vars as they are actually setters for the class....

davidhurley:
One example of my code calls are listed below...I am loading a custom fields view within another view on a separate model. One of my own library classes loads this function.  I'm updating the definition in JView and I'm looking into the modification of the constructor as a possible solution as well as recommended.

```
    $customFields = $customFieldModel->loadCustomFields($type, $id, $edit);
    $config['name'] = 'customfield';
    $customfieldView = new JView($config);
    $layout = self::getCustomFieldLayout($customfieldView, $type, $edit, $customLayout);
    $customfieldView->setLayout($layout);
    $customfieldView->assignRef('customFields', $customFields);
    $customfieldView->assignRef('type', $type);
    return $customfieldView;
```

davidhurley:
Upon updating the constructor to use $this->name instead of $this->_name the function does work as anticipated. 

I am unsure what if any other implications this change would make. Obviously, $this->name is not a deprecated variable ;) so perhaps it is a better solution. I have things open and can submit an updated pull request once a conclusion is reached on which function would be better to modify. 

Ideas?

chdemko:
See #672

chdemko:
Fixed in #1144

PR #691, opened 2012-01-01
Improper use of lock and unlock and racing condition in tablenested

marcel-kanter:
This is a change according to trackers:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27569
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27562

Reason for the changes are the work of LOCK TABLES SQL statement. it works differnt then intented by the php code here. Lock has no "Incremental" behaviour instead it is "Release-If-Any and Set-New". So in tablenested the tables got unlocked after modifing the assets

Implemented for SQL Server too (was missing).

The lock has to be made in combination for the both tables. Is the same bebaviour on all databases. mysql here as an example: http://dev.mysql.com/doc/refman/5.0/en/lock-tables.html is worth a reading.

Rules for Lock Release

When the table locks held by a session are released, they are all released at
the same time. A session can release its locks explicitly, or locks may be
released implicitly under certain conditions.
-  A session can release its locks explicitly with UNLOCK TABLES.
-  If a session issues a LOCK TABLES statement to acquire a lock while
  already holding locks, its existing locks are released implicitly before the
  new locks are granted.
-  If a session begins a transaction (for example, with START TRANSACTION), an
  implicit UNLOCK TABLES is performed, which causes existing locks to be released.
  (For additional information about the interaction between table locking and
  transactions, see Section 12.3.5.1, “Interaction of Table Locking and
  Transactions”.)

So a change in the lock/unlock functions are needed too.

### DIFF
--- a/libraries/joomla/database/database.php
+++ b/libraries/joomla/database/database.php
@@ -1204,9 +1204,9 @@ abstract class JDatabase implements JDatabaseInterface
 	 *
 	 * @param   string  $tables  A list of tables to lock.
 	 *
-	 * @return  boolean  true on success, false on failure
+	 * @return  JDatabase  Returns this object to support chaining.
 	 *
-	 * @since   XXX
+	 * @since   11.4
 	 * @throws  JDatabaseException
 	 */
 	public abstract function lockTables($tables);
@@ -1565,7 +1565,7 @@ abstract class JDatabase implements JDatabaseInterface
 	/**
 	 * Unlocks tables in the database.
 	 *
-	 * @return  boolean  true on success, false on failure
+	 * @return  JDatabase  Returns this object to support chaining.
 	 *
 	 * @since   11.4
 	 * @throws  JDatabaseException
--- a/libraries/joomla/database/database.php
+++ b/libraries/joomla/database/database.php
@@ -1200,16 +1200,16 @@ abstract class JDatabase implements JDatabaseInterface
 	}
 
 	/**
-	 * Locks a table in the database.
+	 * Locks tables in the database.
 	 *
-	 * @param   string  $tableName  The name of the table to unlock.
+	 * @param   string  $tables  A list of tables to lock.
 	 *
-	 * @return  JDatabase  Returns this object to support chaining.
+	 * @return  boolean  true on success, false on failure
 	 *
-	 * @since   11.4
+	 * @since   XXX
 	 * @throws  JDatabaseException
 	 */
-	public abstract function lockTable($tableName);
+	public abstract function lockTables($tables);
 
 	/**
 	 * Execute the SQL statement.
@@ -1565,7 +1565,7 @@ abstract class JDatabase implements JDatabaseInterface
 	/**
 	 * Unlocks tables in the database.
 	 *
-	 * @return  JDatabase  Returns this object to support chaining.
+	 * @return  boolean  true on success, false on failure
 	 *
 	 * @since   11.4
 	 * @throws  JDatabaseException
--- a/libraries/joomla/database/database/mysql.php
+++ b/libraries/joomla/database/database/mysql.php
@@ -457,22 +457,28 @@ class JDatabaseMySQL extends JDatabase
 		return mysql_insert_id($this->connection);
 	}
 
-	/**
-	 * Locks a table in the database.
-	 *
-	 * @param   string  $table  The name of the table to unlock.
-	 *
-	 * @return  JDatabaseMySQL  Returns this object to support chaining.
-	 *
-	 * @since   11.4
-	 * @throws  JDatabaseException
-	 */
-	public function lockTable($table)
-	{
-		$this->setQuery('LOCK TABLES ' . $this->quoteName($table) . ' WRITE')->query();
+  /**
+   * Locks tables in the database.
+   *
+   * @param   string  $tables  A list of tables to lock.
+   *
+   * @return  boolean  true on success, false on failure
+   *
+   * @since   XXX
+   * @throws  JDatabaseException
+   */
+  public function lockTables($tables)
+  {
+    foreach($tables as $i => $table)
+    {
+      $tables[$i] = $this->quoteName($table);
+    }
 
-		return $this;
-	}
+    $this->setQuery('LOCK TABLES ' . implode(' WRITE, ', $tables) . ' WRITE');
+    $this->query();
+
+    return true;
+  }
 
 	/**
 	 * Execute the SQL statement.
@@ -832,18 +838,19 @@ class JDatabaseMySQL extends JDatabase
 		return $error ? false : true;
 	}
 
-	/**
-	 * Unlocks tables in the database.
-	 *
-	 * @return  JDatabaseMySQL  Returns this object to support chaining.
-	 *
-	 * @since   11.4
-	 * @throws  JDatabaseException
-	 */
-	public function unlockTables()
-	{
-		$this->setQuery('UNLOCK TABLES')->query();
+  /**
+   * Unlocks tables in the database.
+   *
+   * @return  boolean  true on success, false on failure
+   *
+   * @since   11.4
+   * @throws  JDatabaseException
+   */
+  public function unlockTables()
+  {
+    $this->setQuery('UNLOCK TABLES');
+    $this->query();
 
-		return $this;
-	}
+    return true;
+  }
 }
--- a/libraries/joomla/database/database/mysql.php
+++ b/libraries/joomla/database/database/mysql.php
@@ -462,9 +462,9 @@ class JDatabaseMySQL extends JDatabase
 	 *
 	 * @param   string  $tables  A list of tables to lock.
 	 *
-	 * @return  boolean  true on success, false on failure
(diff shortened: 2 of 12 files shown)


marcel-kanter:
How to fix the test log? Are the functions _unlock and _lock of JTable tested? Then the test need to be removed
Is the test multithreaded?
Is it possible to get an output of the checkstyle?

realityking:
Why did you change them to always return true instead of allowing to chain them with other methods?

chdemko:
Whe did you change the installer library?

marcel-kanter:
@realityking For some reason in Joomla Exceptions and Errors are used to communicated normal things like the failure of functions. Exceptions are used to inform the caller about abnormal ( ;) ) events normally (an dead database is such a abnormal thing - but not the result of an query). Since SQL LOCK is a blocking query this would never fail.

@chdemko This should not be in this pull request. I'm new to git and somehow this got into this request. I hope i can remove these changes from this pull request. The reason for the change anyway is that i'm working on a library which has some dependancies so i need to check it before installing. The other thing is that the other extension types have support for it too so why not the rest of them?

mbabker:
@marcel-kanter please test #237 for adding that support for package extensions (trying to not duplicate efforts)

marcel-kanter:
@mbabker Thanks, I'll try that. Hmm. Just realised that completes the set. Thank you.

Somehow i cant find a link to manage the pull request on github... :S

elkuku:
I have a **personal** (non official) instance of the pull tester:
http://elkuku.github.com/pulltester/pulls/691

marcel-kanter:
@elkuku does your tester update autmaticly? committed some changes but tester did not update...

elkuku:
No, sorry mine is updated manually, but the official one is updated every three hours AFAIK.

elkuku:
... but I have updated it now ;)

marcel-kanter:
@mbabker The changes of the installer have moved to the branch called installer

chdemko:
@marcel-kanter Any news on this?

marcel-kanter:
No news on that. For correct use we need to change the prototypes of the general database class. I have no clue in which components use the database look. I think this is an real issue and we should let the change impact on the developers of these components. To be honest i thought that the changes in this pull will be introduced into joomla 2.5 since this is a major change - but  they were not implemented. Maybe in 3.0 release?

The next thing is that a complete look (even for reading) is necessairy. I found more places where looks are needed and an multi-user write access (what is normal for a cms) may corrupt data integrity. Generally speeking this is every class in which a read-modify access to the database is done.

The deeper i dig into the code of joomla, the more i think a complete re-design of some classes need to be done. This may sound a bit agressive but it should not - i know for shure how such projects are born and evolving... For this reason i startet to write a library with more or less compatible classes that plug onto the current joomla library. These classes i'll test exessivly- When these are ready for the public i'll introduce them to the public joomla library.

chdemko:
@marcel-kanter Could we close this pull request for the moment? Feel free to reopen it if new code is ready to test.

marcel-kanter:
ok, then close it.
how about the redesign?

chdemko:
@marcel-kanter you can propose code to implement this redesign but keeping backward compatibility

marcel-kanter:
Hmm. This means that the wrong way is kept in some classes... For correcting these errors a break with current design may be needed thats why i would introduce this to a new major release.

realityking:
Could you line out what changes are needed so we could discuss them before you spend a lot of time implementing it? I hate it when people spend a lot of time on implementing something that can't be accepted for some reason.

marcel-kanter:
Currently i have not much time to spent on that. Basicly a review of all classes file by file would lead to good results. A few things i got already:
- Alter lock and unlock of the database interface. Argument for lock should be an array for the table names and lock type.
- Remove all components that are not really necessairy for the platform and build packages for them. Then the user can choose which to install. For e.g. there are classes in joomla.database.tables that are part of components.
- Make the framework extentable via "application parts". The admin interface, the installer or the site of such things. Currently the site stuff resides on the same folder as the framework.
- The publish function has a somewhat strange name since it changes state (think of and trashed, archived!) and the state is stored in a table row called "publish" (in newer components "state")
- Review the plugin/Event system for state changes you get en event for all rows. For other events before and after the action (e.g. content display)
- Some functions dont use their arguments (e.g. the id in an table row) and return only the result from the class and not the correct value.
- Rework the asset/acl system. No fall back to some default values of root or component level (which is currently the case). This leads to security problems since then the user may be granted rights that he dont have. The implementation of the access levels itself are somewhat strange. "core.edit" means the right to edit the object itself. yes!?! in some component is also allows the user to edit the childs. - I have to take a closer look onto this to give you a more detailed answer.
- The predefined functions controller/model/view/table need some rework. publish in an table for example takes an array of PKs, delete only one PK.
- Some functions behave unnatural. Deleting an non-existent PK generates an error. Why? The purpose of delete is to remove an PK. So the model may be imformed about that but it should be happy. "I want that element to be removed! Oh. Nice it is not there. Fine Nothing to do." The table should be the work horse and the model+ controller the jokey. A good horse only runs where the jockey it says and only jumps so high it has to...
- Rewrite the installer to equalise the different types of install packets.

I can give you more points for this list. But this takes time. Some (maybe the most) of the points show a break with current implementation would be needed for correction...

realityking:
I misunderstood your comment - I suppose so did Christophe - but I was more asking about the lock/unlock table thing since that's the context of this pull request. Some of the other things on your list sound interesting as well but we gotta be careful with backwards compatibility, there's just now a big discussion about it on one of the CMS lists.

marcel-kanter:
Oh. Coming back to the context of this pull request. I have to admit that i did take a look into this pull request because i thought that it was clear what is the problem and that the last commit of it is regarded as an proposure for solving this. I'm a bit unhappy on how this went...

Please read the description of the LOCK statement of any SQL database (start with mysql then go on with any other). Then you should clearly see what is wrong with the current implementation.
To help you: Look at bullet #2 of the pull request description ( If a session issues a LOCK TABLES statement to acquire a lock while already holding locks, its existing locks are released implicitly before the new locks are granted. )
This is completely different to the current (and thus wrong) joomla attempt (There is also no way to emulate the joomla implementation)!

An example e.g. from delete function. The function modifies tables "list" (the nested list) and the table "assets". 
First it tries to lock the table "list" with _lock(). That is ok. After this it checks for trackassets and then it tries to lock the table "assets" with _lock(). This will work since SQL LOCK is blocking. _BUT_ the table "list" is UNLOCKED from this time. Thus for the following SQL statements the table list is completely unprotected! You see the problem?

The solution:
Remove of _lock and _unlock from table WITHOUT any successor. (As I did already in this pull request.)
Remove lockTable from database. (As i did already on this pull request
Add lockTables with array of the tables and the lock type to database interface (I did this already, but now it needs a change since the user of the function needs to set the lock type too.)

This solution breaks the backward compatibility (what is good since any class using this implementation has an error too).

Btw. The order of the update of the tables is wrong too. If you set up the FOREIGN KEY constraints correctly there should be an error since. list.asset_id references to assets.id and the row in assets is removed first - resulting in an reference to an non-existent element. Add this to the complete-redesign list above...

marcel-kanter:
The same thing happens with the JTableNested store function. In a but more extented verion, since it calls store of JTable which uses lock. At the beginning of store in JTable the current lock of the table is released and an new lock is requested. If one user is re-arranging the nodes (e.g. re-ordering) and another user is creating new nodes then the data in the table may become corrupted since the lft, rgt and level data can change.

marcel-kanter:
I have an solution for that problem. It a bit tricky, since there is now way to check granted locks, and will still break the backwards-compatiblity. I'll implement and test it on my own framework and then port it to joomla and re-open this pull request.

alxcube:
Hi! I am now working on some class that extends JTable, and discovered the same issue. Mysql really releases all table locks when another LOCK TABLES query is executed.  And also when parent method or some method that can be called directly or from another method calls _unlock() -- the table locks are realeased too. So here I propose some solution:

1. create new method JTable::_getTablesForLock() that should return array of tables which should be locked for the top level _lock method:
```
protected function _getTablesForLock() {
		$tables		= array($this->_tbl => 'WRITE');
		
		if ($this->_trackAssets) {
			$tables['#__assets']	= 'WRITE';
		}
		
		return $tables;
	}
```

2. create static property to hold locks counter:
`protected static $_locks_counter = 0`

3. modify _lock and _unlock methods:

```
protected function _lock() {
		if (self::$_lock_counter == 0) {
			$lockTables = $this->_getTablesForLock();
			$query		= 'LOCK TABLES ';
			$tables		= array();
			foreach ($lockTables as $table => $mode) {
				$tables[]	= $table . ' ' . $mode;
			}
			$query		.= implode(',', $tables);
			$this->_db->setQuery($query)->execute();
		}
		
		self::$_lock_counter++;
		
		return true;
	}
	
	protected function _unlock() {
		self::$_lock_counter--;
		
		if (self::$_lock_counter == 0) {
			$this->_db->setQuery('UNLOCK TABLES')->execute();
		}
		
		return true;
	}
```

PR #696, opened 2012-01-02
Make Installer package class vars public

mbabker:
Per discussion at http://groups.google.com/group/joomla-dev-platform/browse_thread/thread/e31b5078cb2816db, we need these class vars to be public otherwise PHP Fatal Errors are thrown preventing update/uninstall of Library and Package extension types.

### DIFF
--- a/libraries/joomla/installer/extension.php
+++ b/libraries/joomla/installer/extension.php
@@ -24,7 +24,7 @@ class JExtension extends JObject
 	 * @var    string
 	 * @since  11.1
 	 */
-	protected $filename = '';
+	public $filename = '';
 
 	/**
 	 * Type of the extension
@@ -32,7 +32,7 @@ class JExtension extends JObject
 	 * @var    string
 	 * @since  11.1
 	 */
-	protected $type = '';
+	public $type = '';
 
 	/**
 	 * Unique Identifier for the extension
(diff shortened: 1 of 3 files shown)


chdemko:
In fact we could have these attributes as public and read-only by only adding

```
/**
 * Magic method to get a protected field
 *
 * @param   mixed  $name  Name of the protected field to retrieve.
 *
 * @return  mixed  The protected field
 *
 * @since   11.4
 */
public function __get($name)
{
    if (property_exists($this, $name))
    {
        return $this->$name;
    }
    else
    {
        return null;
    }
}
```

to these classes (keeping the protected keyword)

mbabker:
I'm fine with either approach.  But, we need something otherwise we've
broken the installer for 2 (probably lesser used) extension types.

On 1/2/12 4:41 PM, "Christophe Demko"
<reply+i-2705216-d286499e6f3635c0b264e343d502e81306f4b08f-368545@reply.gith
ub.com> wrote:

pasamio:
We don't need to mark them as read only. Protect it from what? The class itself does nothing beyond set a few variables, it doesn't read them once it's set them or do anything fancy. We don't gain anything by marking it read only because the only thing that will interact with these fields post construction is all external! If anything it makes life difficult by obstructing access for no apparent reason to be able to update the values if need be for what ever reason (perhaps someone writes a serialisation mapping class).

PR #699, opened 2012-01-02
Add some simple HTML5 support to JDocument.

realityking:
Another split off #683.

After a discussion with Christophe I think this is to tightly couple to the CMS. This will do, we can do the template magic in JApplicationSite.

### DIFF
--- a/libraries/joomla/document/html/html.php
+++ b/libraries/joomla/document/html/html.php
@@ -94,6 +94,14 @@ class JDocumentHTML extends JDocument
 	protected $_caching = null;
 
 	/**
+	 * Set to true when the document should be output as HTML%
+	 *
+	 * @var    boolean
+	 * @since  12.1
+	 */
+	private $_html5 = null;
+
+	/**
 	 * Class constructor
 	 *
 	 * @param   array  $options  Associative array of options
@@ -303,6 +311,35 @@ class JDocumentHTML extends JDocument
 		$this->_custom[] = trim($html);
 
 		return $this;
+	}
+
+	/**
+	 * Returns whether the document is set up to be output as HTML5
+	 *
+	 * @return  Boolean true when HTML5 is used
+	 *
+	 * @since   12.1
+	 */
+	public function isHtml5()
+	{
+		return $this->_html5;
+	}
+
+	/**
+	 * Sets whether the document should be output as HTML5
+	 *
+	 * @param   bool  $state  True when HTML5 should be output
+	 *
+	 * @return  void
+	 *
+	 * @since   12.1
+	 */
+	public function setHtml5($state)
+	{
+		if (is_bool($state))
+		{
+			$this->_html5 = $state;
+		}
 	}
 
 	/**
--- a/libraries/joomla/document/html/renderer/head.php
+++ b/libraries/joomla/document/html/renderer/head.php
@@ -62,7 +62,13 @@ class JDocumentRendererHead extends JDocumentRenderer
 		$tagEnd = ' />';
 		$buffer = '';
 
-		// Generate base tag (need to happen first)
+		// Generate charset when using HTML5 (should happen first)
+		if ($document->isHtml5())
+		{
+			$buffer .= $tab . '<meta charset="' . $document->getCharset() . '" />' . $lnEnd;
+		}
+
+		// Generate base tag (need to happen early)
 		$base = $document->getBase();
 		if (!empty($base))
 		{
@@ -74,7 +80,7 @@ class JDocumentRendererHead extends JDocumentRenderer
 		{
 			foreach ($tag as $name => $content)
 			{
-				if ($type == 'http-equiv')
+				if ($type == 'http-equiv' && !($document->isHtml5() && $name == 'content-type'))
 				{
 					$content .= '; charset=' . $document->getCharset();
 					$buffer .= $tab . '<meta http-equiv="' . $name . '" content="' . htmlspecialchars($content) . '" />' . $lnEnd;


libraries/joomla/document/html/renderer/head.php
```diff
@@ -60,8 +60,14 @@ public function fetchHead(&$document)
 		$tab = $document->_getTab();
 		$tagEnd = ' />';
 		$buffer = '';
+
```
elkuku: whitespace ;)

piotr-cz:
great!
what do you think about more html5 stuff like simplified < script > and < link > tags?
That doesn't add much in terms of functionality but html is more readable & elegant :)

libraries/joomla/document/html/renderer/head.php
realityking: Fixed. Thanks.

realityking:
@piotr-cz
What do you mean be simplify script elements? I don't think they can be simplified with HTML5.
For the link element you probably mean to take off the "/". I'm kinda against this because someone may wanna use the platform (or even the CMS) to output XHTML5 - and it doesn't really hurt either.

bigbangireland:
@realityking I think @piotr-cz means that the type attribute is unnecessary for style, script and links when they are default values in HTML5.

```
<style type="text/css">
to
<style>

<script type="text/javascript">
to
<script>

<link rel="stylesheet" href="css/style.css" type="text/css" />
to
<link rel="stylesheet" href="css/style.css">

<script src="js/scripts.js" type="text/javascript"></script>
to
<script src="js/scripts.js""></script>
```

bigbangireland:
http://www.w3.org/TR/html5/links.html#link-type-stylesheet
http://www.w3.org/TR/html5/semantics.html#attr-style-type
http://www.w3.org/TR/html5/scripting-1.html#attr-script-type

chdemko:
I think it will be for P!12.x (need to release P!11.4 for now)

realityking:
Sure. We can back port it to the CMS if we want to later on.

piotr-cz:
yes, I've had in mind exactly what **bigbangireland** gave example of, now since we have html5 switch (is it going to be set in plugin/ configuration.php/ somewhere else?).
I'd also love to use some of the html5 forms functionality in JForm (I found the `placeholder` attribute extremely handy) but this would require proper discussion and probably it doesn't hurt to implement it later this year when more users move to html5 compatible browsers (ref: http://www.html5rocks.com/en/tutorials/forms/html5forms/). Anyway this isn't part of this pull request

bigbangireland:
A detailed discussion on HTML5 is needed well before Joomla! 3.0.

As @piotr-cz says JForm can not currently handle many new attributes. eg. "pattern" could be useful when used with a JS class. 

Greater thought is also needed in regards to microdata and handling tags (aside, section, nav, etc) on modules.

piotr-cz:
Question: from where to switch the $html5 variable? For me it would make most sense to use `JConfig`, as it's now in the platform.

**bigbangireland** > I've been thinking about exactly this for a while, I guess we could start a discussion on Platform group to share implementation experiences and thoughts (Sorry to write it here, but it's not possible to message you on GitHub).

bigbangireland:
@piotr-cz there was a discussion on the ideas forum a few months back. goo.gl/g1md6

Something like this:
templates/TEMPLATE_NAME/templateDetails.xml

```
    <html5>true</html5> or <doctype>html5</doctype>
```

I've temporarily added my email address to my public profile.

realityking:
I don't really wanna go into a lengthy discussion about the CMS stuff here so just a couple of points:

The way I see it the isHTML5() check is mainly for platform related stuff, including JForm. However we have to be careful that everything keeps working when isHTML5 returns false. So I'd like to take us the conservative approach. One things that we could due for starters is output the required attribute when isHTML5() returns true.

I really don't see this used for components or modules. I think template overrides are way better here. Of course nothing stops an extension dev from using it in this way.

If this gets committed I'll submit code for JApplicationSite and JApplicationAdmin that lets you set the value trough a parameter of the template. I originally had that code right in JDocument but it couples the platform to tightly to the CMS. The global configuration (JConfig) is a really poor place for it since you may have both HTML5 and XHTML templates (e.g. one admin one site or two different site templates)

piotr-cz:
**realityking** > Thanks this explanation, I get it now. You are correct about not using JConfig.

fevangelou:
We (JoomlaWorks) run a quick check on core code to see how it could be possible for Joomla! to read a template param (e.g. <doctype>html5</doctype>) and decide whether it should convert its output. I'm afraid this is not possible as it seems components may have already executed and therefore we cannot have full control of the output.

It's a lot easier to do this with a system plugin and it's what we're going for. **Check the source code** of this page to see the plugin in action: http://dev.joomlaworks.net/

The code is actually pretty small and conforms to this: http://dev.w3.org/html5/html4-differences/

libraries/joomla/document/html/html.php
```diff
@@ -94,6 +94,14 @@ class JDocumentHTML extends JDocument
 	protected $_caching = null;
 
 	/**
+	 * Set to true when the document should be output as HTML%
+	 *
+	 * @var    boolean
+	 * @since  11.1
+	 */
+	private $_html5 = null;
```
elkuku: If I understood right, even private member vars should go without a leading underscore. see this [comment](https://github.com/joomla/joomla-platform/pull/564#issuecomment-3100294). I think they are ugly ;)

realityking: Either way works for me, it should just be written down some place so we eventually get a consistent code style. (The good thing about private variables is however that we can rename them at will)

realityking:
Any news on this?

PR #702, opened 2012-01-03
Allow template and library extension types to use scriptfiles

marcel-kanter:
This pull request makes changes to the template and library extension installer to allow for a scriptfile to be included on a template or a library.

This will allow them to have a preflight method, for example to check for any dependencies, such as another extension's presence or a minimum Joomla! version.

The scripts stay together with the rest of the files in extension_root with the other files.


marcel-kanter:
Shall we fix the use of '/' in the files and replace them with DS?

libraries/joomla/installer/adapters/library.php
```diff
@@ -22,6 +22,15 @@
 class JInstallerLibrary extends JAdapterInstance
 {
 	/**
+	 * A path to the PHP file that the scriptfile declaration in
+	 * the manifest refers to.
+	 *
+	 * @var    string
+	 * @since  XXX
```
mbabker: This should be the current platform version being developed towards, currently 11.4

libraries/joomla/installer/adapters/library.php
```diff
@@ -111,6 +120,47 @@ public function install()
 			$this->parent->setPath('extension_root', JPATH_PLATFORM . '/' . implode(DS, explode('/', $group)));
 		}
 
+		// If there is an manifest class file, lets load it; we'll copy it later (don't have dest yet)
+		$manifestScript = (string) $this->manifest->scriptfile;
+		if($manifestScript)
```
mbabker: There should be a space after if

marcel-kanter:
I'll fix the style issues.
Shall we fix the use of '/' in the files and replace them with DS?

mbabker:
We're leaning more towards / versus DS since PHP is able to handle both slashes as separators just fine, including mixed use.

mbabker:
Not "official", but has a list of most of your code style errors as of yesterday - http://elkuku.github.com/pulltester/pulls/702

marcel-kanter:
Is it possible to trigger the style check manually? Since i've corrected the style errors in the last 2 commits and they are still reported? Just take the if in line 125 as an example. there is no missing space.
Used your checker already ;)

mbabker:
If you have phpcs installed, yes.  There's some info on developer.joomla.org to help you.

-Michael

mbabker:
I sent you a pull request updating the branch to the current master commit as well as fixing the code style errors.

marcel-kanter:
Pulled It. Tnx

libraries/joomla/installer/adapters/library.php
elinw: 12.1 I guess.

libraries/joomla/installer/adapters/library.php
```diff
@@ -191,28 +299,12 @@ public function install()
...
-		$this->set('name', $name);
-		$this->set('element', $element);
-		$installer = new JInstaller; // we don't want to compromise this instance!
-		$db = $this->parent->getDbo();
-		$db->setQuery('SELECT extension_id FROM #__extensions WHERE type="library" AND element = "' . $element . '"');
-		$result = $db->loadResult();
-		if ($result)
-		{
-			// Already installed, which would make sense
-			$installer->uninstall('library', $result);
```
pasamio, line 213: what is the reason to change this behaviour?

marcel-kanter, line 213: Install and Update are the same with a little but relevant difference: Install should run when it is not installed and update should run when there is already data oder files present.
1. This is what a creator of an library would expect: Install does install and update is update and NOT re-install.
2. The old code always removed the files before an re-install to do an update (which could be incremental) or there are maybe files that the install-script generated dynamicly and the update may need these files. Some inis for example...
3. Since the creator of the extension can do anything in the script, we should not asume that the library is only files anymore. So we should give the control to the creator of the library. Btw.  I'm already thinking about where to store configuration of my new library.
4. The installers for the other extensions are correct already. So why keep this incorrect behaviour here?

realityking:
The pull is not in a mergeable state.

PR #566, opened 2011-11-28
Eliminate check for admin application for using toolbar.

elinw:

PR #703, opened 2012-01-03
Restore the admin application check for toolbar.

elinw:
This undoes a previous change (#566) which had backward compatibility issues.


realityking:
I know it was this way before but we can prevent running some code when we write it this way:

```
if ($app->isAdmin() && $path = JApplicationHelper::getPath('toolbar'))
```

pasamio:
Let's just put it back the way it was and do a more thorough job of fixing the class later.

PR #705, opened 2012-01-04
JDate breaks valid DateTime input

bembelimen:
JDate assumes that every number is an unix timestamp, but according to the PHP documentation (see http://www.php.net/manual/en/datetime.formats.date.php) it is possible to use an 8 digit number as valid parameter which is not a timestamp (see ISO8601 Notations). This number consists the year, the month and the day without any separators (e.g. 20120101 for 2012/01/01). It's also valid to use only the year (four digits) as parameter.

So the first step is, that Joomla! should only handle numbers with more than 8 digits as unix timestamp.
The second step should be the removing of the whole test, because DateTime can handle unix timestamps by itself, if the programmer prepend an "@" char to the number (see http://www.php.net/manual/en/datetime.formats.compound.php "Localized Notations"). But this needs further testing, if there are any conflicts in the Joomla! environment. So this requests fixs only the "8 digit problem" and mark the test itself as deprecated.

### DIFF
--- a/libraries/joomla/utilities/date.php
+++ b/libraries/joomla/utilities/date.php
@@ -113,9 +113,16 @@ class JDate extends DateTime
 			}
 		}
 
-		// If the date is numeric assume a unix timestamp and convert it.
+		/**
+		 *  @deprecated
+		 *  This test is obsolete, because DateTime supports timestamps
+		 *  if there is a prepended "@" char, so everyone should handle
+		 *  this issue while calling JDate and not JDate itself
+		 */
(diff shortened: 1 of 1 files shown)


pasamio:
Wouldn't this break backwards compatibility for anyone with a date with 8 or less digits in it?

bembelimen:
Yep, some small numbers will be broken. So do you think, it would be better to mark the check as "deprecated" and don't change the function at all. And then remove the check complete in the following Joomla! version?

LouisLandry:

PR #707, opened 2012-01-05
Sqlexporter

elkuku:
This will fix the xml creation and add the ability to export also the data.


elkuku:
I fixed the failure (hopefully) but I am lost with the remaining 2 errors.
The function is broken and I can not understand how the unit tests could have passed before in the first place..

A quick test on the Joomla! CMS:

``` php
<?php
$exporter = new JDatabaseExporterMySQLi;

$result = (string) $exporter->setDbo(JFactory::getDbo())->from('#__users')->asXml();

// alternative:
// $result = (string) JFactory::getDbo()->getExporter()->from('#__users');

var_dump($result);
```

This gives a bunch of warnings like
`Notice: Trying to get property of non-object in .../libraries/joomla/database/database/mysqlexporter.php on line 176`

because JDatabase::getTableColumns() is called without the second parameter, so it will only return the field **types** but no objects.

The resulting XML file looks like this:

```
<table_structure name="#__users">
   <field Field="" Type="" Null="" Key="" Extra="" />
   <field Field="" Type="" Null="" Key="" Extra="" />
   <field Field="" Type="" Null="" Key="" Extra="" />
...
```

So the unit tests are definitely "faking" something here.. but where ?

ghost:
<?php
error_reporting(-1);
include_once 'libraries/import.php';
function dg($var){
    echo '<pre>'.htmlspecialchars(print_r($var, true)).'</pre>';
}
$db = jfactory::getDBO();
$e = $db->getExporter()->from($db->getTableList());
dg("$e");

this is how I do on manual test.

eddieajau:
Nikolai, I've patched up a few things and removed the need for the inspector class here:
https://github.com/eddieajau/joomla-platform/tree/elkuku-sqlexporter

Would you like to merge that into this branch?

elkuku:
It's a pleasure ;)

elkuku:
Hmm, might I place a small RFC here, somewhat related to the sqlexporter ?
I am currently playing in a multidb branch with options for multiple databases. The first goal was to have separate options for the CMS installation process, so I started in `libraries/cms`.
Then I wrote some basic importers and ended up with the crazy idea that the installer could accept xml files instead of sql files to install extensions. I did a small POC (included in the branch) and It seems to work :)

It looks like this: https://github.com/elkuku/joomla-cms/tree/multidb/libraries/cms/database

Now the question is if this actually makes any sense and, if so, to which part of the library should I send my pull request...

chdemko:
I had also this idea: define a common xml database schema and use xslt to translate this schema into mysql, postgresql, sqlazure, ...

chdemko:
See
http://xml2ddl.berlios.de/
http://erw.dsi.unimi.it/
http://www.omg.org/mda/mda_files/Model-Driven_Architecture.pdf
http://www.omg.org/spec/XMI/2.1.1/

eddieajau:
@elkuku in terms of renaming the classes, I'd put something on list because there are a few different thoughts the I know of regarding how that should be done (I don't care which one, just that we can all decide on one, hehe).  Regarding the installation app accepting XML, that would be awesome.  I originally created the import/export format for extension developers so they can just include the XML with their latest version of the extension, and the importer will just work out the diff without having complicated incremental upgrade files.  I expect the CMS could do the same thing (I'm not sure if the new repair db feature uses it or not, but it would probably become almost obsolete if the XML import/export was used).

@chdemko would you mind reviewing this pull as both Nikolai and I have worked on it?

libraries/joomla/database/database/mysqliexporter.php
```diff
@@ -54,7 +54,7 @@ public function check()
 	 *
 	 * @since   11.1
 	 */
-	public function setDbo(JDatabaseMySQLi $db)
```
chdemko, line 57: Why did you change this?

chdemko:
Except the comment above, all is ok for me. Should implement test with option 'with-data'

libraries/joomla/database/database/mysqliexporter.php
eddieajau, line 57: Oops.  Will have to fix that.

elkuku, line 57: It produces a PHP strict warning since the signature differs from the base class.

elkuku:
Sorry for the late reply..
As for the unit test: I have to confess, that I am a bit lost on how the database unit tests are supposed to work.. I did the first shy step and was honoured with a fatal error. I use the following chaining for exporting the data: `$rows = $this->db->setQuery($query)->loadObjectList();`
Unfortunately the Mockdatabasethingy's::setQuery() method returns void  - a clear show stopper. I tried to return a new JDatabaseGlobalMock, implementing a LoadObjectList() method that returns a FooBar object - but I feel that this is not the right way - any advise on this would be highly appreciated ;)

LouisLandry:

PR #709, opened 2012-01-05
Adds uninstall support for related media files

matrikular:
While un-/installing a few extension types, i noticed that media files won't be catched for the uninstall process of libraries.

### DIFF
--- a/libraries/joomla/installer/adapters/library.php
+++ b/libraries/joomla/installer/adapters/library.php
@@ -302,6 +302,7 @@ class JInstallerLibrary extends JAdapterInstance
 			}
 		}
 
+		$this->parent->removeFiles($xml->media);
 		$this->parent->removeFiles($xml->languages);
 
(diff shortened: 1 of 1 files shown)


pasamio:
Are the other types of installers that are accepting the media type work properly or was it just libraries?

matrikular:
Nope, everything else seems fine.

matrikular:
So, this patch didn't make it into 2.5 because?

mbabker:
This was committed to the Platform after the 11.4 tag, which is what CMS 2.5 is based on.  It's entirely possible to backport this into the CMS for the 2.5.1 release in March; just need to submit a bug report at http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemBrowse&tracker_id=8103 adding the code (you can do a pull request on joomla/joomla-cms with the code).

matrikular:
Ok, thanks for the explanation.

PR #713, opened 2012-01-06
Fixes issue raised in http://joomlacode.org/gf/project/joomla/tracker/?a...

sseshachala:
...ction=TrackerItemEdit&tracker_item_id=27630

Please suggest if there are alternatives.

### DIFF
--- a/libraries/joomla/database/database/sqlsrv.php
+++ b/libraries/joomla/database/database/sqlsrv.php
@@ -684,6 +684,23 @@ class JDatabaseSQLSrv extends JDatabase
 	}
 
 	/**
+	 * This function replaces a string identifier <var>$prefix</var> with the string held is the
+	 * <var>tablePrefix</var> class variable.
+	 *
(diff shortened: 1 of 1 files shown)


libraries/joomla/database/database/sqlsrv.php
```diff
@@ -684,6 +684,23 @@ public function query()
...
+	 * @param   string  $prefix  The common table prefix.
+	 *
+	 * @return  string  The processed SQL statement.
+	 *
+	 * @since   11.1
+	 */
+	function replacePrefix($sql, $prefix = '#__')
+	{
+		$newSQL = parent::replacePrefix($sql, $prefix);
+		return str_replace("#__", $this->tablePrefix, $newSQL);
```
jonnsl, line 700: If someone create some content(create a new article for example) with "#__" somewhere, it will be replaced by the prefix, and that is a serious security issue.

sseshachala:
ok - will add the condition.

PR #719, opened 2012-01-06
Joomla CMS [#27630] Issue with replacePrefix -- valid only for SQLServer 

sseshachala:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27630

sorry for the confusion.

### DIFF
--- a/libraries/joomla/database/database/sqlsrv.php
+++ b/libraries/joomla/database/database/sqlsrv.php
@@ -682,7 +682,6 @@ class JDatabaseSQLSrv extends JDatabase
 
 		return $this->cursor;
 	}
-	
 	/**
 	 * This function replaces a string identifier <var>$prefix</var> with the string held is the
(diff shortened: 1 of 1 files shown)


hooduku:
Can any one please tell me where is that 1 warning and 1 error? I know the warnings should be less than 200 or 199.

pasamio:
Line 685 has white space at the end, not sure where the extra is warning is triggered from with a quick examination.

If you install phpcs via PEAR (http://pear.php.net/package/PHP_CodeSniffer/), executing the following in the root of the platform should tell you what is wrong:

phpcs --standard="build/phpcs/Joomla" libraries/joomla/database/database/sqlsrv.php

elkuku:
The warning refers to a missing scope modifier for the function replacePrefix()
The white space in line 685 is still present ;)

sseshachala:
Can this merged in to the platform please?

PR #722, opened 2012-01-07
Fix category field to correctly manage edit.state permissions in combina...

elinw:
Fix category field to correctly manage edit.state permissions in combination with other permissions.

Currently when a user has edit for an item but not edit.state the category field returns a blank list of options. There are also situations where the user may have create in other categories but not edit.state in the category of the current item. In that case (contrary to current behavior) the option to change category should not be given but only the current category should be shown.

### DIFF
--- a/libraries/joomla/form/fields/category.php
+++ b/libraries/joomla/form/fields/category.php
@@ -45,6 +45,7 @@ class JFormFieldCategory extends JFormFieldList
 		$options = array();
 		$extension = $this->element['extension'] ? (string) $this->element['extension'] : (string) $this->element['scope'];
 		$published = (string) $this->element['published'];
+		$name = (string) $this->element['name'];
 
 		// Load the category options for a given extension.
 		if (!empty($extension))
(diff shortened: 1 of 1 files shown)


elinw:
I actually want to add that I don't know if this belongs in the platform except to the extent that  it already integrates permissions. I'm not sure that category by itself shouldn't just be a list of categories and then somehow the user can apply permission rules to it. But it is what it is right now, it's paying attention to some permissions but not others and that is making it broken for the ways that it is used. Part of me thinks that the cms should just have its own field extended from the base category field that works in ways that make sense in context.

libraries/joomla/form/fields/category.php
```diff
@@ -67,17 +68,44 @@ protected function getOptions()
...
 
-				foreach ($options as $i => $option)
+				// For new items we want a list of categories you are allowed to create in.
+				if (!$this->form->getValue($name))
 				{
-					// To take save or create in a category you need to have create rights for that category
-					// unless the item is already in that category.
-					// Unset the option if the user isn't authorised for it. In this field assets are always categories.
-					if ($user->authorise('core.create', $extension . '.category.' . $option->value) != true)
+					foreach ($options as $i => $option) {
```
elkuku, line 74: The opening curly brace should be on a new line.

PR #726, opened 2012-01-08
JText::plural(), dynamic plurals for arbitrary numbers

WebMechanic:
This oneliner patch allows "dynamic" pluralization of language keys for specific numeric values, eliminating the need to write and assign a callback for `getPluralSuffixes`. Essentially allowing any "magic" number passed into `JText::plural()` to result in more meaningful (semantic, human) phrases if a developer or template/UI designer wishes to do so.

Sample usage, numbers (7, 14, 30) are made up to illustrate the idea:
``` # standard behaviour
 STATISTICS="Statistcs"
 STATISTICS_0="No Statistcs available"
 STATISTICS_1="Todays Statistcs"
 STATISTICS_MORE="All Time Statistcs"
 # enhanced behavior
 STATISTICS_7="Weekly Statistcs"
 STATISTICS_14="Bi-Weekly Statistcs"
 STATISTICS_30="Monthly Statistcs"

``````
In code:
```echo JText::plural('Statistics'); // "Statistics"
echo JText::plural('Statistics', 0); // "No Statistcs Statistics"
echo JText::plural('Statistics', 1); // "Current Statistics"
echo JText::plural('Statistics', 7); // "Weekly Statistics"
echo JText::plural('Statistics', 14); // "Bi-Weekly Statistics"
echo JText::plural('Statistics', 30); // "Monthly Statistics"
echo JText::plural('Statistics', 42); // "All Time Statistics"```
``````

### DIFF
--- a/libraries/joomla/methods.php
+++ b/libraries/joomla/methods.php
@@ -228,6 +228,7 @@ class JText
 			// Try the key from the language plural potential suffixes
 			$found = false;
 			$suffixes = $lang->getPluralSuffixes((int) $n);
+			array_unshift($suffixes, (int) $n);
 			foreach ($suffixes as $suffix)
 			{
 				$key = $string . '_' . $suffix;
--- a/libraries/joomla/methods.php
+++ b/libraries/joomla/methods.php
@@ -228,6 +228,8 @@ class JText
 			// Try the key from the language plural potential suffixes
 			$found = false;
 			$suffixes = $lang->getPluralSuffixes((int) $n);
+			$suffixes[] = (int) $n;
+			array_unshift($suffixes, (int) $n);
 			foreach ($suffixes as $suffix)
 			{
 				$key = $string . '_' . $suffix;

### CODE REVIEW

**Comment by infograf768** [comment]
This has to be checked for "normal" JPlural breakage.

**Comment by WebMechanic** [comment]
@infograf768 thanks for your reply. I'm not sure what "JPlural" refers to. Did I miss a class?
the existing UnitTests apparently didn't break ;) and I tested this by means of using it in a current project where any of _0, _1 and _MORE are used "the normal way" and my component and some modules leveraging the ability of the new.

The current "behavior" of `JText::plural()` remains unaffected, including the CMS defaults that make `xx-YY.localise.php::getPluralSuffixes()` add explicit checks for the _0 _1 and _MORE suffixes and the implied fallback to the non-suffixed key. If `STATISTICS_7` isn't present in the list of language keys, it'll try `STATISTICS_MORE` as usual or eventually pick `STATISTICS`.

What this patch adds is also considering _specific_ number literals other than just 0 and 1 to become translatable _if present_ in the language array. Those numbers don't "stick" or have any side effects, as they're considered only for the very call of `JText::plural()` that gets this number passed in.

I don't know if some `getPluralSuffixes()` overrides for languages such as Welch, Irish or Czec already feature any smart code to support their special numbering (range) schemas by either introducing other "magic key suffixes" such as _MORE or any "math" to handle that.
However, even if they do use, let's say `_10`, to translate phrases for values "more than 10" as opposed to "equal to 10", that behavior would remain unaffected 'cos the number 10 is already considered a legitimate keyname suffix in that particular language anyway, and has been assigned a specific meaning and behavior for it; just like _MORE in general.
The patch would then just "unnessesarily" add 10 (again) as a possible suffix to the very end of a list that already considered this number in the first place.

As a consequence ("improvement") of this patch, `xx-YY.localise.php::getPluralSuffixes()` could effectively get loose of the two if-statements testing for `_0`and `_1`, and leave `_MORE` as the default alternative for the "uncounted" (pluralisation) keys,  'cos those two numbers (0, 1) are now covered implicitly by this patch.
It's a tiny "inprovement" as we'd exchange two local if-statements checking these two numers several hundred times in a page's life-cycle in favor of a more flexible one-liner ;)

**Comment by infograf768** [comment]
I meant JText::plural(), was just a shortcut. ;)

Here is one the more complex implementation of getpluralsuffixes we have for now (Scottish Gaelic)

public static function getPluralSuffixes($count) {
        if ($count == 0 || $count > 19) {
            $return =  array('0');
        }
        elseif($count == 1 || $count == 11) {
               $return =  array('1');
        }
        elseif($count == 2 || $count == 12) {
               $return =  array('2');
        }
        elseif(($count > 2 && $count < 12) || ($count > 12 && $count < 19)) {
                $return =  array('FEW');
        }
        return $return;
     }

**Comment by elkuku** [comment]
nplurals=4; plural=(n==1 || n==11) ? 0 : (n==2 || n==12) ? 1 : (n > 2 && n < 20) ? 2 : 3 [ref](http://translate.sourceforge.net/wiki/l10n/pluralforms)

**Comment by WebMechanic** [comment]
@infograf768 thanks for the Gaelic example; I had that in mind, but wrote Welch instead :) my bad.

@elkuku thanks, interesting reference and bookmarked :)
However, that's essentially the same expression but using ternaries and using '3' but 'FEW' at the end.

**This patch is not about pluralisation!** It's not intended to solve **any** _pluralisation rules_ for any given language on this planet, be it english, gaelic or klingon.
It is supposed to allow _one specific_ language key for _one specific_ number to bypass/override _on purpose_ whatever happens to be the default pluralization rule[s] for that language. E.g. in most cases that'd be xxx_MORE, and it'd remain effective for any number >= 2, same as xxx_FEW would continue to match 3-11 and such in Gaelic.
The latter has in fact more exceptions to those rules, in particular for multiples of 20, and whether $n is used as a cardinal or oridinal number. The grammar then not only ought to change, you have to use completly different words.
http://www.akerbeltz.org/beagangaidhlig/gramar/grammar_numerals.htm

Just like the number 12 is equal to "one dozen" in many languages, it's equally common to call a time period of "7 days" also "a week", "30 days" are known as "one month", or 512 "bits" x 2 = 1024 == "1 kilobyte" etc.

The current system would require to use an distinct "basekey" for any of these or perform some other odd trickery evalualting the number first and then decide which key to use:
`$dough .= ($eggs == 12) ?JText::_( 'DOZEN EGSS') : JText::plural('EGGS', $eggs);`

Or consider `JText::plural('DAYS', $n)` where $n is any of 0, 1, 5, 7, 14
"DAYS" = "Jours"
"DAYS_1" = "Un jour"
"DAYS_MORE" = "%d jours"
For both 7 and 14 it would return "7 jours" or "14 jours" respectively, whereas with the patch applied **and** adding the appropriate language key to one's .ini file, it _may_ now return:
"DAYS_7" = "Une semaine"
"DAYS_7" = "One week"
"DAYS_7" = "Eine Woche"

As for the Scots: That behavior won't change as well, **unless** anyone puts a language key such as "DAYS_7" into a gaelic language file **on purpose**, otherwise `JText::plural('DAYS', 7)` would result in "DAYS_FEW" and pick that key like it did before, or in lack thereof just "DAYS".

**Comment by WebMechanic** [comment]
Any feedback from someone in charge would be appreciated.
thanks.

File: libraries/joomla/methods.php [hunk 1]
```diff
@@ -228,6 +228,7 @@ public static function plural($string, $n)
 			// Try the key from the language plural potential suffixes
 			$found = false;
 			$suffixes = $lang->getPluralSuffixes((int) $n);
+			$suffixes[] = (int) $n;
```
chdemko: use

```
array_unshift($suffixes, (int)$n);
```

to give priority to the more specific rule

chdemko: if you add the array_unshift, you have to delete this line

WebMechanic: argh... dammit :-) Had a "bad merge" on my side prior pushing the changeset; I'll amend this commit (again).

PR #731, opened 2012-01-10
Joomla CMS [#27664] SQL Server - Language column in grids displays as "Undefined" if item language set to All 

sseshachala:
The patch fixes the issue raised in http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27664

### DIFF
--- a/libraries/joomla/application/component/modellist.php
+++ b/libraries/joomla/application/component/modellist.php
@@ -133,6 +133,14 @@ class JModelList extends JModel
 			$this->setError($this->_db->getErrorMsg());
 			return false;
 		}
+		//For SQLServer - language has to be trimmed.
+		foreach($items as $item)
+		{
+			if(isset($item->language))
+			{
+				$item->language = trim($item->language);
+			}
+		}
 
 		// Add the items to the internal cache.
 		$this->cache[$store] = $items;


sseshachala:
The reason we have to trim -
In SQLServer/SQLAzure - the language column is of ncharacter data type with size as 7. This will insert white spaces if there are characters less than 7.

sseshachala:
We cannot modify the data type to nchar - nvarchar - since MS SQLServer recommends to use nchar/
I am not sure if we can modify in JDatabaseSQLSrv. We get the row object in a generic way and cannot trim the column name explicitly.
If there are alternatives, please do recommend.

realityking:
First of all the issue is due to this code that is actually in the list view (the layout to be precise).

```
<?php if ($item->language=='*'):?>
    <?php echo JText::alt('JALL','language'); ?>
<?php else:?>
    <?php echo $item->language_title ? $this->escape($item->language_title) : JText::_('JUNDEFINED'); ?>
<?php endif;?>
```

Since we can't change all them we have to move further up and the first thing they'd all have in common is indeed JModelList. If we don't wanna couple the platform to the CMS in this way (on prevent iterating over the items twice) I think it would actually be an acceptable solution to just change all the core layouts. Extension will have to be adapter for database abstraction anyways.

Since this difference in behavior is common to all fields that are char(mysql)/nchar(mssql) maybe it should be fixed at the database driver level? This would mean modifying JDatabaseSQLSrv::fetchObject() and probably JDatabaseSQLSrv::fetchAssoc() and JDatabaseSQLSrv::fetchArray() too trim all fields.

This would make using MS SQL Server more transparent for developers, but I suspect there are performance implications.

sseshachala:
I dont think it is wise to do it at the database level without substantial perf. hit especially if we have to do trim on all columns.

PR #732, opened 2012-01-10
Be positive: Allow experimental database drivers to be installed to the application

elkuku:
This will allow the installation and testing of new experimental database drivers.

With the current approach we have a list of hard coded "supported" database drivers. <del>Select lists will return this even if the driver is not supported by the server.</del>

It is also impossible to install drivers for other databases, even if they are actually "supported" by the server _and_ the application (namely PostgreSQL and SQLite)

The currently available database drivers in the platform as of version 11.4 are all supported by the CMS, so I believe there is no need to have them hard coded.

This patch will change the list from "supported to "unsupported" (currently none), so if there appears a database driver in one of the upcoming platform releases that gets merged into the CMS and is not supported by the CMS its name can be added to the list.

Experimental drivers (so far)
- PostgreSQL: https://github.com/gpongelli/joomla-cms/tree/postgresql
- SQLite: https://github.com/elkuku/joomla-cms/tree/sqlite

CMS tracker: [#27690](http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27690)

### DIFF
--- a/libraries/joomla/form/fields/databaseconnection.php
+++ b/libraries/joomla/form/fields/databaseconnection.php
@@ -36,6 +36,8 @@ class JFormFieldDatabaseConnection extends JFormFieldList
 	 *
 	 * This method produces a drop down list of available databases supported
 	 * by JDatabase drivers that are also supported by the application.
+	 * Applications may specify either a comma separated list of "supported" or "unsupported"
+	 * database drivers.
 	 *
 	 * @return  array    The field option objects.
 	 *
@@ -52,10 +54,18 @@ class JFormFieldDatabaseConnection extends JFormFieldList
 		 */
 		$unsupported = explode(',', $this->element['unsupported']);
 
+		$supported = ((string) $this->element['supported']) ? explode(',', $this->element['supported']) : array();
+
 		// This gets the connectors available in the platform and supported by the server and the application.
 		foreach (JDatabase::getConnectors() as $connector)
 		{
-			if(in_array($connector, $unsupported))
+			if (in_array($connector, $unsupported))
+			{
+				// The connector is not supported by the application
+				continue;
+			}
+
+			if (count($supported) && ! in_array($connector, $supported))
 			{
 				// The connector is not supported by the application
 				continue;
--- a/libraries/joomla/form/fields/databaseconnection.php
+++ b/libraries/joomla/form/fields/databaseconnection.php
@@ -40,38 +40,28 @@ class JFormFieldDatabaseConnection extends JFormFieldList
 	 * @return  array    The field option objects.
 	 *
 	 * @since   11.3
-	 * @see		JDatabase
+	 * @see     JDatabase
 	 */
 	protected function getOptions()
 	{
-		// Initialize variables.
-		// This gets the connectors available in the platform and supported by the server.
-		$available = JDatabase::getConnectors();
-
 		/**
-		 * This gets the list of database types supported by the application.
+		 * This gets the list of database types unsupported by the application.
 		 * This should be entered in the form definition as a comma separated list.
-		 * If no supported databases are listed, it is assumed all available databases
+		 * If no unsupported databases are listed, it is assumed all available databases
 		 * are supported.
 		 */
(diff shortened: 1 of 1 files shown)


realityking:
Would it be possible to support both, a whitelist and a blacklist? (mutually exclusive)

If I write an app that only supports mysql and mysqli I shouldn't have to change my forms every time the platform adds a new driver.

elkuku:
Sounds like a valid request ;)

libraries/joomla/form/fields/databaseconnection.php
```diff
@@ -36,42 +36,42 @@ class JFormFieldDatabaseConnection extends JFormFieldList
...
 		 * This should be entered in the form definition as a comma separated list.
-		 * If no supported databases are listed, it is assumed all available databases
+		 * If no unsupported databases are listed, it is assumed all available databases
 		 * are supported.
 		 */
-		$supported = $this->element['supported'];
-		if (!empty($supported))
+		$unsupported = explode(',', $this->element['unsupported']);
+
+		$supported = ((string) $this->element['supported']) ? explode(',', $this->element['supported']) : array();
```
realityking, line 63: Why does this line check if the argument is actually a string and line 55 not? they basically do the same thing.

libraries/joomla/form/fields/databaseconnection.php
```diff
@@ -36,42 +36,42 @@ class JFormFieldDatabaseConnection extends JFormFieldList
...
 	 */
 	protected function getOptions()
 	{
-		// Initialize variables.
-		// This gets the connectors available in the platform and supported by the server.
-		$available = JDatabase::getConnectors();
-
 		/**
-		 * This gets the list of database types supported by the application.
+		 * This gets the list of database types unsupported by the application.
```
realityking, line 50: Comments have to be changed to reflect the new code.

libraries/joomla/form/fields/databaseconnection.php
elkuku, line 63: if an empty string is set, line 55 will produce an array with one empty item - not a problem in the $unsupported array, but for the $supported array..

BTW, is does not check but cast because $this->element is a SimpleXMLElement ;)

libraries/joomla/form/fields/databaseconnection.php
elkuku, line 50: Fixed. Thanks ;)

elinw:
So as I discussed on the CMS tracker, I guess my feeling is that it is not possible in any way to list all possible databases that you don't support. At the same time if you are an advanced enough user to be testing database drivers shouldn't you be advanced enough to edit the places the field is used to add to the list? 
I think we need to understand that some applications will not support all databases and will have for example Oracle specific code or MySQL specific code not to mention non relational types. Now, I do think that adding a black list that would basically be used not for all hypothetical databases but for those databases for which there are drivers and which are supported on your server that are not supported by the application makes some sense. But that doesn't really deal with the scenario Nikolai proposed which is adding a new driver to the libraries.
I very much like the idea of being able to easily add to  the default drivers, and maybe I'm overthinking though. 

I'm also wondering whether you are seeking a CMS focused solution in the platform. My feeling is that perhaps we can in the CMS have a way of designating certain drivers as experimental and not fully supported (i.e. we don't promise that they will work in all situations).  Even in a release we could say this driver is a beta. Right now the field is written in a way that if no supported databases are specified we assume all of the databases with drivers are allowed. As stated earlier, we currently support all of the drivers in the platform, so for now we could just leave off the supported element and solve this problem in the future more generally. I have no idea in the world how the CMS would test on Oracle to the standard of testing that we assume, we can barely manage on MS databases or PostGres.

elinw:
Actually just wanted also to mention that as Nikolai acknowledged on the CMS tracker the current implementation does not return options that are not available on the server.

chdemko:
@elkuku Any news on this?

elkuku:
No.
Seems that it won't be accepted, so I am going to close it myself ;(

PR #734, opened 2012-01-11
Remove the remaining occurences of DS except for its definition and JParameter.

realityking:
This gets us 90% of the way on removing DS. As soon as we remove JParameter it won't be used by the platform anymore and we can decide to remove it.

### DIFF
--- a/libraries/joomla/filesystem/folder.php
+++ b/libraries/joomla/filesystem/folder.php
@@ -661,7 +661,6 @@ abstract class JFolder
 	 */
 	public static function makeSafe($path)
 	{
-		//$ds = (DS == '\\') ? '\\/' : DS;
 		$regex = array('#[^A-Za-z0-9:_\\\/-]#');
 		return preg_replace($regex, '', $path);
 	}
--- a/tests/suite/joomla/cache/storage/JCacheStorageMock.php
+++ b/tests/suite/joomla/cache/storage/JCacheStorageMock.php
@@ -7,7 +7,7 @@
  * @license     GNU General Public License version 2 or later; see LICENSE
  */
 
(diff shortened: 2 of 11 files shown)


tests/suite/joomla/cache/storage/JCacheStorageMock.php
```diff
@@ -7,7 +7,7 @@
  * @license     GNU General Public License version 2 or later; see LICENSE
  */
 
-JLoader::register('JCacheStorage', JPATH_PLATFORM.DS.'joomla'.DS.'cache'.DS.'storage.php');
+JLoader::register('JCacheStorage', JPATH_PLATFORM . 'joomla/cache/storage.php');
```
elkuku: missing /

realityking: Fixed. Thanks.

libraries/joomla/filesystem/folder.php
```diff
@@ -661,7 +661,6 @@ public static function listFolderTree($path, $filter, $maxLevel = 3, $level = 0,
 	 */
 	public static function makeSafe($path)
 	{
-		//$ds = (DS == '\\') ? '\\/' : DS;
 		$regex = array('#[^A-Za-z0-9:_\\\/-]#');
```
pasamio, line 664: I just realised something, does this mean you can't have folders with periods in them? Like hidden folders?

piotr-cz, line 664: What about non-latin characters (ie. ěščłż), those are permitted in filesystem?
I discourage using them by all means, but I know some people will do it anyway - I have such experience with CMS users who have permission to use media folder.

realityking, line 664: Kinda a different topic :)

chdemko:

realityking:
I fixed the conflicts.

PR #738, opened 2012-01-12
Update libraries/joomla/html/html/menu.php

vinset:


elinw:
http://code.joomla.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27702

PR #740, opened 2012-01-12
JUser accesslevel and usergroup caching off empty object var and not object id

krileon:
Current implement caches based off object variable being empty or not. This however does not function if you build the user object before a user logs in as it then populates the variable with guest accesslevels and usergroups. Once login is triggered if the object is persisted then it persists with those guest accesslevels and usergroups and prevents menu items for example displayed only to registered users from displaying.

Proposed implemented keeps caching intact, but rather then caching if variable is populated or not it caches based off the object id. This prevents duplicate calls to JAccess::getAuthorisedViewLevels per page load, but maintains accuracy of user accesslevels and usergroups when transitioning from guest to logged in.

### DIFF
--- a/libraries/joomla/user/user.php
+++ b/libraries/joomla/user/user.php
@@ -431,9 +431,12 @@ class JUser extends JObject
 			$this->_authLevels = array();
 		}
 
-		if (empty($this->_authLevels))
+		static $_authLevels = array();
+
+		if (!isset($_authLevels[$this->id]))
 		{
 			$this->_authLevels = JAccess::getAuthorisedViewLevels($this->id);
+			$_authLevels[$this->id] = $this->_authLevels;
 		}
 
 		return $this->_authLevels;
@@ -452,9 +455,12 @@ class JUser extends JObject
 			$this->_authGroups = array();
 		}
 
-		if (empty($this->_authGroups))
+		static $_authGroups = array();
+
+		if (!isset($_authGroups[$this->id]))
 		{
 			$this->_authGroups = JAccess::getGroupsByUser($this->id);
+			$_authGroups[$this->id] = $this->_authGroups;
 		}
 
 		return $this->_authGroups;


mahagr:
For example having system plugin that calls:
    JFactory::getUser()->getAuthorisedViewLevels(); 
    JFactory::getUser()->getAuthorisedGroups(); 
WILL break login from your site as it makes Joomla to think that the user has only guest access levels and user groups.

In the real life, this bug breaks login inside Kunena when CB integration has been enabled. The bug was introduced in here: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=23336

There are also some other ways to fix this bug, like clearing cached values from the user object during JApplication::login() or creating a new object for registered user (later one may cause other troubles, though).

elinw:
Since this is a backward compatibility break can we please at minimum return to what was there before?

mahagr:
I'm not sure how patch from @krileon breaks backwards compatibility. It does the same as Joomla 1.7.2, but fixes ACL caching that broke the old behavior in 1.7.3.

mahagr:
Authentication loads the new user like this:

/plugins/user/joomla/joomla.php:

```
if ($id = intval(JUserHelper::getUserId($user['username'])))  {
    $instance->load($id);
    return $instance;
}
```

I don't understand why the code is getting instance from the current user (usually id=0, but nothing prevents it from being existing user), as JUser object seems to contain no variables that will be kept from guest to the logged in user -- except these new protected variables (and isRoot), which will give user the wrong access levels and groups.

So to me the code should have: $instance = new JUser; which makes login to be effective only after current request. But...

Getting instance will change JUser::getInstance() to contain the new logged in user, but on the other hand it also affects JUser::getInstance(0) inside the current request. If this behavior is needed (Joomla seems to store new values into session), it's much safer to clear cached values inside JUser::load().

The last option seems to be the best one: if we change JUser::load() to clear cached/calculated values, the effect is the same as in this pull request, but without extra layer of caching. It will not introduce any backward incompatible changes (from J1.7.2), as the result is the same as with no caching, which was the old behavior.

elinw:
No I meant the problem was a break from 1.7.2.

elinw:
So the change Matias referenced was done before 1.6. Has this issue been present throughout this release cycle?

mahagr:
The issue has been present from Joomla 1.7.3 -- at least we haven't got any bug reports before Joomla 1.7.3 and Kunena 1.7.1 went out -- now there are a few. In Kunena the logic has been the same from Kunena 1.6.3, which was the first version that run on Joomla 1.6.

We have worked around the bug in the upcoming Kunena 1.7.2, but the issue is clear -- you cannot call these functions before you login:
JFactory::getUser()->getAuthorisedViewLevels();
JFactory::getUser()->getAuthorisedGroups(); 
as Joomla will clone the cached information to the new user. Also emergency root account flag gets copied, so I see that this is a potential security bug.

So which way to fix this bug? There are currently 3 proposals. @LouisLandry, @eddieajau?

mahagr:
Pull Request #1356 is separate issue to this pull request. This issue is just present only during authentication as cached values aren't cleared from the JUser object when user id changes.

Another pull request deals with issue where JFactory::getUser() doesn't return user at all. But both bugs seem to have introduced at the same time -- when the function was changed to allow better caching.

realityking:
I can see how this breaks but I wonder why would you access JUser::getAuthorisedViewLevels() or JUser::getAuthorisedGroups() in an event that fires before the user is logged in?

mahagr:
I'n this case it was a 'mistake' in a plugin that called a function which happened to call the JUser functions. But I still consider this to be a bug, as it is perfectly legal to call JUser::getAuthorisedViewLevels() in a plugin -- for example in advanced caching solution. 

As several extensions handle login by themselves (including CB and Kunena) such caching solution has no way to know that user is getting logged in just after the event fires.

elinw:
Wouldn't you do it for the guest group and level?  Also #1474 relates to the issue of guest status staying in place after login in some situations.

mahagr:
Commit #1474 doesn't fix this particular issue, but is related to this one.

All the issues come from reusing existing JUser object instead of creating a new one. Whenever someone adds a new feature (= caching) to JUser object they do not assume that object will remain the same before and after login -- or between two requests. So if you forget to clear the new protected variable during login, there is no guarantee that those variables have the correct values for the current user. 

Maybe JUser object should have a function called cleanCache() to make sure that all protected variables (and session) reset their values to the default?

ianmacl:
So this one has sat a long time.  I think rather than adding the static we should make the privates arrays and just store the data in there by array directly, if that makes sense.  If you (or somebody) makes those changes I would be happy to merge them.  I'm going to close the request - ask for it to be reopened when that is done.

Denitz:
Please check new patch in
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=23336

Now we use static caching of auth levels, groups and categories separated per
user ID, so we don't have issues with reloading JUser instance with new user
data.

PR #743, opened 2012-01-12
Remove deprecated JDatabase* code

mbabker:
Removed from JDatabase\* classes is most of the 12.1 deprecated code.  Items that remain are getErrorNum, getErrorMsg, stdErr, $errorNum, and $errorMsg.


realityking:
+1 from me

(Just of of curiosity, why the many commits?)

mbabker:
Was pretty much testing one at a time to make sure they weren't still in
use and the unit tests didn't break (as evidenced by the one oops
commit...)

On 1/12/12 11:38 PM, "Rouven Weßling"
<reply+i-2824426-f88893890721c36090d68a78c46e72bb5d4cd1dc-368545@reply.gith
ub.com> wrote:

tests/includes/mocks/JDatabaseMock.php
```diff
@@ -34,12 +34,10 @@ public static function create($test)
 	{
 		// Collect all the relevant methods in JDatabase.
 		$methods = array(
-			'addQuoted',
 			'connected',
 			'debug',
```
chdemko: Is this now deprecated ?

realityking: It's actually removed in this pull, he probably overlooked it. the replacement is setDebug().

mbabker: Fixed.

eddieajau:
Rock on :)

PR #744, opened 2012-01-12
Remove JParameter and JElement

mbabker:


chinhpv:
Hi Michael,
I have a question about this, why the JElement class is removed in Joomla 3, and when i need this class, i should use what class to replaced JElement.
Regards

pasamio:
In general JForm has replaced the functionality provided by JParameter. Additionally many tasks that used JParameter can transparently use JRegistry.

infograf768:
JRegistry indeed works great to retrieve a plugin parameter in a component. Updated wiki.

PR #748, opened 2012-01-14
There is no reason to protect

romacron:
protected function dayToString($day, $abbr = false)
protected function monthToString($month, $abbr = false)


elinw:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27722

romacron:
thanks for fast changing !!!

PR #749, opened 2012-01-14
Fix of getQuery.

gpongelli:
Added a $query object to database class that holds JDatabaseQuery object, created and returned by getQuery() .
Aligned getQuery to return only JDatabaseQuery object.

### DIFF
--- a/libraries/joomla/database/database.php
+++ b/libraries/joomla/database/database.php
@@ -115,6 +115,12 @@ abstract class JDatabase implements JDatabaseInterface
 	protected $offset = 0;
 
 	/**
+	 * @var    JDatabaseQuery  Current query object created and returned by getQuery.
+	 * @since  12.1
+	 */
+	protected $query = null;
+
+	/**
 	 * @var    mixed  The current SQL statement to execute.
 	 * @since  11.1
 	 */
--- a/libraries/joomla/database/database/mysql.php
+++ b/libraries/joomla/database/database/mysql.php
@@ -288,11 +288,11 @@ class JDatabaseMySQL extends JDatabase
 	}
 
 	/**
-	 * Get the current or query, or new JDatabaseQuery object.
+	 * Get the current query object or a new JDatabaseQuery object.
 	 *
-	 * @param   boolean  $new  False to return the last query set, True to return a new JDatabaseQuery object.
+	 * @param   boolean  $new  False to return the current query object, True to return a new JDatabaseQuery object.
 	 *
-	 * @return  mixed  The current value of the internal SQL variable or a new JDatabaseQuery object.
+	 * @return  JDatabaseQuery  The current query object or a new object extending the JDatabaseQuery class.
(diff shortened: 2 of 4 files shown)


pasamio:
So this means you can't use getQuery to get the last set query for debugging purposes and you're assuming people are always using a query object which again may not be the case, someone could easily set a string and this would break.

gpongelli:
I've only aligned specific getQuery to JDatabase's comment, changed in [this commit (6th November 2011)](https://github.com/joomla/joomla-platform/blob/ff29b837bf1421a01447756a675fc37b868b66cd/libraries/joomla/database/database.php#L657) .

If you wish get sql string from JDatabaseQuery object, you've simply to call <code>(string) $db->getQuery()</code> .

You're right about setting strings, but sooner I think will be added JDatabaseQuery functionality to remaining clauses and use of setQuery without JDatabaseQuery will be deprecated .

I've added these changes because, inside my postgresql's version of cms, I need to get table name from INSERT INTO query to emulate mysql_insert_id, that task is simple when using JDatabaseQuery ($query->get('insert')->getElements()) and more difficult when was set a string.

eddieajau:
I think at the platform level we want to allow developers to use raw SQL.  Not every application is going to want to use the query object.  A simple CLI may not want to bother with the query object.

gpongelli:
I've added this patch because while I was trying to add postgresql driver inside cms I've encountered an issue with insertid call.
Inside insertid I've a getQuery to get last INSERT INTO query, but getQuery returns a null object, not a string nor a JDatabaseQuery object.
So I've added this constraint to use JDatabaseQuery, as expected watching docblock, but it was made as temporary fix waiting to know why getQuery returns null value.

Ing. Gabriele Pongelli

AVVERTENZE AI SENSI DEL D.LGS. 196/2003
Le informazioni contenute in questo messaggio di posta elettronica e negli eventuali files allegati, sono da considerarsi strettamente riservati. Il loro utilizzo è consentito esclusivamente al destinatario del messaggio, per le finalità indicate nel messaggio stesso. Qualora riceveste per errore questo messaggio, Vi preghiamo cortesemente di darcene notizia all'indirizzo e-mail di cui sopra e di procedere alla distruzione del messaggio stesso, cancellandolo dal Vostro sistema; costituisce comportamento contrario ai principi dettati dal D.lgs. 196/2003 il trattenere il messaggio stesso, divulgarlo anche in parte, distribuirlo ad altri soggetti, copiarlo, od utilizzarlo per finalità diverse.
This electronic transmission is strictly confidential and intended solely for the addresses. It may contain information which is covered by legal, professional or other privilege. If you are not the intended addressee, you must not disclose, copy or take any action in reliance of this transmission. If you have received this transmission in error, please notify us and delete the received data as soon as possible.

Il giorno 20/gen/2012, alle ore 08:32, Andrew Eddiereply@reply.github.com ha scritto:

PR #750, opened 2012-01-14
Fixing default value issue in arguments

fastslack:
The true value is true only in PHP, but in CLI it returns 1 (int). What about if i need to use the number 1 into the argument too?
Ex.

$ ./JCli --testing
Returns 1
$ ./JCli --testing 1
Returns 1

I need to filter this, so i changed the true (bool) to 'true' (string) so can help me to filter it into the CLI script.


pasamio:
Aren't you just shifting the problem for someone who might be wanting to test for 'true'?

pasamio:
Alternatively, have you tried using the triple equals (===) to do a type check?

fastslack:
Yes, but is the same type. If you try this:

``` php
$argv = $this->input->get('argv', 0);
$type = gettype($argv);
echo "Value: {$argv} -- {$type} \n";
```

It returns:

``` bash
$ ./CliScript --argv 59
Value: 59 -- string 

$ ./CliScript --argv 
Value: 1 -- string 
```

WebMechanic:
I think that's just how a shell works. There's no notion of "integer" oder "boolean" data types in a shell context so PHP gets what the shell provides: strings.

``` <?php
// $ php -f retval.php 1 "2"
print_r($argv);
echo gettype($argv[1]), PHP_EOL;
echo gettype($argv[2]), PHP_EOL;
exit( count($argv) );
```

gives this result in plain vanilla php-cli:

`````` Array
(
    [0] => retval.php
    [1] => 1
    [2] => 2
)
string
string```
If you need on/off switches see http://de.php.net/manual/en/function.getopt.php for a little more control.
Maybe you can teach JInputCLI to accept some options for `getopt()` and make use of it?
``````

**Comment by fastslack** [comment]
Sorry, Sam was right, i can tell to the filter what type i like to use... 

```
$value = $this->input->get('argv', 0, 'int');
$type = gettype($value);
```

Returns

```
string(7) "integer"
```

I have a question about this... why this is not automatically calling gettype() ?

**Comment by WebMechanic** [comment]
> why this is not automatically calling gettype() ?
- for what general purpose?
- at what point
- based on which assumption(s)?

It's up to the developer to decide how (incomming) data's supposed to be transformed, validated, and used.

One might have a very good reason in an application to use an argument that consists of a character such as '1' either as a number, a string or a boolean. Just because something **looks** like a number or a sequence of characters that incidently form the word "true" doesn't mean anything.
`$ php app.php -- -queue=true`

``` PHP
// perform tasks according to '-queue'
$tasks = str_split($queue);
$transaction = false;
if (in_array('t', $tasks)) {
    $transaction = startTransaction();
}
if (in_array('r', $tasks)) {
    resetCache();
}
if (in_array('u', $tasks)) {
    updateStuff();
}
if ($transaction == true) {
    endTransaction();
}
if (in_array('e', $tasks)) {
    sendEmail();
}
```

fastslack:
Thanks for all the explanation... closing this pull request.

Regards

fastslack:
I need to reopen this pull request because seems that the JInput::clean method change the type of the input. 

With this patch 8b77168, the devs have the control if they like to cleanup the input or not. If you do:

``` php
$args = $this->input->get('args', 0, 'int', false);
echo gettype($args);
```

It returns:

``` bash
$ ./CliScript --args 1
string

$ ./CliScript --args 
boolean
```

realityking:
Instead of passing 'int' you could pass 'none' if you don't want any filtering. I'm gonna close this, if there's still an issue please post.

PR #754, opened 2012-01-16
Add Union method

elinw:
This adds a union method to JDatabaseQuery along with unionDistinct which provides a proxy to the union method with distinct. This is based on the method that was in the JXtended libraries but updated for the current implementation of JDatabaseQuery. Tests included.

### DIFF
--- a/libraries/joomla/database/query.php
+++ b/libraries/joomla/database/query.php
@@ -244,11 +244,6 @@ abstract class JDatabaseQuery
 	protected $union = null;
 
 	/**
-	 * @var    JDatabaseQueryElement  The unionDistinct element.
-	 * @since  12.1
-	 */
-	protected $unionDistinct = null;
-
 	/**
 	 * Magic method to provide method alias support for quote() and quoteName().
 	 *
@@ -343,12 +338,11 @@ abstract class JDatabaseQuery
 					$query .= (string) $this->order;
 				}
 
-				if ($this->union)
-				{
-					$query .= (string) $this->union;
-				}
-
 				break;
+
+			case 'union':
+				$query .= (string) $this->union;
+			break;
 
 			case 'delete':
 				$query .= (string) $this->delete;
@@ -356,7 +350,7 @@ abstract class JDatabaseQuery
 
 				if ($this->join)
 				{
-					// special case for joins
+					// Special case for joins
 					foreach ($this->join as $join)
 					{
 						$query .= (string) $join;
@@ -1230,6 +1224,7 @@ abstract class JDatabaseQuery
 	 * Usage:
 	 * $query->union('SELECT name FROM  #__foo')
 	 * $query->union('SELECT name FROM  #__foo','distinct')
+	 * $query->union(array('SELECT name FROM  #__foo','SELECT name FROM  #__bar'))
 	 *
 	 * @param   JDatabaseQuery  $query     The JDatabaseQuery object to union.
 	 * @param   boolean         $distinct  True to only return distinct rows from the union.
@@ -1239,28 +1234,42 @@ abstract class JDatabaseQuery
 	 *
 	 * @since   12.1
 	 */
-	public function union($query, $distinct = false, $glue = ' ')
+	public function union($query, $distinct = false, $glue = '')
 	{
 
-		// Clear any ORDER BY clause in unioned query.
+		// Clear any ORDER BY clause in UNION query
+		// See http://dev.mysql.com/doc/refman/5.0/en/union.html
 		if (!is_null($this->order))
 		{
 			$this->clear('order');
 		}
 
-		// Apply the distinct flag to the union if set.
+		// Set up the DISTINCT flag, the name with parentheses, and the glue.
 		if ($distinct)
 		{
-			$this->union[] = new JDatabaseQueryElement('UNION DISTINCT', $query, $glue);
-
-			return $this;
+			$name = 'UNION DISTINCT ()';
+			$glue = ')' . PHP_EOL . 'UNION DISTINCT (';
 		}
 		else
 		{
-			$this->union[] = new JDatabaseQueryElement('UNION', $query, $glue);
+			$glue = ')' . PHP_EOL . 'UNION (';
+			$name = 'UNION ()';
 
-			return $this;
 		}
+		// Great the JDatabaseQueryElement if it does not exist
+		if (is_null($this->union))
+		{
+
+				$this->union = new JDatabaseQueryElement($name, $query , "$glue");
+		}
+		// Otherwise append the second UNION.
+		else
+		{
+			$glue = '';
+			$this->union->append($query);
+		}
+
+		return $this;
 	}
 
 	/**
@@ -1276,7 +1285,7 @@ abstract class JDatabaseQuery
 	 *
 	 * @since   12.1
 	 */
-	public function unionDistinct($query, $glue = ' ')
+	public function unionDistinct($query, $glue = '')
 	{
 		$distinct = true;
 
--- a/libraries/joomla/database/query.php
+++ b/libraries/joomla/database/query.php
@@ -1265,4 +1265,24 @@ abstract class JDatabaseQuery
 		}
 		return $this;
 	}
+		/**
+	 * Add a query to UNION DISTINCT with the current query. Simply a proxy to Union.
+	 *
+	 * @param   object   The JDatabaseQuery object to union.
+	 *
+	 * @return  mixed    The JDatabaseQuery object on success or boolean false on failure.
+	 *
+	 * @since   12.1
+	 */
+	public function unionDistinct($query)
+	{
+		if (!$query instanceof JDatabaseQuery)
+		{
+			return false;
+		}
+
+		// Apply the distinct flag to the union if set.
+
+		return $this->union($query,'distinct');
+	}
 }
--- a/libraries/joomla/database/query.php
+++ b/libraries/joomla/database/query.php
@@ -345,12 +345,12 @@ abstract class JDatabaseQuery
 
 				if ($this->union)
 				{
-					$query .= (string) $this->union = array();
+					$query .= (string) $this->union;
 				}
 
 				if ($this->unionDistinct)
 				{
-					$query .= (string) $this->unionDistinct = true;
+					$query .= (string) $this->unionDistinct;
 				}
 
 				break;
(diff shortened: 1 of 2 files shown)


libraries/joomla/database/query.php
```diff
@@ -1200,4 +1232,57 @@ public function __clone()
...
+	public function unionDistinct($query)
+	{
+		if (!$query instanceof JDatabaseQuery)
+		{
+			return false;
+		}
+
+		// Apply the distinct flag to the union if set.
+
+		return $this->union($query, 'distinct');
```
mbabker: Since the $distinct param on union is documented to be boolean, wouldn't it be better to do `return $this->union($query, true);`?

libraries/joomla/database/query.php
```diff
@@ -1200,4 +1232,57 @@ public function __clone()
...
+		if (!$query instanceof JDatabaseQuery)
+		{
+			return false;
+		}
+
+		// Clear any ORDER BY clause in unioned query.
+		$query->clear('order');
+
+		// Apply the distinct flag to the union if set.
+		if ($distinct !== null)
```
mbabker: $distinct should be boolean (true/false), so why not:

`if ($distinct)
{
    return ' UNION DISTINCT' . $query;
}
else
{
    return ' UNION ' . $query;
}`

libraries/joomla/database/query.php
```diff
@@ -1200,4 +1232,57 @@ public function __clone()
...
+	 * Add a query to UNION with the current query.
+	 *
+	 * @param   object   $query     The JDatabaseQuery object to union.
+	 * @param   boolean  $distinct  True to only return distinct rows from the union.
+	 *
+	 * @return  mixed    The JDatabaseQuery object on success or boolean false on failure.
+	 *
+	 * @since   12.1
+	 */
+	public function union($query, $distinct = null)
```
mbabker: Why not default $distinct to false since it should be boolean?

elinw:
Good idea.

elinw:
I thought about it and I think actually using null and 'distinct' make more sense. Null + any value make a Boolean, it does not have to be true/false 1/0, those actually make it three possible conditions, null, true and false. But more importantly I think $query->('something','distinct') makes for more readable code than $query->('something','true') although the latter will work as well.  I'm not wedded to it if people feel strongly otherwise, however.

Should these be in a __toString test also?

mbabker:
I agree it makes for a better read, but the way it is documented, an IDE like PhpStorm will flag `$query->distinct('something', 'distinct')` as having an invalid param (expected Boolean, got string type message).  Not a big deal as you said, but the details make a difference.

I'd do a __toString test.  There's been a couple of issues I've had to fix that probably could have been avoided if it were tested.

realityking:
It think it should be just true/false too. If I'm reading the code right, setting false would actually mean true right now.

elinw:
Okay I think this will work for everyone :).

libraries/joomla/database/query.php
```diff
@@ -1200,4 +1232,61 @@ public function __clone()
...
+
+	/**
+	 * Add a query to UNION with the current query.
+	 * Multiple unions each require separate statements.
+	 *
+	 * Usage:
+	 * $query->union('SELECT name FROM  #__foo')
+	 * $query->union('SELECT name FROM  #__foo','distinct')
+	 *
+	 * @param   object   $query     The JDatabaseQuery object to union.
```
realityking: It would be helpful to some IDEs to put JDatabaseQuery instead of object here.

libraries/joomla/database/query.php
```diff
@@ -1200,4 +1232,61 @@ public function __clone()
...
+	 * @param   boolean  $distinct  True to only return distinct rows from the union.
+	 * @param   string   $glue      The glue by which to join the conditions.
+	 *
+	 * @return  mixed    The JDatabaseQuery object on success or boolean false on failure.
+	 *
+	 * @since   12.1
+	 */
+	public function union($query, $distinct = false, $glue = ' ')
```
realityking: I'm a big fan on type hinting so I'd put JDatabaseQuery in front of $query but that's just me.

libraries/joomla/database/query.php
```diff
@@ -1200,4 +1232,61 @@ public function __clone()
...
+		}
+	}
+
+	/**
+	 * Add a query to UNION DISTINCT with the current query. Simply a proxy to Union with the Distinct clause.
+	 *
+	 * Usage:
+	 * $query->unionDistinct('SELECT name FROM  #__foo')
+	 *
+	 * @param   object  $query  The JDatabaseQuery object to union.
```
realityking: It would be helpful to some IDEs to put JDatabaseQuery instead of object here.

libraries/joomla/database/query.php
```diff
@@ -1200,4 +1232,61 @@ public function __clone()
...
+	 * @param   string  $glue   The glue by which to join the conditions.
+	 *
+	 * @return  mixed   The JDatabaseQuery object on success or boolean false on failure.
+	 *
+	 * @since   12.1
+	 */
+	public function unionDistinct($query, $glue = ' ')
```
realityking: My comment above about type hinting applies here too.

elinw:
Our docblock style does not use that for @param, which if you think about it would really destroy the layouts. That's why we put the information in the description. We only put the class name as the type for the return.

realityking:
We do that in a number of places (for the benefit of some IDE), take the JHttp package as an example. But yes with such long classes it won't be the prettiest thing.

elinw:
I would only do that in the case  that there is a single @param tag. The goal is to have reading the source code directly easy and in an IDE as universal as possible. If we start catering to specific IDEs that is a long terrible road in which there will never be satisfaction since there will always be new ones and new releases of old ones we'd have to update thousands lof lines of doc blocks more or less quarterly.  Rather the goal should be to have the IDEs cope with Joomla correctly since that is what they are supposed to do.  If they aren't I would file a bug report there or even better work on Joomla specific plugins.

AmyStephen:
@realityking - agree, it's also an approach I see @LouisLandry taking in his newer work. 

https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/application/web.php#L27 

It ends up being very helpful as you are coding due to object property and method hints which are getting to be an industry norm in better IDEs.

eddieajau:
Yes, please replace "object" with the correct class name.  I don't know of any IDE's that will have a problem when you use the correct class name (and if they do, stop using it).  The current practice _is_ to add hinting to the params as well in the newer code as mentioned above.  @AmyStephen I assume phpStorm handles that ok?

AmyStephen:
Yes, as @mbabker indicated above, it works fine with phpStorm.

Thanks.

elinw:
Andrew are we planning to do that in all of the files?
Because that is a style decision that should not be done piece meal but one big run through the whole code base. Also that change should be documented in the style docs?

it's reading it in source that is the issue given that other params will now possibly be far far away from their types.

Also am glad to knwo that we are not asking people to make changes because of the stated reason of personal preferences of one individual as above.

eddieajau:
Yes, a note in the style manual is appropriate as is progressive updating of docblocks where applicable.
.

elinw:
Oh man where did that error come from.

elinw:
Should be good now.

libraries/joomla/database/query.php
```diff
@@ -1200,4 +1232,61 @@ public function __clone()
...
+		}
+
+		// Apply the distinct flag to the union if set.
+		if ($distinct)
+		{
+			return $this->union = new JDatabaseQueryElement('UNION DISTINCT', $query, $glue);
+		}
+		else
+		{
+			return $this->union = new JDatabaseQueryElement('UNION', $query, $glue);
```
eddieajau: Why are you not supporting chaining with the return values?

eddieajau:
Some general comments:
1. I don't think you need the unionDistinct variable (the alias method is fine).
2. I think the union method should return $this to support chaining.
3. I think the union method should add the element to a stack, e.g. `$this->union[] = new ...` to allow for multiple unions.

elinw:
/headslap, i was thinking about multiple unions initially, got sidetracked.
Should I also be adding Union to test__toStringSelect() ?

eddieajau:
Yes, would be worth adding it to the test.

elinw:
I'm going to change the test to use a data provider and not use the inspector, I think, but I believe the tests themselves are covering what they need to cover. I handled the multi union in a different way than Andrew suggested but I think that the expected results in the tests are what he was looking for and this is delivering them. Let me know if that isn't the case.

elinw:
So I talked to Ian about this a bit and I think a data provider doesn't give a lot of advantage. Also, I think i will work on updating away from the inspector for the whole class as a separate pull request.

libraries/joomla/database/query.php
```diff
@@ -1200,4 +1215,79 @@ public function __clone()
...
+	 *
+	 * @param   JDatabaseQuery  $query     The JDatabaseQuery object to union.
+	 * @param   boolean         $distinct  True to only return distinct rows from the union.
+	 * @param   string          $glue      The glue by which to join the conditions.
+	 *
+	 * @return  mixed    The JDatabaseQuery object on success or boolean false on failure.
+	 *
+	 * @since   12.1
+	 */
+	public function union($query, $distinct = false, $glue = '')
```
realityking: Could you add the type hint for $query? All new API uses it now.

libraries/joomla/database/query.php
```diff
@@ -1200,4 +1215,79 @@ public function __clone()
...
+			$name = 'UNION DISTINCT ()';
+			$glue = ')' . PHP_EOL . 'UNION DISTINCT (';
+		}
+		else
+		{
+			$glue = ')' . PHP_EOL . 'UNION (';
+			$name = 'UNION ()';
+
+		}
+		// Great the JDatabaseQueryElement if it does not exist
```
realityking: Nitpicking, you probably mean get not great.

libraries/joomla/database/query.php
```diff
@@ -1200,4 +1215,79 @@ public function __clone()
...
+	 * $query->unionDistinct('SELECT name FROM  #__foo')
+	 *
+	 * @param   JDatabaseQuery  $query  The object to union.
+	 * @param   string          $glue   The glue by which to join the conditions.
+	 *
+	 * @return  mixed   The JDatabaseQuery object on success or boolean false on failure.
+	 *
+	 * @since   12.1
+	 */
+	public function unionDistinct($query, $glue = '')
```
realityking: Please add the type hint here too.

libraries/joomla/database/query.php
```diff
@@ -1304,4 +1319,79 @@ public function __clone()
...
+		else
+		{
+			$glue = ')' . PHP_EOL . 'UNION (';
+			$name = 'UNION ()';
+
+		}
+		// Get the JDatabaseQueryElement if it does not exist
+		if (is_null($this->union))
+		{
+				$this->union = new JDatabaseQueryElement($name, $query, "$glue");
```
pasamio, line 1365: Stupid question, why is $glue double quoted instead of just being directly used?

PR #756, opened 2012-01-17
Line 293 : syntax issue in multidb

sseshachala:
$case_when .= $query->charLength('c.alias') . '!=0'; modified to    $case_when .= $query->charLength('c.alias');

Not sure if this was done for a reason.

### DIFF
--- a/libraries/joomla/application/categories.php
+++ b/libraries/joomla/application/categories.php
@@ -290,7 +290,7 @@ class JCategories
 		// Right join with c for category
 		$query->select('c.*');
 		$case_when = ' CASE WHEN ';
-		$case_when .= $query->charLength('c.alias') . '!=0';
+		$case_when .= $query->charLength('c.alias');
 		$case_when .= ' THEN ';
 		$c_id = $query->castAsChar('c.id');
 		$case_when .= $query->concatenate(array($c_id, 'c.alias'), ':');


gpongelli:
Sudhi, 
I've added that comparison everywhere because PostgreSQL's CASE WHEN clause has to compare boolean expression, as you can see [here](http://www.postgresql.org/docs/9.1/static/plpgsql-control-structures.html) .
Maybe we have to thought together about db-independent CASE WHEN clause, maybe in JDatabaseQuery, maybe not.
Tell me your thoughts.

Eng. Gabriele Pongelli.

sseshachala:
this is breaking in SQLServer and sure will break in Oracle.
May be it is a good idea to introduce "case when" in JDatabaseQuery and submit a pull request.

Thanks
sudhi

gpongelli:
Can you tell me which is SQLServer's correct syntax of CASE WHEN, in a similar manner of PostgreSQL one ?
I don't know if a JDatabaseQuery element will solve this difference until is known syntax differences.

Eng. Gabriele Pongelli

gpongelli:
For Oracle database, I've found [official documentation about CASE WHEN](http://docs.oracle.com/cd/B28359_01/appdev.111/b28370/case_statement.htm#sthref1650), as you can see there and inside example links, there are two accepted syntax for Oracle:

<pre><code>CASE case_operand
    WHEN when_operand THEN statement;
    WHEN when_operand THEN statement;
    ...
    ELSE statement;
END CASE;</code></pre>

and

<pre><code>CASE
    WHEN boolean_expression THEN statement
    WHEN boolean_expression THEN statement
    ...
    ELSE statement;
END CASE;</code></pre>


These examples are very similar to PostgreSQL's one, so what is SQLServer's correct syntax for CASE WHEN ?

sseshachala:
very similar to Oracle
http://msdn.microsoft.com/en-us/library/ms181765.aspx

Please note - the SQLServer support is already part of the core. So we cannot break that code to support Postgres SQL. I would humbly request to revert back this change and revisit the case when in JDatabaseQuery appropriately  after careful thinking all possibilities.

Thanks
Sudhi

infograf768:
This patch does not break mysqli on CMS

gpongelli:
It seems like Oracle and PostgreSQL , just a question on how my changes can break SQLServer.
Inside your link I've found "Searched CASE Expression" , where WHEN evaluates a boolean expression so in your code where is that boolean expression?

I added '!=0' to have that requested boolean expression, to compare a char length (it is an integer) to another integer value.

Edit:
I think your code has a conceptual error, charLength has to return integer value, not boolean one that you return with adding "IS NOT NULL" statement, as you can see here
https://github.com/gpongelli/joomla-platform/blob/master/libraries/joomla/database/database/sqlsrvquery.php#L119

Try your cms without "IS NOT NULL" to see if it works with '!=0' or not, because I think that PostgreSQL doesn't work with "IS NOT NULL" .

Eng. Gabriele Pongelli

gpongelli:
Just a correction: charLength returns a string that is used to evaluate the string's character length, not an integer, nor a string containing null check.

Eng. Gabriele Pongelli

infograf768:
To Platform Committers
This is apparently is quite urgent.

sseshachala:
IMHO - we need to get the code into platform since it is breaking the SQLServer support in CMS which is already part of core. SQLServer is part of supported database and Postgress is not as of today. May be we can address the issue in a better way at the framework level rather than include here to break support for SQLServer.

infograf768:
We can' wait anymore for the RC1 release.
Committed in CMS

sseshachala:
This needs to be fixed in platform too. We will address additional concerns raised in CMS 3.0.

eddieajau:
So, do we have agreement that the current patch is ok for all databases?  @infograf768 there is no urgency to commit.  The next release of the platform is 12.1 at the end of March.

gpongelli:
Is ok for databases already present in platform and cms repository.
About PostgreSQL this patch is not ok.

I've talked with sudhi and he will change sqlsrv's charLength behavior (removing 'IS NOT NULL' from inside charLength return string) for 3.0 .
Without that 'IS NOT NULL' and don't applying this patch it's reasonable to think that cms works for sqlsrv and PostgreSQL too.

Eng. Gabriele Pongelli

eddieajau:
Ok, so as far as the platform is concerned (CMS 3.0 will probably run on the 12.x series of the platform anyway), we should reject this patch?

infograf768:
@eddieajau 
The point Sudhi is making is that we are more and more introducing differences between platform and CMS.
It would be better IMHO to commit to CMS another patch presented here that works really for all db and accepted, this to limit discrepancies.

gpongelli:
@Andrew
You don't have to reject this patch for 2.5, but this code will be changed again for 3.0 , when will be added PostgreSQL (rigth?).
If you think to add PostgreSQL for 2.5 (as beta too) you've to reject this patch and fixing sqlsrv's charLength as told before . 
But this means that it's added some solution before on cms and then apply to next platform release.

Eng. Gabriele Pongelli

sseshachala:
@andrew : For now - please commit the patch. For CMS 3.0 - we will need to fix char_length method for both SQLSrv and mysql as well.
That way for 2.5 - we can keep cms-platform in sync with respect to db changes and do understand - platform will never be same as CMS.

Thanks
sudhi

gpongelli:
You've to fix charLength only for sqlsrv, I test my changes with PostgreSQL database and with mysql database and the last still works.
That '!=0' doesn't break mysql queries.

Is maybe possible to change platform code before in cms (as this case for sqlsrv's charLength) and then report this change inside platform? 

Eng. Gabriele Pongelli

sseshachala:
I think - am not explaining this right.
SQLServer checks "is not null" and !=0 wont work on sqlserver.
If I fix in sqlserver - everywhere we have case when - we will need to check the db and apply the condition appropriately.
this is more realistic
1. Fix in the all db to include the '!= 0' in the db - so the code where case when is used does not need to check
Blame it on vendors not supporting ANSI SQL. Char_Length is implemented in specific ways for all databases.

eddieajau:
Guys, this is the Platform, not the CMS.  It's got nothing to do with 2.5's or 3.0's.  The CMS is just one downstream user.  We are currently in the 12.x version and if that is the one that the CMS would be using anyway.  If you are "fixing" it anyway, I don't need to apply this patch - that is my understanding.

realityking:
What I'm not understanding what about this code is breaking the CMS now but not in the future?

elkuku:
**Suggestion:**
As char_length can be used in both "select" and "case" statements, and the syntax to use may be vendor specific, what if we implement a new function that deals with this "anomaly" ?
something like:

``` php
<?php
JDatabaseQueryPostgreSQL

public function whenCharLength($field, $operator = '', $condition = '')
{
    if('' == $operator)
    {
        // Postgres special for boolean true comparison
        $condition = ' != 0';
    }

    return 'CHAR_LENGTH(' . $field . ')'.$operator.$condition;
}
```

The changes in CMS would be:
e.g. components/com_contact/models/contact.php

```
-  $case_when .= $query->charLength('a.alias');
+  $case_when .= $query->whenCharLength('a.alias');
```

This might allow more flexibility.

gpongelli:
@elkuku 
it is reasonable that each driver implements this function instead of charLength.

elkuku:
Well, it has to be implemented in the extending classes only when necessary (e.g. Postgres). The base class function could be:

``` php
<?php
JDatabaseQuery

public function whenCharLength($field, $operator = '', $condition = '')
{
    return 'CHAR_LENGTH(' . $field . ')'.$operator.$condition;
}
```

If this is OK for the extending classes (e.g. MySQL and SQLServ) there is no need to implement it.
Also it is meant as an additional function, not as a replacement for charLength().

gpongelli:
No, it's not what I've thought, I'll try to explain myself better.

This is actual charLength code for each driver:

MySQL & MySQLi & SQLAzure & PostgreSQL use [JDatabaseQuery::charLength](https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/database/query.php#L450) inside query.php, inherited by JDatabaseQueryMySQL, JDatabaseQueryMySQLi, JDatabaseQuerySQLAzure and JDatabaseQueryPostgreSQL .
Called code in these cases is

<pre><code>public function charLength($field)
{
    return 'CHAR_LENGTH(' . $field . ')';
}</code></pre>

<br>
SQLSrv uses [JDatabaseQuerySQLSrv::charLength](https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/database/database/sqlsrvquery.php#L125) .
Called code is

<pre><code>public function charLength($field)
{
    return 'DATALENGTH(' . $field . ') IS NOT NULL';
}</code></pre>

<br>
Can you see where is the difference ?
I've added no code for my driver and SQLSrv adds 'IS NOT NULL' instead , so this what means ?
SQLSrv returns a string that, when it will be executed by database engine, it returns a <b>bool value</b> because of not null comparison, when other drivers' call will be executed they return an integer that is <b>string's length</b>.
Again, can you see where is the difference ?

Just to be sure I've tried to execute this query in PostgreSQL
<code>SELECT CHAR_LENGTH(name) IS NOT NULL FROM #__assets</code>
and it works as expected.
The problem is NOT about 'IS NOT NULL' words per se, but use it in an extern context than other drivers' call one.
If Sudhi wrote charLength <b>without 'IS NOT NULL'</b> but adding this string every time query is built, we hadn't this issue now because this is focal point of what we're talking about.

<br>
Now, your proposal of adding two extra field is really good, but I wish it will be JDatabaseQuery::charLength next definition

<pre><code>public function charLength($field, $operator = '', $condition = '' )
{
    return 'CHAR_LENGTH(' . $field . ') ' . $operator . ' ' . $condition;
}</code></pre>


While SQLSrv definition will be

<pre><code>public function charLength($field, $operator = '', $condition = '' )
{
    return 'DATALENGTH(' . $field . ') ' . $operator . ' ' . $condition;
}</code></pre>

so everyone can call 
<code>$query->charLength(aColumn)</code>
and use returning string as numerical value, while calling
<code>$query->charLength(aColumn, 'IS', 'NOT NULL')</code>
getting 
<code>CHAR_LENGTH(aColumn) IS NOT NULL</code>
or
<code>DATALENGTH(aColumn) IS NOT NULL</code>
to use a bool value.

Have you got my point of view?

Sorry for this long reply, 
Eng. Gabriele Pongelli.

eddieajau:
Can we put this topic on the mailing list and I'll close this pull until we have a clear direction on how to solve this issue.

PR #757, opened 2012-01-17
Remove deprecated methods in JLanguage

mbabker:
Per roadmap, removed deprecated methods from JLanguage.


realityking:
Looks fine to me.

PR #758, opened 2012-01-17
Added JDatabaseQuery function to extract part of timestamp.

gpongelli:
This function is used to extract part of date from timestamp in a database independent manner.
Added MySQL test of element string conversion.

Usage:
$query->day($query->quoteName('dateColumn'))
for MySQL will return "DAY( dateColumn )"

$query->month($query->quoteName('dateColumn'))
for MySQL will return "MONTH( dateColumn )"

$query->year($query->quoteName('dateColumn'))
for MySQL will return "YEAR( dateColumn )"

$query->hour($query->quoteName('dateColumn'))
for MySQL will return "HOUR( dateColumn )"

$query->minute($query->quoteName('dateColumn'))
for MySQL will return "MINUTE( dateColumn )"

$query->second($query->quoteName('dateColumn'))
for MySQL will return "SECOND( dateColumn )"

### DIFF
--- a/libraries/joomla/database/query.php
+++ b/libraries/joomla/database/query.php
@@ -724,6 +724,27 @@ abstract class JDatabaseQuery
 	}
 
 	/**
+ 	 * Used to get a string from date column for part (day, month, year) extraction.
+	 *
+	 * Usage:
+	 * $query->select($query->getPartOfDate($query->quoteName('dateColumn'), 'day'));
+	 * $query->select($query->getPartOfDate($query->quoteName('dateColumn'), 'month'));
+	 * $query->select($query->getPartOfDate($query->quoteName('dateColumn'), 'year'));
+	 *
+	 * @param   string  $date  Date column containing part to be extracted.
+	 * @param   string  $part  String of what part of date will be extracted.
+	 *
+	 * @return  string  Returns required string to extract part of date.
+	 *
+	 * @since   12.1
+	 */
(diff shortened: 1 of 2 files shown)


chdemko:
Would it be simpler to add these functions (and thus restrict the 'part' value)

```
public function year($date)
public function month($date)
public function day($date)
public function hour($date)
public function minute($date)
public function second($date)
public function microsecond($date)
```

?

Allowing to precise the 'part', developers could use some mysql functions that could not exist in other db engines

gpongelli:
Yes it could be simpler.
I've added only that function because of same word use in each database's documentation, but this proposal is interesting.

Ing. Gabriele Pongelli

eddieajau:
Are there differences in the way different databases handle the date parts?

gpongelli:
Yes, an example extracting year from a date between MySQL and PostgreSQL .

Mysql: YEAR(dateColumn)
PostgreSQL: EXTRACT(YEAR FROM dateColumn)

Eng. Gabriele Pongelli

eddieajau:
Ok, in that case I agree with Christophe.  I think it would be better to add the extra methods.

gpongelli:
I did the changes locally but I haven't pushed them yet, I'll do soon.

Eng. Gabriele Pongelli.

gpongelli:
Pushed last commits.

Eng. Gabriele Pongelli.

eddieajau:
Nice job. Merged.

PR #759, opened 2012-01-17
Remove deprecated code from JRegistry

mbabker:
Per the roadmap, this removes deprecated code in JRegistry.  A couple of additional notes as well:
- Removed commented out methods in JRegistryTest as they had been implemented already elsewhere in the file
- Made a change in JUser::getParameters that I'm not 100% on that calls `$this->_params->loadSetupFile()`, but based on reviewing the old JParameter::loadSetupFile method, I'm fairly certain that the actual intended behavior of this call has been restored (load the file instead of set a boolean true as was happening with JRegistry::loadSetupFile)


tests/suite/joomla/registry/JRegistryTest.php
```diff
@@ -84,17 +84,6 @@ public function testDef()
 	}
 
 	/**
-	 * @todo Implement testGet().
-	 */
-	/*public function testGet()
```
chdemko: The get method is not deprecated

tests/suite/joomla/registry/JRegistryTest.php
```diff
@@ -475,18 +403,6 @@ public function testMerge()
...
-		$this->markTestIncomplete(
-		'This test has not been implemented yet.'
-		);
-	}*/
-
-	/**
-	 * Test the JRegistry::set method.
-	 * @deprecated	1.6
-	 */
 	public function testSet()
```
chdemko, line 406: The set method is not deprecated

mbabker, line 406: This is a commented out test method that is implemented in the method immediately below these deleted lines.
- Michael

Please excuse any errors, this message was sent from my iPhone.

tests/suite/joomla/registry/JRegistryTest.php
mbabker: This too was a duplicate test method that was commented out.  Implemented test is at https://github.com/mbabker/joomla-platform/blob/8204612487005d2a6dfd57fcc65afc2513b0861b/tests/suite/joomla/registry/JRegistryTest.php#L167

eddieajau:
Good stuff!

PR #760, opened 2012-01-17
Modify JDatabaseSQLSrv::getVersion() to use sqlsrv_server_info

mbabker:
Per the PHP docs, the sqlsrv_server_info appears to be the equivalent PHP function to mysql_get_server_info, so I've modified the getVersion method to use this to get the appropriate information.

@sseshachala, any objections from you?


sseshachala:
This is good to have. No objections.

PR #761, opened 2012-01-20
Joomla CMS [#27780] Fixed bug causing duplicate template styles in form field

hieblmedia:
Duplicate template styles if other type of extension exists with the same name.

I have added a filter with type=template for the query.

See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27780

### DIFF
--- a/libraries/joomla/form/fields/templatestyle.php
+++ b/libraries/joomla/form/fields/templatestyle.php
@@ -72,7 +72,7 @@ class JFormFieldTemplateStyle extends JFormFieldGroupedList
 		}
 		$query->join('LEFT', '#__extensions as e on e.element=s.template');
 		$query->where('e.enabled=1');
-		$query->where('e.type=' . $db->quote('template'));
+		$query->where($db->quoteName('e.type') . '=' . $db->quote('template'));
 
 		// Set the query and load the styles.
 		$db->setQuery($query);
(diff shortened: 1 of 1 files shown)


hieblmedia:
Also affected is the styles view in the component, see:
https://github.com/joomla/joomla-cms/pull/93

libraries/joomla/form/fields/templatestyle.php
```diff
@@ -72,6 +72,7 @@ protected function getGroups()
 		}
 		$query->join('LEFT', '#__extensions as e on e.element=s.template');
 		$query->where('e.enabled=1');
+		$query->where('e.type=' . $db->quote('template'));
```
chdemko: could you use

```
$query->where($db->quoteName('e.type') . '='  . $db->quote('template'));
```

because 'type' could be a reserved word in some db engine

infograf768:
Test negative. See Tracker.

eddieajau:
The fix looks good.  Since this is a CMS only field, would you also mind moving it from /libraries/joomla/ to /libraries/cms/.  Thanks in advance.

hieblmedia:
I've added the file in the cms library.
See: https://github.com/joomla/joomla-cms/pull/93#commits-pushed-122c84d

I think on the platform this file should be deleted only in futher versions, for lecacy?

realityking:
The currently plan is not to actually move these files until Joomla 3.0. We just move classes we consider cms to the libraries/cms folder inside the platform repro.

hieblmedia:
Ok.

But now i can't commit in this pull request. I have created a new fork for a other pull request. 
I am very new on github... It is required and recommended to create a new branch in my fork for each pull request?

hieblmedia:
I have created a new Pull Request: https://github.com/joomla/joomla-platform/pull/814

PR #762, opened 2012-01-20
JLanguage: Avoid BOM error when debug, as BOM is OK when using parse_ini

infograf768:
JLanguage: Avoid BOM error when debug, as BOM is OK when using parse_ini

### DIFF
--- a/libraries/joomla/language/language.php
+++ b/libraries/joomla/language/language.php
@@ -871,6 +871,11 @@ class JLanguage extends JObject
 			while (!$stream->eof())
 			{
 				$line = $stream->gets();
+				// Avoid BOM error as BOM is OK when using parse_ini
+				if($lineNumber == 0)
+				{
+					$line = str_replace("\xEF\xBB\xBF",'',$line);
+				}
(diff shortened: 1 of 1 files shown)


infograf768:
Tracker:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27783

libraries/joomla/language/language.php
```diff
@@ -871,6 +871,11 @@ protected function parse($filename)
 			while (!$stream->eof())
 			{
 				$line = $stream->gets();
+				// Avoid BOM error as BOM is OK when using parse_ini
+				if($lineNumber == 0)
```
chdemko: add a blank between if and (

libraries/joomla/language/language.php
```diff
@@ -871,6 +871,11 @@ protected function parse($filename)
...
+				{
+					$line = str_replace("\xEF\xBB\xBF",'',$line);
```
chdemko: add a blank after each comma

PR #769, opened 2012-01-20
Deprecate JView::escape().

realityking:
JView::escape() should go the way of the Dodo.

There is little gained by being able to switch between htmlspecialchars() and htmlentities() (I can't think of a situation where I'd use it). It's also a bit inconsistent since modules don't have anything similar.

Last but not least, call_user_func() is comparably slow so there should be a little speed gain when all layouts change.

### DIFF
--- a/libraries/joomla/application/component/view.php
+++ b/libraries/joomla/application/component/view.php
@@ -174,14 +174,7 @@ class JView extends JObject
 	 * Callback for escaping.
 	 *
 	 * @var string
-	 */
-	protected $escape = 'htmlspecialchars';
-
-	/**
-	 * Callback for escaping.
-	 *
-	 * @var string
-	 * @deprecated use $escape declare as private
+	 * @deprecated 13.3 Will be removed
 	 */
 	protected $_escape = 'htmlspecialchars';
 
@@ -435,6 +428,7 @@ class JView extends JObject
 	 * @return  mixed  The escaped value.
 	 *
 	 * @since   11.1
+	 * @deprecated 13.3 Use htmlspecialchars()
 	 */
 	public function escape($var)
 	{
@@ -640,6 +634,7 @@ class JView extends JObject
 	 * @return  void
 	 *
 	 * @since   11.1
+	 * @deprecated 13.3 This will be removed without a replacement
 	 */
 	public function setEscape($spec)
 	{


eddieajau:
-1 don't support this one at this time.

AmyStephen:
I'd probably go ahead and deprecate this and either add a method to JFactory that can be access by Templates and Modules, along with Components, or just use the PHP methods. This method being available for only components probably isn't a good thing.

eddieajau:
There is no reason why modules can't use jview. In fact, I've started doing it in custom work.

AmyStephen:
@AndrewEddie - 

Couple of questions, then. What benefit is there to JView::escape over the two native PHP methods htmlspecialchars() and htmlentities()? (Is it for defining unicode-friendly escaping?) 

And, if JView's escape method can be used and there is benefit beyond the native PHP methods, should the CMS do so? A problem I see is, with the exception of the module class suffix which is escaped using htmlspecialchars, there is no escaping in any of the Joomla core modules. This is true for 1.5 and 1.6. Since modeling core code is a recommended learning approach, this practice then is followed with many community extensions. 

Bottom line, it would be good to settle on one way of escaping output and making certain rendered output is always escaped in that manner.

eddieajau:
It's just good encapsulation practice and it's easier to train a layout designer, not necessarily familiar with the importance of all the arguments in htmlspecialchars, nor familiar with the difference between that and htmlentities, to use $this->escape($foo). Bottom line is it allows the view to upgrade the escaping technology without changing layouts in whatever application is using the platform. It's immaterial whether cms modules do or don't use this, I just think it's a valid part of the view and one you see in many frameworks and textbooks.  Aside from that, if you want to use htmlspecialchars with three arguments go for it, it's opt in, and, there are far more important things to be debating about the mvc paradigm than this little method.

AmyStephen:
Module output is not escaped at all.

eddieajau:
That may be, but it's something to take up with the cms.

realityking:
TBH I didn't expect to start this kind of debate. Andrew, is your issue with the entire patch or would you be OK with JView::escape() just calling htmlspecialchars() with the appropriate parameters?

AmyStephen:
Done. https://github.com/joomla/joomla-cms/issues/95

Thanks @eddieajau and @realityking

eddieajau:
@realityking my only issue is that I think it's premature to do this given we are going to completely review the mvc anyway.

eddieajau:
Actually, to be more specific, yes I agree with killing the setEscape method and just have escape as an alias for htmlspecialchars.

eddieajau:
Do you mind if we park this for now?

realityking:
Sorry I didn't respond earlier, yes I'm fine with tabling this.

PR #771, opened 2012-01-20
Remove deprecated code in Installer package

mbabker:
- Removes getOverwrite and getUpgrade methods
- No longer allows install tag as valid XML root tag in extension installation


eddieajau:
Thanks.

PR #772, opened 2012-01-20
Remove deprecated code from JDate

mbabker:
- Deprecated code removed
- Test case now extends JoomlaDatabaseTestCase to allow for testing in isolation


eddieajau:
Thanks.

PR #773, opened 2012-01-20
Deprecate JUser::getParameters()

mbabker:
Based on Hannes' response at http://groups.google.com/group/joomla-dev-platform/t/45fe515e53940db2, this deprecates the JUser::getParameters() method and removes the `$usertype` property from JUser as this was the only place this property was still in use.


eddieajau:
Thanks

PR #777, opened 2012-01-21
JInstaller improvements/adjustments

hieblmedia:
1. The package adapter executes the install method twice. This can insert dublicate database entries.
   See: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27802
2. On some install adapters a check is_a($updateElement, 'JXMLElement') return always true and force an update. This is not desirable.
3. The template install adapter never return an existing extension id (setQuery was not set). This forces and update also and create dublicate database entries in the extensions table (also template styles). Updated query with exception and direct rollback.

### DIFF
--- a/libraries/joomla/installer/adapters/language.php
+++ b/libraries/joomla/installer/adapters/language.php
@@ -183,7 +183,7 @@ class JInstallerLanguage extends JAdapterInstance
 			// Update function available or
 			// Update tag detected
 			if ($this->parent->isUpgrade() || ($this->parent->manifestClass && method_exists($this->parent->manifestClass, 'update'))
-				|| is_a($updateElement, 'JXMLElement'))
+				|| $updateElement)
 			{
 				return $this->update(); // transfer control to the update function
 			}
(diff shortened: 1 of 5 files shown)


hieblmedia:
The template installer issue is very important to fix. (commit: f2eefd8)

Becuase, 
if a template is already installed the, database entry of a (random?) core extension would be overwritten. This is very very bad.

hieblmedia:
Hm i have this exception handling from the plugin adapter.
You mean to remove the new 'try...catch' part?

realityking:
Not remove it, supplement it with the legacy error handling. I just had a quick glance at the plug-in adapter, I think we gotta fix that one too.

hieblmedia:
Can you explain me an example? I don't know what is to do for error legacy support.

realityking:
Here's an example of using both methods: https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/database/table.php#L528 (up to line 546)

chdemko:
@hieblmedia There remains 3 errors for the code style

hieblmedia:
Code style errors fixed

PR #786, opened 2012-01-23
Am I missing something... 

Dj83:
Hi can someone please brief me and explain why JRequest is being used all over the Platform (+CMS) still?

Do all the files in the Platform need updated? (those using JRequest?)

Am I on the wrong repo? I am looking for 12.X

Thanks

### DIFF
--- a/libraries/joomla/application/application.php
+++ b/libraries/joomla/application/application.php
@@ -278,7 +278,10 @@ class JApplication extends JObject
 		$router = $this->getRouter();
 		$result = $router->parse($uri);
 
-		JRequest::set($result, 'get', false);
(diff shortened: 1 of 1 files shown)


mbabker:
It's still in use because the work hasn't been done yet to replace all uses of JRequest with their replacement (JInput for most, still gotta figure out checkToken).

Also, JInput shouldn't be directly accessed as JRequest has been in the past.  You should get it from the application (`JFactory::getApplication()->input`, but in this case since we're in JApplication, `$this->input` should work).  So, change the `JInput::set($key, $value);` line in your diff to be `$this->input->set($key, $value);`

PR #787, opened 2012-01-23
Allow JFormFieldCalendar to accept a JDate object as value.

realityking:
First fall out from toying around with the UCM.

We used to just take the date as a string and send it to the form field which converts it into a JDate object if needed. JContent however already gives you a JDate object so the form field should just be able use it.

### DIFF
--- a/libraries/joomla/form/fields/calendar.php
+++ b/libraries/joomla/form/fields/calendar.php
@@ -79,34 +79,35 @@ class JFormFieldCalendar extends JFormField
 		$config = JFactory::getConfig();
 		$user = JFactory::getUser();
 
-		// If a known filter is given use it.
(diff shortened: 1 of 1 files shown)


chdemko:
@realityking  there are unit tests failure

realityking:
I know, I haven't had time to figure out why they fail.

chdemko:
@realityking See https://gist.github.com/1746576

realityking:
Thanks for looking into this. Could you explain why my approach isn't working. As far as I can tell JDate should be able to handle the timezone change.

chdemko:
I can't remember why. I've also added some unit test for the JDate case

chdemko:
@realityking do you want to continue on this issue?

realityking:
I'm closing this for now, might pick it up later again.

PR #788, opened 2012-01-23
More Work on JDaemon

robschley:
Added an option to keep a JDaemon application in the foreground.
Fixed some issues with trying to handle undefined signals in JDaemon.
Fixed an issue with forked JDaemon applications trying to manage the PID file.


elkuku:
It seems you are not running the code sniffer ;) May I humbly suggest to set "Remove trailing white space on save" in your IDE ? http://elkuku.github.com/pulltester/pulls/788

robschley:
Should be fixed now. It wasn't trailing whitespace, it was line endings because I copied for a gist I think.

elkuku:
Interesting - first time the sniffer lies - apologies ;)

PR #791, opened 2012-01-24
Deprecating JRequest in favour of JInput usage

Dj83:
Beginning the process of switching from JRequest to JInput in the Platform. If all files are not ready to use JInput please someone throw me a heads up... otherwise I will continue to change them as I see them.

Thanks all

### DIFF
--- a/libraries/joomla/application/application.php
+++ b/libraries/joomla/application/application.php
@@ -278,7 +278,9 @@ class JApplication extends JObject
 		$router = $this->getRouter();
 		$result = $router->parse($uri);
 
-		JRequest::set($result, 'get', false);
+		foreach($result as $key=>$value){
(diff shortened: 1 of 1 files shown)


libraries/joomla/application/application.php
```diff
@@ -278,7 +278,9 @@ public function route()
 		$router = $this->getRouter();
 		$result = $router->parse($uri);
 
-		JRequest::set($result, 'get', false);
+		foreach($result as $key=>$value){
```
robschley, line 281: To comply with the style rules, this should be rewritten to be:

foreach ($result as $key => $value)
{
...

realityking:
@Dj83 If you fix the code style issues (http://elkuku.github.com/pulltester/pulls/791) that were found we can get this merged.

PR #792, opened 2012-01-24
Fix bug in countMenuChildren

katalystsol:
$query was generating an undefined variable error notice and then a fatal error notice because it is not defined previously in that method. After changing that on my site, which I had just upgraded to 2.5.0, it started working again.

### DIFF
--- a/libraries/joomla/document/html/html.php
+++ b/libraries/joomla/document/html/html.php
@@ -493,7 +493,7 @@ class JDocumentHTML extends JDocument
 			$active = $menu->getActive();
 			if ($active)
 			{
-				$query = dbo->getQuery(true);
+				$query = $dbo->getQuery(true);
 				$query->select('COUNT(*)');
 				$query->from('#__menu');
 				$query->where('parent = ' . $active->id);
--- a/libraries/joomla/document/html/html.php
+++ b/libraries/joomla/document/html/html.php
@@ -493,11 +493,11 @@ class JDocumentHTML extends JDocument
 			$active = $menu->getActive();
 			if ($active)
 			{
-				$query->getQuery(true);
-				$query->select('COUNT(*)');
-				$query->from('#__menu');
-				$query->where('parent = ' . $active->id);
-				$query->where('published = 1');
+				$dbo->getQuery(true);
+				$dbo->select('COUNT(*)');
+				$dbo->from('#__menu');
+				$dbo->where('parent = ' . $active->id);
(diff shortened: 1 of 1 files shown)


libraries/joomla/document/html/html.php
```diff
@@ -493,11 +493,11 @@ public function countMenuChildren()
 			$active = $menu->getActive();
 			if ($active)
 			{
-				$query->getQuery(true);
```
mbabker, line 496: This line should read `$query = $dbo->getQuery(true);`.  The next 4 lines shouldn't need to be changed afterwards.

katalystsol, line 496: Thanks. I just tested that on my site and it works with only changing the one line. Do I need to change this pull request then?

mbabker, line 496: Yes.  Just update the file again and push the change and the pull request
will automatically update.

On 1/24/12 6:37 PM, "Don Cranford"
<reply+p-380497-462b1bc4a0c192e2c9834bb809e9c307f2312f33-368545@reply.githu
b.com> wrote:

katalystsol, line 496: I updated the file. Thanks for your help.

katalystsol:
@joomla-jenkins ... I'm not sure what you're referring to. If I did something wrong, let me know. I'd like to learn how to work with this... The docs.joomla didn't say anything about a test log.

mbabker:
That is an automated test account to run the unit tests and code style rules.  The unit tests most likely failed because of the first change you made.  It looks like the tests will still fail as you've forgotten the $ on the `$query = $dbo->getQuery(true);` line.

katalystsol:
Ok. Thanks. I've corrected the line. I'm guessing the 165 warnings are from the rest of the file?

mbabker:
The full repository.  All warnings that are being worked on.

On 1/24/12 8:55 PM, "Don Cranford"
<reply+i-2953661-0b060b7a03a0026bdc57d3cbe398a0ef3e45461f-368545@reply.gith
ub.com> wrote:

eddieajau:
Good catch. Merged.

imarklee:
Hey, this patch doesn't seem to be included in 2.5.6 .

realityking:
This is fixed in Joomla 3.0 alpha 1. If you need the fix in an older version please file a bug on the CMS tracker and link to this pull request.

PR #794, opened 2012-01-25
JInput implements Serializable

robschley:
Updated JInput to implement Serializable so that you can serialize an input and it will copy the current request environment into the object before serializing. This allows you to unserialize the object and retrieve the original request environment.


LouisLandry:
This is kick-ass. :+1:

PR #797, opened 2012-01-26
Deprecate JController::authorise

mbabker:
Without the other code from JController that has been removed as deprecated, JController::authorise() really has no function any longer.  I've deprecated it and removed its only use in JController::execute().

### DIFF
--- a/libraries/joomla/application/component/controller.php
+++ b/libraries/joomla/application/component/controller.php
@@ -695,8 +695,7 @@ class JController extends JObject
 		// Record the actual task being fired
 		$this->doTask = $doTask;
 
-		$retval = $this->$doTask();
-		return $retval;
+		return $this->$doTask();
 	}
 
 	/**
--- a/libraries/joomla/application/component/controller.php
+++ b/libraries/joomla/application/component/controller.php
@@ -469,28 +469,11 @@ class JController extends JObject
 	 * @return  boolean  True if authorised
 	 *
 	 * @since   11.1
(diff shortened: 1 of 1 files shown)


libraries/joomla/application/component/controller.php
```diff
@@ -712,16 +695,8 @@ public function execute($task)
...
-		{
-			$retval = $this->$doTask();
-			return $retval;
-		}
-		else
-		{
-			return JError::raiseError(403, JText::_('JLIB_APPLICATION_ERROR_ACCESS_FORBIDDEN'));
-		}
+		$retval = $this->$doTask();
+		return $retval;
```
jonnsl: return $this->$doTask();
???

mbabker: I left the conditional code as is, but it honestly looks funny anyways as `$this->$doTask();`... why the $ in front of doTask?

jonnsl: http://php.net/manual/en/functions.variable-functions.php
I was trying to say that you don't need to store the value returned by the function in a variable just to return that value on the next line, you can do in just one line

``` php
return $this->$doTask();
```

mbabker: Went ahead and changed it.

LouisLandry:
Thanks a bunch for this work you are doing Michael.  It's very much appreciated -- and we'll get through the other requests as soon as we can.

PR #798, opened 2012-01-26
Improved charLength function for multidb environment.

gpongelli:
Added two extra parameter used to compare charLength integer returning value.
Second parameter is comparison's operator, third parameter is integer value to compare charLength with.

### DIFF
--- a/libraries/joomla/database/database/sqlsrvquery.php
+++ b/libraries/joomla/database/database/sqlsrvquery.php
@@ -126,7 +126,7 @@ class JDatabaseQuerySQLSrv extends JDatabaseQuery
 	 */
 	public function charLength($field, $operator = null, $condition = null)
 	{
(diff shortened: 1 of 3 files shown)


libraries/joomla/database/database/sqlsrvquery.php
```diff
@@ -116,15 +116,17 @@ public function castAsChar($value)
...
 	 *
 	 * @return  string  The required char length call.
 	 *
 	 * @since 11.1
 	 */
-	public function charLength($field)
+	public function charLength($field, $operator = null, $condition = null)
 	{
-		return 'DATALENGTH(' . $field . ') IS NOT NULL';
+		return 'DATALENGTH(' . $field . ')' . (isset($operator) ? ' ' . $operator . ' ' . $condition : '');
```
oc666: You should check also if $condition is null before using it.
You'll never know how 3rd party extension would call this function.

PR #803, opened 2012-01-29
by default JDatabaseMySQL::getTableColumns() second arguments is 'true',...

ghost:
... make it 'false' to get fields as expected


elkuku:
I also found this bug while playing with this class - Seems like a "won't fix" - see: #707 :(

chdemko:
There is an issue while trying to merge:

"Your merge target has changed since we cached the merge commit - try the merge again "

Could you update your branch to the current head?

ghost:
that's the problem. I can't do merge either ;(
What should i do?

chdemko:
@signotorez Could you please close this pull request and open a new one?

PR #804, opened 2012-01-29
PHP_EOL for prettier purpose

ghost:
adapted title ;)


realityking:
Do we really want this? Personally I think it's annoying to handle CR or CRLF files.

But this does make the code prettier ;)

ianmacl:
If we're going to change it I think I would rather see it changed in favour of detecting the encoding and adapting rather than depending on the OS's line endings which may or may not match the file.

chdemko:
@signotorez Any news on this (see ianmacl comment)?

ghost:
Not sure sir, like 'realityking' said prettier.
since code should have consistency (IMO). theres will be more replacement between "\n" to PHP_EOL in libraries. i.e in utilities/xmlelement.php
back again. Do we really*2 need this?
i'll close this pull if unnecessary ;)

robschley:
This pull can't be merged automatically anymore. Please update and reopen if appropriate.

PR #805, opened 2012-01-30
Issue #27898

laoneo:
Added a check if the buffer from the document is really an array. Issue
number is #27898 on joomlacode.

### DIFF
--- a/libraries/joomla/cache/cache.php
+++ b/libraries/joomla/cache/cache.php
@@ -549,7 +549,8 @@ class JCache extends JObject
 
 		// Get the modules buffer before component execution.
 		$buffer1 = $document->getBuffer();
-		if(!is_array($buffer1)){
+		if (!is_array($buffer1))
+		{
 			$buffer1 = array();
 		}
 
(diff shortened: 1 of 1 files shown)


laoneo:
The cache class needs an array from $buffer1 = $document->getBuffer(); but when a Ajax request is done with a json response the $buffer1 is a string.

libraries/joomla/cache/cache.php
```diff
@@ -549,6 +549,9 @@ public static function setWorkarounds($data, $options = array())
 
 		// Get the modules buffer before component execution.
 		$buffer1 = $document->getBuffer();
+		if(!is_array($buffer1)){
```
realityking: You need a space after the if and the { goes on a new line.

libraries/joomla/cache/cache.php
```diff
@@ -605,6 +608,9 @@ public static function setWorkarounds($data, $options = array())
 			// @todo Check if the following is needed, seems like it should be in page cache
 			// Get the module buffer after component execution.
 			$buffer2 = $document->getBuffer();
+			if(!is_array($buffer2)){
```
realityking: Same as above.

LouisLandry:
Thanks a bunch!

PR #813, opened 2012-01-30
Implement a Database Iterator for iterating on queries

chdemko:
This pull request implement a Database Iterator for all SQL engines. The use is:

```
$dbo = JFactory::getDbo()->setType('stdClass');
$dbo->setQuery($dbo->getQuery(true)->select('*')->from('#__users'));
$dbo2 = clone $dbo;
foreach ($dbo as $i => $row)
{
    $query2 = $dbo2->getQuery(true)->select('*')->from('#__user_usergroup_map')->where('user_id = ' . (int) $row->id);
    $dbo2->setQuery($query2);
    foreach ($dbo2 as $i2 => $row2)
    {
        var_dump($row->id, $row2->group_id);
    }
}
```


realityking:
Cool stuff Christophe! I do have one small concern. Should we really take away access to the fetch\* methods without warning? Maybe it's better to add a deprecate note that access will be protected in release X. (I wouldn't add it do the deprecated log since the methods will be called hundreds of time's from inside JDatabase.)

chdemko:
@realityking, access is moved from protected to public (not the inverse)

pasamio:
Why not make JDatabase and it's derivatives just implement Iterator directly? Effectively it has to do this anyway, you're just creating a proxy object to it.

chdemko:
@pasamio: it is a very good idea but it leads to implement the cloning of objects as I wanted to allow nested loops query. Code has been updated for this purpose.

realityking:
To me keeping this separate makes sense. I can see us using other interfaces (e.g. countable, serializeable, jsonserializeable, maybe even array access) here as well.

Maybe we should have something like a JDatabseResult object and a method in JDatabase that returns that object. One could than either use JDatabase::setQuery() and make another function call to get the result object or use the API originally proposed by Christophe.

chdemko:
@realityking, we can add multiple interface to the JDatabase class. In fact Sam's remark is relevant: the purpose of iterating through a database is to browse the query results

libraries/joomla/database/database.php
```diff
@@ -54,7 +54,7 @@ public static function test();
 	 * @var    string
 	 * @since  11.1
 	 */
-	public $name;
+	public static $name;
```
ianmacl, line 57: This is an API change.  Is there a particular reason for it?  I don't get it.  See also all the other properties that are being made static - why?

tests/suite/joomla/database/database/JDatabaseMySQLTest.php
```diff
@@ -608,4 +608,250 @@ public function testUpdateObject()
...
+	 * @return  void
+	 *
+	 * @since   12.1
+	 */
+	public function testForEach()
+	{
+		// Testing 'alloc' type without specific key
+		$this->object->setType('assoc')->setKey(null)->setQuery($this->object->getQuery(true)->select('title')->from('jos_dbtest'));
+		$results = array(
+			array('title' => 'Testing'),
```
ianmacl: It isn't clear to me what this is testing against.  It might be worthwhile to document the stubs with the tests.

tests/suite/joomla/database/database/JDatabaseMySQLTest.php
```diff
@@ -608,4 +608,250 @@ public function testUpdateObject()
...
+		foreach ($this->object as $i => $result)
+		{
+			$this->assertThat(
+				$result,
+				$this->equalTo($results[$i]),
+				__LINE__
+			);
+		}
+
+		// Testing 'stdClass' type without specific key
```
ianmacl: Each new test should be in a separate method.

libraries/joomla/database/database.php
chdemko, line 57: Since the cloning could be possible, it's for avoiding useless duplicated properties

tests/suite/joomla/database/database/JDatabaseMySQLTest.php
chdemko: this is testing the pattern

```
foreach ($dbo as $key => $row)
```

which calls rewind, current, key, next and valid methods of the Iterator interface

tests/suite/joomla/database/database/JDatabaseMySQLTest.php
chdemko: I will split the tests in several methods

chdemko:
@ianmacl I've added backward compatibility for old instance properties and I've reorganized unit tests

eddieajau:
I really like the idea in principle, but I'm feeling like we need to do the iteration on a result set and not the db object (and that gets around your cloning problem).  To be honest it would also be nice to be able to iterate over "pages" of data as well.  I'll have a think about it because I need this for work :)

chdemko:
@eddieajau this was my primary idea (see https://github.com/joomla/joomla-platform/pull/813#issuecomment-3734335). I can come back to the original idea (implementing a JDatabaseIterator class). What do you mean by "pages" of data ?

chdemko:
@eddieajau See #903

chdemko:
Closing this one. See #903

PR #815, opened 2012-02-01
Update libraries/joomla/database/database.php

AmyStephen:
databaseexception.php no longer exists, has been renamed to exception.php.


realityking:
Good catch. Actually we can remove that entire line since JDatabaseException is loaded by the autoloader when necessary.

PR #818, opened 2012-02-03
Switch JApplicationweb to Use JLoader::Register

elinw:
Mixing jimport and JLoader in JWebApplication is causing some conflicts, so this just switches the jimports to JLoader::Register.


elinw:
Ugh

PR #819, opened 2012-02-03
Change jimport to JLoader::register in JApplicationWeb.

elinw:
Mixing Jimport and Jloader in the same class can lead to problems so it's better to just consistently use JLoader.


chdemko:
@elinw What problems are raised by mixing JLoader::register and jimport ? + pull request cannot be merged

elinw:
I'm going to close for now.

PR #822, opened 2012-02-04
Add missing semi-colon

mbabker:
Per http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27843

### DIFF
--- a/libraries/cms/form/field/media.php
+++ b/libraries/cms/form/field/media.php
@@ -65,7 +65,7 @@ class JFormFieldMedia extends JFormField
 			$script[] = '	function jInsertFieldValue(value, id) {';
 			$script[] = '		var old_id = document.id(id).value;';
 			$script[] = '		if (old_id != id) {';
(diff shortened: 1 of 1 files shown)


libraries/cms/form/field/media.php
```diff
@@ -65,7 +65,7 @@ protected function getInput()
 			$script[] = '	function jInsertFieldValue(value, id) {';
 			$script[] = '		var old_id = document.id(id).value;';
 			$script[] = '		if (old_id != id) {';
```
chdemko, line 67: is the leading ';' useless  ?

mbabker, line 67: Where at?

chdemko, line 67: line 67

chdemko, line 67: oops: my fault, I was misreading. Forgot that

PR #824, opened 2012-02-04
Make JDatabaseQuery subclasses autoloadable.

realityking:
There should be full backwards compatibility due to the lines added in import.php.


chdemko:
See #610

robschley:
Yes, I just spent a bunch of time updating that code. Let's not merge this yet.

realityking:
Ok. Let's see this as a minimum solution in case the other doesn't work out for some reason. (This one should be 100% B/C BTW)

robschley:
Merged. Thanks!

PR #827, opened 2012-02-05
Add tests for JTableLanguage

mbabker:
This pull adds unit tests for JTableLanguage.  The DDL is sync'd to the CMS #__languages table, so the pull tester will be broken since the sitename and ordering columns are added.


eddieajau:
Branch needs a re-sync.

mbabker:
Rebased to current staging.

mbabker:
Added a update file `jos_languages_update.sql` to `tests/tmp` with the schema changes so that those who are set up to run the unit tests locally can update their databases.

realityking:
Thank you!

PR #829, opened 2012-02-05
fix for issue #27946

jamesmoey:


chdemko:
The groups of a user are now of the form:
array($groupId1 => $groupId1, $groupId2 => $groupId2, ...)

See
https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/table/user.php#L112
https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/table/user.php#L162

You can use that code  for setUserGroups
https://gist.github.com/1751107

jamesmoey:
Do you want me to do the changes and do another pull request? addUserToGroup method on the helper class need to be change too then.

chdemko:
Change only your code in your branch and push it on your remote repo

chdemko:
@jamesmoey Any news on this?

PR #830, opened 2012-02-06
Add minimum supported version information to database drivers

mbabker:
This adds the minimum supported version information to the database drivers and a method to retrieve that string.  The values currently used are MySQL 5.0.4 (the CMS minimum version at present) and SQL Server 2008 R2 (internal version 10.50.1600.1, per a discussion I had with Sudhi yesterday).  I've also modified the getVersion method for SQL Server to just return the version string instead of the full array the sqlsrv_server_info method returns.


libraries/joomla/database/database.php
```diff
@@ -153,6 +153,12 @@ public static function test();
 	protected static $instances = array();
 
 	/**
+	 * @var    string  The minimum supported database version.
+	 * @since  12.1
+	 */
+	protected $dbMinimum;
```
chdemko: You can define this property static

libraries/joomla/database/database.php
```diff
@@ -600,6 +606,18 @@ public function getLog()
...
 	/**
+	 * Get the minimum supported database version.
+	 *
+	 * @return  string  The minimum version number for the database driver.
+	 *
+	 * @since   12.1
+	 */
+	public function getMinimum()
+	{
+		return $this->dbMinimum;
```
chdemko: in that case, use

```
return static::$dbMinimum
```

libraries/joomla/database/database/mysql.php
```diff
@@ -52,6 +52,12 @@ class JDatabaseMySQL extends JDatabase
 	protected $nullDate = '0000-00-00 00:00:00';
 
 	/**
+	 * @var    string  The minimum supported database version.
+	 * @since  12.1
+	 */
+	protected $dbMinimum = '5.0.4';
```
chdemko: static

libraries/joomla/database/database/sqlsrv.php
```diff
@@ -50,6 +50,12 @@ class JDatabaseSQLSrv extends JDatabase
 	protected $nullDate = '1900-01-01 00:00:00';
 
 	/**
+	 * @var    string  The minimum supported database version.
+	 * @since  12.1
+	 */
+	protected $dbMinimum = '10.50.1600.1';
```
chdemko: static

chdemko:
We could add a method to JDatabase

```
public function isSupported()
{
    return version_compare($this->getVersion(), static::$dbMinimum) >= 0;
}
```

And define in JDatabase

```
abstract public function getVersion();
```

mbabker:
getVersion is already abstract in JDatabase, working the other changes right now.

mbabker:
Changes pushed up.

PR #839, opened 2012-02-08
add curl and socket to update process of Joomla

oc666:
Due to requests of users to be able update joomla when allow_url_fopen is block (bug tracker #27852 & https://groups.google.com/forum/#!msg/joomla-dev-framework/lP1ql5K8Rdw/ZuVRB8JL_SAJ), this pull request add the ability to use curl & socket. 
The pull request change the way the update communicate, and it's using the new JHttpTransport layer of the platform.
The default prefer way to transport is curl and it can be change (see JUpdater::getAvailableDriver).
I've added isAvailable method to JHttpTransport interface, and each sub-class implemented this method accordingly (include in the pull request).
Please review.


libraries/joomla/http/transport.php
```diff
@@ -42,4 +42,11 @@ public function __construct(JRegistry &$options);
 	 * @since   11.3
 	 */
 	public function request($method, JUri $uri, $data = null, array $headers = null, $timeout = null, $userAgent = null);
+
+	/**
+	 * method to check if http transport layer available for using
+	 * 
+	 * @return bool true if available else false
+	 */
+	static public function isAvailable();
```
realityking: In JArchive we use isSupported(), probably good to be consistent here. Also please add a @since tag. Last but not least I think we standardized on having the access keyword first, so in this case it should be

```
 public static function isSupported();
```

libraries/joomla/updater/updater.php
```diff
@@ -218,4 +211,22 @@ public function update($id)
 		}
 		return false;
 	}
-}
+	
+	public static function getAvailableDriver(JRegistry $options, $default = 'curl') {
```
realityking: This is useful far beyond updating/installing. Maybe this should be placed in a factory (or JFactory) or JHttp. I bet others will have a stronger opinion on where to put it.

Also you're missing the docblock here.

libraries/joomla/http/transport/stream.php
```diff
@@ -177,4 +177,13 @@ protected function getResponse(array $headers, $body)
...
+	/**
+	 * method to check if http transport stream available for using
+	 * 
+	 * @return bool true if available else false
+	 */
+	static public function isAvailable() {
+		return function_exists('fopen') && is_callable('fopen');
+	}
+}
\ No newline at end of file
```
realityking: Not your fault but all filles should have a new line at the end.

libraries/joomla/updater/updater.php
```diff
@@ -218,4 +211,70 @@ public function update($id)
...
+	 * @param   JRegistery $options the option for creating http transport object
+	 * @param   String     $default the preffered method to use
+	 *
+	 * @return  JHttpTransport Interface sub-class
+	 *
+	 * @since   12.1
+	 */
+	public static function getAvailableDriver(JRegistry $options, $default = 'curl') {
+		$available_adapters = self::getHttpTransports();
+		// check if there is available http transport adapters
```
realityking: I don't think the transports are supposed to be used independent of JHttp.

While I looked at that I noticed that JHttp's constructor is a bit weird, you actually have to pass it the transport object. Maybe that's something worth improving.

libraries/joomla/updater/update.php
```diff
@@ -296,17 +298,14 @@ public function loadFromXML($url)
 		xml_set_element_handler($this->xml_parser, '_startElement', '_endElement');
 		xml_set_character_data_handler($this->xml_parser, '_characterData');
 
-		while ($data = fread($fp, 8192))
+		if (!xml_parse($this->xml_parser, $response->data))
```
pasamio, line 301: Does this potentially introduce a memory fault for large update files? Instead of chunking the updates, it's all read into memory at once.

libraries/joomla/updater/updater.php
```diff
@@ -218,4 +211,70 @@ public function update($id)
...
+	 * Finds an available http transport object for communication
+	 *
+	 * @param   JRegistery $options the option for creating http transport object
+	 * @param   String     $default the preffered method to use
+	 *
+	 * @return  JHttpTransport Interface sub-class
+	 *
+	 * @since   12.1
+	 */
+	public static function getAvailableDriver(JRegistry $options, $default = 'curl') {
```
pasamio: I'd suggest that this isn't a function of the updater but a function of the JHttp package instead and it'd be better to have a JHttpFactory class instead which would construct things and determine this information. Particularly given it's also being used in the installer package as well, instead of putting this into the updater, if you want to get a JHttp, then you should ask the JHttp package - not JUpdater.

As an aside I'd also change it from a default of a single item to an array. Then you go through the array in order of preference to select the adapter. If you only have one preference this still works as you have it now however it also gives you the ability to specify multiple. Additionally if you have transports defined elsewhere in the system but registered with JLoader, having an array could permit you to do a class_exists check and see if they exist as well and then testing them that way which makes this thing flexible without requiring people to put transports in the library.

libraries/joomla/updater/update.php
oc666, line 301: This should be take care in the jhttptransport layer. 
Currently (and before this commit) the layer doesn't supporting chunking.

libraries/joomla/updater/updater.php
oc666: I've moved this function to JHttp for refactoring (loose coupling).

oc666: Add the ability to input few adapters when using jhttp.
In addition I've removed socket class from being use when update the J! core, cause update.joomla.org doesn't suppot socket (fsockopen).

oc666: I've found that the socket have a bug - timeout doesn't input well on connected (null instead of number).
After fixing this, move back socket support to the updater classes.

libraries/joomla/http/http.php
```diff
@@ -183,4 +183,73 @@ public function trace($url, array $headers = null)
...
+	 *
+	 * @return  array    An array of available transport handlers
+	 *
+	 * @since   12.1
+	 * @todo make this function more generic cause the behaviour taken from cache (getStores)
+	 */
+	public static function getHttpTransports()
+	{
+		jimport('joomla.filesystem.folder');
+		$basedir = __DIR__ . '/../http/transport';
```
realityking: This can be simplified to

```
$basedir = __DIR__ . '/transport';
```

libraries/joomla/http/http.php
```diff
@@ -183,4 +183,73 @@ public function trace($url, array $headers = null)
...
+
+		$names = array();
+		foreach ($handlers as $handler)
+		{
+			$name = substr($handler, 0, strrpos($handler, '.'));
+			$class = 'JHttpTransport' . ucfirst($name);
+
+			if (!class_exists($class))
+			{
+				include_once $basedir . $name . '.php';
```
realityking: Is this necessary? I'd think the autoloader would pick up the classes-

libraries/joomla/http/http.php
```diff
@@ -183,4 +183,73 @@ public function trace($url, array $headers = null)
...
+			}
+
+			if (call_user_func_array(array($class, 'isSupported'), array()))
```
realityking: Since we're using PHP 5.3 now we can just call the static method on a variable class (http://php.net/manual/en/language.oop5.static.php) Just have to remember to change this back when back porting to the CMS.

libraries/joomla/http/http.php
oc666: fixed

libraries/joomla/http/http.php
oc666: remove unnecessary code.

libraries/joomla/http/http.php
oc666: fixed

libraries/joomla/http/http.php
```diff
@@ -183,4 +183,71 @@ public function trace($url, array $headers = null)
...
+			$available_adapters = $default;
+		}
+		// check if there is available http transport adapters
+		if (!count($available_adapters))
+		{
+			return false;
+		}
+		foreach ($available_adapters as $adapter) {
+			$class = 'JHttpTransport' . ucfirst($adapter);
+			if (call_user_func_array(array($class, 'isSupported'), array())) {
```
realityking: The same as line 244 applies here too. Also why are you checking it twice?

libraries/joomla/http/http.php
```diff
@@ -183,4 +183,71 @@ public function trace($url, array $headers = null)
...
+		foreach ($handlers as $handler)
+		{
+			$name = substr($handler, 0, strrpos($handler, '.'));
+			$class = 'JHttpTransport' . ucfirst($name);
+
+			/**
+			 * on J!2.5 (PHP 5.2) the condition should be:
+			 * call_user_func_array(array($class, 'isSupported'), array())
+			 */
+			if ($class::isSupported())
```
realityking: Isn't the check in getAvailableDriver() enough?

libraries/joomla/http/http.php
```diff
@@ -183,4 +183,71 @@ public function trace($url, array $headers = null)
...
+	 * Finds an available http transport object for communication
+	 *
+	 * @param   JRegistery $options the option for creating http transport object
+	 * @param   mixed      $default adapter (string) or queue of adapters (array) to use
+	 *
+	 * @return  JHttpTransport Interface sub-class
+	 *
+	 * @since   12.1
+	 */
+	public static function getAvailableDriver(JRegistry $options, $default = null) {
```
realityking: I'll leave it to the other wether this belongs into JHttp or a separate factory, I'd be fine with this.

libraries/joomla/http/http.php
```diff
@@ -183,4 +183,71 @@ public function trace($url, array $headers = null)
 	{
 		return $this->transport->request('TRACE', new JUri($url), null, $headers);
 	}
+
+	/**
+	 * Finds an available http transport object for communication
+	 *
+	 * @param   JRegistery $options the option for creating http transport object
+	 * @param   mixed      $default adapter (string) or queue of adapters (array) to use
+	 *
```
realityking: The docblock comment for $default is missing.

realityking:
You're almost there. phpcs reports 20 style errors, I could spot only one from looking at it. I placed a comment there.

oc666:
I could not install the phpcs. I've followed the instructions here: http://docs.joomla.org/Joomla_CodeSniffer#Netbeans but I'm getting the next error: "the file /usr/bin/phpcs isn't the right code sniffer batch"
Can someone upload for me the php-cs output?

libraries/joomla/http/http.php
oc666: fixed

libraries/joomla/http/http.php
oc666: fixed

libraries/joomla/http/http.php
oc666: fixed

realityking:
Here you go

```
FILE: ...Applications/MAMP/htdocs/joomla-platform/libraries/joomla/http/http.php
FOUND 9 ERROR(S) AFFECTING 6 LINE(S)
  40 | ERROR | The variable names for parameters $transport (2) and $adapters
     |       | (3) do not align
  40 | ERROR | The comments for parameters $transport (2) and $adapters (3) do
     |       | not align
  40 | ERROR | Expected 2 spaces after the longest variable name
 191 | ERROR | Expected 2 spaces after the longest type
 191 | ERROR | Expected 2 spaces after the longest variable name
 198 | ERROR | Opening brace should be on a new line
 208 | ERROR | Please start your comment with a capital letter; found "// check
     |       | if there is available http transport adapters"
 213 | ERROR | Expected "foreach (...)\n...{\n"; found "foreach (...) ...{\n"
 226 | ERROR | Whitespace found at end of line


FILE: ...ns/MAMP/htdocs/joomla-platform/libraries/joomla/http/transport/curl.php
FOUND 2 ERROR(S) AFFECTING 2 LINE(S)
 183 | ERROR | Whitespace found at end of line
 191 | ERROR | Opening brace should be on a new line


FILE: .../MAMP/htdocs/joomla-platform/libraries/joomla/http/transport/socket.php
FOUND 2 ERROR(S) AFFECTING 2 LINE(S)
 235 | ERROR | Expected "if (...)\n"; found "if (...) "
 264 | ERROR | Opening brace should be on a new line


FILE: .../MAMP/htdocs/joomla-platform/libraries/joomla/http/transport/stream.php
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
 188 | ERROR | Opening brace should be on a new line


FILE: ...tions/MAMP/htdocs/joomla-platform/libraries/joomla/installer/helper.php
FOUND 3 ERROR(S) AFFECTING 3 LINE(S)
 49 | ERROR | Instanciating new classes without parameters does not require
    |       | brackets.
 57 | ERROR | Expected "if (...)\n...{...}\n...else\n"; found "if (...)
    |       | ...{...}\n...else\n"
 59 | ERROR | Whitespace found at end of line

FILE: ...htdocs/joomla-platform/libraries/joomla/updater/adapters/collection.php
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
 229 | ERROR | Instanciating new classes without parameters does not require
     |       | brackets.


FILE: .../htdocs/joomla-platform/libraries/joomla/updater/adapters/extension.php
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
 167 | ERROR | Instanciating new classes without parameters does not require
     |       | brackets.


FILE: ...cations/MAMP/htdocs/joomla-platform/libraries/joomla/updater/update.php
FOUND 1 ERROR(S) AND 3 WARNING(S) AFFECTING 4 LINE(S)
 287 | ERROR   | Instanciating new classes without parameters does not require
     |         | brackets.


FILE: ...ations/MAMP/htdocs/joomla-platform/libraries/joomla/updater/updater.php
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
 215 | ERROR | Please end your files with an empty line.

```

oc666:
Thanks Rouven
Fixed these issues.

realityking:
Another round

```
FILE: ...Applications/MAMP/htdocs/joomla-platform/libraries/joomla/http/http.php
FOUND 2 ERROR(S) AFFECTING 2 LINE(S)
 191 | ERROR | Expected 2 spaces after the longest variable name
 228 | ERROR | Whitespace found at end of line

FILE: ...cations/MAMP/htdocs/joomla-platform/libraries/joomla/updater/update.php
FOUND 1 ERROR(S) AND 3 WARNING(S) AFFECTING 4 LINE(S)
 287 | ERROR   | Instanciating new classes without parameters does not require
     |         | brackets.
```

oc666:
Thanks again, Rouven
Fixed.

realityking:
One last thing, we discussed this and the change to JHttp should go into a separate factory.

oc666:
Hey Rouven,
I'm really confused now. 
First of all, where did you discussed about this?
My confusing is about the next questions:
Isn't it well-formed refactoring?
What required to refactoring? getAvailableDriver? 
What is the best approach for refactoring in J! core?
Are you aware to the fact that it works the same as the cache works? (same refactoring; see getStores method) 
Isn't it more important to get over #27984 first? (Make the auto-update process more reliable across different hosts)

Thanks and sorry for inconvience, this is for good cause.

realityking:
Sorry for the confusion, we discussed this on Skype. The reason why it should be done right in the first place is that if we move it after it's been in a release we'll have to keep the old API around. Don't want that.

Not refactoring, factory. (http://en.wikipedia.org/wiki/Factory_method_pattern) It essentially means to add another class the package called JHttpFactory that deals with creating the class with the right options. It's kinda like the JFactory just for one package.

There isn't really an example of a factory (besides JFacotry) in the platform right now, so you couldn't know. It's a design pattern we'll slowly move to. The only example I can think of is the JContentFactory (https://github.com/eBaySF/joomla-platform/blob/content/libraries/joomla/content/factory.php) but that one is quite a bit more complex.

If you need any more infos just shout.

oc666:
If you're discussing about stuff that I'm implementing, it would be better and more convience to corporate with me those discussions.

Is JHttpFactory just creating instance of JHttp according for some simple input? For example, new JHttpFactory('auto') would create instance of JHttp automatically based on the available one in the server. Correct me if I'm wrong.
What are the base things (properties & methods) this factory require? Wouldn't it better to implement first JObjectFactory (or something like that)?
Why we can't commit this change and after that use factory?

Thanks

realityking:
Sorry, took me a while to get back. In the middle of my exams, but today is Joomla day ;)

I'd really prefer to get this right the first time so we don't have to lug around a legacy API.

What would you use JObjectFactory for?

Anyway you'd need to move getHttpTransports() and getAvailableDriver() to JHttpFactory.

Lastly add a method like this:

```
public static function getHttp(JRegistry $options = null, $adapters = null)
{
   return new JHttp($options, self::getAvailableDriver($options, $adapters));
}
```

piotr-cz:
Hi, I've found one issue related to JHttp and this is kind of related to this pull request. Let's say I'm using `JHttpTransportSocket`.

When sending `JHttp::post()`, there is no way to choose timeout (defaults to null), so all connections are returning errors immediately (tested on slower connection).

Now, I'm a bit confused about JHttp::options and JHttpTransport::options. In case the first one would be intended to hold `timeout` and `userAgent` options, all commands in JHttp should pass them (kind of like in `interface JHttpTransport`):

```
public function post($url, $data, array $headers = null)
    {
        return $this->transport->request('POST', new JUri($url), $data, $headers, $this->getOption('timeout'), $this->getOption('useragent'));
    }
```

Should I create new pull request? Or I'm completely wrong somewhere?

realityking:
@piotr-cz I stumbled about the same thing when I took a stab at implementing direct to file downloading. My thinking would be that the arguments for timeout and userAgent should be removed and be dealt with in the options (the transports of an options object too). I was also wondering why the options are a JRegistry and not an array like most classes. I suggest to bring this up on jplatform so the original authors of that package (Louis and Ian if I remember correctly) can chime in.

realityking:
@oc666 
Code looks good to me, I'm gonna pull this as soon as you fix the style errors.

```
FILE: ...lications/MAMP/htdocs/joomla-platform/libraries/joomla/http/factory.php
FOUND 4 ERROR(S) AFFECTING 3 LINE(S)
 30 | ERROR | Expected 2 spaces after the longest type
 30 | ERROR | Expected 2 spaces after the longest variable name
 31 | ERROR | The comments for parameters $options (1) and $adapters (2) do not
    |       | align
 34 | ERROR | Missing @return tag in function comment

FILE: ...Applications/MAMP/htdocs/joomla-platform/libraries/joomla/http/http.php
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
 40 | ERROR | Superfluous doc comment at position 3
```

libraries/joomla/http/factory.php
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	protected static $instances = array();
+
+	/**
+	 * method to recieve Http instance.
+	 *
+	 * @param   JRegistry  $options   Client options object.
+	 * @param   mixed      $adapters  Adapter (string) or queue of adapters (array) to use for communication
+	 *
+	 * return   JHttp      Joomla Http class
```
realityking: You missed the @ before return. This probably caused the last two errors.

piotr-cz:
@realityking Do you mean Platform development Group?
Yeah, In my opinion this is a typical misuse of JRegistry. I hate that you cannot unset it's keys.
So timeout and userAgent options should be passed to constructor of JHttpTransport.

realityking:
Yes I mean the google Group.

oc666:
someone can tell me what is the style error, please?

realityking:
```

FILE: ...lications/MAMP/htdocs/joomla-platform/libraries/joomla/http/factory.php
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
 33 | ERROR | Return comment requires a blank newline after it
```

realityking:
Thanks for sticking with this :)

PR #844, opened 2012-02-09
Reworked Database Package

robschley:
This is mostly Omar's database refactoring work. It was getting untenable to keep it up to date with all the changes so I just did an export and then imported the files over an up to date checkout. This probably warrants a fair bit of review as it touches a lot of files.


realityking:
Haven't really looked trough it yet, but for started some of the @since tags should read 12.1, not 11.3.

elinw:
Couldn't we do this in a way that preserves history more completely? These aren't really new files for the most part. i.e. can't we create the new and then git mv the old?

realityking:
I agree, moving stuff without preserving the history isn't good.

#824 would take care of the query classes, since this is collapsed into one commit now anyways it should be easy to create a new pull for this after merging #824.

On a different note, do we really wanna rename JDatabase to JDatabaseDriver? Somehow that doesn't feel right.

elinw:
I also have a question about the class names that include SQL. Is that going to fool the autoloader when it breaks things up by camel case? e.g. https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/database/database/mysql.php#L23

mbabker:
Odds are the class will be renamed at some point.  The JDatabaseQuery classes were recently renamed, e.g. https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/database/query/mysql.php#L19

elinw:
But should we just do that now as part of this clean up /reorganization? That's  the question.

PR #848, opened 2012-02-09
Update PHPMailer to version 5.2.1.

realityking:


libraries/phpmailer/phpmailer.php
```diff
@@ -53,7 +53,7 @@ class PHPMailer {
    * Sets the CharSet of the message.
    * @var string
    */
-  public $CharSet           = 'utf-8';
+  public $CharSet           = 'iso-8859-1';
```
mbabker, line 56: This should remain utf-8 (I know iso-8859-1 is PHPMailer's default); at least that's how we've used it for quite some time.

chdemko, line 56: I disagree with @mbabker : since we use the JMail class, we only have to declare

```
public $CharSet = 'utf-8';
```

in our JMail class and leave PHPMailer unchanged

chdemko:
BTW, we have to upgrade to mootools 1.4.4 and to simplepie 1.2.1

libraries/phpmailer/phpmailer.php
mbabker, line 56: Well, I ended up changing it when I did the 5.2 update, so if that works better, I'm all for it.

realityking:
I have MooTools high on my list, I wanna do the update together with the dropping of the 1.2 compatibility layer.

As for simplepie, can't we just remove it from the platform? We don't really use it but for one method in the factory, I'd just leave it up to the CMS to include it.

libraries/phpmailer/phpmailer.php
realityking, line 56: Fixed.

PR #851, opened 2012-02-10
Remove extraneous ) in SQL, fix bug #28008

j16sdiz:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28008

### DIFF
--- a/libraries/joomla/table/session.php
+++ b/libraries/joomla/table/session.php
@@ -132,7 +132,7 @@ class JTableSession extends JTable
 		$query = $this->_db->getQuery(true);
 		$query->delete();
 		$query->from($this->_db->quoteName($this->_tbl));
-		$query->where($this->_db->quoteName('time') . ' < \'' . (int) $past . '\')');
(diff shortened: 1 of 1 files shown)


libraries/joomla/table/session.php
```diff
@@ -132,7 +132,7 @@ public function purge($maxLifetime = 1440)
 		$query = $this->_db->getQuery(true);
 		$query->delete();
 		$query->from($this->_db->quoteName($this->_tbl));
-		$query->where($this->_db->quoteName('time') . ' < \'' . (int) $past . '\')');
+		$query->where($this->_db->quoteName('time') . ' < ' . $this->quote($past));
```
chdemko, line 135: sorry for the mistake, it should be

```
$this->_db->quote($past)
```

elinw:
I think when it is an (int) you don't use quote.

chdemko:
@elinw +1

chdemko:
Fixed in code

PR #854, opened 2012-02-10
Move JFormFieldUser to the CMS.

realityking:


chdemko:
I would replace

```
```

by

```
defined('JPATH_BASE') or die;
```

in libraries/{joomla/form/fields → cms/form/field}/user.php

eddieajau:
There's probably a case for a user form field, but this one is so tightly coupled to the CMS it's not funny.

PR #855, opened 2012-02-10
Remove code that was made obsolete by the autoloader.

realityking:


tests/suite/cms/form/field/JFormFieldHelpsiteTest.php
```diff
@@ -20,6 +20,7 @@ class JFormFieldHelpsiteTest extends JoomlaTestCase
 	 */
 	protected function setUp()
 	{
+		require_once JPATH_PLATFORM . '/cms/form/help/help.php';
```
mbabker: This shouldn't have 'form' in the path.

realityking: Thanks. Fixed.

PR #861, opened 2012-02-12
Remove legacy code

jonnsl:


realityking:
A couple of things

-There are multiple public methods removed that weren't deprecated before. Moving them to JInstaller - which makes sense to me - should happen in a backward compatible fashion, like you did with parseXMLInstallFile().
-Removing the toolbar code in JApplicationHelper is one of the things we wanna do, but I'd like to know where in the CMS it's gonna end up (which class, which method) before we remove it.

jonnsl:
Indeed the parseXMLLangMetaFile method was not supposed to be remove, my mistake.
The JApplicationHelper::getPath is not used anymore since 1.6, in favor of the new MVC pattern, so I think that there is no need to deprecate that before remove it.
The toolbar code in JApplicationHelper wont be moved to the cms, it was some legacy code to generate the toolbar before the mvc pattern.

realityking:
So far we've tried to avoid to remove methods without prior warning, I don't see the harm to keep getPath() around as a deprecated method.

As for the toolbar, I realize this gotta move to the CMS. I'd just prefer to know where before we scratch it out. (I maintain a fork of Joomla that keeps working with the current platform)

realityking:
Ok I figured out where to put the toolbar code.

realityking:
I dove some more into this and discovered that this is actually a legacy feature back from Joomla 1.0 and should probably be completely dropped.

jonnsl:
what about [this](https://github.com/jonnsl/joomla-platform/blob/94c30d86f969516759ac82c906b0c95dd1cb7e81/libraries/joomla/application/component/helper.php#L332-L340) (another 1.0 legacy code)
I think that this should be replaced simple by

``` php
$path = JPATH_COMPONENT . '/' . $file . '.php';
```

realityking:
I agree, lets try that. We'll have a CMS 3.0 alpha and if lots of stuff starts breaking we could still revert and add a deprecation notice instead.

LouisLandry:
Can you add a JLog::add() entry in the deprecated/moved `JApplicationHelper::parseXMLInstallFile` so we can get this thing merged?

libraries/joomla/application/helper.php
```diff
@@ -166,128 +166,12 @@ public static function addClientInfo($client)
...
-	}
-
-	/**
 	 * Parse a XML install manifest file.
 	 *
 	 * XML Root tag should be 'install' except for languages which use meta file.
 	 *
+	 * @deprecated    12.1
+	 * @note    Use JInstaller::parseXMLInstallFile instead.
 	 * @param   string  $path  Full path to XML file.
```
chdemko: Move the  param just after the comment 

```
* Parse a XML install manifest file.
*
* XML Root tag should be 'install' except for languages which use meta file.
*
* @param   string  $path  Full path to XML file.
*
* @deprecated    12.1  Use JInstaller::parseXMLInstallFile instead.
```

libraries/joomla/installer/installer.php
```diff
@@ -1924,7 +1924,7 @@ public function isManifest($file)
 	 */
 	public function generateManifestCache()
 	{
-		return json_encode(JApplicationHelper::parseXMLInstallFile($this->getPath('manifest')));
+		return json_encode(JInstaller::parseXMLInstallFile($this->getPath('manifest')));
```
chdemko: use self::parseXMLInstallFile

jonnsl:
done

eddieajau:
Good stuff! Thanks.

PR #867, opened 2012-02-13
Removed caching from module list

ghost:
Since modules can be published based on time (publish_up / publish_down), the caching on the initial published module list makes no sense.

The caching key ($cacheid) only takes into consideration the $Itemid, $groups, $clientId, $lang.
However, the cached module list is time based.

This means that the publish up/down settings have no effect as long as the  list is still cached.
So that is a bug.

When looking at what is actually cached, there is not much to be gained by the caching as it is. It is only a database query and some basic checks on the returned object list.

So taking above into account (that caching causes the timed based bug and that there is little to no gain in caching), the solution is to simply take out the caching altogether.


realityking:
I agree that there's a bug nut I'd love to see some benchmarks on a site with a lot of modules since all modules are loaded in that query if I read it correctly.

ghost:
No, as far as I can tell only the module list that is returned from the database is stored in that cache.
Not the rendered modules themselves.

Advanced Module Manager and MetaMod Pro have been overruling this module helper file as it is with one that (among other things) doesn't use this cache.
We have seen no noticeable effects on page load time due to this non-use of module list cache.
Thousands of people are using these extensions and we have had no reports of pageload time problems related to this.

In what I have been able to test locally, there is no noticeable difference in load time (debug info) and a small decrease in memory usage when commenting out the $cache->store line.

chdemko:
@nonumber The pull request cannot be merged

ghost:
Because?

realityking:
It's not in a mergeable state due to other changes made, you have to update it.

ghost:
That's because merging sucks if you change indentation

ghost:
Hmmm, don't know why it is showing 5 files in the diff now, as I only commited changes to the module helper.php file.

LouisLandry:
Would you mind rebasing this?

try: `git pull --rebase git://github.com/joomla/joomla-platform.git master`

That will likely clean up the odd file changes and also get rid of the merge commits.

ghost:
I guess Git doesn't like me. Done that. All files no up-to-date. No commits/changes open.

LouisLandry:
OK, I'll take a look when I get home.  @eddieajau would you take a look at the access changes for me.

elinw:
Sometimes it is just easier to make a new branch and do a diff between the old and new, apply differences to your new. You do lose some of your own commit history that way but you only have one commit anyway.

ghost:
Created new fork and pull request: https://github.com/joomla/joomla-platform/pull/934

PR #871, opened 2012-02-13
Improved test coverage on JFormField classes

mbabker:


realityking:
Why are you removing JFormFieldCheckboxesTest? Otherwise great stuff.

eddieajau:
That JFormFieldCheckboxesTest is really doing nothing anyway. Not hard to put it back in at a later date.

mbabker:
@realityking

It was a duplicate test file.  There is also a JFormFieldCheckboxTest, which uses the correct class name.

realityking:
@mbabker 

There's both a JFormFieldCheckbox and a JFormFieldCheckboxes.

mbabker:
Wow, don't know how I missed that one.  I was going off the file list
directly in PhpStorm while working on this.  Chalk that one up to aging
eyes ;-)

On 2/14/12 1:17 PM, "Rouven Weßling"
<reply+i-3208922-6d25447d12c9b7e6ea85ea3459dc7c56d1b83ee1-368545@reply.gith
ub.com> wrote:

PR #872, opened 2012-02-13
Add basic tests for JDatabaseSQLSrv

mbabker:
PHPUnit natively doesn't support database operations when testing with SQL Server, so without writing our own test class, there isn't much we can test.  Basically, if it doesn't require a query, it appears we can test it (so, basically testing that the class is returned for chaining, the queries are built correctly with proper quoting, etc.).

I've implemented this test class the same way as the MySQL(i) test classes in that the class should be skipped if a config file isn't present (verified locally that this works as intended).

I've also renamed the MySQL DDL and added a DDL for SQL Server should future testing support become available.


eddieajau:
Merged.

PR #873, opened 2012-02-13
Remove $_SERVER and $_ENV variables from JInput serialize.

eddieajau:


robschley:
JInputCli has its own serialize method that probably needs to be modified as well.

Issue #866, opened 2012-02-13
JLoggerCallback


okonomiyaki3000:
I'd like to suggest a new type of JLogger which can be instantiated with a callback function (passed in the $options array, of course) and which simply passes $entry to that function when addEntry is called.

okonomiyaki3000:
Although this sounds simple enough, you'd want it to work with the lambdas (available since 5.3) in addition to the old 'create_function' and 'array' callback types. Currently this doesn't work because JLog::addLogger serializes $options to use it as a signature. PHP's lambdas can not be serialized. Trying to do so produces errors. So this can either be done with the caveat that only the old-school approach will work or some change could be made to JLog. I'm not sure how to fix it at the moment though. Would need a way to uniquely identify a lambda.

okonomiyaki3000:
It seems that object comparison works as you'd hope with Closure objects. So, there is a way to support lambdas. A special case would need to be added to JLog::addLogger so that it checked a particular key on the $options array for the presence of a Closure object. If it finds one, it would need to swap it for some unique id for serialization then swap it back in afterward. 

I'm just talking to myself, maybe?

elinw:
We're waiting for some code and tests to see how it would work.

okonomiyaki3000:
It's included in this pull request: https://github.com/joomla/joomla-platform/pull/874

Let me know if you think it needs more tests or anything.

elinw:
Okay in that case you can close this issue and just leave that one open.
#874

okonomiyaki3000:
Over and out.

PR #874, opened 2012-02-14
new JLogger type: JLoggerCallback (issue #866)

okonomiyaki3000:
This adds a new JLogger type that simply passes a log entry to a user defined callback function. Callback can be in any of the callback formats normally used by PHP, including lambda (available since PHP 5.3).

In order to properly support lambda functions (Closure objects), a minor modification to JLog was required.


libraries/joomla/log/log.php
```diff
@@ -13,7 +13,7 @@
 
 JLoader::register('LogException', JPATH_PLATFORM . '/joomla/log/logexception.php');
 
-JLoader::discover('JLogger', __DIR__ . '/loggers');
+JLoader::discover('JLogger', dirname(__FILE__) . '/loggers');
```
mbabker: Platform 12.1 requires PHP 5.3 minimum, and `__DIR__` is valid in 5.3.  Can you change this back?

libraries/joomla/log/log.php
```diff
@@ -419,4 +433,28 @@ protected function findLoggers($priority, $category)
...
 	}
+	
+	/**
+	 * Method to track closure objects passed as callbacks 
+	 *
+	 * @param   Closure  $closure  A Closure object
+	 *
+	 * @return  string  A unique identifier for the passed object
+	 *
+	 * @since   11.5
```
mbabker: `@since` tags should be for 12.1 for new development.  There won't be a 11.5 release.

libraries/joomla/log/log.php
```diff
@@ -419,4 +433,28 @@ protected function findLoggers($priority, $category)
...
+		
+		$id = array_search($closure, $closures);
+		
+		if ($id === false)
+		{
+			$closures[] = $closure;
+			$id = count($closures) - 1;
+		}
+		
+		return '_CLOSURE_ID_'.$id;
```
mbabker: To match the code style rules, this line should be `return '_CLOSURE_ID_' . $id;`

okonomiyaki3000:
OK, that should do it.

okonomiyaki3000:
Where can I see details of this checkstyle analysis?

realityking:
You have to run phpcs on your own computer. The code style definition is in the build/phpcs folder.

okonomiyaki3000:
Fixed the 9 errors but phpcs didn't turn up any warnings for me.

eddieajau:
Just trying to understand what's being proposed.  Why would you use a callback function when you can just as easily make your own logger class?

okonomiyaki3000:
It is surely simpler to write a function than a class. Especially with the lambda support in PHP 5.3, you can get a lot of versatility with very little effort. Maybe I've written too much Javascript but I really like the idea of extending functionality by just passing a function. 

I imagine that someone might use this in case their Joomla! installation is part of a larger ecosystem of apps and they'd like to hand off logging to some other part of that system. Or perhaps during some testing phase you find that you need some specific logging functionality on a temporary basis. Of course, you can write a class for these cases but just passing a function may be faster/easier/more fun?

eddieajau:
Ok, fair enough.  I'm not sure I'd use it, but I can see where you are coming from.  You have some testing errors to fix and we'd want some additional unit testing to verify it all works as advertised :)

eddieajau:
Oh, and if you can give me some explanatory text, I'll put it in the dev manual (http://developer.joomla.org/manual/chap-Joomla_Platform_Manual-Log.html) unless you want to have a crack at that as well.

Thanks in advance.

okonomiyaki3000:
I don't think these testing errors have anything to do with me. I get the same errors whether I run the tests on master or staging. Shouldn't master run without errors if I haven't touched it?

libraries/joomla/log/loggers/callback.php
```diff
@@ -0,0 +1,69 @@
...
+	 *
+	 * @since   12.1
+	 * @throws  LogException
+	 */
+	public function addEntry(JLogEntry $entry)
+	{
+		// Pass the log entry to the callback function
+		if (!call_user_func($this->callback, $entry))
+		{
+			throw new LogException;
```
robschley: There should be a message for every exception thrown.

robschley:
I think that if we're going to add this we really need some tests for it. I suspect this is the type of feature that not many people will use so if there are no tests it is likely to get neglected.

libraries/joomla/log/loggers/callback.php
okonomiyaki3000: In fact, I think we don't need an exception here at all. I put it in because the FormattedText logger throws one (also without a message) but, on second thought, it doesn't make much sense in this case. Any problem to just remove it?

okonomiyaki3000:
I've also written some documentation for this that should be suitable for the manual. What should I do with that?

okonomiyaki3000:
Hey guys, anything happening here?

chdemko:
The pull request cannot be merged (libraries/joomla/log/loggers has been renamed to libraries/joomla/log/logger)

okonomiyaki3000:
OK, all updated to the new locations

libraries/joomla/log/log.php
```diff
@@ -188,9 +188,23 @@ public static function addLogger(array $options, $priorities = self::ALL, $categ
 		}
 		$options['logger'] = strtolower($options['logger']);
 
+		// Special case - if a Closure object is sent as the callback (in case of JLoggerCallback)
+		// Closure objects are not serializable so swap it out for a unique id first then back again later
+		if (isset($options['callback']) && is_a($options['callback'], 'closure'))
+		{
+			$callback = $options['callback'];
+			$options['callback'] = self::getClosureUniqueId($options['callback']);
```
chdemko: You can use spl_hash_object($options['callback']); and drop the getClosureUniqueId function

libraries/joomla/log/logger/callback.php
```diff
@@ -0,0 +1,70 @@
...
+ * @package     Joomla.Platform
+ * @subpackage  Log
+ *
+ * @copyright   Copyright (C) 2005 - 2012 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE
+ */
+
+defined('JPATH_PLATFORM') or die;
+
+jimport('joomla.log.logger');
```
chdemko: This is useless

libraries/joomla/log/logger/callback.php
```diff
@@ -0,0 +1,70 @@
...
+ * Joomla! Callback Log class
+ *
+ * This class allows logging to be handled by a callback function.
+ * This allows unprecedented flexibility in the way logging can be handled.
+ *
+ * @package     Joomla.Platform
+ * @subpackage  Log
+ * @since       12.1
+ */
+class JLoggerCallback extends JLogger
```
chdemko: Use JLogLogger (JLogger is deprecated)

libraries/joomla/log/logger/callback.php
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Joomla! Callback Log class
+ *
+ * This class allows logging to be handled by a callback function.
+ * This allows unprecedented flexibility in the way logging can be handled.
+ *
+ * @package     Joomla.Platform
+ * @subpackage  Log
+ * @since       12.1
```
chdemko: 12.2

libraries/joomla/log/logger/callback.php
```diff
@@ -0,0 +1,70 @@
...
+	 * @since  12.1
+	 */
+	protected $callback;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param   array  &$options  Log object options.
+	 *
+	 * @since   12.1
```
chdemko: 12.2

libraries/joomla/log/logger/callback.php
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	/**
+	 * Method to add an entry to the log.
+	 *
+	 * @param   JLogEntry  $entry  The log entry object to add to the log.
+	 *
+	 * @return  boolean  True on success.
+	 *
+	 * @since   12.1
```
chdemko: 12.2

tests/suites/unit/joomla/log/loggers/JLoggerCallbackTest.php
```diff
@@ -0,0 +1,184 @@
...
+require_once JPATH_PLATFORM . '/joomla/log/logger/callback.php';
+require_once __DIR__ . '/stubs/callback/inspector.php';
+require_once __DIR__ . '/stubs/callback/helper.php';
+
+/**
+ * Test class for JLoggerCallback.
+ *
+ * @package     Joomla.Platform
+ * @subpackage  Log
+ * @since       12.1
```
chdemko: 12.2

tests/suites/unit/joomla/log/loggers/stubs/callback/helper.php
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Sample callbacks for the JLog package.
+ */
+
+/**
+ * Helper class for JLoggerCallbackTest
+ *
+ * @package     Joomla.UnitTest
+ * @subpackage  Log
+ * @since       12.1
```
chdemko: 12.2

chdemko:
Please add @since 12.2 to all your test functions

libraries/joomla/log/logger/callback.php
chdemko: Could you please remove this line?

chdemko:
Can you please make all references to 12.1 to be 12.2 and add <code>@since</code> to your test methods

PR #878, opened 2012-02-14
Update docblocks based on PhpStorm inspection

mbabker:
Browsed through many of the docblocks and updated the described types of various params and returns as appropriate.

### DIFF
--- a/libraries/joomla/database/query.php
+++ b/libraries/joomla/database/query.php
@@ -132,6 +132,10 @@ class JDatabaseQueryElement
  * @package     Joomla.Platform
  * @subpackage  Database
  * @since       11.1
+ *
+ * @method      string  q   Alias for quote method
+ * @method      string  qn  Alias for quoteName method
+ * @method      string  e   Alias for escape method
  */
 abstract class JDatabaseQuery
 {
@@ -685,7 +689,7 @@ abstract class JDatabaseQuery
 	 * @return  string  The escaped string.
 	 *
 	 * @since   11.1
-	 * @throws  DatabaseError if the internal db property is not a valid object.
+	 * @throws  JDatabaseException if the internal db property is not a valid object.
 	 */
 	public function escape($text, $extra = false)
 	{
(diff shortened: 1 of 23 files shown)


libraries/joomla/database/query.php
```diff
@@ -132,6 +132,10 @@ public function __clone()
  * @package     Joomla.Platform
  * @subpackage  Database
  * @since       11.1
+ *
+ * @method      string  q   Alias for quote method
+ * @method      string  qn  Alias for quoteName method
+ * @method      string  e   Alias for escape method
```
robschley, line 138: Shouldn't there be a dollar sign before the variable name?

mbabker, line 138: Looking at the phpDocumentor page, doesn't look like it's needed.  If
anything, it would be qn() or qn(string $name) (for example) based on the
examples in their documentation.  As is, in PhpStorm, the method now
displays as valid, but instead of taking me to either the __call method or
the real method, I'm quick linked to the class docblock.

http://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_t
ags.method.pkg.html

On 2/14/12 9:57 PM, "Rob Schley"
<reply+p-450093-acf37c32cb324435b8a814deafe9d4654554123d-368545@reply.githu
b.com> wrote:

robschley, line 138: Woops. Should read more carefully, it's a method not a param.

elkuku:
Wow this is pretty cool stuff, didn't knew about the possibility to use the docblocks for type hinting magic functions..
Based on this, it would be pretty cool to also document the __get() methods like this:
http://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.property.pkg.html

So, is this your next step, or would it make sense for me to invest some time on this ?

P.S. PHPStorm rocks :)

mbabker:
If you'd like to take a whack at it, by all means, go for it.  I just kinda hop around and decide "hey, this is what I'm going to look at today" ;-)

PR #880, opened 2012-02-15
Move JArchive to it's own package.

realityking:
Add type hinting to JArchive.
Add the JArchiveExtractable interface.


LouisLandry:
Can you please drop the dependency on JObject?  There is absolutely no reason for it to exist.

realityking:
I got a pull request in the works that deals with JObject for a number of classes (https://github.com/realityking/joomla-platform/commit/665f971efe8605c7a5b001e5821ee0429c121605), didn't even think about it when I made this pull. Changed that ;)

PR #838, opened 2012-02-08
fixed memcache and add new memcached driver

oc666:
fix bug #26514 and add memcached driver (#27979 in the feature tracker)

PR #881, opened 2012-02-15
Add JGithubCommits class

mbabker:
Added class to work with GitHub commits API per:
- http://developer.github.com/v3/repos/commits/
- http://developer.github.com/v3/git/commits/

### DIFF
--- a/libraries/joomla/github/commits.php
+++ b/libraries/joomla/github/commits.php
@@ -104,9 +104,9 @@ class JGithubCommits extends JGithubObject
 	/**
 	 * Method to delete a comment on a commit.
 	 *
-	 * @param   string   $user  The name of the owner of the GitHub repository.
-	 * @param   string   $repo  The name of the GitHub repository.
-	 * @param   string   $id    The ID of the comment to edit.
+	 * @param   string  $user  The name of the owner of the GitHub repository.
+	 * @param   string  $repo  The name of the GitHub repository.
+	 * @param   string  $id    The ID of the comment to edit.
 	 *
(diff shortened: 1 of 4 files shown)


libraries/joomla/github/commits.php
```diff
@@ -0,0 +1,357 @@
...
+	 * @param   array   $parents  Array of the SHAs of the commits that were the parents of this commit.
+	 *                            If omitted or empty, the commit will be written as a root commit.
+	 *                            For a single parent, an array of one SHA should be provided.
+	 *                            For a merge commit, an array of more than one should be provided.
+	 *
+	 * @return  object
+	 *
+	 * @since   12.1
+	 */
+	public function create($user, $repo, $message, $tree, $parents = array())
```
realityking: Since the rest of the package is type hinted so should probably $parents.

mbabker:
The 3 additional checkstyle warnings are coming from #838 which looks intentional (they match the errors in the memcache driver).

LouisLandry:
Very nice Michael... very nice.

oc666:
Hey
I'm wondering what is used for?
Thanks

mbabker:
The GitHub package is a group of classes that can be used in Joomla! code to interact with GitHub's API's.  The most notable use of the package right now is the pull tester (the @joomla-jenkins comments in pull requests).

This specific class doesn't have an implementation in code right now, but it adds the possibility to interact with the GitHub API's commits data points.

oc666:
Michael, Thanks for reply

PR #884, opened 2012-02-16
Fix plugin uninstall

mbabker:
In removing some of the legacy code from the plugin installation adapter, some of the file checks weren't properly handled, which is causing uninstall for plugins to fail since, as is, the XML manifest isn't being found.  This fixes that.

### DIFF
--- a/libraries/joomla/installer/adapters/plugin.php
+++ b/libraries/joomla/installer/adapters/plugin.php
@@ -581,34 +581,10 @@ class JInstallerPlugin extends JAdapterInstance
 		// Set the plugin root path
 		$this->parent->setPath('extension_root', JPATH_PLUGINS . '/' . $row->folder . '/' . $row->element);
 
+		$this->parent->setPath('source', $this->parent->getPath('extension_root'));
+
 		$this->parent->findManifest();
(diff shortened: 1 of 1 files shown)


realityking:
I noticed this comment:

```
// Try for Joomla 1.5 type queries
```

I've seen it in some of the other adapters too. Is that still accurate? Do we want to keep that code?

mbabker:
I've read those comments as well, and I don't know if the comment is
inaccurate or if it is truly parsing 1.5 style code in those cases.  It
seems like what is actually happening is what should be happening just
reading the code, but maybe there was another parameter somewhere that got
dropped?  I'd have to dust off a 1.5 trunk checkout and read the code to
verify that.

On 2/16/12 8:28 PM, "Rouven Weßling"
<reply+i-3249497-a529be643e56901671181fa7167980457ad17977-368545@reply.gith
ub.com> wrote:

mbabker:
Ya, it definitely looks like we need to clean up the comments at some point soon.  The 1.5 comments (at least comparing the component adapter) are referring to when the adapters had checks in place for 1.0 style installations versus 1.5 style and when the 1.0 code was dropped, the comments weren't changed.

PR #889, opened 2012-02-18
Make JApplication::getMenu() and JApplication::getPathway() static to prevent strict errors.

realityking:
The menu one causes a strict error in the CMS, pathway is just proactive.


LouisLandry:
You can't reference $this in a static function.  Also, it is not a good trend to make these things static when they don't have to be.  A better way to solve the strict error would be to not call the methods statically.

realityking:
Damn I knew there was a reason why it isn't static when I looked the last time. When I looked today I missed the $this.

I'm closing this pull request until I come up with a better fix.

PR #891, opened 2012-02-19
Improvements to the system autoloader.

LouisLandry:
- Add support for numeric digits in class names with the system autoloader.
- Add ability to register multiple lookup paths with the system autoloader.

To register a new filesystem location for Joomla libraries (`J` prefixed classes) with the system autoloader is as simple as:

``` php
// Register my library base path for Joomla platform libraries.
JLoader::registerPrefix('J', '/path/to/my/joomla/libraries');
```

To register a set of libraries with a different class prefix (in this example `C`) it is as simple as:

``` php
// Register my libraries.
JLoader::registerPrefix('C', '/path/to/custom/libraries');
```

### DIFF
--- a/libraries/loader.php
+++ b/libraries/loader.php
@@ -235,9 +235,9 @@ abstract class JLoader
 	 * set to true then any registered lookups for the given prefix will be overwritten with the current
 	 * lookup path.
 	 *
-	 * @param   string   $class  The class prefix to register.
-	 * @param   string   $path   Absolute file path to the library root where classes with the given prefix can be found.
-	 * @param   boolean  $reset  True to reset the prefix with only the given lookup path.
+	 * @param   string   $prefix  The class prefix to register.
+	 * @param   string   $path    Absolute file path to the library root where classes with the given prefix can be found.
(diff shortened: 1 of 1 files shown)


pasamio:
Read my mind! I was thinking about this just this morning about needing a way third party libraries could register their own prefix to get the same advantage as the Joomla classes

libraries/loader.php
```diff
@@ -249,28 +294,42 @@ public static function setup()
...
-		{
-			// Split the class name (without the J) into parts separated by camelCase.
-			$parts = preg_split('/(?<=[a-z])(?=[A-Z])/x', substr($class, 1));
+	/**
+	 * Load a class based on name and lookup array.
+	 *
+	 * @param   string  $class   The class to be loaded (wihtout prefix).
+	 * @param   array   $lookup  The array of base paths to use for finding the class file.
+	 *
+	 * @return  void
```
eddieajau, line 312: After the fact, but this method doesn't return void.

http://php.net/manual/en/function.include.php

See example 5

PR #892, opened 2012-02-19
Add more phpcs sniffs

realityking:
This adds quite a few more sniffs, I made about one commit per sniff so if there are any we don't want I can easily remove them from this pull.

I also added some XML to exclude our 3rd party libraries so we can now safely run the sniffer on the whole libraries folder. Jenkins should probably changed to check that folder. This picked up some style errors in the phpmailer lang file.

I also added our first CSS sniffs. There's currently an error with the last-line-must-be-blank sniff so jenkings probably shouldn't check those files just yet, the CSS sniffs themselves pass just fine.

Lastly I fixed 2 warnings that were there before so now there should be less warnings.

CC @elkuku


eddieajau:
Done.

PR #894, opened 2012-02-19
Reduce dependency on JObject

realityking:
Special pull request for @LouisLandry. ;)

The following classes no longer extend JObject
JBrowser
JClientLdap
JDocumentRenderer
JFeedEnclosure
JFeedImage
JFeedItem
JFilterInput
JProfiler
JOpenSearchImage
JOpenSearchUrl
JToolBar

Also deprecate JObject::__toString().

### DIFF
--- a/libraries/joomla/archive/archive.php
+++ b/libraries/joomla/archive/archive.php
@@ -167,7 +167,7 @@ class JArchive
 	 *
 	 * @param   string  $type  The type of adapter (bzip2|gzip|tar|zip).
 	 *
-	 * @return  object   JObject
+	 * @return  object  JArchiveExtractable
 	 *
 	 * @since   11.1
 	 */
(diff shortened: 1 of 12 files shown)


eddieajau:
The only thing I disagree with is deprecating JObject::__toString.  We have some changes in the UCM stuff that propose changing it to return a JSON representation of the object (helps with caching and serialization).  Otherwise, happy with the rest of it.

realityking:
Wouldn't that be a breaking change anyways? Also as far as I've seen the UCM stuff doesn't even touch or extend JObject.

I talked to Rob about this a while back and he meant everything in JObject is fair game.

eddieajau:
It's stuff we've worked on since, and yes it would be a breaking change, technically (though, the end result of a breaking change and deprecation is more or less the same in the long run).  However, the toString only returns the name of the class which is next to useless but allows for a workaround in existing and future code.

eddieajau:
Actually, I'll merge it and then we can haggle over reinstating __toString :)

PR #895, opened 2012-02-20
Allows JLanguage::load() to accept an array of $basePath[s] to search

WebMechanic:
... for a given language file w/ optional fallback to the default language.

Signature changes the default and internal behavior of "$lang = null" to "$lang = true".
The change is BC and has no effect for the "11.4" behavior.

The reason behind this is to avoid bulk calls to $lang->lang() as found en masse in the CMS to find an appropriate .ini file in different known locations.

### DIFF
--- a/libraries/joomla/language/language.php
+++ b/libraries/joomla/language/language.php
@@ -700,11 +700,11 @@ class JLanguage extends JObject
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


chdemko:
The unit test are failing

WebMechanic:
I know, but I'm too busy to take care of this now.
closing.

PR #896, opened 2012-02-20
Fix failing unit tests.

realityking:


realityking:
I was so bold to merge this myself since this fixes only failing tests and doesn't touch shipping code.

PR #898, opened 2012-02-20
Implement JAccess::getActionsFromFile and JAccess::getActionsFromData method

chdemko:
This implements JAccess::getActionsFromFile using an xml file and an xpath parameter and JAccess::getActionsFromData using an xml string or a JXMLElement and an xpath parameter.

Tests are now fully covered.

### DIFF
--- a/libraries/joomla/access/access.php
+++ b/libraries/joomla/access/access.php
@@ -423,7 +423,7 @@ class JAccess
 	 *
 	 * @since   11.1
 	 *
-	 * @deprecated  12.3  Use JAccess::getActionsFromFile or JAcces::getActionsFromData instead.
+	 * @deprecated  12.3  Use JAccess::getActionsFromFile or JAccess::getActionsFromData instead.
 	 *
 	 * @codeCoverageIgnore
 	 *
--- a/libraries/joomla/access/access.php
+++ b/libraries/joomla/access/access.php
@@ -245,17 +245,6 @@ class JAccess
 		$db->setQuery($query);
 		$result = $db->loadColumn();
 
-		// Get the root even if the asset is not found and in recursive mode
-		if ($recursive && empty($result))
-		{
-			$query = $db->getQuery(true);
-			$query->select('rules');
-			$query->from('#__assets');
-			$query->where('parent_id = 0');
-			$db->setQuery($query);
-			$result = $db->loadColumn();
-		}
-
 		// Instantiate and return the JAccessRules object for the asset rules.
 		$rules = new JAccessRules;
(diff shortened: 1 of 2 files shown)


eddieajau:
The (OR a.parent_id=0) is causing slow queries (like home page is totally inaccessible) on sites with a large number of groups (100 or so). The table indexes are not tuned for it and it would be helpful to know why the condition is there because a 0 parent_id means the whole asset table has lost integrity.

elinw:
Isn't there a difference between having a parent id of 0 and not being in the assets table at all? If an asset is not found  shouldn't it fall back to parent_id =1 and the default actions?

Andrew, I think there is a huge question out there about what to do when the assets table has lost its integrity  there are many many instances where it happens in real life, not because of anything internal to the API but because of ways that people mess with their databases and sometime'ts try to get around the API (or haven't fully understood it). If the assets table is borked huge numbers of queries basically slow to a halt. I think there has to be some way that when a messed up asset table is detected rebuilding the tree is triggered (rather than simply skipping around the problem).

elinw:
Actually i should clarify that the root has parent id of 0 in both the asset table and the categories table, so it is a valid value if it exists only once and really is the root.

chdemko:
@elinw it is done by https://github.com/chdemko/joomla-platform/blob/2456b71274c0022871f9210cb1122a1fd06cad61/libraries/joomla/access/access.php#L249

elinw:
Yes I saw that, I just wanted to claifyr what I said earlier because we have a really really serious performance impact of  problems related to that specifically when asset tables end up with 2 or more records with parent ids of 0 which seems to .be very common in migrated databases;. I hadn't really noticed that part of the code before but it explains some of the serious problems that we see.

eddieajau:
@elinw I think the long term solution is to move away from the asset table and bring the permissions into objects, which is what we are trying to do in UCM.  I think the asset table was a good step in the process, but it has it's problems.  In the mean time the way to restore integrity is basically to open up each "content" and save it again.  It's certainly not practical to do, and I'm not sure if that's practical to automate.  A stop-gap could be that anything where the asset name is like, for example, com_content. and has a parent id could be re parented to the component's asset entry.  Not perfect but more easily automated (just select all the components {parent == root}, then select all their assets with parent_id = 0; fix the parent_id; delete what's left).

mbabker:
Just out of curiosity (now that I'm looking at porting the code back to CMS 2.5), why the change from the `$section` param in getActions() to having to specify the full `$xpath` in the new methods.

elinw:
We need to handle this as a b/c break and I don't think it has been documented at all has it?

PR #900, opened 2012-02-20
Refactor database package

eddieajau:
This is mostly Omar's database refactoring work (and Rob's getting it back up to date). It was getting untenable to keep it up to date with all the changes so I just did an export and then imported the files over an up to date checkout. This probably warrants a fair bit of review as it touches a lot of files.


libraries/joomla/database/driver.php
```diff
@@ -0,0 +1,1664 @@
...
+			}
+
+			// If the class still doesn't exist we have nothing left to do but look at the next type.  We did our best.
+			if (!class_exists($class))
+			{
+				continue;
+			}
+
+			// Sweet!  Our class exists, so now we just need to know if it passes it's test method.
+			if (call_user_func_array(array($class, 'test'), array()))
```
oc666, line 220: We can use $class->test if we are using at least PHP 5.3

libraries/joomla/database/driver.php
```diff
@@ -0,0 +1,1664 @@
...
+	/**
+	 * Splits a string of multiple queries into an array of individual queries.
+	 *
+	 * @param   string  $sql  Input SQL string with which to split into individual queries.
+	 *
+	 * @return  array  The queries from the input string separated into an array.
+	 *
+	 * @since   11.1
+	 */
+	public static function splitSql($sql)
```
oc666, line 295: Someone can explain what this function used for? example would be great.

mbabker, line 295: During install and update, this function is used to separate multiple queries in an extension's .sql file into single queries to be executed one at a time.

vietvh, line 295: We are having _splitQueries() (in the installation folder) which do the same task.

mbabker, line 295: The installation app itself is in need of some work (Rouven and Nikolai have both been doing some cleanup and experimenting).  This is just one of those instances where the duplicate code needs removed.

mbabker:
The JDatabase mock needs updated.  This is why the tests are failing.

LouisLandry:
This still isn't tracking file history appropriately.  I'm gonna take a crack at it and see if we can get that done a little better.

chdemko:
see #803

eddieajau:
Closing this pull. See #906

PR #903, opened 2012-02-20
Implement Countable Database Iterators for iterating on queries

chdemko:
This pull request implement Countable Database Iterators for all SQL engines. The use is:

```
$dbo = JFactory::getDbo();
$iterator = $dbo->getIterator($dbo->getQuery(true)->select('*')->from('#__content')->execute());
foreach ($iterator as $row)
{
// Deal with $row
}
$count = count($iterator);
```

for results queries

### DIFF
--- a/libraries/joomla/database/driver.php
+++ b/libraries/joomla/database/driver.php
@@ -201,7 +201,7 @@ abstract class JDatabaseDriver extends JDatabase implements JDatabaseInterface
 				continue;
 			}
 
-			// Sweet!  Our class exists, so now we just need to know if it passes it's test method.
+			// Sweet!  Our class exists, so now we just need to know if it passes its test method.
 			// @deprecated 12.3 Stop checking with test()
 			if ($class::isSupported() || $class::test())
 			{


eddieajau:
Can we wait until the database refactor is finalize before doing all this stuff.  I want to get into DB improvements myself, but it will be complicated if the goal posts keep moving :)

eddieajau:
Christophe, I like where you are heading with.  I've made a slight adjustment to separate the result set (I call it a cursor) more. See:

https://gist.github.com/1874789

The driver would still be responsible for executing the query, but once it has done that, you can access the cursor object however you want.  What do you think?

chdemko:
@eddieajau Ok, I'm waiting #906 for being committed before continuing that way

eddieajau:
Ok, ping me and we can discuss.

eddieajau:
Actually, thinking about it, why don't we have the iterator always use the assoc array or object return, and then the load\* methods can do a conversion on the fly

eddieajau:
Ok, I've dropped the gist back to just supporting objects (https://gist.github.com/1874789). The reason is I think converting the result on the fly is better than having all the nested conditions, for example:

```
foreach ($dbi as $key => $object)
{
    // But I want an array.
    $array = (array) $object;
}
```

That leaves the engine specific iterator to duplicate just the fetch object and the free result methods which I think is acceptable.  The driver methods loadColumn, loadAssocList and loadObjectList can be converted to use the iterator, but the loadResult, loadAssoc and loadObject methods can remain unchanged.

chdemko:
@eddieajau Pull request updated. (I will be offline until 5 of March)

LouisLandry:
:heart:

libraries/joomla/database/driver.php
```diff
@@ -671,6 +671,61 @@ public function getQuery($new = false)
...
+	 * @return  JDatabaseIterator  A new database iterator.
+	 *
+	 * @since   12.1
+	 * @throws  RuntimeException
+	 */
+	public function getIterator($class = 'stdClass', $column = null)
+	{
+		// Derive the class name from the driver.
+		$class = 'JDatabaseIterator' . ucfirst($this->name);
+
```
robschley: Why have a $class option if you are just going to set the class in the first line of the function?

libraries/joomla/database/driver.php
```diff
@@ -671,6 +671,61 @@ public function getQuery($new = false)
...
+	 * @return  JDatabasePages  A new database pages iterator.
+	 *
+	 * @since   11.1
+	 * @throws  RuntimeException
+	 */
+	public function getPages($limit, $class = 'stdClass', $column = null)
+	{
+		// Derive the class name from the driver.
+		$class = 'JDatabasePages' . ucfirst($this->name);
+
```
robschley: Same question about the $class argument here...

chdemko:
@robschley It was a mistake in my code (updated now)

chdemko:
@robschley @eddieajau @LouisLandry @realityking 

The branch is ready for review

docs/manual/en-US/chapters/packages/database.xml
```diff
@@ -0,0 +1,52 @@
...
+    <title>Introduction</title>
+
+    <para>The <emphasis>Database</emphasis> package is designed to manage the operations of data management through the use of a generic database engine.</para>
+
+  </section>
+
+  <section>
+    <title>Iterating on results</title>
+
+    <para>The <code language="PHP (HTML)">JDatabaseIterator</code> class allows to iterate over database results
```
elinw: Instead of "to iterate" would you say "iteration" please.

docs/manual/en-US/chapters/packages/database.xml
```diff
@@ -0,0 +1,52 @@
...
+    </para>
+    <para>It allows also to count the results.
+      <programlisting language="PHP (HTML)"><![CDATA[$count = count($iterator);]]></programlisting>
+    </para>
+  </section>
+
+  <section>
+    <title>Iterating on pages</title>
+
+   <para>The <code language="PHP (HTML)">JDatabasePages</code> class allows to iterate over database results pages
```
elinw: Same as above "iteration over" instead of "to iterate over"

docs/manual/en-US/chapters/packages/database.xml
```diff
@@ -0,0 +1,52 @@
...
+$pages = $dbo->getPages($dbo->getQuery(true)->select('*')->from('#__content')->execute(), 10);
+foreach ($pages as $page)
+{
+	foreach ($page as $row)
+	{
+		// Deal with $row
+	}
+}]]></programlisting>
+    </para>
+    <para>It allows also to count the pages.
```
elinw: It also allows counting of pages.

libraries/joomla/database/iterator.php
```diff
@@ -0,0 +1,233 @@
...
+
+			// Update the number of fetched object
+			$this->_fetched++;
+		}
+	}
+
+	/**
+	 * Rewinds the iterator.
+	 *
+	 * This iterator cannot be rewinded.
```
elinw: rewound.

English is such a pain.

libraries/joomla/database/pages.php
```diff
@@ -0,0 +1,182 @@
...
+	public function count()
+	{
+		// Return the number of pages
+		return $this->_count;
+	}
+
+	/**
+	 * Rewinds the iterator.
+	 *
+	 * This iterator cannot be rewinded.
```
elinw: rewound.

libraries/joomla/database/iterator.php
```diff
@@ -0,0 +1,233 @@
...
+ */
+abstract class JDatabaseIterator implements Countable, Iterator
+{
+	/**
+	 * The database cursor.
+	 *
+	 * @var    mixed
+	 * @since  12.1
+	 */
+	private $_cursor;
```
realityking: If we'd make this protected instead of private we wouldn't have to pass the cursor to fetchCount(), fetchObject() and freeResult(). Not sure if there's a benefit to this approach I'm overlooking.

libraries/joomla/database/iterator.php
```diff
@@ -0,0 +1,233 @@
...
+
+	/**
+	 * The class of object to create.
+	 *
+	 * @var    string
+	 * @since  12.1
+	 */
+	private $_class;
```
realityking: Same as above but only applies to fetchObject().

libraries/joomla/database/pages.php
```diff
@@ -0,0 +1,182 @@
...
+defined('JPATH_PLATFORM') or die;
+
+/**
+ * Database Pages Class.
+ *
+ * @package     Joomla.Platform
+ * @subpackage  Database
+ * @since       12.1
+ */
+abstract class JDatabasePages implements Iterator, Countable
```
realityking: Do we actually need subclasses of this? All the concrete implementations are empty, the database specific work seems to be abstracted into the Iterator.

To get the correct iterator we could pass the iterator object to JDatabasePages in JDatabaseDriver::getPages().

Of course if there's database specific stuff that will be added later to JDatabasePages than this approach makes more sense.

realityking:
This will be awesome to have!

libraries/joomla/database/iterator.php
chdemko: private properties are easier to change as they are not accessible from 3rd party classes

libraries/joomla/database/pages.php
chdemko: You're true, we could add a generic JIteratorPages able to iterate over any iterators.

realityking:
Commits for both my proposed changes:
Pages: https://github.com/realityking/joomla-platform/commit/9a2e68acf417f274fe38cabc3f9a797829bd12bf
Iterator: https://github.com/realityking/joomla-platform/commit/5b8a465685437e32889e3d5e4887152ada66beb4

Feel free to cherry pick them.

chdemko:
I have removed the 'pages' classes as it could be replaced by a generic JIteratorPages class (to be defined later, but not in this pull request)

LouisLandry:
@chdemko You forgot to remove the getPages() method from JDatabaseDriver.

PR #904, opened 2012-02-20
Fix typo

elinw:


elinw:
I might as well be completely annoying.

PR #906, opened 2012-02-21
Reorganized and refactored database package.

LouisLandry:
Most of this work was done by Omar Ramos.  Rob and Andrew took a crack at getting it all up to date and ready to be merged, and now so have I.  In this branch the tests all the tests are passing and things seem pretty good.  There may be some extra things we need to get done for backward compatibility or whatnot, but I feel it is something we can get done after getting this stuff in there.


elinw:
Sweet, hits all the places I was worried about being impacted. :)

chdemko:
Everything seems ok to me except the file tests/suite/joomla/database/database/JDatabaseMySQLTest.php:114

```
$this->isInstanceOf('JDatabaseMySQL'),
```

ought to be

```
$this->isInstanceOf('JDatabaseDriverMysql'),
```

The tests on the database have to be redesigned anyway.

LouisLandry:
Well, I purposefully didn't redesign the tests because I wanted to ensure as much backward compatability as possible, and one great way of doing that is to maintain the same tests and swap out the implementation.  That's what we have here.  I think once we get this thing merged we can do a whole lot of things to improve upon stuff ... that's always the case, but that was my rationale on this for now.

Quite honestly when it is time to redesign I'd like to get rid of the inspectors all together and use reflection for accessing non-public members.  I'd like to add new tests for classes that have been rearranged as well.  I do think however that it makes a lot of sense for us to keep the old DB tests around in as original a format as possible until the 12.1 release simply to ensure that we don't go and break backward compatibility on a fundamental part of the code base more than we have to.  And to be honest, referencing old class names is a part of that in my opinion.

eddieajau:
@mbabker has some fixes here:
https://github.com/mbabker/joomla-platform/commits/database

chdemko:
@LouisLandry In fact the unit tests do not success if I have a tests/config_mysql.php file (This is why I propose this modif in the unit test)

LouisLandry:
Good catch @chdemko.  @mbabker, I pulled in your fixes.  Thanks.

libraries/joomla/database/driver.php
```diff
@@ -0,0 +1,1654 @@
...
+			}
+
+			// Derive the class name from the type.
+			$class = str_ireplace(array('.php', 'sql'), array('', 'SQL'), 'JDatabaseDriver' . ucfirst(trim($type)));
+
+			// If the class doesn't exist, let's look for it and register it.
+			if (!class_exists($class))
+			{
+				// Derive the file path for the driver class.
+				$path = __DIR__ . '/driver/' . $type;
```
realityking: Is this piece of code even needed? All classes will already be loaded by the autoloader due to the class_exists() call in line 196.

libraries/joomla/database/database.php
```diff
@@ -9,1612 +9,33 @@
 
 defined('JPATH_PLATFORM') or die;
 
-jimport('joomla.filesystem.folder');
+// Deprecation warning.
+JLog::add('JDatabase is deprecated, use JDatabaseDriver instead.', JLog::NOTICE, 'deprecated');
```
realityking: Since JDatbaseDriver extends JDatabase this message can't be avoided, it probably makes more sense to rely on the messages inside the methods.

libraries/joomla/database/factory.php
```diff
@@ -0,0 +1,240 @@
...
+		$class = 'JDatabaseDriver' . ucfirst(strtolower($options['driver']));
+
+		// If the class still doesn't exist we have nothing left to do but throw an exception.  We did our best.
+		if (!class_exists($class))
+		{
+
+			// Legacy error handling switch based on the JError::$legacy switch.
+			// @deprecated  12.1
+
+			if (JError::$legacy)
```
realityking: Since JDatabaseFactory is  completely new class do we actually need legacy error handling?

libraries/joomla/database/importer.php
```diff
@@ -0,0 +1,22 @@
...
+defined('JPATH_PLATFORM') or die;
+
+/**
+ * Joomla Platform Database Importer Class
+ *
+ * @package     Joomla.Platform
+ * @subpackage  Database
+ * @since       12.1
+ */
+abstract class JDatabaseImporter
```
realityking, line 19: What's the point of this class if there are no methods or members defined? Wouldn't an interface make more sense anyways?

LouisLandry, line 19: The point is both a common ancestor as well as a place to put common functionality as we develop it into the future.  Consider it a downpayment on a better architecture.

libraries/joomla/database/factory.php
LouisLandry: Fixed.

libraries/joomla/database/database.php
LouisLandry: Good point.  Fixed.

libraries/joomla/database/driver.php
LouisLandry: Removed.  Thanks.

PR #516, opened 2011-11-12
UT 3/10: Adding unittests for application package

Hackwar:
Fixing issues in JController and JPathway

PR #907, opened 2012-02-21
UT 1/4: Adding unittests for application package

Hackwar:
Fixing issues in JController and JPathway

This is a fixed version of pull request #516


realityking:
The ApplicationException doesn't exist anymore so we don't need a test for it.

Also JInput has move to its own package and the unit tests should go into the corresponding folder.

realityking:
Still got 3 test failures.

Hackwar:
Should be fixed now.

Hackwar:
I also changed JPathway and tested it completely now. For that I added an Inspector class for JApplicationHelper, which will ultimately be used for testing that class and several other classes, too. (f.ex. JRouter)

realityking:
Thanks!

PR #909, opened 2012-02-22
New Crypt package.

LouisLandry:
This adds a new crypt package to the Joomla Platform to replace and enhance the old JSimpleCrypt library.  JSimpleCrypt has been deprecated in favor of JCrypt.  There are currently 4 "cipers" implemented for JCrypt though any number of them could be added down the road.  It should be noted that besides the weak encryption carried over from JSimpleCrypt all of the other ciphers use the PHP mcrypt extension for cryptography.

I'm sure this thing isn't perfect, and I'm sure there will people with lots of improvements and criticisms, but this gives us an opportunity to really address encrypting/decrypting data within the platform in a way that is extensible and easy to use.


libraries/joomla/crypt/cipher/rijndael256.php
```diff
@@ -0,0 +1,118 @@
...
+	public function encrypt($data, JCryptKey $key)
+	{
+		// Validate key.
+		if ($key->type != 'rijndael256')
+		{
+			throw new InvalidArgumentException('Invalid key of type: ' . $key->type . '.  Expected rijndael256.');
+		}
+
+		// Generate an initialisation vector.
+		$iv = mcrypt_create_iv(mcrypt_get_iv_size(MCRYPT_RIJNDAEL_256, MCRYPT_MODE_ECB));
```
jonnsl: Any reason to use ECB instead of CBC?

LouisLandry: Not that I can remember off the top of my head.  I actually wrote this a couple of months ago.  Do you have a specific reason to use one or the other?

jonnsl: I'm not a security expert but I think that CBC is safer (http://www.adayinthelifeof.nl/2010/12/08/encryption-operating-modes-ecb-vs-cbc/).
and by the way, ECB don't need an initialisation vector so this line is useless

LouisLandry: Ah ... solid point.  I'll fix this stuff up.  Thanks.

LouisLandry:
OK, that is much better.  Thanks for the reminder @jonnsl

chdemko:
Tests are running successfully here

PR #911, opened 2012-02-23
Adding a default value for checkbox field

chdemko:
Checkbox field miss a default value.


elinw:
Shouldn't this be documented in the minor compatibility change list since it will potentially change behaviors of forms for people with existing checkbox fields?

PR #913, opened 2012-02-23
renew pull from #803

ghost:
by default JDatabaseMySQL::getTableColumns() second arguments is 'true'


ianmacl:
Can you elaborate more on what this solves?

chdemko:
@ianmacl see #803 and #707

LouisLandry:
@ianmacl The issue is that the exporter needs more than just the name of the field to put together an XML document describing the schema.  By default the second argument is true which means that the result will not include whether the field is nullable, what the default value is, etc.  Explicitly setting the second argument ($typeOnly) to false, we get the rest of the field information out of the DB, which allows for correct XML being generated.

ianmacl:
Fair enough.  I suppose we should go ahead and merge it then unless there are other objections.

PR #914, opened 2012-02-23
Fix the modal behavior.

realityking:
Modals now require MooTools More because with 1.3 Hash moved to More.
It may be possible to fixe Squeezebox but for this fixes the immediate issue.


realityking:
Certainly didn't expect this to be tested.

PR #918, opened 2012-02-23
Split methods.php into two files.

realityking:
It always bugged me that these two classes that belong into different packages sit in the same file.


AmyStephen:
+1

PR #919, opened 2012-02-24
Change JDatabaseDriver, JCacheStorage and JSessionStorage to use isSupported() instead of test().

realityking:
This makes this the same across all packaged and isSupported() is just much more descriptive than test(). I also changed all the tests to use dynamic access to static methods instead of call_user_func_array() since we're know in PHP 5.3 land.


robschley:
You've got some issues to fix ;)

PR #920, opened 2012-02-24
Fix a regression in the returned groups in load in the user table

elinw:
See CMS tracker #28125 fix by Mark Dexter.

### DIFF
--- a/libraries/joomla/table/user.php
+++ b/libraries/joomla/table/user.php
@@ -109,7 +109,7 @@ class JTableUser extends JTable
 			$this->_db->setQuery($query);
 
 			// Add the groups to the user data.
-			$this->groups = $this->_db->loadAssocList('id', 'id');
+			$this->groups = $this->_db->loadAssocList('title', 'id');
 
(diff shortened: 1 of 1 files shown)


realityking:
The commit that caused this is https://github.com/joomla/joomla-platform/commit/297caf113d4c7261104a7c79c06d3624ee015dc5

This was a fix for the following issue: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27046

elinw:
See also https://groups.google.com/forum/?fromgroups#!topic/joomlabugsquad/CH5apeXluNs

I really don't know what to say, I personally think this was an out of bounds change given that there are no tests and no discussion on list and no documentation,  but it's in a released version so by definition the maintainers accepted it as a valid change. 

I'm going to leave this open for someone else to decide, and suggest getting  Mark's take on it.

infograf768:
Decision in CMS has been to not change this behavior for now and just the docblock

PR #928, opened 2012-02-26
Detailed Documentation for Inline Commenting

elinw:


elinw:
These are the basic rules I have used for editing comments.

Issue #930, opened 2012-02-26
JMail addAttachment ignores parameters (filename, encoding, type) 


brandtOSS:
The addAttachment method of the JMail class does not pass on the parameters
$name, $encoding and $type to its PHPMailer parent class.
This makes it impossible to e.g. specify another file name for an attachment.

The patch: https://github.com/brandtOSS/joomla-platform/commit/4c173d20e06578ab9da32468e00da85bd1c530e2

realityking:
Closing, this is dealt with in #931.

PR #931, opened 2012-02-26
Update libraries/joomla/mail/mail.php

brandtOSS:
The addAttachment method of the JMail class does not pass on the parameters
$name, $encoding and $type to its PHPMailer parent class.
This makes it impossible to e.g. specify another file name for an attachment.


realityking:
Opening a pull request automatically opens an issue too so I closed #930.

brandtOSS:
Thanks, did not know that, sorry. I updated the issue description.

LouisLandry:
Thanks!

PR #934, opened 2012-02-27
Removed caching from module list

ghost:
[this is a re-request from the faulty: https://github.com/joomla/joomla-platform/pull/867]

Since modules can be published based on time (publish_up / publish_down), the caching on the initial published module list makes no sense.

The caching key ($cacheid) only takes into consideration the $Itemid, $groups, $clientId, $lang.
However, the cached module list is time based.

This means that the publish up/down settings have no effect as long as the list is still cached.
So that is a bug.

When looking at what is actually cached, there is not much to be gained by the caching as it is. It is only a database query and some basic checks on the returned object list.

So taking above into account (that caching causes the timed based bug and that there is little to no gain in caching), the solution is to simply take out the caching altogether.

Regarding performance:
As far as I can tell only the module list that is returned from the database is stored in that cache.
Not the rendered modules themselves.

Advanced Module Manager and MetaMod Pro have been overruling this module helper file as it is with one that (among other things) doesn't use this cache.
We have seen no noticeable effects on page load time due to this non-use of module list cache.
Thousands of people are using these extensions and we have had no reports of pageload time problems related to this.

In what I have been able to test locally, there is no noticeable difference in load time (debug info) and a small decrease in memory usage when commenting out the $cache->store line.

### DIFF
--- a/libraries/joomla/application/module/helper.php
+++ b/libraries/joomla/application/module/helper.php
@@ -298,91 +298,80 @@ abstract class JModuleHelper
 		$lang = JFactory::getLanguage()->getTag();
 		$clientId = (int) $app->getClientId();
 
-		$cache = JFactory::getCache('com_modules', '');
-		$cacheid = md5(serialize(array($Itemid, $groups, $clientId, $lang)));
+		$db = JFactory::getDbo();
 
-		if (!($clean = $cache->get($cacheid)))
+		$query = $db->getQuery(true);
+		$query->select('m.id, m.title, m.module, m.position, m.content, m.showtitle, m.params, mm.menuid');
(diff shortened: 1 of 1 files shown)


jonnsl:
you definitely need to change your text editor.

ghost:
Using PhpStorm. Nothing wrong with that.

jonnsl:
never mind, I thought you changed the indentation to spaces. my mistake.

LouisLandry:
Thanks a bunch for getting this cleaned up and in a workable state.

ghost:
ah, yes, very good! I think I accidentally removed that after trying to get a clean github fork and push for the nth time :)

PR #936, opened 2012-02-28
Fix JModuleHelper::_load() broken in #934.

realityking:
JModuleHelper::getModules() relies on the index starting at 0.

### DIFF
--- a/libraries/joomla/application/module/helper.php
+++ b/libraries/joomla/application/module/helper.php
@@ -373,6 +373,9 @@ abstract class JModuleHelper
 
 		unset($dupes);
 
+		// Return to simple indexing that matches the query order.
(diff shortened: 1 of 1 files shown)


LouisLandry:
Two things. 1. Can we move this lib to the CMS tree?  2. Can you write a small unit test to make sure we don't get another regression?

realityking:
1. JDocument depends on this (JDocumentHtml to be exact) I don't see how we could decouple it without breaking B/C.
2. My environment is beyond repair as far as db dependent unit tests go but maybe someone else can step in.

LouisLandry:
Ok, this has been simplified greatly.  Looks good.  Thanks Rouven.

PR #937, opened 2012-02-28
Do some early spring cleaning.

realityking:
This a mix of things that I didn't wanna create individual pull requests for:
-Remove the few occurrences of die(), we decided to just use die
-Add a few type hints
-Remove _a lot_ of the legacy PHP call-by-value of objects ampersands 
-Change a couple of docblocks to be more informative regarding the type of variables


robschley:
Merged. Thanks!

PR #939, opened 2012-02-29
Calendar readonly="true" status prevents loading custom calendars. 

infograf768:
date is always displayed as Gregorian.

### DIFF
--- a/libraries/joomla/html/html.php
+++ b/libraries/joomla/html/html.php
@@ -864,7 +864,7 @@ abstract class JHtml
 			$attribs = JArrayHelper::toString($attribs);
 		}
 
-		if (!$disabled)
+		if ((!$readonly) && (!$disabled))
 		{
 			// Load the calendar behavior
 			JHtml::_('behavior.calendar');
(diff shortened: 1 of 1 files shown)


realityking:
Why do you wann load a GUI calendar when it's read only?

infograf768:
Because JHtml::_('date', $value) is not used in this case and the date displayed is the db Gregorian date.
Example in CMS: edit an article. The Modified Date field will always display a Gregorian date.
In Persian for example, it should display the Jalali date.

Letting read only = "true", after the patch, displays correctly the date without the calendar icon.

(Custom calendars in J use js)
I welcome  a better solution.

realityking:
After a brief discussion with JM we figured out the issue isn't actually the JHtml::_('date', $value) call (which is done without this patch too) but that the date conversion for those other calendars actually happens in JS, so when we don't load the JS it isn't converted.

The whole system is very ugly, but I can come up with a better one. Anyone else wiser?

chdemko:
I think we can go on with that one

pasamio:
So why not convert the date in PHP and change the field behaviour to output a hidden field so that saving still works and instead display a disabled input field as usual.

pasamio:
More importantly this will introduce a bug where "read only" calendars will become writeable again.

infograf768:
Hmm, this creates issues when changing dates in the calendars when using custom calendars

PR #940, opened 2012-02-29
Fix a regression from #791.

realityking:
This required a small addition to the JInput API that should be helpful for others as well.

### DIFF
--- a/libraries/joomla/input/input.php
+++ b/libraries/joomla/input/input.php
@@ -209,6 +209,26 @@ class JInput implements Serializable
 	}
 
 	/**
+	 * Define a value. The value will only be set if there's no value for the name or if it is null.
+	 *
+	 * @param   string  $name   Name of the value to define.
+	 * @param   mixed   $value  Value to assign to the input.
+	 *
+	 * @return  void
+	 *
+	 * @since   12.1
+	 */
+	public function def($name, $value)
+	{
+		if (isset($this->data[$name]))
+		{
+			return;
+		}
+
+		$this->data[$name] = $value;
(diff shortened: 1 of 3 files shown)


libraries/joomla/input/input.php
```diff
@@ -183,15 +183,21 @@ public function getArray(array $vars, $datasource = null)
...
+	 * @param   boolean $overwrite  If true and an existing key is found, the value is overwritten, otherwise it is ignored.
 	 *
 	 * @return  void
 	 *
 	 * @since   11.1
 	 */
-	public function set($name, $value)
+	public function set($name, $value, $overwrite = true)
 	{
+		if (!$overwrite && array_key_exists($name, $this->data))
```
LouisLandry: isset() is substantially faster than array_key_exists() ... and this function may get called a few times in an execution cycle.  Might consider changing it.

LouisLandry:
Convention-wise we've always had ->def() for this behavior.  Not that I think that's necessarily better, but it is worth noting.

Also, it would be great to add a couple of small unit tests just to make sure the functionality doesn't get lost somewhere down the line.

realityking:
Either works for me. I followed how JRequest worked but I'm open to change it if we agree on def() as a convention.

I'll whip up some unit tests when I'm back home.

realityking:
Either works for me. I followed how JRequest worked but I'm open to change it if we agree on def() as a convention.

I'll whip up some unit tests when I'm back home.

robschley:
I like def() better than having another parameter in set. For what it's worth.

PR #941, opened 2012-02-29
Adding missing unittest stubs for the platform

Hackwar:
This pull request adds missing unittest stubs to the test suite. Hopefully I caught all that were overlooked at some point or another. It also moves unittests from classes that have been moved to their correct position. (specifically in the database package)


tests/suite/joomla/database/JDatabaseDriverTest.php
```diff
@@ -0,0 +1,419 @@
+<?php
+
+require_once dirname(__FILE__) . '/../../../../libraries/joomla/database/driver.php';
```
realityking: Could you change these to be paths off JPATH_PLATFORM. Else this is maintenance heel when we move files around.

robschley:
Aren't we supposed to be trying to get the tests to comply with the coding style guide?

Hackwar:
I thought I already caught all those. Now those should be fixed.

@Rob None of the unittests are in the platform coding style. Should that be changed?

LouisLandry:
Closing this since the unit testing tree has been completely revamped.

PR #942, opened 2012-03-01
fix bug #28039

oc666:
The bug produce cause preg_match not checked for return value. 
The bug report: PHP Notice:  Undefined offset: 1
Cause it tries to find the second item in the matches ($version) array, where is declared from preg_match function before the line error.
See more here:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28039

### DIFF
--- a/libraries/joomla/environment/browser.php
+++ b/libraries/joomla/environment/browser.php
@@ -245,8 +245,7 @@ class JBrowser
 				 * but only if version is > 9.80. See: http://dev.opera.com/articles/view/opera-ua-string-changes/ */
 				if ($this->_majorVersion == 9 && $this->_minorVersion >= 80)
 				{
-					preg_match('|Version[/ ]([0-9.]+)|', $this->_agent, $version);
-					list ($this->_majorVersion, $this->_minorVersion) = explode('.', $version[1]);
+					$this->identifyBrowserVersion();
 				}
 			}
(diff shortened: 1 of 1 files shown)


libraries/joomla/environment/browser.php
```diff
@@ -434,6 +430,25 @@ public function getPlatform()
 	}
 
 	/**
+	 * Set browser version, not by engine version
+	 * Fallback to use when no other method identify the engine version
+	 * 
+	 * @return null
```
chdemko: Replace null by void

libraries/joomla/environment/browser.php
```diff
@@ -434,6 +430,25 @@ public function getPlatform()
...
+	 */
```
chdemko, line 437: Add a @ since 12.1

libraries/joomla/environment/browser.php
```diff
@@ -434,6 +430,25 @@ public function getPlatform()
...
+		{
+			list ($this->_majorVersion, $this->_minorVersion) = explode('.', $version[1]);
+			return;
+		}
+		// Can't identify browser version
+		$this->_majorVersion = 0;
+		$this->_minorVersion = 0;
+		JLog::add("Can't identify browser version. Agent: " . $this->_agent, JLog::NOTICE);
+	}
+
```
chdemko: Remove extra tab

PR #948, opened 2012-03-01
Make doc block match what is actually present.

elinw:
There was a change made to this library that made this doc block description incorrect. Because user group name is not required to be unique it can't be used as a key in an array returned by loadAssocList.  See CMS tracker #28125 (corresponding change already committed).


elinw:
For a more detailed discussion see https://groups.google.com/d/topic/joomlabugsquad/CH5apeXluNs/discussion

If you think a different solution would be better please let us know.

eddieajau:
Thanks.

elinw:
Wow fastest merge ever :P

PR #952, opened 2012-03-02
public access to JDatabaseDriver::getDatabase

ChristianEhlscheid:
see: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28116

### DIFF
--- a/libraries/joomla/database/driver.php
+++ b/libraries/joomla/database/driver.php
@@ -508,13 +508,13 @@ abstract class JDatabaseDriver extends JDatabase implements JDatabaseInterface
 	}
 
 	/**
-	 * Gets the name of the database used by this conneciton.
+	 * Gets the name of the database used by this connection.
 	 *
 	 * @return  string
 	 *
 	 * @since   11.4
 	 */
-	protected function getDatabase()
+	public function getDatabase()
 	{
 		return $this->_database;
 	}


eddieajau:
Can you give me an example of the use case? The method was included specifically to allow for compatibility in the MSSQL driver. It wasn't intended to be used outside the driver class.  That said, if there's a good reason to change it, it can be.

ChristianEhlscheid:
Hi,

I want to use it in a component installer. e.g

<pre>
class com_mycomponentInstallerScript {

    private $db = null;

    function __construct() {
        $this->db = JFactory::getDbo();
    }

    // i know thats MYSQL specific ... but since the component will only ever run on MySQL anyway ...
    // the database name has to be specified in the where clause, otherwise false posivites could be reported if the MYSQL user has access to several 
    // databases and the index already exists in one of these ...
    private function dbIndexExists($table, $index) {
        $table = $this->db->getPrefix() . $table;
        $dbname = $this->db->getDatabase();
        $query = 'SELECT COUNT(*) FROM INFORMATION_SCHEMA.STATISTICS WHERE ';
        $query .= 'table_schema = ' . $this->db->quote(dbname) . ' AND ';
        $query .= 'table_name = ' . $this->db->quote($table) . ' AND index_name = ' . $this->db->quote($index);
        $this->db->setQuery($query);
        $count = $this->db->loadResult();
        return $count > 0;
    }

   // update function for component ..
    function update( $parent ) {
      if (!$this->dbIndexExists('sometable', 'someindex')) {
        // create the index ...
      }   
    }
}
</pre>

It's really a minor issue.
Ideally such functions (Index detection) should exist on the specific database driver classes, but since they aren't there i have to code my own in the meantime ...

pasamio:
Instead of making this change, I'd put in the change to get something like that added to the main JDatabaseDriver API and provide the MySQL implementation you've got as the pull request. Both are going to become available around the same time and the later option provides more functionality to everyone.

eddieajau:
Just a thought - I think you can use getTableKeys instead and modify the loadObjectList call to:

```
$keys = $this->loadObjectList('Key_name');
```

and change the docblock to:

```
@return  array  An associative array of the index specifications for the table in the form, keyed on the index name.
```

Then you could do something like this in your component:

   private function dbIndexExists($table, $index) {
   {
        return array_key_exists($index, $this->db->getTableKeys($table)))
    }

The only other driver that implements this is Oracle so we'll have to work out how to get the index name for it.

How does that sound?

pasamio:
Closing this pull request due to lack of response.

PR #953, opened 2012-03-02
Use language pack name as 'element' name instead of language pack

Chraneco:
At the moment the installer adapter for languages uses the language tag as 'element' name when storing a language pack in #__extensions.

This means that language packs for extensions (non-core language packs) cannot be installed properly like it was in Joomla! 1.5 as this will override the database entry for the core language pack in #__extensions.

With this patch the language pack name will be used as 'element' name, so this problem cannot occur anymore.

### DIFF
--- a/libraries/joomla/installer/adapters/language.php
+++ b/libraries/joomla/installer/adapters/language.php
@@ -101,6 +101,7 @@ class JInstallerLanguage extends JAdapterInstance
 		// Set the extensions name
 		$name = JFilterInput::getInstance()->clean((string) $this->manifest->name, 'cmd');
 		$this->set('name', $name);
+		$this->set('element', $name);
 
 		// Get the Language tag [ISO tag, eg. en-GB]
 		$tag = (string) $this->manifest->tag;
@@ -246,7 +247,7 @@ class JInstallerLanguage extends JAdapterInstance
 		$row = JTable::getInstance('extension');
 		$row->set('name', $this->get('name'));
 		$row->set('type', 'language');
-		$row->set('element', $this->get('tag'));
+		$row->set('element', $this->get('element'));
 
 		// There is no folder for languages
 		$row->set('folder', '');
@@ -266,7 +267,7 @@ class JInstallerLanguage extends JAdapterInstance
 
 		// Clobber any possible pending updates
 		$update = JTable::getInstance('update');
-		$uid = $update->find(array('element' => $this->get('tag'), 'type' => 'language', 'client_id' => '', 'folder' => ''));
+		$uid = $update->find(array('element' => $this->get('element'), 'type' => 'language', 'client_id' => '', 'folder' => ''));
 		if ($uid)
 		{
 			$update->delete($uid);
@@ -305,6 +306,7 @@ class JInstallerLanguage extends JAdapterInstance
 		$name = (string) $this->manifest->name;
 		$name = JFilterInput::getInstance()->clean($name, 'cmd');
 		$this->set('name', $name);
+		$this->set('element', $name);
 
 		// Get the Language tag [ISO tag, eg. en-GB]
 		$tag = (string) $xml->tag;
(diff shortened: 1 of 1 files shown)


pasamio:
Why use the name? Why not use a packagename tag or something like that?  Name, at least for me in 2.5, is "English (United Kingdom)" and the Urdu pack I grabbed was "Urdu (Pakistan)". This would make the element set to something completely different to "en-GB" or "ur-PK".

Have you tested uninstalling a partial language pack in this way as well? Or even a full language pack? Given the tag name is used for uninstalling the extension, changing this setting would break uninstalling any language pack near as I can tell from a simple code review (again "en-GB" != "English (United Kingdom)").

infograf768:
Installing extensions specific lang files is done in 2.5 using the type="file"
The language adapter in installer library contains some obsolete code which has to be deleted.

Chraneco:
> This would make the element set to something completely different to "en-GB" or "ur-PK".

Yes, this is intended. It is required that something else than the langage tag is used so that different extension language packs for the same language can be distinguished.

> changing this setting would break uninstalling any language pack near as I can tell from a simple code review

I know that this would break backward compatibility, but as far as I know this is allowed for Joomla! 3.0.

> Installing extensions specific lang files is done in 2.5 using the type="file"

Is this documented somewhere? According to the code in the languages installer, it should work the way like it was in Joomla! 1.5.
(So that the installer can notify the user when he tries to install an extension specific language pack for a language for which the core pack isn't installed)

pasamio:
Which one of these doesn't fit:

```
article
atomic
banip
beez5
beez_20
bluestork
cache
canonicalization
categories
codemirror
com_admin
com_banners
com_cache
com_categories
com_checkin
com_config
com_contact
com_content
com_cpanel
com_finder
com_installer
com_languages
com_login
com_mailto
com_media
com_menus
com_messages
com_modules
com_newsfeeds
com_plugins
com_redirect
com_search
com_ssomanager
com_templates
com_users
com_weblinks
com_wrapper
contactcreator
contacts
content
debug
emailcloak
English (United Kingdom)
extensionupdate
finder
geshi
gmail
hathor
highlight
http
image
ip
joomla
joomlaupdate
languagecode
languagefilter
ldap
lib_gforge
lib_jauthtools_openid
lib_jauthtools_sso
lib_jauthtools_token
lib_jauthtools_usersource
loadmodule
log
logout
mod_articles_archive
mod_articles_categories
mod_articles_category
mod_articles_latest
mod_articles_news
mod_articles_popular
mod_banners
mod_breadcrumbs
mod_custom
mod_feed
mod_finder
mod_footer
mod_languages
mod_latest
mod_logged
mod_login
mod_menu
mod_multilangstatus
mod_popular
mod_quickicon
mod_random_image
mod_related_items
mod_search
mod_sso
mod_stats
mod_status
mod_submenu
mod_syndicate
mod_title
mod_toolbar
mod_users_latest
mod_version
mod_weblinks
mod_whosonline
mod_wrapper
newsfeeds
none
openid
p3p
pagebreak
pagenavigation
phpmailer
phputf8
pkg_joomla
profile
readmore
recaptcha
redirect
remember
sef
session
Urdu (Pakistan)
simplepie
simplesso
sso
sync
tinymce
tokenlogin
typec
vote
weblinks
```

Element is not supposed to be a "name", it's supposed to be a uniquely identifiable element without funky characters. At the moment that is the language code. If you intend on changing this you need to create something similar.

Additionally, you still don't address the issue that language uninstall is completely broken by this pull request. Have you tested this bit of code?

infograf768:
"According to the code in the languages installer, it should work the way like it was in Joomla! 1.5."

This is what I meant by obsolete code in the adapter:

// Either we are installing a core pack or a core pack must exist for the language we are installing.
        if (!$this->_core)
        {
            if (!JFile::exists($this->parent->getPath('extension_site') . '/' . $this->get('tag') . '.xml'))
            {
                $this->parent
                    ->abort(JText::sprintf('JLIB_INSTALLER_ABORT', JText::sprintf('JLIB_INSTALLER_ERROR_NO_CORE_LANGUAGE', $this->get('tag'))));
                return false;
            }
        }

realityking:
The code mentioned by JM was now removed in #1386. So I'm closing this one.

infograf768:
For info on non-core language packs, see this recent doc:
http://docs.joomla.org/Making_non-core_language_packs_in_2.5

PR #954, opened 2012-03-02
Added support for custom column names

tampe125:
So, here we are again.
Since my previous pull request wasn't mergeable, i've created another one.

I've changed the JTable class so we can have custom "special" names, such as "ordering", "hits" and so on.

### DIFF
--- a/libraries/joomla/table/table.php
+++ b/libraries/joomla/table/table.php
@@ -39,6 +39,13 @@ abstract class JTable extends JObject
 	protected $_tbl_key = '';
 
 	/**
+	* Array with alias for "special" columns such as ordering, hits etc etc
+	*
+	* @var    array
+	*/
+	protected $columnAlias = array();
+
+	/**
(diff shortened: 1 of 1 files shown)


realityking:
I understand the intention but I think we should try to standardise these things more instead of making them configureable. Convention over configuration.

tampe125:
Mhm... no.

I think that a framework should be very flexible, there should be some
conventions, but they should be used only when there are no configurations.
I think Zend Framework won't force you with conventions: if there are no
options set, it will use the default ones (ie conventions).

Custom columns name have been added on FrameworkOnFramework (a Nicholas
project where I gave some contributions) and everything is working like a
charm.
It solved some problems when we both had to convert some projects.
I actually think this is a resource, not a problem.
Il giorno 19/lug/2012 01:42, "Rouven Weßling" <
reply@reply.github.com>
ha scritto:

realityking:
We currently have no interest to do this so I'm closing here.

PR #958, opened 2012-03-03
Added capability to use differet glue in WHERE clause.

gpongelli:
First where() call doesn't need glue, the other calls can have this set and it's referred to actual condition.
This change permit to call a sequence like this:
<code>$q->where('a = 1')->where('b = 1', 'OR')->where('c = 1', 'AND');</code>
that will be translated in
<code>WHERE a = 1 OR b = 1 AND c = 1</code>

It's now possible to have more complex $condition as following
<code>$q->where('a = 1')->where('(b = 1 OR c = 4)')->where('c = 7', 'OR');</code>
that is translated to
<code>WHERE a = 1 AND (b = 1 OR c = 4) OR c = 7</code>

### DIFF
--- a/libraries/joomla/database/query.php
+++ b/libraries/joomla/database/query.php
@@ -1444,35 +1444,31 @@ abstract class JDatabaseQuery
 	 * Usage:
 	 * $query->where('a = 1')->where('b = 2');
 	 * $query->where(array('a = 1', 'b = 2'));
+	 * $query->where('a = 1')->where('b = 2', 'OR');
 	 *
 	 * @param   mixed   $conditions  A string or array of where conditions.
-	 * @param   string  $glue        The glue by which to join the conditions. Defaults to empty.
-	 *                               Note that the glue isn't set on first use and can be changed
+	 * @param   string  $glue        The glue by which to join the conditions. Defaults to AND glue.
+	 *                               Note that the glue isn't set on first call and can be changed
 	 *                               for each call.
 	 *
 	 * @return  JDatabaseQuery  Returns this object to allow chaining.
 	 *
 	 * @since   11.1
 	 */
-	public function where($conditions, $glue = '')
+	public function where($conditions, $glue = 'AND')
 	{
+		if (is_array($conditions))
+		{
+			$conditions = implode(' ' . strtoupper($glue) . ' ', $conditions);
+		}
+
 		if (is_null($this->where))
 		{
-			$glue = strtoupper($glue);
-			$this->where = new JDatabaseQueryElement('WHERE', $conditions, " $glue ");
+			$this->where = new JDatabaseQueryElement('WHERE', $conditions, ' ');
 		}
 		else
 		{
-			if (is_array($conditions))
-			{
-				$im = strtoupper($glue) . ' ';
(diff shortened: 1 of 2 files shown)


gpongelli:
See #956 .

gpongelli:
Running entire test suite on this branch and on master one I receive same result: 
Tests: 2905, Assertions: 4024, Failures: 5, Errors: 139, Incomplete: 554, Skipped: 135.
So these changes doesn't add failures or errors.

Eng. Gabriele Pongelli.

AmyStephen:
@gpongelli - 

Appreciate your efforts on this. Unfortunately, even with system tests that cover an intended use, it still is possible for developers to unknowingly create ambiguous where clauses. 

This approach unfortunately has this type of problem:
https://github.com/joomla/joomla-platform/pull/956#issuecomment-4302563

Maybe give some thought to creating a grouping mechanism? 

$q->where('begin')->where('a = 1')->where('b = 1', 'OR')->q->where('end')->where('c = 1', 'AND');

could translate to:

WHERE (a = 1 OR b = 1) AND c = 1

That type of approach could be useful and produce predictable results.

An idea, anyway.

Thanks, again, for all of your work.

Updated note: even with this type of approach, a developer could still use the glue incorrectly to create ambiguous where clauses. But, there would at least be a way to create logical groupings correctly. That would put a solution at par with working with native SQL in that it could be used correctly - or incorrectly. To me, that would be satisfactory.

gpongelli:
As I told you before, again, I think that it's simpler to write
$q->where('a = 1')->where('(b = 1 AND c = 1)', 'OR')
to obtain what you want instead of add what you're thinking of with these "grouping mechanism" that, in my opinion, complicate uselessly platform code and create unnecessary rules for 3pd to be used correctly.
Correct grouping mechanism is the use of parenthesis inside WHERE condition, and everyone who wrote a web application need to know himself/herself how create an SQL query.

That type of problem is solved by designing and writing correctly any SQL query.
For me, you're creating a problem that doesn't exist.

Actually there's not a really need for this patch, because you can always write
$q->where('a = 1 OR (b = 1 AND c = 1)')
it works, it's portable to each driver, it's faster, it's the simplest solution.

Eng. Gabriele Pongelli.

AmyStephen:
@gpongelli said: 

"Actually there's not a really need for this patch, because you can always write
$q->where('a = 1 OR (b = 1 AND c = 1)')
it works, it's portable to each driver, it's faster, it's the simplest solution."

We are in agreement, then. That's what I think too. https://github.com/joomla/joomla-platform/pull/956#issuecomment-4306363

eddieajau:
I'm going to close this pull request so we are just evaluating on solution, and 956 got in first. Amy's right that we need to limit the ambiguity, but there is merit in finding an API to do that.

gpongelli:
I know there's not good solution to limit the ambiguity yet.
I thought that, in case of similar pull request, you look at that solve issue in a cleaner way, not choosing with "who's the first".

My 2 cents,
Eng. Gabriele Pongelli

eddieajau:
No, we don't choose features on who was first - but I don't want to be talking about the same issues in two places, that's all :)  I think your solution is good but let's see if there are other ideas to make it better.

PR #960, opened 2012-03-04
Wrong usage of "/" instead of "DIRECTORY_SEPARATOR"

Naouak:
It was creating malformed path for Windows OS.

### DIFF
--- a/libraries/joomla/installer/adapters/library.php
+++ b/libraries/joomla/installer/adapters/library.php
@@ -117,7 +117,7 @@ class JInstallerLibrary extends JAdapterInstance
 		}
 		else
 		{
-			$this->parent->setPath('extension_root', JPATH_PLATFORM . '/' . implode(DIRECTORY_SEPARATOR, explode('/', $group)));
+			$this->parent->setPath('extension_root', JPATH_PLATFORM . DIRECTORY_SEPARATOR . implode(DIRECTORY_SEPARATOR, explode('/', $group)));
 		}
 
 		/*


realityking:
Does the path actually break or does it just look weird? PHP should automatically change / to \ when accessing the filesystem on Windows.

Naouak:
Then why bother doing the explode/implode thing on the right side of the line ?

eitheir it use DIRECTORY_SEPARATOR everywhere or it use only "/".

realityking:
That's a valid question and I suspect it's not needed. Since I don't have a windows environment set up I can't test things like this but it would be great if someone did.

Naouak:
PHP will treat "/" as correct for path. It won't generate any errors but you for code consistency, it would be better to either use only "/" or DIRECTORY_SEPARATOR.

elinw:
We've really moved to '/' but most likely there are still some residual workarounds from when it couldn't be used.

realityking:
Since there seems to be no issues I'm going to close this. It'd be great if someone could confirm whether that workaround is needed or not.

ianmacl:
My hunch is the line should be:
$this->parent->setPath('extension_root', JPATH_PLATFORM . '/' . implode('/', explode(DIRECTORY_SEPARATOR, $group)));

What we have now is, as Naouak correctly pointed out, is inconsistent.  This isn't causing issues but should perhaps be looked at for consistency sake.

realityking:
Why even explode it?

My hunch would be that this would work just fine:

```
$this->parent->setPath('extension_root', JPATH_PLATFORM . '/' . $group);
```

If you output the path at some point it may not be pretty on windows, but I'm not sure if JPATH_PLATFORM is either.

ianmacl:
Well considering no bugs have been reported in actual use, that's probably safe to do.  The problems that stem from directory separator issues happen when you try to do string comparison on paths.

PR #917, opened 2012-02-23
PostgreSQL driver for Joomla! platform [clean version].

gpongelli:
This driver implements inherited function from JDatabaseDriver class and add some other useful function to be able to use Joomla! on PostgreSQL database.
There is also a class that inherits from JDatabaseQuery to be able to create query object using PostgreSQL dialect.
These two classes are tested by other two classes under "test" folder.

PostgreSQL's added functions:
- getAlterDbCharacterSet, set database encoding
- getCreateDbQuery, returns a query string to create a database using $option array members
- getRandom, get a random number
- getStringPositionSQL, returns string's position inside another string
- getTableSequences, returns an array of table's sequences information
- releaseTransactionSavepoint, release given savepoint during transaction
- showTables, lists all table in database
- transactionSavepoint, creates transaction savepoint

Overridden functions:
- connected
- dropTable
- escape
- execute
- fetchArray
- fetchAssoc
- fetchObject
- freeResult
- getAffectedRows
- getCollation
- getNumRows
- getQuery
- getTableColumns
- getTableCreate
- getTableKeys
- getTableList
- getVersion
- insertid
- insertObject
- lockTable
- renameTable
- replacePrefix
- select
- setUTF
- transactionCommit
- transactionRollback
- transactionStart
- unlockTables
- updateObject

PostgreSQL database query added functions:
- limit, a possible replace for limit in setQuery
- offset, a possible replace for limit in setQuery
- forShare, lock table/row during SELECT
- forUpdate, lock table/row during SELECT
- getInsertTable, used in "indertid()" to retrieve last INSERT INTO's table name
- noWait, no wait a locked table
- returning, an INSERT INTO optional clause to returns last insert id

Added also Postgresql's exporter and importer classes and their test classes.

PR #961, opened 2012-03-04
PostgreSQL driver for Joomla! platform [clean version _ v2].

gpongelli:
This driver implements inherited function from JDatabaseDriver class and add some other useful function to be able to use Joomla! on PostgreSQL database.
There is also a class that inherits from JDatabaseQuery to be able to create query object using PostgreSQL dialect.
These two classes are tested by other two classes under "test" folder.

PostgreSQL's added functions:
- getAlterDbCharacterSet, set database encoding
- getCreateDbQuery, returns a query string to create a database using $option array members
- getRandom, get a random number
- getStringPositionSQL, returns string's position inside another string
- getTableSequences, returns an array of table's sequences information
- releaseTransactionSavepoint, release given savepoint during transaction
- showTables, lists all table in database
- transactionSavepoint, creates transaction savepoint

Overridden functions:
- connected
- dropTable
- escape
- execute
- fetchArray
- fetchAssoc
- fetchObject
- freeResult
- getAffectedRows
- getCollation
- getNumRows
- getQuery
- getTableColumns
- getTableCreate
- getTableKeys
- getTableList
- getVersion
- insertid
- insertObject
- lockTable
- renameTable
- replacePrefix
- select
- setUTF
- transactionCommit
- transactionRollback
- transactionStart
- unlockTables
- updateObject

PostgreSQL database query added functions:
- limit, a possible replace for limit in setQuery
- offset, a possible replace for limit in setQuery
- forShare, lock table/row during SELECT
- forUpdate, lock table/row during SELECT
- getInsertTable, used in "indertid()" to retrieve last INSERT INTO's table name
- noWait, no wait a locked table
- returning, an INSERT INTO optional clause to returns last insert id

Added also Postgresql's exporter and importer classes and their test classes.


gpongelli:
This branch should be easily merged inside master, instead of #917 (and it's parent #532) that could be closed without merging.

gpongelli:
any log ?

sfeigl:
Probably caused by still using mysql for the postgresql tests because config.php still has mysql as database driver selected.

Before changing test config to postgresql driver, I got exactly the same error / failure counts. 

It would probably correct and consistent to use a separate config file like config_postgresql.php for postgresql tests and skip these tests when it is not present - like it is done for the other DBMS.

If you still need logs, feel free to request them.

gpongelli:
you're right, but as you can see from [this commit of previous pull 917](https://github.com/gpongelli/joomla-platform/commit/23e0e1678554dca495200d462ec3644fc1ffed0f) I've removed configuration file because someone told me inside mailing list to do this.

Which are your test result after changing configuration to postgresql driver ?
Which other database driver skip tests if it's not present its configuration file ?

Thank you,
Eng. Gabriele Pongelli.

sfeigl:
Ok as I see now, that recently changed. On Thursday that was the case for mysql drivers. Now the testing architecture radically changed and you will probably have to adapt to the current head. Now, the DBMS specific config files seem to be gone. - at least I cannot find them at the moment.

Running the junit tests on the checkout of Thursday, most errors are invalid date errors. Probably by not using "null date" but a hardcoded 0000-00-00 00:00:00. I might have a more detailed look at this when I have more spare time

gpongelli:
That hardcoded date were put inside <code>joomla-platform/tests/suite/joomla/access/stubs/S01.xml</code> .

Thank you,
Eng. Gabriele Pongelli.

LouisLandry:
Hi Gabriele,

I spent some time last night and rebased your branch over master, along with squashing all of your commits into one.  In doing this I also moved things around quite a bit to line up with the current test scaffolding.  We still have work to do generally on making the joomla testing tree more robust and such, but that is a much larger task.

You can see my work here: https://github.com/LouisLandry/joomla-platform/compare/merging

The one thing I didn't do was really test the PostgreSQL driver itself.  I didn't have a postgres instance setup and didn't wanna go through that last night.  Every other test case class passes completely though, and it is all lined up with how things are in the test tree now.  You've done a masterful job with the postgresql support... well done.  If you are OK with what I've done here, you could pull it down, make sure the driver tests are working (_You'll need to uncomment the postgresql DSN string in `phpunit.xml` in the repository root_), and submit it as a pull request.  If you want to simply use it as a reference to get everything lined up yourself that's fine as well.  I just know that you've been working on this a long time and trying to get it done against an ever moving target.  I know that can be frustrating, so I wanted to put in some effort to help you out.

We are very close.

gpongelli:
Hi Louis,
it's not really easy start a pull request and, after some time, re-do all again because base structure was changed meanwhile (I did this three times, each time for a different change), and yes, it's frustrating.
This is not a criticism directed to a person, the problem maybe could be the procedure of doing this kind of job and it's management (merging this driver before moving each driver files and now their testing files would have avoided me to send two extra pull request), but I don't have the skill to discuss about management of project like this.
Sorry for this "extra" and off topic part.

Thank you for this job, I'll pull it and I'll do again tests, maybe I'll add importer and exporter test files for older database version (sublassing actual class and change mock's "getVersion" inside setUp). Query class doesn't change between versions and driver class can be tested on older version only running it.
I'll do this next days.

Have you solved problem with <code>joomla-platform/tests/suite/joomla/access/stubs/S01.xml</code> file and hardcoded null value ?

Eng. Gabriele Pongelli.

LouisLandry:
Yes, I know it's frustrating.  You had the unfortunate situation of coming in with this at a time when the database package was being re-thought/re-built and also the unit tests being reworked.  Interestingly, the entire year before that part of the codebase was very stable and would have been easier to do probably.  I don't know.  Ultimately it is what it is, and hopefully we are in a better place moving forward.

I'm not aware of a problem with the access package stub off the top of my head, but all the tests are running and passing for me now.

gpongelli:
I'm working on a branch coming from yours, test on Postgresql driver returns me OK .
I've found only driver test class while I've created library's classes and test classes for database query, for importer and exporter classes too, where are they ?

Eng. Gabriele Pongelli .

LouisLandry:
Everything except the driver tests themselves (which actually connect to pgsql/mysql/etc) are in the main test suite: `tests/suites/unit/joomla/database/database/`.

The driver test itself is at: `tests/suites/database/driver/postgresql/JDatabasePostgresqlTest.php`.

LouisLandry:
OK, merged the other one.  Thanks a ton.

PR #962, opened 2012-03-05
Fixes to the SQLite driver.

LouisLandry:
Setting the SQLite driver to use backticks for name quoting while we figure out a more robust way of handling this cross platform.  Also fixing a few issues with the SQLite driver around escaping and getting a list of table names.


robschley:
Merged. Thanks!

PR #963, opened 2012-03-05
Uncouple the Filesystem Path Class from JUser. 

AmyStephen:
A temporary file name is created using a random value in order to test if the script owns the path. Currently, this process invokes a method from JUser to build a random value. The patch changes that process to instead use mt_rand() so as to uncouple the Path Class from JUser.


LouisLandry:
Thanks Amy.  Good catch!

PR #966, opened 2012-03-07
JError Warnings and Exceptions to JLog::add Statements

AmyStephen:
I'm not sure what I did - but, I'm going to make this pull request to get feedback.

In some cases, the lines are not properly indented. I used phpStorm's reformatting command in the JControllerAdmin -- hoping that might correct things, but it ddin't.

It also looks like Rob Schley's commit is reversed. https://github.com/AmyStephen/joomla-platform/commit/7ea7f52bdbf510bf0282c6e3e12a399bd90e8069#diff-0 

I've never had line formatting issues before so I think I need advise on what I need to do.

Sorry.

### DIFF
--- a/build/phpcs/Joomla/Sniffs/Commenting/SingleCommentSniff.php
+++ b/build/phpcs/Joomla/Sniffs/Commenting/SingleCommentSniff.php
@@ -51,7 +51,7 @@ class Joomla_Sniffs_Commenting_SingleCommentSniff implements PHP_CodeSniffer_Sni
 		* The exception to this is if the preceding line consists of a single open bracket.
 		*/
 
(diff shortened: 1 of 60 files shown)


eddieajau:
I'm not quite sure what you did either. There seems to be a lot of check style errors introduced. I think it might be easier to start with a new branch. Not sure why JControllerAdmin is showing such a dramatic change.

AmyStephen:
I know why JControllerAdmin is showing the change -- I used the "Format" option to see if it would clear the problem up for that specific file. It changed all the lines (which is odd -- typically, it would only correct lines not properly formatted).

The one thing I did differently than my usual approach is that I used the "Export to Mac" option after I forked. 

I guess I'll start over, try a line, and see.

Thanks.

PR #968, opened 2012-03-07
JError Warning and Notice to JLog

AmyStephen:


libraries/legacy/application/application.php
```diff
@@ -271,7 +270,7 @@ public function route()
 
 		foreach ($result as $key => $value)
 		{
-			$this->input->def($key, $value);
+			$this->input->set($key, $value);
```
realityking: You'd be undoing a recent fix with this, please use def().

libraries/joomla/session/storage.php
```diff
@@ -65,7 +65,7 @@ public static function getInstance($name = 'none', $options = array())
 				}
 				else
 				{
-					// No call to JError::raiseError here, as it tries to close the non-existing session
+					// Attempt to close the non-existing session
 					jexit('Unable to load session storage class: ' . $name);
```
realityking: Should we throw an exception here?

AmyStephen:
I'm going to redo this since those idea files got in there.

Rouven's changes will be in the new PR.

PR #970, opened 2012-03-08
JError Exception and Warning to JLog

AmyStephen:
Replace previous PR which had idea files.


libraries/joomla/application/module/helper.php
```diff
@@ -334,7 +334,7 @@ protected static function &_load()
 
 		if ($db->getErrorNum())
```
eddieajau, line 335: This if block can be removed because the driver will throw an exception before it gets to it.

libraries/joomla/database/driver/pdo.php
```diff
@@ -370,22 +348,8 @@ public function execute()
...
-					throw new Exception('JDatabaseDriverPDO::query: ' . $this->errorNum . ' - ' . $this->errorMsg, 500);
-				}
-				return false;
-			}
-			else
-			{
-				JLog::add(JText::sprintf('JLIB_DATABASE_QUERY_FAILED', $this->errorNum, $this->errorMsg), JLog::ERROR, 'database');
-				throw new RuntimeException($this->errorMsg, $this->errorNum);
-			}
+			JLog::add(JText::sprintf('JLIB_DATABASE_QUERY_FAILED', $this->errorNum, $this->errorMsg), JLog::ERROR, 'database');
```
eddieajau: This call to JLog isn't in the other drivers so let's remove it.

Strike that - it's all over the place so let's _include_ it in all the drivers in the same location.  Thanks.

libraries/joomla/form/fields/componentlayout.php
```diff
@@ -118,7 +118,7 @@ protected function getInput()
 			// Check for a database error.
```
eddieajau, line 118: Any of these checks on $db->getErrorNum() can now be dropped because an exception is already thrown.

eddieajau:
A fews little things in there to look at but otherwise a nice job.

libraries/joomla/database/driver/pdo.php
AmyStephen: Hm. Looks like I accidentally removed the RuntimeException.

libraries/joomla/form/fields/componentlayout.php
AmyStephen, line 118: Andrew - 

This makes me wonder. Really, this type of fix probably should have gone with the "throw error" work. As I search the core code fro $db->getErrorNum, some are associated with JError Exception/Warning and other times, not.

As an example, JFactory has the $db->getErrorNum IF statement in the createDbo method, but there is no JError::Warning statement at that location. If $db->getErrorNum() has no bearing due to the throw error inside the database class, what should be happening at that location instead? Is it possible that since there is no catch processing, this might explain why Joomla gets that looping when the database connection doesn't succeed? It is also appears in ModelList, Database, etc., without the JError.

I can certainly remove it but -- stepping back, I wonder if there is a bigger issue here? How and where should catch processing take place? It might make sense to some recommendations/standards in place and then to walk through the code and make certain error handling consistently treated. Right now, I suspect you would really like catch processing where these JError statements were. If we remove them, then we also lose that marker.

Should we get this PR in, and then follow with a catch processing PR?

AmyStephen:
@LouisLandry - 

@ElinWaring indicated the two of you are already working on this. https://groups.google.com/d/msg/joomla-dev-platform/eH1E0vFbrUo/LtQNa9VLmmoJ

Can you use this work? Or, should I just close this PR?

Thanks!

libraries/joomla/database/driver/pdo.php
robschley: The exception should be kept and the log entry should be removed.

AmyStephen: Definitely. That was a mistake I made. Just waiting to see if Louis is already working on this.

Thanks Rob.

AmyStephen: OK, what I removed was the JError side, which is okay.

What remains still has a RuntimeException:

if (!is_object($this->connection))
{
    JLog::add(JText::sprintf('JLIB_DATABASE_QUERY_FAILED', $this->errorNum, $this->errorMsg), JLog::ERROR, 'database');
    throw new RuntimeException($this->errorMsg, $this->errorNum);
}

@robschley are you saying that log entry should be removed?

AmyStephen: @eddiejau - 

The execute method only appears in three of the drivers (mysql, pdo, and sqlsrv.) I changed two (pdo and sqlsrv) removing the JError side of the logic -- and now it matches the MySQL method. (which has both the log and the throw error, as I posted above for Rob.)

All other drivers extend one of these three and therefore have the same code in the execute method they use.

Unless I am missing something, I think this is consistent for every driver (after this PR).

eddieajau: @AmyStephen yes, remove to the JLog message.

libraries/joomla/form/fields/componentlayout.php
eddieajau, line 118: I agree, let's not worry about the getErrorNum stuff in this pull.  As for catching, it's the responsibility of the developer to add the try-catch.  It's just something you need to learn to handle when working with an OOP framework.

AmyStephen:
In case there is a question, this PR should be ready to use. If there are other issues, please let me know. When the request is either processed or closed, I can use the fork to help with https://github.com/joomla/joomla-platform/pull/973#issuecomment-4443039 

Appreciate an update, either way.

libraries/joomla/application/component/modeladmin.php
```diff
@@ -1063,7 +1063,7 @@ public function saveorder($pks = null, $order = null)
 
 		if (empty($pks))
 		{
-			return JError::raiseWarning(500, JText::_($this->text_prefix . '_ERROR_NO_ITEMS_SELECTED'));
```
eddieajau, line 1066: Amy, I just noticed we are turning warnings and notices in a few places.  Can you revert those cases and I'll work out how to fix them.  Thanks.

AmyStephen, line 1066: @eddieajau - Happily - except, I am afraid I don't know what you mean by "turning warnings and notices?" Sorry about that - but if you can explain it for me, I'll certainly try. :) 

(+ your nick is hard to type. Why not use AndrewEddie or bring back Masterchief? Do we get to vote on this? ;- ) )

eddieajau, line 1066: Sorry, I meant "returning".  See how this line is `return JError:: ...`.  So, just leave those as `return JError::raiseWarning( ... )` etc.

Can't change my username sorry ;)

AmyStephen, line 1066: Gotcha! Yup, that's not good. Will try to get those lines out of the commit a little later today, Andrew.

eddieajau, line 1066: Just a friendly follow up on how this is going. No pressure :)

AmyStephen:
I'm going to start over on this.

PR #971, opened 2012-03-08
Revamping Unit Tests

LouisLandry:
One of of my frustrations with the test suite for a while has been the necessity to spend time setting up an environment to run the simplest of tests.  Now that we have an SQLite database driver I thought it would be great to have the base test suite not require any setup or configuration to run.  You simply clone the repository and call `phpunit` to have the suite run.

To achieve this a few things had to be done:
- `tests/ddl.sql` schema file for SQLite was created
- `tests/includes/JoomlaDatabaseTestCase.php` was modified to utilize an SQLite memory database.
- To execute the URL to the JHttp test stub file must be set either as a PHP constant `JTEST_HTTP_STUB` or as an operating system environment variable `JTEST_HTTP_STUB`, otherwise those tests will be skipped.
- `tests/config.dist.php` was removed since a tests configuration isn't necessary to run the default tests.
### Testing database drivers.

With these changes database drivers have been broken out into their own path: `tests/suites/database` where tests can be written independent of the main suite.  If you look in the phpunit.xml.dist file you will find a section commented out with PHP constant definitions in it.  To enable testing of specific drivers you simply add the relevant constant with a DSN string for the database instance you want the tests to run against.  The commented out examples are the same default configuration that has been present.
### Nice side effects.

One of the nice side effects of this exercise (and going forward) is the use of a non-MySQL engine for discovering cross-database problems.  Several bugs and limitations were uncovered during the process and continuing down this path will only help in this area.


AmyStephen:
Awesome.

PR #973, opened 2012-03-08
Convert thrown exception messages to plain text in HTML and Form packages

eddieajau:
Removed the use of JText in hard thrown exception messages that are runtime errors (something the developer needs to fix).

### DIFF
--- a/libraries/joomla/html/html.php
+++ b/libraries/joomla/html/html.php
@@ -117,12 +117,12 @@ abstract class JHtml
 
 				if (!class_exists($className))
 				{
-					throw new InvalidArgumentException(JText::sprintf('JLIB_HTML_ERROR_NOTFOUNDINFILE', $className, $func), 500);
+					throw new InvalidArgumentException(sprintf('%s not found.', $className), 500);
 				}
 			}
 			else
 			{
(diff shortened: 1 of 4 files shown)


libraries/joomla/html/html.php
```diff
@@ -138,7 +138,7 @@ public static function _($key)
 		}
 		else
 		{
-			throw new InvalidArgumentException(JText::sprintf('JLIB_HTML_ERROR_NOTSUPPORTED', $className, $func), 500);
+			throw new InvalidArgumentException(JText::sprintf('%s::%s not found.', $className, $func), 500);
```
elkuku: I think you missed a JText here ;)

AmyStephen:
@eddieajau - good to see this, it answers one of my questions about the separation. If you want help doing this in other packages, I am happy to do so.

eddieajau:
Ooops. Thanks. Fixed.

@AmyStephen that would be great if you can go through the other thrown exceptions.

AmyStephen:
@eddieajau Great - will do so as soon as my fork spot is available when this pull request is either processed or closed. https://github.com/joomla/joomla-platform/pull/970

PR #977, opened 2012-03-10
Android Tablet Detection

elinw:
This pull request adds detection of most Android tablets to the platform detection in JApplicationWebClient. 
It is highly useful to detect mobile devices that are larger than smart phones because often it is preferable to render the desktop version of a site rather than the mobile version.

This makes the detection for Android more like that available for iOS devices, which already differentiate between iPhone and iPad for example. 

Detection for Android tablets is more complex than for iOS since there are many manufacturers and they each create their own UA strings. In addition, most detection methods for Android tablets work only when the default WebKit browser is used, not if users have installed a different browser. This code attempts to overcome some of these issues although detection will still not be perfect.

The basic decision rules are that, given that Android is in the UA string:
- Following Google recommendations strings that do not include Mobile are assumed to be from tablets (http://googlewebmastercentral.blogspot.com/2011/03/mo-better-to-also-detect-mobile-user.html)
- Android 3.x (Honeycomb) is only used for tablets
- Some alternative browsers (e.g. Opera, Fennec) add or allow users to add Tablet to the UA string.
- Kindle Fire's UA string when it is in non Silk/Android mode includes Mobile contrary to the Google recommendation, but it will always include Silk. Since among tablets it is fairly common it is worth including it as a specific device.

This retains the general approach of this class of keeping the methods lightweight with developers who need more detailed detection (based on screen size for example) able to extend it.


elinw:
Note that the test for opera fails because of an issue in the platform detection for Opera since the version is not next to the browser name.

elinw:
Okay the test passes now. What happened was that Opera as usual has to be different and was breaking our previous assumption that the browser name would only be present once in the UA string.

eddieajau:
Merged. Thanks.

PR #980, opened 2012-03-12
Corrected phone nubmer cleanup regex

danyaPostfactum:
Corrected regex in $cleanvalue = preg_replace('/[+. -()]/', '', $value) to fix problem with phone numbers, containing "-" symbol

### DIFF
--- a/libraries/joomla/form/rules/tel.php
+++ b/libraries/joomla/form/rules/tel.php
@@ -86,7 +86,7 @@ class JFormRuleTel extends JFormRule
 			 * 7 and 15 digits inclusive and no illegal characters (but common number separators
 			 * are allowed).
 			 */
-			$cleanvalue = preg_replace('/[+. -(\)]/', '', $value);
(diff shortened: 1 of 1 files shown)


elinw:
Could you add a test for the issue this solves please? Of if you  give me the string that passes/doesn't pass now I can do it.

danyaPostfactum:
Regex must cleanup formatted phone number by removing such allowable symbols as whitespace, dot, plus and hyphen.
So, such string as '+7 (914)-388.24.85' should be cleaned as '79143882485' and pass next regex. String containing other sybmols than hese 4 allowable symbols, shouldn't be cleanuped fully, and shouldn't pass numbers-testing regex.

The regex was broken due to unscreened hyphen.

g1smd:
GB number such as +44(0)20 3000 5555 should be cleaned to +442030005555 removing the zero, but Italian number such as +39 (0)55 678 3456 should be cleaned to +390556783456 retaining the zero. Italian numbers include the zero even when dialled from abroad.

PR #984, opened 2012-03-13
PostgreSQL driver for Joomla! platform [clean version _ v3].

gpongelli:
This driver implements inherited function from JDatabaseDriver class and add some other useful function to be able to use Joomla! on PostgreSQL database.
There is also a class that inherits from JDatabaseQuery to be able to create query object using PostgreSQL dialect.
These two classes are tested by other two classes under "test" folder.

PostgreSQL's added functions:
- getAlterDbCharacterSet, set database encoding
- getCreateDbQuery, returns a query string to create a database using $option array members
- getRandom, get a random number
- getStringPositionSQL, returns string's position inside another string
- getTableSequences, returns an array of table's sequences information
- releaseTransactionSavepoint, release given savepoint during transaction
- showTables, lists all table in database
- transactionSavepoint, creates transaction savepoint

Overridden functions:
- connected
- dropTable
- escape
- execute
- fetchArray
- fetchAssoc
- fetchObject
- freeResult
- getAffectedRows
- getCollation
- getNumRows
- getQuery
- getTableColumns
- getTableCreate
- getTableKeys
- getTableList
- getVersion
- insertid
- insertObject
- lockTable
- renameTable
- replacePrefix
- select
- setUTF
- transactionCommit
- transactionRollback
- transactionStart
- unlockTables
- updateObject

PostgreSQL database query added functions:
- limit, a possible replace for limit in setQuery
- offset, a possible replace for limit in setQuery
- forShare, lock table/row during SELECT
- forUpdate, lock table/row during SELECT
- getInsertTable, used in "indertid()" to retrieve last INSERT INTO's table name
- noWait, no wait a locked table
- returning, an INSERT INTO optional clause to returns last insert id

Added also Postgresql's exporter and importer classes and their test classes.

### DIFF
--- a/libraries/joomla/database/driver.php
+++ b/libraries/joomla/database/driver.php
@@ -972,9 +972,12 @@ abstract class JDatabaseDriver extends JDatabase implements JDatabaseInterface
 		static $cursor = null;
 
 		// Execute the query and get the result set cursor.
-		if (!($cursor = $this->execute()))
+		if ( is_null($cursor) )
 		{
-			return $this->errorNum ? null : false;
+			if (!($cursor = $this->execute()))
+			{
+				return $this->errorNum ? null : false;
+			}
 		}
 
 		// Get the next row from the result set as an object of type $class.
@@ -1005,9 +1008,12 @@ abstract class JDatabaseDriver extends JDatabase implements JDatabaseInterface
 		static $cursor = null;
 
 		// Execute the query and get the result set cursor.
-		if (!($cursor = $this->execute()))
+		if ( is_null($cursor) )
 		{
-			return $this->errorNum ? null : false;
+			if (!($cursor = $this->execute()))
+			{
+				return $this->errorNum ? null : false;
+			}
 		}
 
 		// Get the next row from the result set as an object of type $class.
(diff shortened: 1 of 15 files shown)


gpongelli:
See #961 .

gpongelli:
Louis, there are two classes for older database version now added to this branch, how can I add them now?

eddieajau:
Whoot! Nice to see this finally in the tree. Well done :)

gpongelli:
Thank you ;)

robschley:
Awesome! Glad this finally made it.

LouisLandry:
@gpongelli You'd just create a new pull request.  I'd really rather find a way to not have separate drivers for database versions if we can avoid it though.

garyjaybrooks:
Exciting!

elinw:
w00t!

gpongelli:
Thank you ;) !

Eng. Gabriele Pongelli

elkuku:
Congrats Gabriele :)

sparxashish:
Yes, It is too nice and exiciting but have queries.......
my query is same , how to create procedures , i have asked so many forums but i have no answer.......
In this discussion and i have also see some features corresponding JDatabase and it functions but where i can see procedure or function to create........

PR #986, opened 2012-03-14
JSession storage database adapter doesn't work.

LouisLandry:
Fixing up the database session handler to simply use try/catch instead of explicit checks for database connectivity.  This handles the new lazy-connecting methodology of JDatabaseDriver.


realityking:
A big :thumbsup: from me. :smile:

PR #989, opened 2012-03-15
Proposed fix for CMS tracker issue 28156 (only one template position for titled module used)

vovtz:
The first commit represents a code quality fix.

The second is a proposed fix for two tracker issues (which are in fact each other's duplicates).


LouisLandry:
Thanks!

PR #991, opened 2012-03-16
coding style

piotr-cz:
Should we fix such small things or typos in comments?


mbabker:
Yes, comments should be fixed.  The method doc blocks are what is used to generate the api.joomla.org pages, so what you see in the code is what is going there.

PR #995, opened 2012-03-17
Custom Calendar issue when read-only -new request

infograf768:
See
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28175


libraries/joomla/html/html.php
```diff
@@ -829,14 +829,17 @@ public static function calendar($value, $name, $id, $format = '%Y-%m-%d', $attri
...
-			. ($readonly ? ''
-			: self::_('image', 'system/calendar.png', JText::_('JLIB_HTML_CALENDAR'), array('class' => 'calendar', 'id' => $id . '_img'), true));
+			return '<input type="text" title="' . (0 !== (int) $value ? JHtml::_('date', $value) : '') . '" name="' . $name . '" id="' . $id
+				. '" value="' . htmlspecialchars($value, ENT_COMPAT, 'UTF-8') . '" ' . $attribs . ' />'
+				. JHtml::_('image', 'system/calendar.png', JText::_('JLIB_HTML_CALENDAR'), array('class' => 'calendar', 'id' => $id . '_img'), true);
+ 		}
+ 		else
+ 		{
+ 			return '<input type="text" title="' . (0 !== (int) $value ? JHtml::_('date', $value) : '')
+ 				. '" value="' . (0 !== (int) $value ? JHtml::_('date', $value, JFactory::getDbo()->getDateFormat()) : '') . '" ' . $attribs
```
realityking, line 839: Why is JDatabase::getDateFormat() used here? We're not storing anything in the database.

chdemko:
It's only for keeping same visual output

realityking:
Sorry, I still don't get it. Why should the visual representation depend on what database you've selected? What happens if you haven't set up a db connection? Why display a different format in the title and in the input?

realityking:
Pulling this broke jenkins due to style errors: http://build.joomla.org/job/platform/660/checkstyleResult/new/

elinw:
I'm sorry but the question of what happens if there is no database connection was never answered ... there needs to be a check for the existence of a connection.

realityking:
We have since removed the dependency on the database, there really was no reason for it.

infograf768:
we do have a new patch to solve doubled time.
See: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=28535

PR #997, opened 2012-03-18
Joomla CMS [#27526] Media manager doesn't support multi-byte characters 

oc666:
This will add support for multi-byte file name on onsafe method of file class. The default is full backward compatible, and not using multi-byte file name.

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=27526

### DIFF
--- a/libraries/joomla/filesystem/file.php
+++ b/libraries/joomla/filesystem/file.php
@@ -53,17 +53,26 @@ class JFile
 	/**
 	 * Makes file name safe to use
 	 *
-	 * @param   string  $file  The name of the file [not full path]
+	 * @param   string   $filename   The name of the file [not full path]
+	 * @param   boolean  $multibyte  Support for multibyte file-name
 	 *
 	 * @return  string  The sanitised string
 	 *
 	 * @since   11.1
 	 */
-	public static function makeSafe($file)
+	public static function makeSafe($filename, $multibyte = false)
 	{
-		$regex = array('#(\.){2,}#', '#[^A-Za-z0-9\.\_\- ]#', '#^\.#');
-
-		return preg_replace($regex, '', $file);
+		if ($multibyte)
+		{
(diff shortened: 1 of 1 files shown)


realityking:
Wouldn't it make sense to always support multibypte characters? (without the argument)

Also I wonder if phputf8 doesn't have some function that could help you.

oc666:
I left it without support by default for backward compatibility.
If you found phputf8 function that make this trick, please, let me know.

realityking:
I don't even know what your change does, I don't speak regex ;) That's why I haven't pulled it either.

libraries/joomla/filesystem/file.php
```diff
@@ -53,17 +53,26 @@ public static function stripExt($file)
...
-	public static function makeSafe($file)
+	public static function makeSafe($filename, $multibyte = false)
 	{
-		$regex = array('#(\.){2,}#', '#[^A-Za-z0-9\.\_\- ]#', '#^\.#');
-
-		return preg_replace($regex, '', $file);
+		if ($multibyte)
+		{
+			$search = "/\?%*:|\"<>#;()&;, ";
+			return str_replace(str_split($search), '_', $filename);
```
pasamio, line 68: Shouldn't we use something that is UTF-8 safe here? Perhaps a preg_replace with the u modifier to ensure we don't clobber parts of multibyte characters?

LouisLandry:
@oc666 I'd love to see this get merged.  Would you mind putting together a few test cases so that we can verify that everything works as expected?  I do think that we may have an issue with multibyte characters, but trust that you can/will test that out for us.  I'm going to close this until we get a couple of tests to ensure everything works and we don't have any regressions down the road.  **Please** re-open this when you get that sorted out.

oc666:
@LouisLandry, can you guide me to some simple guide or example for Joomla unit test?

Thanks, Ofer

PR #1000, opened 2012-03-19
Special 1000th pull request.

realityking:
To commemorate our 1000th pull request I thought I'd do something special, so here it is.


elinw:
:)

Hmm but probably PHP should be capitalized.

PR #1003, opened 2012-03-19
Remove a workaround for PHP 5.3.0 in JLanguage.

realityking:
This raises the minimum requirement to 5.3.1.

Even Ubuntut 8.0.4 (the oldest still supported LTS) has PHP 5.3.2 and Debian stable has PHP 5.3.3 so I see no harm in requiring 5.3.1.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -16,7 +16,7 @@ Requirements
 ------------
 
 * PHP 5.2.4+ for Platform versions 11.x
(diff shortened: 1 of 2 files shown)


chdemko:
It could be a good idea to change the README.markdown file (requiring PHP 5.3.1)

realityking:
Good point, fixed.

eddieajau:
Nice to get rid of that annoying workaround finally.

PR #1004, opened 2012-03-19
To use HEAD method, CURL_NOBODY should be set to true

piotr-cz:
To use HEAD method, CURL_NOBODY should be set to true. If it's not, curl is waiting for response body, which will never receive and script reaches max_execution_time. Setting CURLOPT_CUSTOMREQUEST makes sense only for methods other than GET or HEAD.

Please refer to 
http://www.php.net/manual/en/function.curl-setopt.php, CURLOPT_NOBODY and 
http://stackoverflow.com/questions/770179/php-curl-head-request-takes-a-long-time-on-some-sites#770200


LouisLandry:
Awesome.  Thanks!

PR #1005, opened 2012-03-19
When sending new requests to same URI, need to close connection first.

piotr-cz:
After response is received, it's meta is set to 'eof', and cannot be reused anymore.

See http://php.net/manual/en/function.stream-get-meta-data.php


LouisLandry:
Thanks!

Issue #834, opened 2012-02-07
Bug in JString::parse_url


ebitester1:
Try this:

```
$url = 'http://a.com/b.php?retUrl='.rawurlencode('http://a.com/b.php?a=123&b=456');

echo '<pre>';
echo "url: $url\n\n";

echo "JString::parse_url():\n";
$parts = JString::parse_url($url);
parse_str($parts['query'], $parts['vars']);
print_r($parts);

echo "native parse_url():\n";
$parts = parse_url($url);
parse_str($parts['query'], $parts['vars']);
print_r($parts);

echo '</pre>';
```

Output:

```
url: http://a.com/b.php?retUrl=http%3A%2F%2Fa.com%2Fb.php%3Fa%3D123%26b%3D456

JString::parse_url():
Array
(
    [scheme] => http
    [host] => a.com
    [path] => /b.php
    [query] => retUrl=http://a.com/b.php?a=123&b=456        // WRONG
    [vars] => Array
        (
            [retUrl] => http://a.com/b.php?a=123        // WRONG
            [b] => 456        // WRONG
        )

)
native parse_url():
Array
(
    [scheme] => http
    [host] => a.com
    [path] => /b.php
    [query] => retUrl=http%3A%2F%2Fa.com%2Fb.php%3Fa%3D123%26b%3D456        // CORRECT
    [vars] => Array
        (
            [retUrl] => http://a.com/b.php?a=123&b=456        // CORRECT
        )

)
```

PR #1006, opened 2012-03-20
Fix issue #834: Bug in JString::parse_url

stefanneculai:
Bug in JString::parse_url

### DIFF
--- a/libraries/joomla/string/string.php
+++ b/libraries/joomla/string/string.php
@@ -958,20 +958,25 @@ abstract class JString
 	 */
 	public static function parse_url($url)
 	{
-		$result = array();
 
-		// Build arrays of values we need to decode before parsing
-		$entities = array('%21', '%2A', '%27', '%28', '%29', '%3B', '%3A', '%40', '%26', '%3D', '%24', '%2C', '%2F', '%3F', '%25', '%23', '%5B',
-			'%5D');
-		$replacements = array('!', '*', "'", "(", ")", ";", ":", "@", "&", "=", "$", ",", "/", "?", "%", "#", "[", "]");
+		$keys = array('scheme' => 0, 'user' => 0, 'pass' => 0, 'host' => 0, 'port' => 0, 'path' => 0, 'query' => 0, 'fragment' => 0);
 
-		// Create encoded URL with special URL characters decoded so it can be parsed
-		// All other characters will be encoded
-		$encodedURL = str_replace($entities, $replacements, urlencode($url));
+		// The u parameter is the key to determine UTF-8 characters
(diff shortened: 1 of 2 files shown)


ianmacl:
Can you please expand on the implications of this?

stefanneculai:
Please take a look over: https://github.com/joomla/joomla-platform/issues/834

elinw:
The autotester isn't running but his is failing a test

JStringTest::testParse_Url
Line: 629 Results should be equal
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'scheme' => 'http'
     'host' => 'joomla.org'
-    'path' => '/mytestpath/Ã�'
-    'path' => '/mytestpath/%C3%88'
  )

.../tests/suites/unit/joomla/string/JStringTest.php:629
/usr/bin/phpunit:46

You should also probably add a test for the scenario in the issue report.

stefanneculai:
Fixed and test added in.

chdemko:
http://a.com/?a=é
ftp://user:abcé@example.com/

does not pass the tests

See https://gist.github.com/2244092

stefanneculai:
Indeed it was a problem with those 2. 

I have done some more tests and it seems that the parse_url from JString does nothing for UTF-8. I mean.. I have tried to return the result of parse_url and all tests pass.

Maybe this function is old and in the meanwhile PHP has solved the problems. I am using PHP 5.3.8..

stefanneculai:
The test that was expected to fail is below.. At my end it does not fail.

$url = 'http://joomla.org/mytestpath/Ãˆ';
$expected = parse_url($url);
// Fix up path for UTF-8 characters
$expected['path'] = '/mytestpath/Ãˆ';
$actual = JString::parse_url($url);
$this->assertEquals($expected, $actual, 'Line: ' . **LINE** . ' Results should be equal');

Could anyone try my last commit with another version of PHP  or on Linux/MAC. I have used Windows.

stefanneculai:
Any comments on this? I have just tested it on both Windows and Linux and calling the parse_url function from PHP is enough.

chdemko:
The pull request cannot be merged. Could you try this url on windows:

```
http://mydomain.com/path/道
```

On linux php 5.4.3,

it output

```
array(3) {
  'scheme' =>
  string(4) "http"
  'host' =>
  string(12) "mydomain.com"
  'path' =>
  string(9) "/path/道"
}
```

which is correct

stefanneculai:
Works fine on Windows too. I have created a new pull request: https://github.com/joomla/joomla-platform/pull/1285

The new one should work to be merged.

chdemko:
See #1285

PR #1007, opened 2012-03-20
Implementing new file system

chdemko:
This pull request implement new file system for Joomla!

``` PHP
<?php
$filesystem = JFilesystem::getInstance();
$file = $filesystem->getFile('/path/to/file');
$file->contents = 'My contents'; // This will store 'My contents' into the file
$file->open('r'); // This will open the file for reading
$string = $file->readLine(); // This will read a line
$file->close(); // This will close the file

// Iterating on files
foreach ($file->iterateLine() as $line)
{
    // Deal with $line
}

$file->delete(); // This will destroy the file
```

See the docs for other features


chdemko:
Reworking unit test for testing remote file system

chdemko:
@AmyStephen 
"This is beautiful. Once we learn how to use JFileSystemElement - that should be very powerful."

Look at the manual (not already complete but will be done next days)

"Any reason not to simply bring in Jxmlelement? Appears to be the only dependency.  "

I don't understand what you are meaning

AmyStephen:
I most definitely will be studying the documentation, appreciate you doing that.

In this package, there is only one external dependency: 

+jimport('joomla.utilities.xmlelement');

Wondering if the single method needed could be included to remove even that dependency? It's a great stand-alone package and something that will be needed time and time again by developers. Having no external dependencies at all will make it that much easier to use - good, good work. Appreciate it very much.

elinw:
I'm confused are you losing all of history?

chdemko:
I have rebased the code to avoid a lot of commit in the platform history

ianmacl:
I don't think we want to add the static method in JFactory.

libraries/joomla/filesystem/element.php
```diff
@@ -0,0 +1,1008 @@
...
+	{
+		return readlink($this->_fullpath);
+	}
+
+	/**
+	 * Sets the element group
+	 *
+	 * @param   int  $group  The new group ID
+	 *
+	 * @return  int|false  Returns the group ID of the element, or FALSE if an error occurs.
```
elinw: Would you make this mixed please. Also please spell out integer and boolean.

libraries/joomla/filesystem/element/directory.php
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * @property-read  RecursiveIteratorIterator  $files        Iterator on files.
+ * @property-read  RecursiveIteratorIterator  $directories  Iterator on directories.
+ *
+ * @method                            bool create(int $permissions)                   create the directory
+ * @method                            bool delete()                                   delete the directory
+ * @method                       int|FALSE copy(JFilesystemElementDirectory $dest)    copy the directory
+ * @method                       int|FALSE copyFromFile(JFilesystemElementFile $src)  copy from a file
+ * @method       RecursiveIteratorIterator files(array $options)                      iterate on files
+ * @method       RecursiveIteratorIterator directories(array $options)                iterate on directories
```
elinw: Looks like you have tabs here.

elinw:
Somewhere deep inside this something is calling /tests/suites/unit/joomla/table/JTableCategoryTest.php"  which is now in the legacy package.

libraries/joomla/filesystem/element/directory.php
chdemko: No tabs in my code

chdemko:
I'm not calling /tests/suites/unit/joomla/table/JTableCategoryTest.php"

elinw:
Oh cool the tests ran again and it's not failing any more so whatever you were calling that was calling JTableCategory probably isn't any more.

libraries/joomla/filesystem/element/file.php
```diff
@@ -0,0 +1,644 @@
...
+	/**
+	 * Constructor
+	 *
+	 * @param   string       $path       Element path.
+	 * @param   JFilesystem  $system     Element file system
+	 * @param   string       $signature  Signature
+	 *
+	 * @since   12.2
+	 */
+	protected function __construct($path, JFilesystem $system, $signature)
```
ianmacl, line 80: This should be public.

libraries/joomla/filesystem/filesystem.php
```diff
@@ -0,0 +1,197 @@
...
+	 *
+	 * @param   string  $prefix   The file system prefix.
+	 * @param   array   $options  The stream context options.
+	 * @param   array   $params   The stream context params.
+	 *
+	 * @link    http://fr.php.net/manual/en/function.stream-context-create.php
+	 *
+	 * @since   12.2
+	 */
+	protected function __construct($prefix,  $options,  $params)
```
ianmacl, line 58: This should be public.

ianmacl:
It took me quite a while to actually understand how this package was structured.  There is an incredibly liberal use of magic getters, setters, callers and static methods and calls that get sent all over.

I think there are some basic elements of the package that are useful and good.  However, I think that the approach taken is over complicated.

There is a lot of effort put into enabling accessing files using different formats using $fileElement->readIni and proxying to static accessors.  I think a better approach would be to create classes that would take the JFilesystemElement or JFilesystem objects as parameters in their constructors and implement that functionality that way.

The other concern I have is some of the non-obviousness of the interface in general.  In my opinion, the way that FileElement->location works is rather unusual.  Using the value TRUE to move the cursor to the end seems odd. Perhaps using actual numbers to set the cursor would be better and it could then be operator on using normal arithmetic.

It wasn't entirely clear to me what the differences were between read and push and write and pull.  For some reason I just am not sure what you are doing there.

ianmacl:
So I think in general we want to rework the filesystem package but I personally don't think this is it.  I'm going to close this until Christophe or somebody can pick this up again and engage in more discussion on it.

PR #1009, opened 2012-03-21
Fixing error in JGithub package

nprasath002:
Jhithub Package  is not handling POST, PATCH , DELETE Requests well.
The root cause of the error is JGithubHttp which extends JHttp instantaite a JHttpTransportSocket HTTP transport which does not handle the JGithub requests.

The JGithub package is working well in 11.4 release because JHttp instantiates JHttpTransportStream transport direcrlty. In the trunk this has been changed as JHttpFactory::getAvailableDriver($this->options) which returns a JHttpTransportSocket.

### DIFF
--- a/libraries/joomla/github/http.php
+++ b/libraries/joomla/github/http.php
@@ -46,8 +46,8 @@ class JGithubHttp extends JHttp
 	 */
 	public function __construct(JRegistry $options = null, JHttpTransport $transport = null)
 	{
-		// Call the JHttp constructor to setup the object.
-		parent::__construct($options, $transport);
+        $this->options   = isset($options) ? $options : new JRegistry;
+        $this->transport = isset($transport) ? $transport : new JHttpTransportStream($this->options);
 
 		// Make sure the user agent string is defined.
 		$this->options->def('userAgent', 'JGitHub/2.0');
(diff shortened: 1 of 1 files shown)


libraries/joomla/github/http.php
```diff
@@ -1,74 +1,74 @@
...
+	 *
+	 * @param   JRegistry       $options    Client options object.
+	 * @param   JHttpTransport  $transport  The HTTP transport object.
+	 *
+	 * @since   11.3
+	 */
+	public function __construct(JRegistry $options = null, JHttpTransport $transport = null)
+	{
+        $this->options   = isset($options) ? $options : new JRegistry;
+        $this->transport = isset($transport) ? $transport : new JHttpTransportStream($this->options);
```
nprasath002, line 50: The overridden constructor instantiates a JHttpTransportStream instead of a JHttpTransportSocket

realityking:
I think you've run into some line ending issue (LF vs CRLF most likely) and are rewriting the entire file. Could you fix that?

I'm fine with pulling this, but I wonder why the other transports don't work? We should try to fix them.

nprasath002:
Hi realityking
I,m new to Github.
I will fix that and ping you.
I,m just getting familiar with things.

realityking:
No worries, just fix it and add a comment here. I'll be notified by github.

PR #1011, opened 2012-03-21
Fix checkstyle errors and warnings in staging branch.

eddieajau:


libraries/joomla/filesystem/folder.php
```diff
@@ -258,7 +258,10 @@ public static function create($path = '', $mode = 0755)
 			if (!$ret = @mkdir($path, $mode))
 			{
 				@umask($origmask);
-				JLog::add('SOME_ERROR_CODE', __METHOD__ . ': ' . JText::_('JLIB_FILESYSTEM_ERROR_COULD_NOT_CREATE_DIRECTORY'), 'Path: ' . $path, JLog::WARNING, 'jerror');
+				JLog::add(
+					'SOME_ERROR_CODE',
```
realityking, line 262: Is that right? Doesn't make sense to me.

AmyStephen, line 262: I tried to get rid of those, but admittedly I was in autopilot mode, just copying messages there.

@eddieajau - thanks for doing this. Getting closer to getting codesniffing and testing in place. With phpStorm v 4, we can use codesniffing in the IDE. That'll be very helpful. I have the beta installed but have not got the Joomla standards in place yet. https://plus.google.com/109030869001230031837/posts/A2BeQNsXheR

realityking:
I'm going to merge this to get the build running again, we can fix JFolder in another pull.

PR #1012, opened 2012-03-21
[#28210] onchange event does not fire at JFormFieldMedia (Regression)

benjaminpick:
See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28210


realityking:
Looking good. Thank you.

PR #1018, opened 2012-03-21
Updated database drivers to attempt to reconnect from database servers that have been disconnected.

robschley:
This fixes an issue where the database server gets disconnected from the PHP connection resource. This can happen in two situations that I am aware of:
- The database server is heavily loaded and drops a connection. 
- The resource was copied into a forked process which happens when forking JDaemon. If the child process closes the connection or finishes execution, the parent resource will be disconnected.

To address these two issues, if a valid cursor is not returned when executing a query, the driver will check if the resource is still connected to the database. If the resource is disconnected, it will attempt to reconnect and execute the query again. If the reconnect fails or if the server still shows as connected, the normal exception will be thrown.

For PDO drivers, this required a better implementation of the connected() method. I attempted to use PDO::getAttribute(ATTR_CONNECTION_STATUS) but that feature is not supported on Sqlite or Oracle, possibly others. To work around that, the connected() method will attempt to execute a simple database query. To prevent contamination of the driver state, the previous query information is backed up and then restored after the query is executed.

### DIFF
--- a/libraries/joomla/database/driver/mysql.php
+++ b/libraries/joomla/database/driver/mysql.php
@@ -447,7 +447,7 @@ class JDatabaseDriverMysql extends JDatabaseDriver
 		$this->errorNum = 0;
 		$this->errorMsg = '';
 
-		// Execute the query.
+		// Execute the query. Error suppression is used here to prevent warnings/notices that the connection has been lost.
 		$this->cursor = @mysql_query($sql, $this->connection);
 
 		// If an error occurred handle it.
--- a/libraries/joomla/database/driver/mysql.php
+++ b/libraries/joomla/database/driver/mysql.php
@@ -448,16 +448,46 @@ class JDatabaseDriverMysql extends JDatabaseDriver
 		$this->errorMsg = '';
 
 		// Execute the query.
-		$this->cursor = mysql_query($sql, $this->connection);
+		$this->cursor = @mysql_query($sql, $this->connection);
 
 		// If an error occurred handle it.
 		if (!$this->cursor)
 		{
-			$this->errorNum = (int) mysql_errno($this->connection);
-			$this->errorMsg = (string) mysql_error($this->connection) . ' SQL=' . $sql;
+			// Check if the server was disconnected.
+			if (!$this->connected())
+			{
+				try
+				{
+					// Attempt to reconnect.
+					$this->connection = null;
+					$this->connect();
+				}
+				// If connect fails, ignore that exception and throw the normal exception.
+				catch (RuntimeException $e)
+				{
+					// Get the error number and message.
+					$this->errorNum = (int) mysql_errno($this->connection);
+					$this->errorMsg = (string) mysql_error($this->connection) . ' SQL=' . $sql;
 
-			JLog::add(JText::sprintf('JLIB_DATABASE_QUERY_FAILED', $this->errorNum, $this->errorMsg), JLog::ERROR, 'databasequery');
-			throw new RuntimeException($this->errorMsg, $this->errorNum);
+					// Throw the normal query exception.
+					JLog::add(JText::sprintf('JLIB_DATABASE_QUERY_FAILED', $this->errorNum, $this->errorMsg), JLog::ERROR, 'databasequery');
+					throw new RuntimeException($this->errorMsg, $this->errorNum);
+				}
+
+				// Since we were able to reconnect, run the query again.
+				return $this->execute();
+			}
+			// The server was not disconnected.
+			else
+			{
+				// Get the error number and message.
+				$this->errorNum = (int) mysql_errno($this->connection);
+				$this->errorMsg = (string) mysql_error($this->connection) . ' SQL=' . $sql;
+
+				// Throw the normal query exception.
+				JLog::add(JText::sprintf('JLIB_DATABASE_QUERY_FAILED', $this->errorNum, $this->errorMsg), JLog::ERROR, 'databasequery');
+				throw new RuntimeException($this->errorMsg, $this->errorNum);
+			}
 		}
 
 		return $this->cursor;
--- a/libraries/joomla/database/driver/mysqli.php
+++ b/libraries/joomla/database/driver/mysqli.php
@@ -293,7 +293,7 @@ class JDatabaseDriverMysqli extends JDatabaseDriverMysql
 		$this->errorNum = 0;
 		$this->errorMsg = '';
 
-		// Execute the query.
+		// Execute the query. Error suppression is used here to prevent warnings/notices that the connection has been lost.
 		$this->cursor = @mysqli_query($this->connection, $sql);
 
 		// If an error occurred handle it.
--- a/libraries/joomla/database/driver/mysqli.php
+++ b/libraries/joomla/database/driver/mysqli.php
@@ -294,16 +294,42 @@ class JDatabaseDriverMysqli extends JDatabaseDriverMysql
 		$this->errorMsg = '';
 
 		// Execute the query.
-		$this->cursor = mysqli_query($this->connection, $sql);
+		$this->cursor = @mysqli_query($this->connection, $sql);
 
 		// If an error occurred handle it.
 		if (!$this->cursor)
 		{
-			$this->errorNum = (int) mysqli_errno($this->connection);
-			$this->errorMsg = (string) mysqli_error($this->connection) . ' SQL=' . $sql;
+			// Check if the server was disconnected.
+			if (!$this->connected())
+			{
+				try
+				{
+					// Attempt to reconnect.
+					$this->connection = null;
+					$this->connect();
+				}
+				// If connect fails, ignore that exception and throw the normal exception.
+				catch (RuntimeException $e)
+				{
+					$this->errorNum = (int) mysqli_errno($this->connection);
+					$this->errorMsg = (string) mysqli_error($this->connection) . ' SQL=' . $sql;
 
-			JLog::add(JText::sprintf('JLIB_DATABASE_QUERY_FAILED', $this->errorNum, $this->errorMsg), JLog::ERROR, 'databasequery');
-			throw new RuntimeException($this->errorMsg, $this->errorNum);
+					JLog::add(JText::sprintf('JLIB_DATABASE_QUERY_FAILED', $this->errorNum, $this->errorMsg), JLog::ERROR, 'databasequery');
+					throw new RuntimeException($this->errorMsg, $this->errorNum);
+				}
+
+				// Since we were able to reconnect, run the query again.
+				return $this->execute();
+			}
+			// The server was not disconnected.
+			else
+			{
+				$this->errorNum = (int) mysqli_errno($this->connection);
+				$this->errorMsg = (string) mysqli_error($this->connection) . ' SQL=' . $sql;
+
+				JLog::add(JText::sprintf('JLIB_DATABASE_QUERY_FAILED', $this->errorNum, $this->errorMsg), JLog::ERROR, 'databasequery');
+				throw new RuntimeException($this->errorMsg, $this->errorNum);
+			}
 		}
 
 		return $this->cursor;
--- a/libraries/joomla/database/driver/pdo.php
+++ b/libraries/joomla/database/driver/pdo.php
@@ -497,31 +497,31 @@ abstract class JDatabaseDriverPdo extends JDatabaseDriver
 	 */
 	public function connected()
 	{
+		// Backup the query state.
+		$sql = $this->sql;
+		$limit = $this->limit;
+		$offset = $this->offset;
+		$prepared = $this->prepared;
+
 		try
 		{
-			// Backup the query state.
-			$sql = $this->sql;
-			$limit = $this->limit;
-			$offset = $this->offset;
-			$prepared = $this->prepared;
-
-			// Run a simple query.
+			// Run a simple query to check the connection.
 			$this->setQuery('SELECT 1');
 			$status = (bool) $this->loadResult();
-
-			// Restore the query state.
-			$this->sql = $sql;
-			$this->limit = $limit;
-			$this->offset = $offset;
-			$this->prepared = $prepared;
-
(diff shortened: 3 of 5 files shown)


AmyStephen:
Not sure what problem you were trying to solve but suppressing MySQL error messages is bad form. Just takes a lot longer to debug problems for those of us using the Joomla Database package.

pasamio:
"Error suppression is used here to prevent warnings/notices that the connection has been lost."

ianmacl:
Also they are logged so it isn't like they are just being eaten up.

AmyStephen:
Yes, I can read Rob's comments, Sam. 

The assumption that it's a connection problem is invalid. For me, a simple syntax error where a selected column didn't exist in the table was the underlying problem. 

Since the assumption is it's a connection issue, that error is thrown away, a connection is attempted and if all is not cool (like maybe you have a syntax error), it will error on having no cursor. 

When the error is thrown, the application returns a zero error code and no message - because the original error is long gone and the connection wasn't a problem.

You get an error. You have no code, no message, no clue.

If you are trying to trap connection errors, better look at the error to make certain that is indeed the problem. If it's not, let the error be thrown so that the application gets good messages on the real problem.

pasamio:
Then why did you ask what the problem that he was trying to solve was if you can read the comment? It seems obvious to me the problem that he was trying to solve is explained so perhaps I missed something.

So the way I read the code is that it runs the query, the query fails, we ask MySQL if it's connected, if it says it isn't we re-connect and attempt to retry. If it does say it's connected it grabs the error number and error message from the connection and then throws an exception.

Near as I can tell it shouldn't attempt to re-connect and "throw away" the error message but instead should do what it was doing before which is get the error code and message from the server and then throw an exception with these details. Or are you saying that the exception never gets thrown?

Is it possible that mysql_ping/mysqli_ping is clearing the result so perhaps all that needs to occur is 456/457 (mysql.php) or 302/303 (mysqli.php) needs to be put back and those lines that are elsewhere removed? Have you tried merely shifting those lines to see if it makes a difference? They'll obviously need some form of error suppression put around them as well because when the connection is disconnected this will cause a notice to be raised.

This seems to be more likely that there is a simple bug that needs to get sorted around making sure everything is run in the right order.

AmyStephen:
Here are some links to the patch that might help explain:
1. The workaround applied for a dropped connection was to suppress the error message (my error was a syntax error): 

https://github.com/robschley/joomla-platform/commit/1971fcbdbe274d0e5f7ba77cf9c481d547ed462e#L0R451
1. and then silently attempt a reconnect to recover (my connection was fine). 

https://github.com/robschley/joomla-platform/commit/1971fcbdbe274d0e5f7ba77cf9c481d547ed462e#L0R457
1. If the connection is not a problem (which it wasn't for me), it falls into the throw error logic  

https://github.com/robschley/joomla-platform/commit/1971fcbdbe274d0e5f7ba77cf9c481d547ed462e#L2R399
1. Which should be okay except the real error was suppressed in case it was a connection problem (which it wasn't).

Looks like the CMS reported similar problems a month ago. 

Hope that helps.

pasamio:
If you're using MySQL, you're saying that moving lines 483 to 485 of mysql.php to be back where they were on line 456/457 just before the `if (!$this->connected())` line doesn't fix the problem for you?

If you're using MySQLi, you're saying that moving lines 327 and 328 of mysqli.php to be back where they were on line 302 and 303 just before the `if (!$this->connected())` line doesn't fix the problem for you?

Are you really telling me a developer of your calibre can't relocate two/three lines?

AmyStephen:
I have no intention of wasting another minute of my life arguing with defensive Joomla developers.

Unsubscribed

eddieajau:
How about both of you stop being defensive and work the problem.

@AmyStephen, I've had lots of syntax errors and never encountered a problem.  Do you have some code and/or conditions that can reproduce the problem (my gut is telling me it's a particular edge case that not everyone may have - I vaguely remember having a similar problem but can't remember the circumstances)?  All the usual suspects like PHP/MySQL version etc would also help.

The bad form is that PHP echo's warnings or errors and that's the reason we need to suppress them (the PECL SOLR connector is another painful example; it'll throw an exception but also generate PHP notices et al - seriously?).  We won't need to do that when PHP gets it's act together and throws exceptions properly one way or the other.

References: 

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29345
https://github.com/joomla/joomla-platform/pull/1554

pasamio:
I seriously think the problem is simple, mysql_ping/mysqli_ping is merely cleaning the error code which is a behaviour that wasn't initially expected. I feel all that needs to happen is to shift those lines as I've stated previously and the problem may be solved. Given #1554, it looks like someone had the same conclusion and now it's only an issue for MySQL driver not MySQLi any more.

AmyStephen:
@eddieajau - 

Yes, it's pretty easy to reproduce. 

I put a quick gist together that you can use MySQLI Database Driver class, execute Method (libraries/joomla/database/driver/mysqli.php ).  https://gist.github.com/4042212

To create a problem just introduce a syntax error in any of the CMS queries.

The displays I added to the execute method in the gist will demonstrate what I explained previously.

Thanks for the background on the problem you were trying to solve. I understand the need to bury notices one can't avoid. Perhaps testing for the specific code that a connection problem presents might be all that is needed to ensure the other errors are caught, not overwritten.

Thanks.

eddieajau:
So, as far as you are aware, this in only reproducible in the CMS (3.0?).  Do you know if it's occurring in standalone platform apps?  And just for good measures, what are your PHP and MySQL versions?

eddieajau:
I can't reproduce the problem.

/Joomla 3/administrator/components/com_content/models/articles.php

Changed 'from' clause to `$query->from('#__contentX AS a');` to introduce a syntax error.

Goto http://localhost/joomla/joomla-cms/administrator/index.php?option=com_content

```
An error has occurred.
 500 SQL=SELECT a.id, a.title, a.alias, a.checked_out, a.checked_out_time, a.catid, a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language, a.hits, a.publish_up, a.publish_down,l.title AS language_title,uc.name AS editor,ag.title AS access_level,c.title AS category_title,ua.name AS author_name FROM jos_contentX AS a LEFT JOIN `jos_languages` AS l ON l.lang_code = a.language LEFT JOIN jos_users AS uc ON uc.id=a.checked_out LEFT JOIN jos_viewlevels AS ag ON ag.id = a.access LEFT JOIN jos_categories AS c ON c.id = a.catid LEFT JOIN jos_users AS ua ON ua.id = a.created_by WHERE (a.state = 0 OR a.state = 1) ORDER BY a.title asc SQL=SELECT a.id, a.title, a.alias, a.checked_out, a.checked_out_time, a.catid, a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language, a.hits, a.publish_up, a.publish_down,l.title AS language_title,uc.name AS editor,ag.title AS access_level,c.title AS category_title,ua.name AS author_name FROM jos_contentX AS a LEFT JOIN `jos_languages` AS l ON l.lang_code = a.language LEFT JOIN jos_users AS uc ON uc.id=a.checked_out LEFT JOIN jos_viewlevels AS ag ON ag.id = a.access LEFT JOIN jos_categories AS c ON c.id = a.catid LEFT JOIN jos_users AS ua ON ua.id = a.created_by WHERE (a.state = 0 OR a.state = 1) ORDER BY a.title asc LIMIT 0, 20 SQL=SELECT a.id, a.title, a.alias, a.checked_out, a.checked_out_time, a.catid, a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language, a.hits, a.publish_up, a.publish_down,l.title AS language_title,uc.name AS editor,ag.title AS access_level,c.title AS category_title,ua.name AS author_name FROM jos_contentX AS a LEFT JOIN `jos_languages` AS l ON l.lang_code = a.language LEFT JOIN jos_users AS uc ON uc.id=a.checked_out LEFT JOIN jos_viewlevels AS ag ON ag.id = a.access LEFT JOIN jos_categories AS c ON c.id = a.catid LEFT JOIN jos_users AS ua ON ua.id = a.created_by WHERE (a.state = 0 OR a.state = 1) ORDER BY a.title asc
```

Latest master branch; Debug: On; Error Reporting: Maximum; PHP: 5.3.10; MySQL: 5.5.18

ianmacl:
It strikes me that the code in the gist does not represent the current code that is in the platform tree.  Specifically https://github.com/joomla/joomla-platform/pull/1554 does not seem to have been merged into your version.  This change stores the error in the database driver object properties before proceeding with the reconnection attempt.  Thus the error message likely is lost.  If you update to the newer version of the platform I suspect that your problem will go away.

AmyStephen:
@eddieajau -

It's not just the CMS. I am producing this out of Molajo, but since I use the database package differently, to make things easier for the Joomla team to reproduce, I tested using the latest version of the CMS. 

Your results are exactly as I would expect -- you have a generic 500 error and a copy of the SQL -- but no clue as to what went wrong.  Try it with my gist and then those echo's I shared should help explain what the issue is.

Specifically, for my test, what I did was add a column that didn't exist to the SELECT (a.nope) to the Admin Articles List query:

```
    $query->select(
        $this->getState(
            'list.select',
            'a.id, a.title, a.alias, a.checked_out, a.checked_out_time, a.catid, a.nope, ' .
            ', a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language, a.hits' .
            ', a.publish_up, a.publish_down'
        )
    );
```

The results I got from the test, are as follows:

```
After $this->cursor - display the untrapped error before it is overwritten by the next DB operation: 
Code: 1064:
Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ' a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language,' at line 1:


There is no cursor and errors are still available since no more DB operations have taken place.
Code: 1064:
Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ' a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language,' at line 1:


The connection is fine. The errors are now gone because the last DB operation -- $this->connection() -- was successful and overlaid the error codes.
Code: 0:
Error: SQL=SELECT a.id, a.title, a.alias, a.checked_out, a.checked_out_time, a.catid, a.nope, , a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language, a.hits, a.publish_up, a.publish_down,l.title AS language_title,uc.name AS editor,ag.title AS access_level,c.title AS category_title,ua.name AS author_name FROM j_content AS a LEFT JOIN `j_languages` AS l ON l.lang_code = a.language LEFT JOIN j_users AS uc ON uc.id=a.checked_out LEFT JOIN j_viewlevels AS ag ON ag.id = a.access LEFT JOIN j_categories AS c ON c.id = a.catid LEFT JOIN j_users AS ua ON ua.id = a.created_by WHERE (a.state = 0 OR a.state = 1) ORDER BY a.state asc: 
```

As you are noting, obviously, the query is going to fail.  The problem is that the actual error code of 1064 and the specific error message are missing. So, you have a generic 500 error - but a code of 0 which should mean all is well, and a partial copy (depending on how long it is) of your SQL.

Hope that helps.

AmyStephen:
Yes, @ianmacl - you are absolutely right - that does patch does fix the problem. My test now looks like this:

```
After $this->cursor - display the untrapped error before it is overwritten by the next DB operation:
Code: 1064:
Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ' a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language,' at line 1:


There is no cursor and errors are still available since no more DB operations have taken place.
Code: 1064:
Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ' a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language,' at line 1:


The connection is fine. The error code is zero now gone because the last DB operation -- $this->connection() -- was successful and overlaid the error codes.
Code: 1064:
Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ' a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language,' at line 1 SQL=SELECT a.id, a.title, a.alias, a.checked_out, a.checked_out_time, a.catid, a.nope, , a.state, a.access, a.created, a.created_by, a.ordering, a.featured, a.language, a.hits, a.publish_up, a.publish_down,l.title AS language_title,uc.name AS editor,ag.title AS access_level,c.title AS category_title,ua.name AS author_name FROM j_content AS a LEFT JOIN `j_languages` AS l ON l.lang_code = a.language LEFT JOIN j_users AS uc ON uc.id=a.checked_out LEFT JOIN j_viewlevels AS ag ON ag.id = a.access LEFT JOIN j_categories AS c ON c.id = a.catid LEFT JOIN j_users AS ua ON ua.id = a.created_by WHERE (a.state = 0 OR a.state = 1) ORDER BY a.state asc: 
```

My apology for not checking trunk. That definitely retains the error and message and throws the right values.

Appreciate everyone's help.

eddieajau:
For what it was worth, it was fixed a month ago when reported :)  That said, I accidentally tested on 3.0 Beta 1 myself but still couldn't replicate no-message so go figure.

AmyStephen:
If you look at my examples, you'll see, there is indeed a message. The problem is that the original, more descriptive message is gone and the error code is missing. But, a generic "something is wrong" message is returned.

Regardless, it is fixed in the platform and the test confirms that the patch retains the original message and code.

For clarity and for what it's worth, it is not fixed in the CMS master which is what I used in my testing example and Ian pointed out as not consistent with the platform. https://github.com/joomla/joomla-cms/blob/master/libraries/joomla/database/driver/mysqli.php#L529

eddieajau:
Oh I see what you mean now.  Got it.  I'm just so used to looking at those messages I know what's going on :)

PR #1019, opened 2012-03-22
label corrupts the standard layout 

romacron:
mono mustend be

### DIFF
--- a/media/system/css/mooRainbow.css
+++ b/media/system/css/mooRainbow.css
@@ -55,9 +55,6 @@
 	border-left-color: #f5f5f5;
 	border-top-color: #f5f5f5;
 }
-label {
-	font-family: mono;
-}
 /* Following are just <label> */
 .moor-rLabel {
 	/*display: none;*/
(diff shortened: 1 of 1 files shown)


media/system/css/mooRainbow.css
```diff
@@ -83,7 +80,7 @@ label {
 	margin-top: 190px;
 	margin-left: 315px;
 }
-span.moor-ballino { /* Style hue ° (degree) !! */
+span.moor-ballino { /* Style hue ï¿½ (degree) !! */
```
ianmacl, line 83: That looks like a file encoding error.  Can you fix that?

realityking:
We'd need to upstream this too https://github.com/CBeloch/mooRainbow

media/system/css/mooRainbow.css
```diff
@@ -55,9 +55,6 @@
 	border-left-color: #f5f5f5;
 	border-top-color: #f5f5f5;
 }
-label {
-	font-family: mono;
-}
```
w00fz, line 60: I think it should be:

``` css
.moor-box label { font-family: mono; }
```

rather than taking it off entirely. The reason I had that as mono is to ensure the labels such as HEX, RGB, etc, won't inherit any of the admin templates font families and sizes.

media/system/css/mooRainbow.css
romacron, line 83: degree is also fixed

romacron:
thanks it is fixed !

PR #1022, opened 2012-03-22
Throw an exception before invalid response is parsed by getResponse()

piotr-cz:
to test: JHttpFactory::getHttp()->head('http://blah')

Parsing invalid response (false, "") throws PHP errors as getResponse() assumes $content is consists headers and content.

### DIFF
--- a/libraries/joomla/http/transport/socket.php
+++ b/libraries/joomla/http/transport/socket.php
@@ -137,6 +137,12 @@ class JHttpTransportSocket implements JHttpTransport
 			$content .= fgets($connection, 4096);
 		}
 
+		// Received empty response (invalid uri, connection reached timeout);
+		if ($content === "")
+		{
+			throw new RuntimeException('HTTP request failed');
+		}
+
 		return $this->getResponse($content);
 	}
 
--- a/libraries/joomla/http/transport/curl.php
+++ b/libraries/joomla/http/transport/curl.php
@@ -134,6 +134,12 @@ class JHttpTransportCurl implements JHttpTransport
 		$content = curl_exec($ch);
 		curl_close($ch);
 
+		// Received empty response (invalid uri, connection reached timeout);
+		if ($content === false)
+		{
+			throw new RuntimeException('HTTP request failed');
+		}
+
 		return $this->getResponse($content);
 	}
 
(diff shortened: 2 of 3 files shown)


piotr-cz:
In case of JHttpTransportStream, 
PHP shows a Warning:
`Warning:  fopen($invaliduri) failed to open stream: HTTP request failed!  in ~libraries\joomla\http\transport\stream.php on line 125`

But I'm not sure if it's the right thing to supress the errors

Edit:
As I'm thinking about it, fopen should supress any PHP warnings.
Typical scenario may be when the response (after processing) is going to be output as JSON. Any PHP warnings are then appended to the response body, deforming the JSON output. 

The check for no HTTP response that I added is supposed to handle it anyway

see: http://php.net/manual/en/function.fopen.php
_If the open fails, an error of level E_WARNING is generated. You may use @ to suppress this warning._

piotr-cz:
There is 1 error in Unit test (http://developer.joomla.org/pulls/pulls/1022.html) but seems like it's triggered by something strange outside this PR

Does this PR have any chance to be merged? There are few other issues in JHttp package but I decided to wait so this one can be still merged

piotr-cz:
JHttpTransportCurl fixed by #1438

pasamio:
You'll need to update to master because of the curl changes and would you mind putting in some code to check $php_errormsg for the stream handler so that we can get some insight at least there why it failed?

libraries/joomla/http/transport/socket.php
```diff
@@ -137,6 +137,12 @@ public function request($method, JUri $uri, $data = null, array $headers = null,
...
 		}
 
+		// Received empty response (invalid uri, connection reached timeout);
+		if ($content === "")
+		{
+			throw new RuntimeException('HTTP request failed');
+		}
+
 		return $this->getResponse($content);
 	}
```
elinw, line 148: What you are doing here is different than what I did in my pull request which relate to the warning from fsockopen()  at line 256 similar to what you did with fopen() in stream.

piotr-cz, line 148: I've tested all transports against invalid url and var_dump'ed the result:
- for socket it was: `""`
- for curl and stream: `false`

So I've checked for empty response in the request method to be able to throw an exception in about the same place as in other transports.

Probably emphasizing an convention may not be optimal here as every transport is different.

LouisLandry:
@piotr-cz would you mind rebasing this so we can potentially getting it merged?  I'm gonna want to give it a test drive for sure, but I'd love to have consistency here.  Can you think of a way we might be able to test it?  

_Note: I'm closing this until it is mergeable again, not because I'm rejecting the work.  Please re-open the request when you get it ready to go._

PR #1025, opened 2012-03-22
Implement the JsonSerializable interface in JRegistry

realityking:
This is would be the first use of one of the new PHP 5.4 features, but it maintains compatibility with PHP 5.3.

How to use:

```
$a = new JRegistry();
$a->set('foo', 'bar');
echo json_encode($a);
```

This isn't immediately useful since PHP 5.4 is very rare for now, but it may be appreciated by someone building a custom app who runs 5.4.

### DIFF
--- a/libraries/import.legacy.php
+++ b/libraries/import.legacy.php
@@ -57,6 +57,12 @@ JLoader::register('JRoute', JPATH_PLATFORM . '/joomla/application/route.php');
 // Register the folder for the moved JHtml classes
 JHtml::addIncludePath(JPATH_PLATFORM . '/legacy/html');
 
+// Register classes for compatability with PHP 5.3
+if (version_compare(PHP_VERSION, '5.4.0', '<'))
+{
+	JLoader::register('JsonSerializable', __DIR__ . '/compat/jsonserializable.php');
+}
+
 // Add deprecated constants
 // @deprecated 12.3
 define('JPATH_ISWIN', IS_WIN);
--- a/libraries/import.php
+++ b/libraries/import.php
@@ -47,6 +47,12 @@ JLoader::setup();
 // Import the base Joomla Platform libraries.
 JLoader::import('joomla.factory');
 
+// Register classes for compatability with PHP 5.3
+if (version_compare(PHP_VERSION, '5.4.0', '<'))
+{
+	JLoader::register('JsonSerializable', __DIR__ . '/compat/jsonserializable.php');
+}
+
 // Register classes that don't follow one file per class naming conventions.
 JLoader::register('JText', JPATH_PLATFORM . '/joomla/language/text.php');
 JLoader::register('JRoute', JPATH_PLATFORM . '/joomla/application/route.php');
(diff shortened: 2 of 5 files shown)


eddieajau:
If we left this to the auto-loader, wouldn't /libraries/json/serializable.php "just work".  If it's PHP 5.4, the autoloader wouldn't be invoked (I'm assuming), and if it's PHP 5.3, the auto-loader will just pick it up??

realityking:
You'd have to put it into /libraries/son/serializable.php (note the missing J in json) and you'd have to change the case from what is used in the php documentation (to JSonSeriazeable) - not sure that's worth saving 3 lines.

eddieajau:
That's unfortunate (the J thing).  I'm not sure how I feel about this one.

robschley:
I've been playing with something along these lines too and I think it would be really helpful to start using JsonSerializable where appropriate. What if we just register the class if PHP < 5.4 in the platform import.php file? That way we don't have to have a check in each class that uses it.

eddieajau:
Works for me.

realityking:
Fine with me. Any preference to where to put the interface (which folder/subpackage)?

robschley:
How about we create a compat package? I'd probably put it at the same level as the joomla/legacy directories.

realityking:
Done as suggested.

libraries/import.legacy.php
```diff
@@ -57,6 +57,12 @@ class_exists('JLoader') or die;
 JLoader::register('JText', JPATH_PLATFORM . '/joomla/language/text.php');
 JLoader::register('JRoute', JPATH_PLATFORM . '/joomla/application/route.php');
 
+// Register classes for compatability with PHP 5.3
+if (version_compare(PHP_VERSION, '5.4.0', '<'))
+{
+	require_once __DIR__ . '/compat/jsonserializable.php';
```
chdemko: Why don't you use

```
JLoader::register('JsonSerializable', __DIR__ . '/compat/jsonserializable');
```

libraries/import.php
```diff
@@ -47,6 +47,12 @@ class_exists('JLoader') or die;
 // Import the base Joomla Platform libraries.
 JLoader::import('joomla.factory');
 
+// Register classes for compatability with PHP 5.3
+if (version_compare(PHP_VERSION, '5.4.0', '<'))
+{
+	require_once __DIR__ . '/compat/jsonserializable.php';
```
chdemko: Idem as above

libraries/import.legacy.php
robschley: +1

realityking: Fixed. Thanks.

libraries/import.php
realityking: Fixed.

chdemko:

libraries/import.legacy.php
realityking: It was missing the .php so it screwed up the include :P

realityking:
All tests should pass now, was just a typo.

Issue #529, opened 2011-11-14
JMail is ignoring name parameter on addAttachment


dami2:
FIle: libraries\joomla\mail\mail.php
Lines: 254 and 258

PR #1026, opened 2012-03-23
Bug fix for Issue #529

diananeculai:
JMail is ignoring name parameter on addAttachment.

### DIFF
--- a/libraries/joomla/mail/mail.php
+++ b/libraries/joomla/mail/mail.php
@@ -279,9 +279,21 @@ class JMail extends PHPMailer
 		{
 			if (is_array($attachment))
 			{
-				foreach ($attachment as $file)
+				if(!empty($name) && sizeof($attachment) != sizeof($name))
 				{
-					parent::AddAttachment($file, $name, $encoding, $type);
+					throw new RuntimeException("The number of attachments must be equal with the number of name");
+				}
+					
+				for ($i=0; $i<sizeof($attachment); $i++)
+				{
+					if (!empty($name))
+					{
+						parent::AddAttachment($attachment[$i], $name[$i], $encoding, $type);
+					}
+					else
+					{
+						parent::AddAttachment($attachment[$i], $encoding, $type);
+					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


realityking:
You have a few code style error you'll need to fix: http://developer.joomla.org/pulls/pulls/1026.html

libraries/joomla/mail/mail.php
```diff
@@ -279,9 +279,21 @@ public function addAttachment($attachment, $name = '', $encoding = 'base64', $ty
...
 			if (is_array($attachment))
 			{
-				foreach ($attachment as $file)
+				if(!empty($name) && sizeof($attachment) != sizeof($name))
 				{
-					parent::AddAttachment($file, $name, $encoding, $type);
+					throw new RuntimeException("The number of attachments must be equal with the number of name");
+				}
+					
+				for ($i=0; $i<sizeof($attachment); $i++)
```
realityking: I think we should use foreach($attachment $key => $file) instead of the for loop and document that $attachment and $name have to have the same keys. This way the keys don't have to be numerical and without gaps.

libraries/joomla/mail/mail.php
```diff
@@ -279,9 +279,21 @@ public function addAttachment($attachment, $name = '', $encoding = 'base64', $ty
...
+				{
+					if (!empty($name))
+					{
+						parent::AddAttachment($attachment[$i], $name[$i], $encoding, $type);
+					}
+					else
+					{
+						parent::AddAttachment($attachment[$i], $encoding, $type);
```
realityking: This doesn't fit the method signature of AddAttachment():

```
public function AddAttachment($path, $name = '', $encoding = 'base64', $type = 'application/octet-stream') {
```

elinw:
Hi Diana,

Even though the rest of the test isn't implemented, would you add a test for AddAttachement?
https://github.com/joomla/joomla-platform/blob/staging/tests/suites/unit/joomla/mail/JMailTest.php

PR #1030, opened 2012-03-23
Deprecating JNode and JTree and switching off code coverage

Hackwar:
JNode and JTree are classes that are not usable. If someone ever extended these classes, he had to rewrite all methods. Thus deprecating it directly for 12.1.

This is a new pull request, since I messed up the two earlier ones. The earlier requests are #513 and #677.


robschley:
Hannes,

You need to update the require_once statements in the top of the tests. The tests fail after the files are moved because they are trying to load the files in the old location. 

PHP Fatal error:  require_once(): Failed opening required '.../libraries/joomla/base/tree.php' (include_path='.:/usr/share/php:/usr/share/pear') in .../tests/suites/unit/legacy/base/JNodeTest.php on line 10

Hackwar:
fixed

ianmacl:
Fatal error: Class 'JNode' not found in .../tests/suites/unit/legacy/base/JNodeTest.php on line 27

Hackwar:
Ok, this gets annoying. How complicated can it be to get this right? Anyway, the tests now pass on my end. Please try again. And if you are at JAB, I owe you all a beer.

realityking:
Could you register the new class locations in import.legacy.php please. Otherwise existing code would have to be updated for the new location.

realityking:
Looks good to me, I'll just wait for the pull tester to run over it.

One last concern (we can address that after the merge, no biggie): deprecated 12.1 makes no sense, this would mean it won't be present in the next release and we might as well delete it now. IMO it should be 12.3.

realityking:
Something fishy is going on with the pull tester, could you push something else in to the branch? You could set the deprecated tags higher for example ;)

elkuku:
Nope, no :fish:es - the pull tester is right - seems that Hannes merged the staging branch when it contained the checkstyle errors ;)

realityking:
It's not so much the errors but the 4 "missing" warnings that left me wondering.

elkuku:
Well those are the result of moving two files to the legacy folder which is not part of the tests (yet) ([ref](https://gist.github.com/2191139)) - I think this should be fixed..

realityking:
That's what I'm thinking but I'd like to be sure ;)

eddieajau:
Checkstyle errors are in libraries/joomla/html/html.php
See http://developer.joomla.org/pulls/logs/1030checkstyle.xml

realityking:
Those are leftovers from when staging was tainted, just pushing something into the branch should fix it.

eddieajau:
Merged. Fixed up removal version manually.

PR #1031, opened 2012-03-24
Rename JDate->_tz to JDate->tz

elkuku:
This will rename a protected member of JDate to comply with the coding standards. It is an attempt to break up #637 into smaller pieces ;)

### DIFF
--- a/libraries/joomla/utilities/date.php
+++ b/libraries/joomla/utilities/date.php
@@ -66,7 +66,7 @@ class JDate extends DateTime
 	 * The DateTimeZone object for usage in rending dates as strings.
 	 *
 	 * @var    DateTimeZone
-	 * @since  11.1
+	 * @since  12.1
 	 */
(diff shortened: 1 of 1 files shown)


libraries/joomla/utilities/date.php
```diff
@@ -65,16 +65,16 @@ class JDate extends DateTime
 	/**
 	 * The DateTimeZone object for usage in rending dates as strings.
 	 *
-	 * @var    object
+	 * @var    DateTimeZone
 	 * @since  11.1
```
eddieajau: Just need to change to @since 12.1

elkuku:
Hmm.. commit --amend works the same way in PHPStorm and in Eclipse - it doesn't :(
Next time I'll use the always reliable command line - promise :v:

robschley:
Merged. Thanks.

PR #1036, opened 2012-03-24
Add JTEST_HTTP_STUB to phpunit config

mbabker:
Adds the JTEST_HTTP_STUB constant, used in JHttpTransportTest, to the phpunit.xml.dist file.  Default path assumes the platform is checked out at your web root.


robschley:
Merged. Thanks.

PR #1040, opened 2012-03-25
Update libraries/joomla/input/input.php

stevenloppe:
Unless I am missing something very obvious this looks like a copy and paste error?


mbabker:
Your change looks right to me, but I've been wrong before.  Could you also update the docblock to remove the info about the params?

realityking:
Thank you!

PR #1042, opened 2012-03-25
Allow JFormFieldInteger to iterate backwards using a negative step.

### LINKED ISSUE #983: issue with JFormFieldInteger
JFormFieldInteger  does not allow me to do this

<input type="integer" first="100" last="0" step="-1" />

I want 100 to be the first number and 0 to be the last but in the code:

for ($i = $first; $i <= $last; $i += $step)

will never run

aaronschmitz:
This code will make fields such as <field type="integer" first="10" last="0" step="-1"... work as expected.
In addition, this code prevents infinite loops that occur when for xml like <field type="integer" first="0" last="0" step="-1"...

Fixes #983


robschley:
You have some code style issues to fix before this can be merged. Make sure you run phpcs over your code before submitting pull requests.

aaronschmitz:
Oops. Thanks for pointing that out. It should be clean now, but let me know if anything else needs changed.

elinw:
Aaron also would you add this use case to the test?
https://github.com/joomla/joomla-platform/blob/staging/tests/suites/unit/joomla/form/fields/JFormFieldIntegerTest.php

eddieajau:
Merged. Thanks.

PR #1043, opened 2012-03-25
Fixed conflict between JGrid and JHtmlGrid when autoloaded.

bembelimen:
Steps to reproduce the error:
1. load JGrid: jimport('joomla.html.grid');
2. call a JHtmlGrid method: JHtml::_('grid.id', $i, $item->id)
3. Now JHtml::_() calls "class_exists" for JHtmlGrid
4. when not found, the Joomla! autoloader tries to find the path for the class JHtmlGrid.
5. the autoloader convert the class JHtmlGrid to the path (libraries/joomla/)html/grid.php
6. This file exists (of course) and will be loaded (again, see step 1), but unfortunately this is the wrong file
7. => Fatal error: Cannot redeclare class JGrid in html\grid.php on line 22

Solution:

The autoloader shouldn't be called in JHtml::_() and JHtml should handle the loading by itself

### DIFF
--- a/libraries/joomla/html/html.php
+++ b/libraries/joomla/html/html.php
@@ -108,7 +108,7 @@ abstract class JHtml
 
 		$className = $prefix . ucfirst($file);
 
-		if (!class_exists($className))
+		if (!class_exists($className, false))
 		{
 			$path = JPath::find(self::$includePaths, strtolower($file) . '.php');
 			if ($path)
--- a/libraries/joomla/html/html.php
+++ b/libraries/joomla/html/html.php
@@ -115,7 +115,7 @@ abstract class JHtml
 			{
 				require_once $path;
 
-				if (!class_exists($className))
+				if (!class_exists($className, false))
 				{
 					throw new InvalidArgumentException(sprintf('%s not found.', $className), 500);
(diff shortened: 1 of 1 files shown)


elkuku:
I would like to confirm that error.
A small POC:

``` php
<?php
jimport('joomla.html.grid');
$foo = new JGrid;
echo JHtml::_('grid.state', 1);
```

Expected result:

```
nothing
```

Actual result:

```
Fatal error: Cannot redeclare class JGrid in html/grid.php on line 22
```

LouisLandry:
Thanks!

realityking:
This actually makes trouble in the CMS since we started relying on the autoloader for the CMS. Shouldn't we just move JGrid to a different package where it can actually be autoloaded?

Also I think it's time to phase out the dynamic nature of JHtml::_. Thanks to the autoloader it's now fairly easy to overwrite classes one by one - we'd loose the overriding of individual methods thou.

LouisLandry:
Move the file if it is going to continue to be called JGrid IMO.  I'm tempted to say move that to the CMS tree as well ... but I'm not quite sure yet.

PR #1044, opened 2012-03-26
Rename protected members in JProfiler

elkuku:
This might be another class that can be safely renamed to comply with the coding standards.


eddieajau:
Good stuff, thanks.  Merged.

PR #1052, opened 2012-03-27
Remove dependency on several classes from the database package.

realityking:
Classes the dependency on is removed:
JError
JObject
JFolder
JTable

The JTable one may be a bit controversial since it meant I had to remove JDatabaseFactory::getTable() but considering it's just a proxy for JTable::getInstance() I don't see it as being worth adding an additional dependency - especially considering that the table package has it's own very large set of dependencies.

I also made JDatabaseException a subclass of RuntimeException. This way catching RuntimeExceptions (which are now used in the database package) will also catch JDatabaseExceptions.

### DIFF
--- a/libraries/joomla/database/driver.php
+++ b/libraries/joomla/database/driver.php
@@ -180,20 +180,22 @@ abstract class JDatabaseDriver extends JDatabase implements JDatabaseInterface
 		// Instantiate variables.
 		$connectors = array();
 
-		// Get a list of types.
-		$types = JFolder::files(__DIR__ . '/driver');
+		// Get an iterator and loop trough the driver classes.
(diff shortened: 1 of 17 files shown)


eddieajau:
Pull tester choked on this one.

realityking:
Yeah, forgot to change the tests to match the code. Hope it works now.

AmyStephen:
I have read a brief comment about changes coming for JTable but nothing more specific, so I not sure what the thinking is in that area. To me, JTable classes are best redefined as models. I agree with separating JTable from the JDatabase package. In my thinking the database package should have no specific table or column names or "trigger" actions. Would prefer to see it more cleanly defined as a utility of database operations, nothing more.

realityking:
Dependency on JError has been removed in #1104, so I've taken that part out.

PR #1054, opened 2012-03-27
Rename protected members in JBrowser

elkuku:


realityking:
That one if fine with me thou ;) Merged.

PR #1055, opened 2012-03-27
Rename a public member in JBuffer

elkuku:
This will also modify the corresponding unit test

### DIFF
--- a/libraries/joomla/utilities/buffer.php
+++ b/libraries/joomla/utilities/buffer.php
@@ -41,9 +41,9 @@ class JBuffer
 	 * Buffer hash
 	 *
 	 * @var    array
-	 * @since  11.1
+	 * @since  12.1
 	 */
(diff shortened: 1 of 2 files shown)


realityking:
Yeah, I'm gonna get the backward compatibility club out on this one. Since this is a public member IMO it warrants some B/C treatment (which most likely means __get and __set)

eddieajau:
Oops, sorry, didn't see your comment there Rouven.  The class seems to be loaded obscurely in the FTP client but never used.  Seems to be a left over from a by-gone era?

realityking:
If that's the case shouldn't we just deprecate it and remove the entire class at a later date? (move to legacy now)

eddieajau:
I think that's appropriate.

realityking:
I just checked and it's actually used quite a bit in JClientFTP so moving it to legacy is not an option.

PR #1057, opened 2012-03-27
Add optional identity to JApplicationBase; make dependency injection more flexible in JApplicationWeb

eddieajau:
Add optional "identity" has been added to JApplicationBase. This is in preparation for being able to support applications that act as web services, or connect with other web services that need to have some sort of identity (most probably a user but other cases are possible).  The loadDispatcher method has been made public and a new loadIdentity method (coupled, for now, to JUser) has been added.  If the application needs either of these dependancies, it simply calls the loader.  Passing no argument loads a default object but dependency injection is allowed by passing an object to the method.  If neither is called, then no dispatcher or identity support is assumed.

The initialise method in JApplicationWeb has been deprecated in favour of making the loadDocument, loadLanguage and loadSession public which allows for better scaling for more optional dependancies in the future. The tri-state behaviour is the same as for the loaders in JApplicationBase.  No call means the object is not supported, a call without argument loads a default object and a call with an argument injects the object into the application.

All dependency loaders are chainable.

### DIFF
--- a/libraries/joomla/application/base.php
+++ b/libraries/joomla/application/base.php
@@ -51,6 +51,7 @@ abstract class JApplicationBase extends JObject
 	 *
 	 * @return  void
 	 *
+	 * @codeCoverageIgnore
 	 * @since   12.1
 	 */
 	public function close($code = 0)
--- a/libraries/joomla/application/base.php
+++ b/libraries/joomla/application/base.php
@@ -21,20 +21,28 @@ jimport('joomla.event.dispatcher');
 abstract class JApplicationBase extends JObject
 {
 	/**
-	 * The application input object.
-	 *
-	 * @var    JInput
-	 * @since  12.1
-	 */
-	public $input = null;
-
-	/**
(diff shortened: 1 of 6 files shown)


realityking:
Could you explain a bit more what $identity is for? Is it supposed to always hold an object for the current user?

eddieajau:
For now, it will hold a JUser object, but later on it will hold a generic "identity" from which JUser will be derived (think UCM for users and groups).  So this could be the logged-in user for a CMS, it could be an authenticated identity using oAuth, it could be a system ("root") user used by a command line application, or it could be ignored completely.  The reason it's going into JApplicationBase is because both JApplicationCli (imagine a backend scheduler service that acting on behalf of actual user accounts, so it injects "your" identity to run as if it was "you") and JApplicationWeb could use it (any number of uses there).

Does that help?

realityking:
Yes it does :)

LouisLandry:
Thanks!

PR #1059, opened 2012-03-28
JText the database connectors in the form field

mbabker:
This allows downstream applications to use a text string for the database options when using the Database Connection form field.  For example, currently when installing the CMS, the option for SQL Server is displayed as "Sqlsrv".  With this change, a language string `SQLSRV="SQL Server"` could be used instead to avoid the use of shorthand driver names.  Of course, if an application decides they don't want to use this feature, then they won't need to with the way the language key is set up.


realityking:
Thanks!

PR #1061, opened 2012-03-28
Stop ignoring files that are not present in /database/driver

elinw:


elinw:
Doing my part to decrease PMD.

realityking:
Good catch, I found those a few days ago too and got a different fix in #1052: https://github.com/joomla/joomla-platform/pull/1052/files#diff-0

I'll leave this open in case mine doesn't get accepted.

chdemko:
#1052 has been merged

PR #1066, opened 2012-03-29
Archive: Exception and return JError Processing

AmyStephen:


realityking:
The changes look good to me. One thing that 

We do have some testing of exceptions just very little and unfortunately not all can even be tested - at least I don't know of a way.

A couple of suggested changes:
-Could you also change JInstallerHelper::unpack() to expect exceptions? Currently it expect false to be returned when an error occurs.
-Please add @throws tags to all methods that throw exceptions

And a question, why are you using LengthException for 'Invalid Zip Data'?

libraries/joomla/archive/zip.php
```diff
@@ -126,18 +126,16 @@ public function extract($archive, $destination, array $options = array())
...
 		{
-			$this->set('error.message', 'Archive does not exist');
-
-			return false;
+			throw new RuntimeException('Archive does not exist');
 		}
 
 		if ($this->hasNativeSupport())
 		{
-			return ($this->_extractNative($archive, $destination, $options)) ? true : JError::raiseWarning(100, $this->get('error.message'));
```
realityking, line 136: You'll need to fix the docblock on the method if we don't return a boolean anymore.

AmyStephen, line 136: Totally forgot docblocks, thanks.

AmyStephen:
@realityking changed those length to runtime exceptions. (Used length due to the checking of the file string length, but your comment made me realize that was silly.) Hit the docblocks, throw tags and adapted returns, as needed.

Please look at the InstallerHelper and let me know if there is a better approach.

Thanks :)

ianmacl:
None of these methods return exceptions so the doc blocks should not claim that they do.

libraries/joomla/archive/archive.php
```diff
@@ -21,11 +21,12 @@ class JArchive
...
 	 * Extract an archive file to a directory.
 	 *
+	 * @static
 	 * @param   string  $archivename  The name of the archive file
 	 * @param   string  $extractdir   Directory to unpack into
 	 *
-	 * @return  boolean  True for success
-	 *
+	 * @return  mixed  Boolean or Exception
+	 * @throws  InvalidArgumentException
```
LouisLandry: You need to put a spacer line after @return just like after the @param group.

Also, I don't think it's going to return an exception, but rather throw it.  If this is the case simply having:

```
@return  boolean  True on success.

@throws  InvalidArgumentException
```

Should get it done.

libraries/joomla/archive/archive.php
```diff
@@ -167,10 +166,10 @@ public static function extract($archivename, $extractdir)
 	 *
 	 * @param   string  $type  The type of adapter (bzip2|gzip|tar|zip).
 	 *
-	 * @return  object  JArchiveExtractable
+	 * @return  mixed object JArchiveExtractable or BadMethodCallException
 	 *
+	 * @throws  BadMethodCallException
```
LouisLandry: The `@throws` tag should come after the `@since` tag.  That last grouping is in alpha order.

libraries/joomla/archive/archive.php
```diff
@@ -188,7 +187,7 @@ public static function getAdapter($type)
 
 			if (!class_exists($class))
 			{
-				throw new Exception(JText::_('JLIB_FILESYSTEM_UNABLE_TO_LOAD_ARCHIVE'), 500);
+				throw new BadMethodCallException('Unable to load archive', 500);
```
LouisLandry, line 191: Nice use of semantic exception here.

LouisLandry:
In general this is great progress Amy.  I left you a couple of line notes with respect to docblocks.  That'll need to get rinsed and repeated all over, but the work is good.  Thanks.

libraries/joomla/archive/archive.php
AmyStephen: Wondered about that. Good.

AmyStephen:
Thanks for the feedback -- getting this first one right will serve as a useful pattern for me for the others. Will get this updated later today when I get a few moments.

AmyStephen:
Please use https://github.com/joomla/joomla-platform/pull/1071

This PR will be closed.

PR #1067, opened 2012-03-29
Fix for issue 183 - allow additional characters in JFolder

aaronschmitz:
Allowed characters based on http://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words

Additional characters allowed in JFolder names: ()[]{}#$^+.'~`!@&=;,
Disallow : in JFolder names


chdemko:
see #1007

aaronschmitz:
Are we depreciating the current file system?

elinw:
Christophe I don't understand the reference. Yes you have a pull request but from what i can tell first, you have not addressed this issue anywhere in your pull request (in fact JFolder and makeSafe both seem to have disappeared).  If you are asking Aaron to make a similar pull request against your repo that's fine but you need to explain that not expect people to figure that out. There's no problem with people touching the same files as long as everyone is using git correctly.

In the mean time this is a highly useful and important bug fix ( : for sure should not be in the regex).

chdemko:
it's just a reference to the new file system because the old one will be deprecated

elinw:
So most of us can't read minds, it's important to explain what you mean when you paste a link.
I'm not sure what the fact that it will be deprecated has to do with the bug fix though.

aaronschmitz:
As far as I can tell https://github.com/joomla/joomla-platform/pull/1007 isn't going to be ready for a while. Do we want to get this fixed before we tag 12.2?

LouisLandry:
Thanks Aaron!

PR #1069, opened 2012-03-29
Fix JForm::load() not replacing form field in same location (Issue #129)

aaronschmitz:
This code causes JForm to replace fields in place when a new definition is loaded to replace the old definition. Duplicate fields will now remain in their original location rather than being moved to the end of the document.

### DIFF
--- a/libraries/joomla/form/form.php
+++ b/libraries/joomla/form/form.php
@@ -712,11 +712,14 @@ class JForm
 				if ($current = $this->findField((string) $field['name'], implode('.', $groups)))
 				{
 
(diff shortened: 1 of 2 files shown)


LouisLandry:
What are the odds of getting you to write unit tests for this?

aaronschmitz:
@LouisLandry what are the odds of me getting to do GSoC with you guys? The two are about the same :D
Just kidding - I'll added them ASAP.

elinw:
Aaron,
 It would be awesome to get this into the cms release, would you want to make the same pull request on the cms repo?

PR #1071, opened 2012-03-30
Archive Package: Exception and JError updates

AmyStephen:
Consolidated all changes. (Reforked so that I can set up my repository for branching better. )

Previously discussed https://github.com/joomla/joomla-platform/pull/1066 (sorry if this is not appropriate, trying to learn to rebase so I don't use this approach in the future.)

### DIFF
--- a/libraries/joomla/archive/archive.php
+++ b/libraries/joomla/archive/archive.php
@@ -27,6 +27,7 @@ class JArchive
 	 * @return  boolean  True for success
 	 *
 	 * @since   11.1
+	 * @throws  InvalidArgumentException
 	 */
 	public static function extract($archivename, $extractdir)
 	{
@@ -149,9 +150,7 @@ class JArchive
 				break;
 
 			default:
-				JLog::add(JText::_('JLIB_FILESYSTEM_UNKNOWNARCHIVETYPE'), JLog::WARNING, 'jerror');
-				return false;
-				break;
+				throw new InvalidArgumentException('Unknown Archive Type');
 		}
 
 		if (!$result || $result instanceof Exception)
@@ -167,10 +166,10 @@ class JArchive
 	 *
 	 * @param   string  $type  The type of adapter (bzip2|gzip|tar|zip).
 	 *
-	 * @return  object  JArchiveExtractable
+	 * @return  object JArchiveExtractable
 	 *
 	 * @since   11.1
-	 * @throws  Exception
+	 * @throws  BadMethodCallException
 	 */
 	public static function getAdapter($type)
 	{
@@ -188,7 +187,7 @@ class JArchive
 
 			if (!class_exists($class))
 			{
-				throw new Exception(JText::_('JLIB_FILESYSTEM_UNABLE_TO_LOAD_ARCHIVE'), 500);
+				throw new BadMethodCallException('Unable to load archive', 500);
 			}
 
 			$adapters[$type] = new $class;
--- a/libraries/joomla/archive/gzip.php
+++ b/libraries/joomla/archive/gzip.php
@@ -50,6 +50,7 @@ class JArchiveGzip implements JArchiveExtractable
 	 * @return  boolean  True if successful
 	 *
 	 * @since   11.1
+	 * @throws RuntimeException
 	 */
 	public function extract($archive, $destination, array $options = array ())
 	{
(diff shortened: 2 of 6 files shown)


libraries/joomla/archive/archive.php
```diff
@@ -21,12 +21,14 @@ class JArchive
 	/**
 	 * Extract an archive file to a directory.
 	 *
+	 * @static
```
realityking: Is there a reason you added this? We removed all of them a while back because doc generators are capable of seeing the strict keyword before function declarations and don't need this tag anymore. It just means we have to maintain the same information in two placed.

libraries/joomla/archive/tar.php
```diff
@@ -76,13 +77,12 @@ public function extract($archive, $destination, array $options = array())
...
 		{
-			$this->set('error.message', 'Unable to read archive');
-			return JError::raiseWarning(100, $this->get('error.message'));
+			throw new RuntimeException('Unable to read archive');
 		}
 
 		if (!$this->_getTarInfo($this->_data))
 		{
-			return JError::raiseWarning(100, $this->get('error.message'));
+			throw new RuntimeException($this->get('error.message'));
```
realityking: This won't work since error.message is never set. I think this was complimentary to line 155/156:

```
 $this->set('error.message', 'Unable to decompress data');
 return false;
```

I think you can simplify the whole thing to just this:

```
 $this->_getTarInfo($this->_data);
```

The exception will just fall trough than.

libraries/joomla/archive/zip.php
```diff
@@ -126,18 +126,16 @@ public function extract($archive, $destination, array $options = array())
 	{
 		if (!is_file($archive))
 		{
-			$this->set('error.message', 'Archive does not exist');
-
-			return false;
+			throw new RuntimeException('Archive does not exist');
```
realityking, line 130: You're missing the throws tag on this method.

libraries/joomla/archive/gzip.php
```diff
@@ -58,31 +58,26 @@ public function extract($archive, $destination, array $options = array ())
 
 		if (!extension_loaded('zlib'))
 		{
-			$this->set('error.message', JText::_('JLIB_FILESYSTEM_GZIP_NOT_SUPPORTED'));
-
-			return JError::raiseWarning(100, $this->get('error.message'));
+			throw new RuntimeException('The zlib extension is not available.');
```
realityking, line 62: You're missing the throws tag on extract() and _getFilePosition() in this file.

libraries/joomla/installer/helper.php
```diff
@@ -104,11 +104,12 @@ public static function unpack($p_filename)
...
-		$result = JArchive::extract($archivename, $extractdir);
-
-		if ($result === false)
+		try {
+			JArchive::extract($archivename, $extractdir);
+		}
+		catch (Exception $e)
 		{
-			return false;
+			return $e->getMessage();
```
realityking: I'd must leave this at return false so we don't change the API. (A message which is a non-empty string would evaluate to true)

libraries/joomla/archive/archive.php
AmyStephen: Probably just thought it was missing. Will remove that - thanks for clarifying.

libraries/joomla/archive/gzip.php
AmyStephen, line 62: Thanks.

libraries/joomla/archive/tar.php
AmyStephen: Excellent.

libraries/joomla/archive/zip.php
AmyStephen, line 130: Check.

libraries/joomla/installer/helper.php
AmyStephen: Good point, Rouven.

libraries/joomla/archive/archive.php
```diff
@@ -188,7 +187,7 @@ public static function getAdapter($type)
 
 			if (!class_exists($class))
 			{
-				throw new Exception(JText::_('JLIB_FILESYSTEM_UNABLE_TO_LOAD_ARCHIVE'), 500);
+				throw new BadMethodCallException('Unable to load archive', 500);
```
realityking, line 190: I don't think that's the correct exception. From the PHP doc:

```
Exception thrown if a callback refers to an undefined method or if some arguments are missing.
```

The bad thing about semantic exceptions is that we can argue about the semantics :P

Maybe this should be an UnexpectedValueException since it will most likely occur when an unknown value for $type is set. If a file is missing it should probably be a RuntimeException.

Argh. Just do what you think makes sense ;)

AmyStephen, line 190: Definitely agree, UnexpectedValueException is a better fit.

LouisLandry:
This was merged manually.  Thanks Amy.

PR #1072, opened 2012-03-30
Base Package: Exception

AmyStephen:
Replace two "return false" with Exceptions. Updated calls to modified method and unit tests.

### DIFF
--- a/libraries/joomla/base/adapter.php
+++ b/libraries/joomla/base/adapter.php
@@ -100,6 +100,8 @@ class JAdapter extends JObject
 	 * @return  boolean  True if successful
 	 *
 	 * @since   11.1
+	 * @throws BadMethodCallException
+	 * @throws InvalidArgumentException
 	 */
 	public function setAdapter($name, &$adapter = null, $options = array())
 	{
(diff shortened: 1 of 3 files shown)


realityking:
Hi Amy,

thanks a lot for working on these. 2 points:
1. Docblocks ;) We need the throws tags
2. JUpdater also extends JAdapter, I haven't checked but I suspect it needs to be changed as well.

AmyStephen:
The updater does use the setAdapter - but it issues a CONTINUE if it the adapter is not set. (So, don't think it needs change.)

https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/updater/updater.php#L95

Got the docblock and accidentally pulled in a cache change. Patience while I try to remove that. =)

Thanks!

AmyStephen:
Well, look at me. I do believe I have learned how to rebase. That will come in handy since I make a lot of mistakes. ;-)

eddieajau:
Just wondering, why is the `return false` being changed?

LouisLandry:
I'd vote for closing this one for now.  I don't know that I think it makes sense to be throwing exceptions instead of returning false in these cases, and there isn't a real issue needing to be solved here.  Let's think it through fully before we make a move.

PR #1073, opened 2012-03-30
Client Package: Change one return JError to Exception

AmyStephen:


libraries/joomla/client/helper.php
```diff
@@ -216,7 +217,7 @@ public static function setCredentialsFromRequest($client)
 			}
 			else
 			{
-				$return = JError::raiseWarning('SOME_ERROR_CODE', JText::_('JLIB_CLIENT_ERROR_HELPER_SETCREDENTIALSFROMREQUEST_FAILED'));
```
eddieajau, line 219: Let's do the class_exists('JError') trick.  If JError is present, turn the JError, otherwise throw the exception.

PR #1074, opened 2012-03-30
Crypt Package: Add Exception to Docblock

AmyStephen:


libraries/joomla/crypt/key.php
```diff
@@ -74,7 +75,7 @@ public function __get($name)
 		}
 		else
 		{
-			trigger_error('Cannot access property ' . __CLASS__ . '::' . $name, E_USER_WARNING);
+			throw new BadMethodCallException('Cannot access property ' . __CLASS__ . '::' . $name);
```
realityking: Not sure about this one. I suspect the idea was to mimic the behavior of PHP when the property doesn't exist. With your change I'd have to add a try catch (while catching BadMethodCallException) around simple property accesses like $key->something. That seems weird to me and would probably be more confusing than helpful.

AmyStephen: Well, the other thing is, I really don't think this logic can ever be reached given how the object class is identified as a property of the method call parameter.

I'll go ahead and remove it based on your point.

libraries/joomla/crypt/key.php
```diff
@@ -65,6 +65,7 @@ public function __construct($type, $private = null, $public = null)
 	 * @return  mixed
 	 *
 	 * @since   12.1
+	 * @throws  BadMethodCallException
```
realityking: You forgot the docblock.

BTW the logic is reached when I try to access some property that just doesn't exist.

realityking:
Thanks Amy!

PR #1075, opened 2012-03-30
CMS Separation: Check if CMS constants are defined

AmyStephen:
For now, this would be helpful when using JURI outside of the CMS. 

(Later, passing in the application path will be helpful)


eddieajau:
Merged - not pretty, but it will get us out of trouble for a while.  Thanks!

AmyStephen:
No, it's not pretty (but I really appreciate it - thanks!)

PR #1076, opened 2012-03-30
Event Package: Change JError return to Exception

AmyStephen:


realityking:
Thanks! Small tipp for the future, it would be great if you could align the throws tag with the other docblock tags.

AmyStephen:
Good point, happy to fix those if you find more little problems like that. I am using branches and have figured the rebase out so feel free to send those back for me to fix. 

Thanks for your direction.

AmyStephen:
So, before I issue a PR, run the unit tests. Is that what I should be doing?

realityking:
That would be preferable yes, but I should have checked http://developer.joomla.org/pulls/pulls/1076.html before I pulled it so this is my fault.

PR #1077, opened 2012-03-30
fixed storing failed with invalid field in JTableAsset when $updateNulls...

vietvh:
... = true

Tracker item: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=28322

### DIFF
--- a/libraries/joomla/table/asset.php
+++ b/libraries/joomla/table/asset.php
@@ -61,6 +61,7 @@ class JTableAsset extends JTableNested
 	public function __construct($db)
 	{
 		parent::__construct('#__assets', 'id', $db);
+		unset($this->alias);
 	}
 
 	/**


elinw:
I actually came across this today but unsetting alias just means that I got an error with the next field. The real problem is that for some reason the category fields are attempting to save in the assets table. Are you instantiating JTableNested directly?

vietvh:
Hi Elin,

What did you mean "error with the next field"

Viet

elinw:
After you unset alias it will want you to unset path and then note and extension. The error message just tells you the first one that fails.

vietvh:
I got a quick fix by manually unset alias field in my class which extends JTable

elinw:
Do you have more fields in your table?

vietvh:
Yes, my table has around 20 fields

elinw:
If you change it to not updatenulls it works okay?

vietvh:
Yes my issue is directly related to $updateNulls

elinw:
So I have tried to reproduce the issue following your instructions in the CMS tracker on a clean database and I could not reproduce. That doesn't mean there is not a bug, just that that method couldn't reproduce it.

I think one question that must be answered is where this save to the alias field is coming from. Alias is not a required field in JTableNested (a property yes, a field no).  Usually the field comes from something like JTableContent or JTableCategory and that is where it would be saved.  JTable manages saving the asset for you if $this->_trackAssets is true, it is not something that you should do directly. 

I have in fact gotten this error one time, as I mentioned in the CMS tracker. I can't remember exactly why I did it, but i had instantiated JTableNested and that was why I said the other day that I think JTableNested should be abstract.

elinw:
So upon some investigation  in some of the cases where this has appeared it seems to be a case of a corrupted asset table. Can you please check your asset table in the case where you are seeing this error? Specifically you might look to see if there is mroe than one record with parent_id of 0. Also as mentioned the other reason this happens is if someone instantiates JTableNested directly which is why it should possibly be made abstract.

elinw:
I think I found the place generating the fatal error the problem is definitely corrupt assets or trying to access the asset table directly in the getRootId() method of JTableNested if the table that is being evaluated does not have an alias field.  Property_exists is not an adequate screen here, we need to know if the column exists before attempting the query. We also might want to allow another column name to be used e.g. name as in the case of JTableAsset.

elinw:
See #1665   @vietvh if you apply that change does it solve your issue?

vietvh:
@elinw Its good for me, closing my pull request

PR #1079, opened 2012-03-30
Cache Package: return JError

AmyStephen:


libraries/joomla/cache/storage.php
```diff
@@ -139,7 +141,7 @@ public static function getInstance($handler = null, $options = array())
 			}
 			else
 			{
-				return JError::raiseWarning(500, JText::sprintf('JLIB_CACHE_ERROR_CACHE_STORAGE_LOAD', $handler));
+				throw new RuntimeException('Cache Storage Class is not known.');
```
chdemko: Keep the sprintf

AmyStephen:
Thanks @chdemko - should be ready now.

libraries/joomla/cache/storage.php
```diff
@@ -114,7 +116,7 @@ public static function getInstance($handler = null, $options = array())
 			$handler = $conf->get('cache_handler');
 			if (empty($handler))
 			{
-				return JError::raiseWarning(500, JText::_('JLIB_CACHE_ERROR_CACHE_HANDLER_NOT_SET'));
+				throw new UnexpectedValueException(sprintf('Unable to load Cache Storage: %s', $handler));
```
realityking: UnexpectedValueException is great here but I don't think the message is accurate. This should only be thrown if no cache handler is set at all.

AmyStephen: Yup. You are right. Will change that.

AmyStephen:
@realityking - done, thanks

realityking:
Thanks!

PR #1080, opened 2012-03-30
Application Package: Exceptions, ->error, return JError

AmyStephen:


libraries/joomla/application/component/controller.php
```diff
@@ -279,7 +280,7 @@ public static function getInstance($prefix, $config = array())
 			}
 			else
 			{
-				throw new InvalidArgumentException(JText::sprintf('JLIB_APPLICATION_ERROR_INVALID_CONTROLLER', $type, $format));
+				throw new RuntimeException('Invalid class file');
```
chdemko: Why did you remove the sprintf ?
You could use something like RuntimeException(sprintf(' .... %s ... %s'), $type, $format));

libraries/joomla/application/component/controller.php
```diff
@@ -290,7 +291,7 @@ public static function getInstance($prefix, $config = array())
 		}
 		else
 		{
-			throw new InvalidArgumentException(JText::sprintf('JLIB_APPLICATION_ERROR_INVALID_CONTROLLER_CLASS', $class));
+			throw new InvalidArgumentException('Invalid Class');
```
chdemko: Same comment

libraries/joomla/application/component/controller.php
```diff
@@ -586,7 +587,7 @@ protected function createView($name, $prefix = '', $type = '', $config = array()
 
 				if (!class_exists($viewClass))
 				{
-					throw new Exception(JText::sprintf('JLIB_APPLICATION_ERROR_VIEW_CLASS_NOT_FOUND', $viewClass, $path), 500);
+					throw new RuntimeException('View Class not found', 500);
```
chdemko: Same comment

libraries/joomla/application/component/controller.php
```diff
@@ -838,7 +839,7 @@ public function getView($name = '', $type = '', $prefix = '', $config = array())
 			}
 			else
 			{
-				throw new Exception(JText::sprintf('JLIB_APPLICATION_ERROR_VIEW_NOT_FOUND', $name, $type, $prefix), 500);
+				throw new RuntimeException('View not found', 500);
```
chdemko: Same comment

libraries/joomla/application/component/model.php
```diff
@@ -449,7 +449,7 @@ public function getTable($name = '', $prefix = 'Table', $options = array())
 			return $table;
 		}
 
-		throw new Exception(JText::sprintf('JLIB_APPLICATION_ERROR_TABLE_NAME_NOT_SUPPORTED', $name), 0);
+		throw new InvalidArgumentException('Table not found');
```
chdemko: Same comment

libraries/joomla/application/component/modellist.php
```diff
@@ -128,11 +128,7 @@ public function getItems()
 		$items = $this->_getList($query, $this->getStart(), $this->getState('list.limit'));
 
 		// Check for a database error.
-		if ($this->_db->getErrorNum())
-		{
-			$this->setError($this->_db->getErrorMsg());
-			return false;
-		}
+		$this->_db->getErrorNum();
```
chdemko: What is the utilisty of this line?

libraries/joomla/application/component/view.php
```diff
@@ -755,7 +755,7 @@ public function loadTemplate($tpl = null)
 		}
 		else
 		{
-			throw new Exception(JText::sprintf('JLIB_APPLICATION_ERROR_LAYOUTFILE_NOT_FOUND', $file), 500);
+			throw new RuntimeException('Layout file not found', 500);
```
chdemko: Keep the sprintf

libraries/joomla/application/module/helper.php
```diff
@@ -334,7 +334,7 @@ protected static function &_load()
 
 		if ($db->getErrorNum())
```
chdemko: is it necessary to keep these lines ?

libraries/joomla/application/router.php
```diff
@@ -136,7 +136,7 @@ public static function getInstance($client, $options = array())
 			}
 			else
 			{
-				throw new Exception(JText::sprintf('JLIB_APPLICATION_ERROR_ROUTER_LOAD', $client), 500);
+				throw new RuntimeException(JText::sprintf('JLIB_APPLICATION_ERROR_ROUTER_LOAD', $client), 500);
```
chdemko: Keep sprintf but delete remove JText

libraries/joomla/application/component/controller.php
AmyStephen: I also changed the exception - will change that back.

libraries/joomla/application/component/modellist.php
AmyStephen: You are right, it needs to go.

libraries/joomla/application/module/helper.php
AmyStephen: Nope. They will never be executed. Good catch.

AmyStephen:
Thanks @chdemko - I believe this is ready now.

libraries/joomla/application/component/controller.php
```diff
@@ -586,7 +587,7 @@ protected function createView($name, $prefix = '', $type = '', $config = array()
 
 				if (!class_exists($viewClass))
 				{
-					throw new Exception(JText::sprintf('JLIB_APPLICATION_ERROR_VIEW_CLASS_NOT_FOUND', $viewClass, $path), 500);
+					throw new RuntimeException(sprintf('View class not found [class, file]: %1, %2', $viewClass, $path), 500);
```
elkuku: %1 should be %1$s as well as 2 ;)

AmyStephen: Thanks @elkuku :)

AmyStephen:
Should be ready with @chdemko and @elkuku 's corrections.

libraries/joomla/application/component/controller.php
```diff
@@ -210,7 +210,8 @@ protected static function createFileName($type, $parts = array())
 	 * @return  JController
 	 *
 	 * @since   11.1
-	 * @throws  Exception if the controller cannot be loaded.
+	 * @throws  InvalidArgumentException
+	 * @throws  RuntimeException
```
realityking: Is there a RuntimeException thrown in this method? I don't see it.

libraries/joomla/application/component/controlleradmin.php
```diff
@@ -94,7 +94,7 @@ public function __construct($config = array())
 			$r = null;
 			if (!preg_match('/(.*)Controller(.*)/i', get_class($this), $r))
 			{
-				throw new Exception(JText::_('JLIB_APPLICATION_ERROR_CONTROLLER_GET_NAME'), 500);
+				throw new RuntimeException('Controller not found', 500);
```
realityking: I think the message is inaccurate, the message from the lang files is the following:

```
JController::getName(): Cannot get or parse class name.
```

libraries/joomla/application/component/controllerform.php
```diff
@@ -92,7 +92,7 @@ public function __construct($config = array())
 			$r = null;
 			if (!preg_match('/(.*)Controller(.*)/i', get_class($this), $r))
 			{
-				throw new Exception(JText::_('JLIB_APPLICATION_ERROR_CONTROLLER_GET_NAME'), 500);
+				throw new RuntimeException('Controller not found', 500);
```
realityking: See above.

libraries/joomla/application/component/helper.php
```diff
@@ -396,10 +396,10 @@ protected static function _load($option)
 
 		self::$components[$option] = $cache->get(array($db, 'loadObject'), null, $option, false);
 
-		if ($error = $db->getErrorMsg() || empty(self::$components[$option]))
```
realityking, line 399: Not sure here, do we need to catch the db error to maintain the same behavior?

libraries/joomla/application/component/model.php
```diff
@@ -235,7 +235,7 @@ public function __construct($config = array())
 
 			if (!preg_match('/(.*)Model/i', get_class($this), $r))
 			{
-				throw new Exception(JText::_('JLIB_APPLICATION_ERROR_MODEL_GET_NAME'), 500);
+				throw new RuntimeException('Model not found', 500);
```
realityking: CMS error is this:

```
JModel::getName(): Can't get or parse class name.
```

libraries/joomla/application/component/model.php
```diff
@@ -393,7 +393,7 @@ public function getName()
 			$r = null;
 			if (!preg_match('/Model(.*)/i', get_class($this), $r))
 			{
-				throw new Exception(JText::_('JLIB_APPLICATION_ERROR_MODEL_GET_NAME'), 500);
+				throw new RuntimeException('Model class not found', 500);
```
realityking: See above.

realityking:
Should JModelAdmin, JControllerAdmin, JModelForm, JControllerForm etc. be even changed? IMO they belong into the CMS and should move to the legacy folder.

libraries/joomla/application/component/controlleradmin.php
AmyStephen: I'll change it back but that's message is too technical IMO for the reality of the problem.

AmyStephen:
@realityking - I agree with your comment that they belong in the legacy folder. In fact, I think the entire MVC should go there since it is to be rebuilt (if I understand correctly).

libraries/joomla/application/component/controller.php
AmyStephen: Nope. Not in there. Removed Docblock line.

libraries/joomla/application/component/helper.php
AmyStephen, line 399: If it's an empty query, it'll be returned and get the empty instance logic.  (I think it's consistent with other usages where we have simply been removing these.)

libraries/joomla/application/component/model.php
AmyStephen: Put geeky-tech-what-the-heck-does-that-mean? message back.

AmyStephen:
@realityking - Thanks for your feedback, changes made. Let me know if you decide specifically on the Helper, left comment for you on that, but no change.

eddieajau:
We should have a pull for the new MVC this week (cross fingers) and all of this stuff will move to the legacy tree, if that helps.

AmyStephen:
Should we go ahead and close this one?

eddieajau:
I think if there is a desire to accept the new MVC pull, it probably puts this to bed.

PR #1086, opened 2012-03-31
Table Package: db error and Exception/JText changes

AmyStephen:
I left JException in this PR since I am not certain what the plan is there and wanted to hear back https://groups.google.com/d/msg/joomla-dev-platform/SJopGknhizs/ssWpyIMazgAJ

We can go back through this and change JException to a PHP Exception even after this PR, if it's desired. More than likely, this will result in an API change since currently a failure is indicated by "false" not an exception. But, it's probably a good time to make that change with the 3.0 CMS release. 

This PR, however, is restricted to removing the $db error processing, getting rid of JText with Exception (and JException) calls, and any JError statments.

Also - IMO, the tables 'package' belongs in the CMS, not the framework. Maybe the entire folder should be moved into legacy and left, as is?

### DIFF
--- a/libraries/joomla/table/asset.php
+++ b/libraries/joomla/table/asset.php
@@ -87,12 +87,7 @@ class JTableAsset extends JTableNested
 		{
 			return false;
 		}
-		// Check for a database error.
-		if ($error = $this->_db->getErrorMsg())
-		{
-			$this->setError($error);
-			return false;
-		}
+
 		return $this->load($assetId);
 	}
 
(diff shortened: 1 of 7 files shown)


eddieajau:
Let's not use JException. I think most of the cases should be UnexpectedValueException and if you can do the sprintf thing that would be great.

AmyStephen:
OK, I will do that, thanks @eddieajau

AmyStephen:
Made the JException changes.

But, just now thinking, I'm guessing we're going to want to go back thru this and add the legacy JError class check. Not up for it tonight but if it's needed, I'll try to get it tomorrow night (along with any other feedback that might appear).

Thanks for your patience.

eddieajau:
No probs.  I think we only need to do the class_exists trick if we are returning a JError.

eddieajau:
Merged. Thanks.

PR #1087, opened 2012-03-31
Mail package: Exception and JError

AmyStephen:


libraries/joomla/mail/mail.php
```diff
@@ -67,23 +67,23 @@ public static function getInstance($id = 'Joomla')
 	/**
 	 * Send the mail
 	 *
-	 * @return  mixed  True if successful, a JError object otherwise
+	 * @return  mixed  True if successful
```
realityking: Please change the type to boolean.

AmyStephen:
Done. Thanks @realityking

realityking:
Now we have the next problem, your pull isn't mergeable anymore because #1096 was merged. Could you update it? Thanks.

AmyStephen:
Yea, that doesn't look good. I did a fetch/merge from the upstream remote http://help.github.com/fork-a-repo/

What should I have done?

AmyStephen:
Not sure how to do a merge (apparently), so for now, please use https://github.com/joomla/joomla-platform/pull/1097 instead 

I am closing this PR.

realityking:
Actually the merge looks pretty good but I'll use #1097.

libraries/joomla/mail/mail.php
```diff
@@ -67,23 +67,23 @@ public static function getInstance($id = 'Joomla')
...
+			throw new RuntimeException('Send function cancelled because mail is disabled.');
 		}
 
 		@$result = parent::Send();
 
 		if ($result == false)
 		{
-			// TODO: Set an appropriate error number
-			$result = JError::raiseNotice(500, JText::_($this->ErrorInfo));
+			throw new RuntimeException('phpMail Send email failed');
```
realityking: Again, don't we loose information if we don't pas the phpMail error along?

AmyStephen: Of course, thank you for seeing this. My apologies.

AmyStephen:
So, that merge was okay? I guess I didn't expect to see all the files changed in the main repository. 

I'll reopen this then, make the fix, and close the other after I commit it. 

thanks.

libraries/joomla/mail/mail.php
```diff
@@ -67,23 +67,23 @@ public static function getInstance($id = 'Joomla')
...
 	 *
 	 * @since   11.1
+	 * @throws  RuntimeException
 	 */
 	public function Send()
 	{
 		if (($this->Mailer == 'mail') && !function_exists('mail'))
 		{
-			return JError::raiseNotice(500, JText::_('JLIB_MAIL_FUNCTION_DISABLED'));
+			throw new RuntimeException('Send function cancelled because mail is disabled.');
```
eddieajau: This is a change in behavior that will affect B/C.  I think what we'll have to do is this:

<pre>if (class_exists('JError'))
{
    return JError::raiseNotice(500, JText::_('JLIB_MAIL_FUNCTION_DISABLED'));
}
else
{
    throw new RuntimeException('Send function cancelled because mail is disabled.');
}</pre>


If the legacy platform is loaded, it will still return a JError.  If it's not loaded, it will throw an exception.  @return can change to "JError only  if the legacy platform is loaded, otherwise an exception is thrown.".

AmyStephen: @eddieajau - 

That brings up a question I have. JError is deprecated 12.1 

https://github.com/joomla/joomla-platform/blob/staging/libraries/legacy/error/error.php#L32

Doesn't that mean (I am not certain) that we take out all JError code at that time?

eddieajau: I think we negotiated with Rouven to remove JError in 13.1 on the condition we could put it in the legacy tree.

AmyStephen:
OK - I think Mail is ready. It has both the legacy and new approaches. 

Thanks for the feedback and help.

libraries/joomla/mail/mail.php
```diff
@@ -67,23 +67,37 @@ public static function getInstance($id = 'Joomla')
 	/**
 	 * Send the mail
 	 *
-	 * @return  mixed  True if successful, a JError object otherwise
+	 * @return  boolean  True if successful
```
eddieajau, line 70: Just need to fix up this return information.

libraries/joomla/mail/mail.php
```diff
@@ -67,23 +67,37 @@ public static function getInstance($id = 'Joomla')
...
 		{
-			// TODO: Set an appropriate error number
-			$result = JError::raiseNotice(500, JText::_($this->ErrorInfo));
+			if (class_exists('JError'))
+			{
+				$result = JError::raiseNotice(500, JText::_($this->ErrorInfo));
+			}
+			else
+			{
+				throw new RuntimeException('Send function cancelled because mail is disabled.');
```
eddieajau, line 99: and we should probably send the $this->ErrorInfo back as the exception message.  Nearly there :)

eddieajau:
Merged. Thanks.

PR #1090, opened 2012-03-31
Form Package: db->error, Exception/JText modifications

AmyStephen:


libraries/joomla/form/form.php
```diff
@@ -301,15 +301,14 @@ public function getField($name, $group = null, $value = null)
...
 	 */
 	public function getFieldAttribute($name, $attribute, $default = null, $group = null)
 	{
 		// Make sure there is a valid JForm XML document.
 		if (!($this->xml instanceof JXMLElement))
 		{
-			// TODO: throw exception.
-
-			return $default;
+			throw new RuntimeException('XML is not an instance of JXMLElement');
```
eddieajau: I would throw this as an UnexpectedValueException like this:

```
throw new UnexpectedValueException(sprint('%s::getFieldAttribute `xml` is not an instance of JXMLElement', get_class($this)));
```

That way it's easier for the developer to home in on where the error is.  Ditto for the other cases.

libraries/joomla/form/form.php
```diff
@@ -1835,7 +1829,7 @@ protected function validateField($element, $group = null, $value = null, $input
 		// Make sure there is a valid JXMLElement.
 		if (!$element instanceof JXMLElement)
 		{
-			return new JException(JText::_('JLIB_FORM_ERROR_VALIDATE_FIELD'), -1, E_ERROR);
+			return new JException('Form field validation error', -1, E_ERROR);
```
eddieajau: I think we can throw an InvalidArgumentException here.

libraries/joomla/form/form.php
```diff
@@ -1880,7 +1874,7 @@ protected function validateField($element, $group = null, $value = null, $input
 			// If the object could not be loaded return an error message.
 			if ($rule === false)
 			{
-				return new JException(JText::sprintf('JLIB_FORM_VALIDATE_FIELD_RULE_MISSING', $type), -2, E_ERROR);
+				return new JException(sprintf('Validation Rule missing: %s', $type), -2, E_ERROR);
```
eddieajau: I think we should throw an UnexpectedValueException here.

libraries/joomla/form/form.php
```diff
@@ -1902,11 +1896,11 @@ protected function validateField($element, $group = null, $value = null, $input
 
 			if ($message)
 			{
-				return new JException(JText::_($message), 1, E_WARNING);
```
eddieajau, line 1905: We should return an InvalidArgumentException in this and the next case.

libraries/joomla/form/form.php
```diff
@@ -1902,11 +1896,11 @@ protected function validateField($element, $group = null, $value = null, $input
...
+				return new JException($message, 1, E_WARNING);
 			}
 			else
 			{
-				return new JException(JText::sprintf('JLIB_FORM_VALIDATE_FIELD_INVALID', JText::_((string) $element['label'])), 1, E_WARNING);
+				return new JException(sprintf('Invalid field: %s', (string) $element['label']), 1, E_WARNING);
```
eddieajau: I think we should change this to:

```
return new InvalidArgumentException((string) $element['label']);
```

The exception type takes care of the "invalid field" in the former message.

libraries/joomla/form/form.php
AmyStephen: @eddiejau - 

I didn't replace any of the JExceptions since those will end up changing the API (I believe). I put a question on list asking how you wanted to handle those. (The only change I made for the JExceptions was to simply replace the JText.)

I'll change this to InvalidArgumentException, just wanted to make you aware of the bigger question I have.

libraries/joomla/form/rules/equals.php
```diff
@@ -44,13 +44,13 @@ public function test($element, $value, $group = null, $input = null, $form = nul
 		// Check that a validation field is set.
 		if (!$field)
 		{
-			return new JException(JText::sprintf('JLIB_FORM_INVALID_FORM_RULE', get_class($this)));
```
eddieajau, line 47: I think this should "throw" and UnexpectedValueException (there is an error in the XML - do not pass Go or collect $200).

libraries/joomla/form/form.php
AmyStephen: Ok.

libraries/joomla/form/form.php
AmyStephen, line 1905: OK on both.

libraries/joomla/form/rules/equals.php
```diff
@@ -44,13 +44,13 @@ public function test($element, $value, $group = null, $input = null, $form = nul
...
+			return new JException(sprintf('Invalid Form Rule: :%s', get_class($this)));
 		}
 
 		// Check that a valid JForm object is given for retrieving the validation field value.
 		if (!($form instanceof JForm))
 		{
-			return new JException(JText::sprintf('JLIB_FORM_INVALID_FORM_OBJECT', get_class($this)));
+			return new JException(sprintf('Invalid Form Object: :%s', get_class($this)));
```
eddieajau: I think this should "throw" an UnexpectedValueException as well, but you can also put a type hint in the method arguments, e.g.:

```
public function test($element, $value, $group = null, $input = null, JForm $form = null)
```

This is unfortunate design in hindsight because $form should really be mandatory, but we can't do that because it's at the end of the list.

eddieajau:
Really good job overall Amy, just a bit of nuance to fix.

libraries/joomla/form/form.php
AmyStephen: Missed this one. Fixing now.

libraries/joomla/form/rules/equals.php
AmyStephen: Missed this one, too, fixing now.

AmyStephen:
All right, I think I have this one updated for @eddieajau 's feedback.

Thanks Andrew.

libraries/joomla/form/form.php
eddieajau: It should be fine because after `validdateField` is called, we do this check:

```
if ($valid instanceof Exception)
```

Since JException extends Exception we should be fine.  The extra arguments in JException will certainly be lost but I think we can cope with that.

AmyStephen: Great, thanks @eddieajau

eddieajau:
Merged. Thanks.

PR #1091, opened 2012-03-31
Installer Package: db->error removal

AmyStephen:


libraries/joomla/installer/adapters/component.php
```diff
@@ -1133,14 +1133,6 @@ public function uninstall($id)
 		$db->setQuery($query);
 		$db->execute();
 
-		// Check for errors.
-		if ($db->getErrorNum())
-		{
-			JLog::add(JText::_('JLIB_INSTALLER_ERROR_COMP_UNINSTALL_FAILED_DELETE_CATEGORIES'), JLog::WARNING, 'jerror');
```
realityking, line 1139: Would it make sense to throw a new exception with this as the message?

libraries/joomla/installer/adapters/component.php
```diff
@@ -1423,18 +1415,7 @@ protected function _removeAdminMenus(&$row)
 		$ids = $db->loadColumn();
 
 		// Check for error
-		if ($error = $db->getErrorMsg())
-		{
-			JLog::add(JText::_('JLIB_INSTALLER_ERROR_COMP_REMOVING_ADMIN_MENUS_FAILED'), JLog::WARNING, 'jerror');
```
realityking, line 1428: Same as above.

libraries/joomla/installer/adapters/component.php
AmyStephen, line 1139: Not sure, that's something that should be discussed though. Everywhere that these $db->getError msg/num values are - an Exception has already been thrown in the database. I assume the thinking is that the originating process will trap for exceptions. 

I'll add a post on list - I think this needs some discussion.

eddieajau, line 1139: Actually the installer will need some more thought, because you need to rollback changes. The installer is notorious for not giving good error messages so I think we'll need a different strategy.  I think the DB error message will be enough to pinpoint failure, but rather than logging the error, I think it would be better to JLog the successes so that a developer can trace through the point of failure.  So, every time we do something, we could do a:

`JLog::add(JText::_('JLIB_INSTALLER_LOG_SOMETHING'), JLog::INFO, 'jinstaller');`

com_installer can then add a message queue and possibly file log listener and then a big try-catch (where catch will rollback).

Does that make sense?

eddieajau:
Merged.  The CMS is going to have to add things like this:

<pre>try {
    $installer->install();
} catch (Exception $e) {
    $installer->abort($e->getMessage());
}</pre>


I also think it's worth doing a progress log but that's something that can be added later.  This pull addresses the immediate problem of getting JError out of the way.

Thanks again Amy.

Issue #660, opened 2011-12-19
Fatal error with JEditor::getButtons


tbela99:
the method [getButtons](https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/html/editor.php#L443) of JEditor contains code that is unsafe 

the following code test existance of a plugin class but do not care whether of not the class is missing. when 

//if this is not true (the plugin may be missing)
if (class_exists($className))
            {
                $plugin = new $className($this, (array) $plugin);
            }

```
        //this will call onDisplay on NULL
        if ($temp = $plugin->onDisplay($editor, $this->asset, $this->author))
        {
            $result[] = $temp;
        }
```

Buddhima:
Got the solution! I'll fix the issue very soon :)

eddieajau:
Sounds like this has been fixed.  It not re-open or resubmit the issue.

PR #1093, opened 2012-04-01
To avoid Fatal error with JEditor::getButtons (for issue #660)

Buddhima:
joomla-platform / libraries / joomla / html / editor.php
Method: getButtons()

Issue: If the $className would not exists (due to plugin class is missing) , assigning to $temp variable becomes unsafe

Solution: Put the assiging inside a try catch block, so that not existance of plugin class would not disturb the flow of foreach loop 

Signed-off-by: Buddhima Wijeweera

### DIFF
--- a/libraries/joomla/html/editor.php
+++ b/libraries/joomla/html/editor.php
@@ -467,7 +467,7 @@ class JEditor extends JObject
 				$result[] = $temp;
 			}
 
-			}
+		}
 
 		return $result;
 	}
--- a/libraries/joomla/html/editor.php
+++ b/libraries/joomla/html/editor.php
@@ -454,20 +454,17 @@ class JEditor extends JObject
 			{
 				$plugin = new $className($this, (array) $plugin);
 			}
-
-                    try
+                        else 
                         {
+                            throw new Exception("Class does not exists "+$className);
+                        }
+
 			// Try to authenticate
-                        // Unsafe operation, might cause an exception if $className not exists                        
                         if ($temp = $plugin->onDisplay($editor, $this->asset, $this->author))
                         {
                             $result[] = $temp;
                         }
-                        }catch (Exception $e)
-                        {
-                            echo 'Exception while authenticating';
-                            echo 'Message ' . $e->getMessage(); //Print the error message
-                        }
+                        
 		}
 
 		return $result;
--- a/libraries/joomla/html/editor.php
+++ b/libraries/joomla/html/editor.php
@@ -426,8 +426,8 @@ class JEditor extends JObject
 	 *
 	 * @return  array
 	 *
-         * @throws	InvalidArgumentException	If plugin required does not exists
-         *   
+     * @throws	InvalidArgumentException	If plugin required does not exists
+     *   
 	 * @since   11.1
 	 */
 	public function getButtons($editor, $buttons = true)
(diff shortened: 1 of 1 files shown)


mbabker:
You need to fix your code style first - http://developer.joomla.org/pulls/pulls/1093.html

Also, instead of echoing the message, you should be returning an error.  Instead of the try/catch method you've proposed, maybe the current if statement would work fine, but add an else so that an Exception is thrown and the caller can handle the error.

libraries/joomla/html/editor.php
```diff
@@ -455,11 +455,18 @@ public function getButtons($editor, $buttons = true)
 				$plugin = new $className($this, (array) $plugin);
 			}
 
+                        try{
 			// Try to authenticate
-			if ($temp = $plugin->onDisplay($editor, $this->asset, $this->author))
-			{
-				$result[] = $temp;
-			}
+                            // Unsafe opearation, might cause an exception if $className not exists
```
elinw: operation

Buddhima: :D a mistake. Thanks for pointing out
Corrected !!!

libraries/joomla/html/editor.php
```diff
@@ -426,6 +426,8 @@ public function setContent($editor, $html)
 	 *
 	 * @return  array
 	 *
+         * @throws	InvalidArgumentException	If plugin required does not exists
+         *
```
chdemko: Please indent correctly

libraries/joomla/html/editor.php
```diff
@@ -454,12 +456,17 @@ public function getButtons($editor, $buttons = true)
...
+			{
+				throw new InvalidArgumentException('Class does not exists: ' . $className);
+			}
 
 			// Try to authenticate
 			if ($temp = $plugin->onDisplay($editor, $this->asset, $this->author))
 			{
 				$result[] = $temp;
 			}
+
```
chdemko: superfluous line

chdemko:

Buddhima:
Hi, can you tell me why it is failing?

chdemko:
you have to merge with the current trunk

elinw:
The issue is not that the pull request can't be merged (that would be shown on the pull tester report with red)  All of the pull requests with ?? in them have the same failure. I'm not sure if it is because they all have a dependency on JHtml or not, but that is possible.  Those test reports were generated on May 5 and have not been updated since.
All of them show this as the failing line
https://github.com/joomla/joomla-platform/blob/staging/tests/suites/unit/joomla/html/JHtmlTest.php#L896

Buddhima:
Thanks Elin for pointing to the correct place :)

elinw:
Okay now you need to update your branch, since the pulltester has rerun and it says it is not able to merge cleanly.

realityking:
This still doesn't merge cleanly. Also I'm wondering under what circumstances this would be triggered.

elinw:
@realityking  if the plugin is missing but you put the button in the field definition, I believe.

@Buddhima will you please rebase this or close it until you get it rebased?

LouisLandry:
@Buddhima the library in question has been removed from the platform and added to the CMS tree.  I'd suggest you submitting this to the CMS repository since it is a very smart and correct change in behavior.  I'm closing this since it is no longer relevant for the platform.  Apologies for not getting to it sooner.

Issue #708, opened 2012-01-05
jimage division by zero


romacron:
function prepareDimensions
from line: 665

if ($scaleMethod == JImage::SCALE_INSIDE)
{
$ratio = ($rx > $ry) ? $rx : $ry;
}
else
{
$ratio = ($rx < $ry) ? $rx : $ry;
}
is width or height 0 an error will be produzed

an Idea: 
For Web-development/Administration in many cases i need an image-stack with same width or same height.
this harmonize the visible s for the website user. 

perhaps we can build in an decision "is only one value 0" the image will be resized to the opposite value.

Buddhima:
Isn't it better to be $rx same as $this->getWidth() if $width = 0 and $ry same as $this->getHeight() if $height=0 ?

romacron:
ive seen it on your stage it solves it, think

romacron:
i had reopen'd the ticket
and changed  https://github.com/romacron/joomla-platform/commit/3c28b42057de12075caea03eed171a1072a64895#diff-0

resize actions if they should scale to Edge must have as opposite value $rx  or $ry an 0 not the $this->getWith()

PR #1095, opened 2012-04-01
Avoid division by zero (for issue #708)

Buddhima:
libraries/joomla/image/image.php
Method: prepareDimensions()

Issue: If zero is given for the one of the arguments $height or $width, following section will produce an error.
Solution: Treat 3 instances where $height=0,$width=0 and both not equals zero

Signed-off-by: Buddhima Wijeweera

### DIFF
--- a/libraries/joomla/image/image.php
+++ b/libraries/joomla/image/image.php
@@ -654,8 +654,8 @@ class JImage
 	 *
 	 * @return  object
 	 *
-         * @throws  LogicException          If width,height or both given as zero   
-         * 
+	 * @throws  InvalidArgumentException	If width,height or both given as zero   
+	 * 
 	 * @since   11.3
 	 * @throws  InvalidArgumentException
 	 */
@@ -675,9 +675,9 @@ class JImage
 			case self::SCALE_OUTSIDE:
 			
 				// Both $height or $width cannot be zero
-				if($width == 0 || $height == 0)
+				if ($width == 0 || $height == 0)
 				{
-					throw new LogicException(' Both height or width cannot be zero ');
+					throw new InvalidArgumentException(' Width or height cannot be zero with this scale method ');
 				}
 			
 				// If both $width and $height are not equals to zero
--- a/libraries/joomla/image/image.php
+++ b/libraries/joomla/image/image.php
@@ -673,29 +673,20 @@ class JImage
 
 			case self::SCALE_INSIDE:
 			case self::SCALE_OUTSIDE:
-                            
-							// To avoid dividing by $width if it is zero
-							if($width == 0 && $height != 0)
-							{
-								throw new LogicException(' Width cannot be zero ');
-								}
-                            
-							// To avoid dividing by $height if it is zero
-							elseif ($width != 0 && $height == 0)
-							{
-								throw new LogicException(' Height cannot be zero ');
-								}
-                            
-							// Both $height and $width are zero
-							elseif($width == 0 && $height == 0){
-								throw new LogicException(' Both height and width cannot be zero ');
-								}
-                            
-							// If both $width and $height are not equals to zero
-                            else{
-				$rx = $this->getWidth() / $width;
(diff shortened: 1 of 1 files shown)


eddieajau:
There are a lot of style errors in this file.  Have a look at this report:

http://developer.joomla.org/pulls/logs/1095checkstyle.xml

Search for the following line:

<file name=".../libraries/joomla/image/image.php">

Any error that is not complaining about an underscore needs to be fixed.  Thanks.

libraries/joomla/image/image.php
```diff
@@ -671,9 +671,30 @@ protected function prepareDimensions($width, $height, $scaleMethod)
...
+                            
+                            // To avoid dividing by $height if it is zero                            
+                            elseif ($width!=0 && $height==0) {
+                               $rx = $this->getWidth() / $width;
+				$ry = $this->getHeight() ; 
+                            }
+                            
+                            // Both $height and $width are zero
+                            elseif($width==0 && $height==0){
+                                $rx = $this->getWidth() / $width;
```
benjaminpick: This cannot be correct: $width is 0, so this will divide by 0.

benjaminpick:
Also, it would be good to explain in the Doc-Header what a zero value means.

Buddhima:
Hi,

What I thought regarding this matter was there should be a value for rx when width =0. And I know when values getting closer to 0 , rx should tends to infinity. Therefore I can't specifically say a value. Then I thought of as a standard, assign rx with the value given by getWidth(). Same thing goes for ry too. 
But I got to know that throwing exceptions from classes like this is allowed.

So my new solution would be this.
Since this this method consider 2 cases for SCALE_INSIDE and SCALE_OUTSIDE there mustn't be a situation where user is given a chance of entering values less than 1. 
Anything designing on top of this should obey that rule. If this zero values are coming from an application, it should be a mistake done by developer.
Another case might be issues with web browser. Following links [1], [2] would show that.

My solution is to throw exceptions in those places where height or  width would be zero.

[1] http://name1price.com/how-to-get-customers-with-singapore-web-design-basic/jquery-tutorials/149-jquery-get-image-height-or-width-returns-0-problem-.html
[2] https://getsatisfaction.com/galleria/topics/could_not_extract_width_height_from_image_in_webkit_browsers 

Special thank goes to Elin for providing necessary guidance.

libraries/joomla/image/image.php
```diff
@@ -654,6 +654,8 @@ protected function getFilterInstance($type)
 	 *
 	 * @return  object
 	 *
+         * @throws  LogicException          If width,height or both given as zero
```
eddieajau: I think you have space indenting on these two lines.

libraries/joomla/image/image.php
```diff
@@ -671,9 +673,20 @@ protected function prepareDimensions($width, $height, $scaleMethod)
 
 			case self::SCALE_INSIDE:
 			case self::SCALE_OUTSIDE:
-				$rx = $this->getWidth() / $width;
-				$ry = $this->getHeight() / $height;
-
+			
+				// Both $height or $width cannot be zero
+				if($width == 0 || $height == 0)
```
eddieajau: Space needed after the `if`.

libraries/joomla/image/image.php
```diff
@@ -671,9 +673,20 @@ protected function prepareDimensions($width, $height, $scaleMethod)
...
+				{
+					throw new LogicException(' Both height or width cannot be zero ');
```
eddieajau: This should be an InvalidArgumentException

```
throw new InvalidArgumentException('Width or height cannot be zero with this scale method.');
```

Also adjust the DocBlock.  Thanks.

Buddhima: thanks  eddieajau ,
Corrected all mistakes and committed !

elinw:
The pull tester says you have white space on three lines.

http://developer.joomla.org/pulls/pulls/1095.html

PR #1096, opened 2012-04-01
Some small fixes to recent commits.

realityking:
Also removes an outdated todo that recently came up in a CMS bug.


libraries/joomla/user/helper.php
```diff
@@ -81,7 +81,7 @@ public static function addUserToGroup($userId, $groupId)
 	 *
 	 * @param   integer  $userId  The id of the user.
 	 *
-	 * @return  mixed  Array on success
+	 * @return  arraz    List of groups
```
elinw: Typo here

realityking: Thanks! Fixed.

PR #1098, opened 2012-04-01
Set JInputFiles constructor

mbabker:
So a while back on the CMS list, it was brought up that you can't use JInputFiles to pull from the `$_FILES` superglobal.  According to Ian (https://groups.google.com/d/topic/joomla-dev-cms/0CR3NKmiBg8/discussion), there should be a constructor similar to that in JInputCookie.  Well, this finally adds that constructor.


eddieajau:
Nice one. Thanks.

Issue #1094, opened 2012-04-01
Write function in stream.php


joaktree:
Hi,

The write function (stream.php) contains a do-while loop in which the string, which is written to the file, is cut up in chunks. 
However, within the loop the string is not shortened. This results in the effect that every cycle the same part of the original string is written to the file instead chunck-by-chunk.

Example current situtation
$string = 'The quick brown fox jumps over the lazy dog';
$chunk = 10;
Written to file: The quick The quick The quick The quick The
# 

In the current code only the value of remaining is set every cycle.

Current code
            else
            {
                // Wrote something
                $remaining -= $res;
            }

I think the code should be like this.
            else
            {
                // Wrote something
                $string     = substr($string, $amount); 
                $remaining -= $res;
            }

Regards,
Niels

elinw:
Niels, 

Why not make the pull request?

joaktree:
Elin,

Call me stupid ... but I don't know what that means "making the pull request" .   :-(
Can you enlighten me what to do?

Niels

aaronschmitz:
@joaktree, http://help.github.com/send-pull-requests/ describes how to create a pull request.
I've created one for this bug for you (https://github.com/joomla/joomla-platform/pull/1099).

elinw:
Hi Niels,

Making a pull request means actually changing the code and asking for it to be merged, as Aaron did. For a small change all you need to do is go to the file, click the Fork and Edit button and then make the change. When you save you'll see instructions on how to make the pull request.

Elin

joaktree:
Hi Elin,

Thanks for this info. I had never done that before, so I wasn''t sure what to do. I should collaborate more in open source projects, so I know better my way around. I will make a pull request the next time I find something in Joomla core. Thanks again.
Niels

PR #1099, opened 2012-04-02
Fix for issue #1094

aaronschmitz:
Cause JStream to actually iterate along the string rather than repeating the beginning.


pasamio:
We take the string by reference and that is going to cause the original string to change which is probably not accepted and widely regarded as a bad idea (see strtok)

pasamio:
To be clear, you're on the right track but we need a better way. It's aiming where possible to avoid making a copy of the string but we may end up having to do this.

elinw:
This isn't merging cleanly anymore, will you please update your branch?

LouisLandry:
This looks good.  @pasamio would you mind looking over this again?

Issue #1027, opened 2012-03-23
JHttp: cannot use timeout and useragent


piotr-cz:
**At this moment it's not possible to specify timeout or useragent string for transports.**

Although these exist in `JHttpTransport->request` interface and all derived transports, are not used in `JHttp` request methods (options, head, get, post, put, trace).

Right now, all transports are using timeout that is set in php settigs (default_socket_timeout, ~60s) which is way too long for some cases as REST or pinging (It's not possible to call the transport request method directly, as $transport is protected in JHttp).

The easiest way out would be to simply extend JHttp request methods to include $timeout and $useragent just as now $headers are.

However, JHttp instance holds $options (which is not used anywhere) and I guess this was intended to be used as a fallback when no $timeout is specified when using methods.

**My suggestion**

**1)  When instantiating JHttp, one can pass timeout and useragent as an option. This is passed to JHttpTransport right now.**

To JHttpTransports we add simple line to pick up those settings, so for JHttpTransportSocket this change from

```
if (!is_numeric($timeout))
{
    $timeout = ini_get("default_socket_timeout");
}
```

to

```
if (!is_numeric($timeout))
{
    $timeout = $this->options->get('timeout', ini_get("default_socket_timeout"));
}
```

*\* 2)  JHttp request methods are extended to pass $timeout and $useragent variables: **
from

```
public function post($url, $data, array $headers = null)
{
    return $this->transport->request('POST', new JUri($url), $data, $headers);
}
```

to

```
public function post($url, $data, array $headers = null, $timeout = null, $userAgent = null)
{
    return $this->transport->request('POST', new JUri($url), $data, $headers, $timeout, $userAgent);
}
```

I've created a topic on Platform group month ago describing exactly this thing:
https://groups.google.com/group/joomla-dev-platform/browse_thread/thread/38182a4a1b54cb9/1d7432371d5cca84

realityking:
We wanna do 1), just no one has gotten around to writing it.

piotr-cz:
Thanks for answering.

I can submit a patch, however I think it makes sense to allow setting timeout and useragent for each request (with fallback to Instance options). Points 1) and 2) in my Issue description are actually two things I'd like to see in the patch (not one or another).

**Scenario:**
I decide to ping an uri (using HEAD) to see if webservice is available or uri is correct before I send out the requests. 

```
$JHttp      = JHttpFactory::getHttp(JRegistry(array('timeout' => 30)));
```

[...]

PR #1101, opened 2012-04-02
JHttp: extending request methods

piotr-cz:
For each request method added variable $timeout with fallback to class options. Added $useragent string that's being read from class options.

See Issue #1027


LouisLandry:
Thanks!

piotr-cz:
I think you may also close the Issue #1027, which is the initial description for this pull request.
I didn't update the docs though: https://github.com/joomla/joomla-platform/blob/staging/docs/manual/en-US/chapters/classes/jhttp.xml

PR #1102, opened 2012-04-02
Separate CMS specific MVC classes from platform

Chraneco:
It looks like that it's not possible at the moment to use the MVC classes of Joomla in a platform applications because there are many dependencies on the CMS in those classes.

I started to do separate that a bit.

It's not perfect yet, but in first tests it works great with standalone applications and with the CMS.

For testing I added a changed 'hello world' application which uses one view and one task (task=test.dosomething) together with two models and a template override (to see how it works).

You can find that application here:
https://github.com/Chraneco/joomla-platform/downloads

Also another package can be found there which includes a changed 'cms.php' file and some additional classes for the 'cms' library. This is necessary for the CMS to use the separated MVC framework of the platform.
For testing just replace the libraries folder of a running Joomla! 2.5.4 installation with the platform of this pull request and include the files and folders of the cms_mvc_classes.zip package.


realityking:
You identified a valid problem but a solution is already in progress. We are going to get a new MVC package soon (hopefully later this week according to Andrew) and even if we don't we'll most likely move the CMS specific classes to the legacy folder.

I'm going to close this pull request for now, if all else fails we can reopen it.

Chraneco:
Hi,

I see, but what a pity that this work was for nothing then. I searched for work in progress about that afore starting, but could not find anything. Could you please post a link to more information about that work in progress?

Moving the CMS specific classes into the legacy folder shouldn't be a problem. The main point I focused on was to separate the MVC framework so that in can be used in a standalone application.

Regards
Chraneco

realityking:
I feel with you and I'll keep this pull in mind in case the bigger project doesn't work out.

I agree we need to find a better way to communicate bigger projects to the community.

In the future I suggest before you undertake a bigger task, to ask on list for feedback on it. I realize that isn't a great process, but it's the best I can come up with right now.

Best regards
Rouven

AmyStephen:
@chraneco - Link to the Joomla Platform list: https://groups.google.com/forum/#!forum/joomla-dev-platform - that's the best way to see if you have the platform team's buy-in before investing a lot of time.

eddieajau:
@Chraneco just following up.  Update the platform and have a look at what we've done with the MVC (now in separate packages).  They should be really easy to use standalone now but if you've got any other suggestions we'd love to hear them.

PR #1104, opened 2012-04-03
Continuation of legacy support.

LouisLandry:
With this pull request several more CMS specific classes have been moved to the legacy tree.  Among them are `JComponentHelper`, `JModuleHelper`, `JController*`, `JModel*` and `JView`.  All references to the deprecated `JException` class have been removed from non-legacy libraries.  All references to the deprecated `JError` class have either been removed or are adjusted to be temporary until the class is finally removed from the legacy tree.
### Unit Tests

The unit test runner has also been updated to support independent running of the legacy test suite.  To run the legacy suite you would use:

```
phpunit -c legacy.xml.dist
```

or if you have a custom file:

```
phpunit -c legacy.xml
```

To run the main unit and database suites you would still simply use:

```
phpunit
```

Other changes to the unit test runner includes a change in configuration to include all non-legacy Joomla Platform libraries in the code coverage report regardless of whether or not the files were naturally included during the test run.  This will impact the reported code coverage numbers, but will reflect a more accurate account of how well our platform code is covered by tests.


piotr-cz:
If those classes are going to be in legacy tree probably deprecated in future, what is going to replace them?

LouisLandry:
I'm assuming you are referring to the MVC classes.  It's been on our roadmap for a while _https://github.com/joomla/joomla-platform/wiki/Roadmap_ to revamp the MVC foundational classes.  What we have is very much coupled to the CMS and it's conventions.  There are at least 3 different incarnations of simplified MVC classes that I'm aware of.  My hope is that we'll get something in before the 12.1 release, but if not there's no harm because you couldn't use the current ones without loading the legacy tree anyway.  

I've got a set of MVC classes that we've used at eBay that we'll be publishing in the next day or so for review if nothing else.  They are very basic and not prescriptive by design so that any application can take them whichever direction it wants to go.  I've no doubt that something will work itself out though.

As for the other classes, they are obviously tightly coupled to the CMS and really didn't belong in the main Platform tree.

piotr-cz:
Thank you for your explaination.
I'm curious to see refactored MVC architecture. I failed to use current one in standalone application.

PR #1105, opened 2012-04-03
Show Media preview

benjaminpick:
Show a preview of the currently chosen image when using the field "media".

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28176


benjaminpick:
Still applies cleanly and works as intended after merging pull request #1151

timoline:
Nice, thx, we needed this for our component.
I only removed the width and height of the preview image.
maybe you can set this as an option too, thx again

benjaminpick:
Because your images are smaller than the fixed width?
(Mine are like 200px large so I can't display them full size. But
obviously, 10px-Images shouldn't be scaled up. Maybe I'll add a new xml
param, yes.)

timoline:
ok, thx
well sometimes you want to show the original size , sometimes they need to be resized (like your example)

we solve this like:
when width=0 and/or  height= 0 ==> original size, 
any other value will resize it to that value

benjaminpick:
@timoline Would you please review my changes and leave a comment on the feature tracker?
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28176

PR #1108, opened 2012-04-03
Rename protected members in JUpdater, JUpdateadapter and extending classes

elkuku:


eddieajau:
Merged. I think this one is pretty safe.

PR #1109, opened 2012-04-03
Rename a protected (and unused) member in JSessionStorageDatabase

elkuku:
This also seems not to be used anywhere..


realityking:
Can't we just remove it?

PR #1111, opened 2012-04-04
Rename protected members in JStream and JStreamString

elkuku:


eddieajau:
Merged. I think this one is reasonably safe.  Anyone extending it knows what they are doing.

PR #1112, opened 2012-04-04
Fix a parse error from #1110

elkuku:


elkuku:
BTW: I think it would make sense to include a call to `php -l` to the PullTester:registered: to check for "simple" parse errors.
Funnily this one was detected by a style sniff that could not parse the if() statement - generally those parse errors are detected by the unit tests - if there are any..
But honestly I don't know where to submit the pull request :disappointed:

realityking:
Thanks you guys for fixing my mistake.

PR #1113, opened 2012-04-04
Rename protected members in JEditor

elkuku:


realityking:
I have to say I'm not a big fan on changing these in classes that are so commonly extended. This change for example will break both editor plug-ins shipping with the CMS and probably all 3rd party ones as well.

eddieajau:
Let's leave this one for now.

PR #1114, opened 2012-04-04
Rename protected members in JDocumentRenderer and extending classes

elkuku:


eddieajau:
I'm pretty sure a few people fiddle with the document package so we might park this one for now.

PR #1115, opened 2012-04-04
Rename protected members in classes extending JCacheController and JCacheStorage

elkuku:
The remaining members in the cache package are public - should they also be renamed ?


eddieajau:
I'm just going to close this one for now (sorry). The cache package needs a major simplification and overhaul and we'll probably just move this version off to the legacy tree or deprecate it.

PR #1117, opened 2012-04-04
Rename protected members and extending classes of JAdapter

elkuku:


realityking:
I'm not sure how commonly JAdapter is used in 3rd party code but considering the classes are only useful if they are extended I'm not sure this is a good idea.

eddieajau:
I think let's park this one for now (unforeseen consequences).

PR #1118, opened 2012-04-04
Rename protected members in JTree and JNode

elkuku:


realityking:
I think these are going to be deprecated and moved to legacy. If we're eventually dropping them I see no reason in changing them now.

elkuku:
Yes, that would be a good reason. There is no reason to change any code that is going to be dropped. So - what would be the plan ?

eddieajau:
Let's just get a pull request for move these two classes to the legacy tree and deprecate them (12.3 is fine).

PR #1126, opened 2012-04-06
Fix and improve the docblocks in the legacy folder.

realityking:
Besides changing the subpackage tags to match the folders the biggest change is to undeprecated the protected class members that start with an underscore.

Those hints - without offering a real alternative - only confuse developers, I don't think we'll want to do a change like this in the legacy folder anyways since those classes are eventually gonna be removed or moved to the CMS.


eddieajau:
Merged.

PR #1130, opened 2012-04-06
images scaled by Edge

romacron:
In Order to force image scaling to an specific Edge we had decided to set for an width or an height an (int)0.
In example you want to transform an image to width=200px you must set the height (the opposite Value) to 0
so you are able to resize an imageStack with same matching Edges.

### DIFF
--- a/libraries/joomla/html/html/behavior.php
+++ b/libraries/joomla/html/html/behavior.php
@@ -684,7 +684,8 @@ abstract class JHtmlBehavior
 						item.type = 'color';
 					} else {
 						new MooRainbow(item,
-						{
+						{       
+							id:item.id,
 							imgPath: '" . JURI::root(true)
 			. "/media/system/images/mooRainbow/',
 							onComplete: function(color) {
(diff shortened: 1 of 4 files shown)


elinw:
See #1095, for discussion of how to handle the issue in 3c28b42.

realityking:
I think you have some problem with github. This pull request has changes that have nothing to do with the description and isn't mergeable.

romacron:
Hello Elin, i have seen your comment. In order that will get the git running for me, i will check the #1095 with the info from andre.

Perhaps someone can tell me how i remove the 
d1672be         i think there must be an imagedestroy inside the class  
188aa37         Merge pull request #1 from romacron/patch-1  
3c2e137         the db instancer called to late for sterr  
01f5a5e         By using more then one Colorpicker in a form the trigger must identif…  
07b9fea         Update media/system/css/mooRainbow.css  
3c28b42         Update libraries/joomla/image/image.php

from the pull request. Better 1 pull request for one change?

eddieajau:
Yeah, you need to make a new branch and put just those changes in there, and then make a new pull request.  Avoid making changes to you fork's master or staging branch.

PR #1131, opened 2012-04-07
Various fixes related to PostgreSQL.

gpongelli:
Various fixes reported from [this pull request](https://github.com/gpongelli/joomla-cms/pull/2) that correct errors coming with PostgreSQL database.
Added insertObject's test.

Eng. Gabriele Pongelli.

### DIFF
--- a/libraries/joomla/database/driver.php
+++ b/libraries/joomla/database/driver.php
@@ -812,10 +812,6 @@ abstract class JDatabaseDriver extends JDatabase implements JDatabaseInterface
 		$fields = array();
 		$values = array();
 
-		// Create the base insert statement.
-		$query = $this->getQuery(true);
-		$query->insert($this->quoteName($table));
-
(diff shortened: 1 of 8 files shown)


libraries/joomla/database/driver.php
```diff
@@ -832,12 +833,15 @@ public function insertObject($table, &$object, $key = null)
 
 			// Prepare and sanitize the fields and values for the database query.
 			$fields[] = $this->quoteName($k);
-			$values[] = $this->quote($v);
+			$values[] = is_numeric($v) ? $v : $this->quote($v);
```
eddieajau: I'm not sure this change is wise.  I seem to remember some databases require quoting of numeric values if the column is a string type (DB2 maybe?).

gpongelli: I've received this change from another pull request, I'm restoring it to previous version.

Eng. Gabriele Pongelli

PR #1132, opened 2012-04-07
Fixed id of the language inputbox in the label tag

Bakual:
The label for the language selector had a wrong "for" attribut, thus clicking on the label didn't select the inputbox.


eddieajau:
Merged. Thanks.

PR #1133, opened 2012-04-08
Add JDatabase::splitSql() for B/C

mbabker:
The splitSql method is a public function, and as such, can be called upon directly from JDatabase.  This adds a proxy method to JDatabaseDriver, similar to the other static methods remaining in JDatabase.


realityking:
Thanks!

PR #1136, opened 2012-04-08
Implement Countable in JDatabaseDriver + clean up

realityking:
This allows us to use

``` php
$db = JFactory::getDbo();
echo count($db);
```

instead of 

``` php
$db = JFactory::getDbo();
echo $db->getCount();
```

As for the clean up, it's mostly shifting deprecated code to JDatabase. It also moves JDatabaseException to legacy since we only keep it around for B/C purposes.


eddieajau:
I get what you are doing but I'm not sure counting a database driver class makes sense.

LouisLandry:
I feel the same way.  Isn't this simply returning the number of queries that have been executed?  That doesn't really seem like an intuitive use of count() to me.  Clever though.

LouisLandry:
All good except the countable interface.  Going to close this and we can revisit another time.

PR #1137, opened 2012-04-09
i think there must be an imagedestroy inside the class

romacron:
othwise it can combines with an __destruct() function

### DIFF
--- a/libraries/joomla/image/image.php
+++ b/libraries/joomla/image/image.php
@@ -754,4 +754,22 @@ class JImage
 
 		return $width;
 	}
+	
+	/**
+	 * Method to give the used Memory for the current handle free
+	 *
+	 * @return  void
+	 * @see     http://de2.php.net/imagedestroy
+	 */
+	public function destroy()
+	{
+		// Make sure the resource handle is valid.
+		if ( !$this->isLoaded() )
+		{
+			throw new LogicException( 'No valid image was loaded.' );
+		}
+        
+		return imagedestroy( $this->handle );
+
+	}
 }


LouisLandry:
Wouldn't it make more sense to just use __destruct() and have people unset($image) when they want the resource freed?

romacron:
okay thats the alternate, so this patch  can be removed from pull request.

elinw:
@romacron 
Are you going to update this? Or make a separate pull request? Currently this will not merge cleanly.

libraries/joomla/image/image.php
```diff
@@ -754,4 +754,22 @@ protected function sanitizeWidth($width, $height)
...
+	public function destroy()
+	{
+		// Make sure the resource handle is valid.
+		if ( !$this->isLoaded() )
+		{
+			throw new LogicException( 'No valid image was loaded.' );
+		}
+        
+		return imagedestroy( $this->handle );
+
```
realityking, line 773: Could you remove that line? Thanks.

libraries/joomla/image/image.php
```diff
@@ -754,4 +754,22 @@ protected function sanitizeWidth($width, $height)
...
+	 */
+	public function destroy()
+	{
+		// Make sure the resource handle is valid.
+		if ( !$this->isLoaded() )
+		{
+			throw new LogicException( 'No valid image was loaded.' );
+		}
+        
+		return imagedestroy( $this->handle );
```
realityking, line 772: Please remove the spaces around you argument. We generally don't use them. (same in line 769 and 767)

realityking:
Unit tests don't pass on this: http://developer.joomla.org/pulls/pulls/1137.html

realityking:
@romacron Are you still interested in fixing this?

romacron:
Hallo realityKing,

fixed in https://github.com/joomla/joomla-platform/pull/1418

realityking:
Thanks, closing this one since #1418 supersedes it.

PR #1138, opened 2012-04-09
ScaleToEdge

romacron:
In Order to force image scaling to an specific Edge we had decided to set for an width or an height an (int)0.

In example you want to transform an image to width=200px you must set the height (the opposite Value) to 0
so you are able to resize an imageStack with same matching Edges.


LouisLandry:
This cannot be merged cleanly.  Additionally I'd really like to see a unit test case proving that the change actually works so that we don't run into regressions down the road.

PR #1139, opened 2012-04-09
Make more classes autoloadable.

realityking:
This pull request makes JDate, JDispatcher and JSimpleCrypt autoloadable.

-JDate is moved to it's own package
-JDispatcher is renamed to JEventDispatcher
-JSimpleCrypt is moved to the legacy folder (JCrypt should be used now)

This also removed a couple of import() calls that aren't needed.


LouisLandry:
Why not move JDispatcher into a dispatcher.php file in the legacy tree so that it can be autoloaded?

eddieajau:
One check style error.  Plus, would you like to move JObject to it's own package as well?

realityking:
@LouisLandry
Done.

@eddieajau
JObject is already its own package, Rob did that weeks ago ;)

PR #1140, opened 2012-04-09
Add type hinting to JFormRule::test(). Fix an E_STRICT error in the process.

realityking:


realityking:
I changed the type hints to SimpleXMLElement because of #1141.

libraries/joomla/form/field.php
```diff
@@ -298,10 +298,10 @@ public function setForm(JForm $form)
 	 *
 	 * @since   11.1
 	 */
-	public function setup($element, $value, $group = null)
+	public function setup(JXmlElement $element, $value, $group = null)
```
LouisLandry: Didn't you want these to be `SimpleXMLElement` ?

realityking: You're right. Fixed.

chdemko:
Cannnot be merged

realityking:
Updated.

eddieajau:
1 cs error and one ut error.

realityking:
Thanks. They should be fixed now.

PR #1141, opened 2012-04-09
Deprecate JXMLElement

realityking:
JXMLElement isn't all that useful anymore. It only has two methods, one is just an alias to an internal method of the parent class. The other can be replaced by a method of the parent class in most cases, in case people need to pretty print they can either move the method to their own code or use DOMDocument to do it for them.

### DIFF
--- a/libraries/joomla/installer/extension.php
+++ b/libraries/joomla/installer/extension.php
@@ -86,13 +86,13 @@ class JExtension extends JObject
 	/**
 	 * Constructor
 	 *
-	 * @param   JXMLElement  $element  A JXMLElement from which to load data from
+	 * @param   SimpleXMLElement  $element  A SimpleXMLElement from which to load data from
 	 *
 	 * @since  11.1
 	 */
-	public function __construct(JXMLElement $element = null)
+	public function __construct(SimpleXMLElement $element = null)
 	{
-		if ($element && is_a($element, 'JXMLElement'))
+		if ($element)
 		{
 			$this->type = (string) $element->attributes()->type;
 			$this->id = (string) $element->attributes()->id;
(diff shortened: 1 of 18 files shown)


libraries/joomla/installer/extension.php
```diff
@@ -86,13 +86,13 @@ class JExtension extends JObject
 	/**
 	 * Constructor
 	 *
-	 * @param   JXMLElement  $element  A JXMLElement from which to load data from
+	 * @param   SimpleXMLElement  $element  A SimpleXMLElement from which to load data from
 	 *
 	 * @since  11.1
 	 */
 	public function __construct(JXMLElement $element = null)
```
chdemko: Change type to SimpleXMLElement

chdemko:
In https://github.com/realityking/joomla-platform/blob/84e1804de4705399a0da16298475cec9ff52c541/libraries/joomla/form/form.php#L678

https://github.com/realityking/joomla-platform/blob/84e1804de4705399a0da16298475cec9ff52c541/libraries/joomla/form/form.php#L853

https://github.com/realityking/joomla-platform/blob/84e1804de4705399a0da16298475cec9ff52c541/libraries/joomla/installer/installer.php#L1893

And in all tests

Replace JXMLElement by SimpleXMLElement

libraries/joomla/installer/extension.php
realityking: Done.

realityking:
I left the tests intact for now to demonstrate B/C, we can change the later.

The 3 files you mentioned I didn't change because that would be a (small) B/C break, if we wanna do that we can do it in a later pull request.

Issue #1116, opened 2012-04-04
grid.php calls isCheckedOut statically


blueboar2:
I use Joomla 2.5.1, and my request is about file libraries/joomla/database/table.php. Now (in latest trunk joomla-platform - it is in libraries/joomla/table/table.php). But neverthless, request is the same. There is a function isCheckedOut in this file. I worked on my site and receive many such messages

"Strict Standards: Non-static method JTable::isCheckedOut() should not be called statically in /srv/www/libraries/joomla/html/html/grid.php on line 198"

isCheckedOut is not static, but there WAS a message in comments before function declaration: "@todo    This either needs to be static or not.". Now, after discussion in CMS bugtracker (http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28326) it was removed.

There is such code in grid.php

if ($row instanceof JTable)
{
$result = $row->isCheckedOut($userid);
}
else
{
$result = JTable::isCheckedOut($userid, $row->checked_out);
}

If $row is not instance of JTable, isCheckedOut is called statically, and such "error" about Strict Standard occurs. What need i do to remove it?

elinw:
You can use the web interface to edit the file and send a pull request.

AmyStephen:
@blueboar2 - 

I'd probably just make it false (if $row isn't an instance of JTable, then it's not checked out).

if ($row instanceof JTable)
{
$result = $row->isCheckedOut($userid);
}
else
{
$result = false;
}

blueboar2:
OK. I agree. Do it.

blueboar2:
I've send pull request https://github.com/joomla/joomla-platform/pull/1143

PR #1143, opened 2012-04-10
grid.php calls isCheckedOut statically

blueboar2:
I use Joomla 2.5.1, and my request is about file libraries/joomla/database/table.php. Now (in latest trunk joomla-platform - it is in libraries/joomla/table/table.php). But neverthless, request is the same. There is a function isCheckedOut in this file. I worked on my site and receive many such messages

"Strict Standards: Non-static method JTable::isCheckedOut() should not be called statically in /srv/www/libraries/joomla/html/html/grid.php on line 198"

As discussed here https://github.com/joomla/joomla-platform/issues/1116 by AmyStephen, i change the code

### DIFF
--- a/libraries/joomla/html/html/grid.php
+++ b/libraries/joomla/html/html/grid.php
@@ -146,7 +146,7 @@ abstract class JHtmlGrid
 		}
 		else
 		{
-			$result = JTable::isCheckedOut($userid, $row->checked_out);
+			$result = false;
 		}
 
 		$checked = '';


blueboar2:
Or, maybe, we need to change JTable::isCheckedOut, because it handles both static and non-static calls. I think, it is not good practice in PHP code.

eddieajau:
Merged.  Thanks.

PR #1148, opened 2012-04-11
Database Factory proposed tuning

AmyStephen:
I like JDatabassFactory, but I think it could be a little more straight forward and reduce overlap with JDatabaseDriver.

In my thinking, JFactoryDatabase should 1) establish a connection to a database 2) save that connection as a static object for later use with queries, importers, exporters, and iterators. 

This pull request might make more clear my points in this https://groups.google.com/forum/?hl=en#!topic/joomla-dev-platform/KkmzoHMcAkE

Usage:

_To create a database connection (which could be a propoerty of the application object)_

$options = array(
    'driver' => 'mysqli',
    'host' => 'localhost',
    'user' => 'root',
    'password' => 'definitely not root because that would be crazy',
    'database' => 'db_name',
    'prefix' => 'xyz_');

$this->db = JDatabaseFactory::getInstance('mysqli', $options);

_To create a query object:_

$this->query = JFactory::getApplication()->get('db')->getQuery();
$this->query->select('id'); // etc

_Same approach as #2 to create an exporter (or importer or iterator)_


realityking:
I not sure about most of these yet but I'm not a fan of moving getIterator() the factory. The way I see the API is that you get a JDatabaseDriver object from the factory for your connection, you query something from the DB and than get the iterator to use the results. I don't see why I should get the iterator from the factory instead of the JDatabaseDriver object, it also means I'd have to keep track of two objects to query the database.

AmyStephen:
Actually, I'm trying to get it to where there is only one object -- instead of many -- and most importantly, that the connection information is only sent one time and then reused.

Right now, if I understand this correctly, you would instantiate your JDatabaseFactory class:

$factory = JDatabaseFactory::getInstance();

Then, save $factory -- since you'll need it later for queries, etc
$savefactory = $factory;

Use the saved factory to create the database connection:

(get the configuration file and load the $options array);
$db = $savefactory->getInstance('mysqli', $options);

Now, you need to save the name and options and the factory since those are used for queries, export and import:

$savename = 'mysqli';
$saveoptions = $options;

Then, when you use factory to create the query, you have to send all of that back in;

$query = $savefactory->getQuery($savename, $saveoptions);

It's too much, IMO. 

Now, if you are saying you want to get the iterator directly from JDatabaseDriver, you have to save that $query connection, too.

IMO - it's easier to only interact with the Factory class -- and only send that information in one time, creating a static instance of that connection for later reuse.

Now - I _will_ use it that way in Molajo, so, I'm not going to throw down on this. I can work around Joomla here. I just think it's a little complex - and unnecessarily so.

AmyStephen:
I should say ($name, $db connection) must be saved so that they can be used as parameters (not name and options). 

In addition, as I said above, JDatabaseFactory connection, too.

I've made my own database factory, so, it's not a concern for me, but I do think this is a bit cumbersome and it might be easier to fix now before it starts getting used.

AmyStephen:
Going to close this - hope to do some documentation on this package and if it still seems odd to me, I'll have a more sensible explanation and PR.

PR #1153, opened 2012-04-15
JLogger changes (Bug that broke logging, LogException, DBO, Priorities in parent)

AmyStephen:
Changes:
- findLoggers is broken, this patch includes a fix that enables the method to start finding loggers again 
- swapped out LogException for an SPL Exception option
- added ability to pass in the DBO for the database logger
- moved redundant list of priorities into parent
- renamed SysLog to Syslog and WC3 to Wc3 (inline with other filename/class naming standards)
- updated unit tests

### DIFF
--- a/libraries/joomla/log/entry.php
+++ b/libraries/joomla/log/entry.php
@@ -81,7 +81,7 @@ class JLogEntry
 		$this->message = (string) $message;
 
 		// Sanitize the priority.
-		if (!in_array($priority, $this->priorities, true))
+		if (!in_array($priority, $this->priorities))
 		{
 			$priority = JLog::INFO;
 		}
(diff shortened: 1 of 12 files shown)


AmyStephen:
JLog::findLoggers is broken -- The single ampersand needs to be two equal signs
https://github.com/AmyStephen/joomla-platform/commit/b7237655bc69925f50e30734bbee75000d526d35

JLog does not work correctly right now. (There is no priority checking).

AmyStephen:
Two tests are failing. One is failing due to the Exception name change and I don't know how to fix it. 

The second is changing due to the findLogger logic and I am not certain how to review the input so that I can determine why the test failed. (It is very possibly an incorrect test.)

AmyStephen:
I am going to break this up into easy to understand pieces.

PR #1154, opened 2012-04-15
Add chaining ability to JTable

elkuku:


libraries/joomla/table/table.php
```diff
@@ -489,7 +493,7 @@ public function load($keys = null, $reset = true)
 			// If empty primary key there's is no need to load anything
 			if (empty($keyValue))
 			{
-				return true;
```
eddieajau, line 492: There's still a return false in load, and the DocBlock would need to be changed.  Is there a risk the chaining here will cause a false positive in terms of backward compatibility?

libraries/joomla/table/table.php
```diff
@@ -542,14 +546,14 @@ public function load($keys = null, $reset = true)
...
 	 *
-	 * @return  boolean  True if the instance is sane and able to be stored in the database.
+	 * @return  JTable  If the instance is sane and able to be stored in the database.
 	 *
 	 * @link    http://docs.joomla.org/JTable/check
 	 * @since   11.1
 	 */
 	public function check()
 	{
-		return true;
```
eddieajau, line 552: Check is often called and return value checked.  I think this might cause a few problems ... you think?

libraries/joomla/table/table.php
```diff
@@ -590,7 +594,7 @@ public function store($updateNulls = false)
 		// If the table is not set to track assets return true.
 		if (!$this->_trackAssets)
 		{
-			return true;
```
eddieajau, line 593: Store has quite a few `return false` statements so I'm not sure we can chain it.

libraries/joomla/table/table.php
```diff
@@ -713,7 +717,7 @@ public function save($src, $orderingFilter = '', $ignore = '')
 		// Set the error to empty and return true.
 		$this->setError('');
 
-		return true;
```
eddieajau, line 716: Same in the save method.  `return false` and other methods like delete.

elkuku:
another day..

PR #1155, opened 2012-04-16
JDate separation from deprecated JDatabase class

AmyStephen:
Allows JDate to be used without requiring the JDatabase class (which is deprecated).

I ran in problems using the JLoggerDatabase option because of the toSQL method in JDate and it's coupling with the JDatabase object. 

It makes it possible to use the database logger without being tied into JFactory. Very helpful while ensuring the current usage continues to work.

### DIFF
--- a/libraries/joomla/date/date.php
+++ b/libraries/joomla/date/date.php
@@ -425,11 +425,21 @@ class JDate extends DateTime
 	 */
 	public function toSql($local = false, JDatabase $dbo = null)
 	{
-		if ($dbo === null)
+		if ($dbo === null && class_exists('JFactory'))
 		{
 			$dbo = JFactory::getDbo();
 		}
-		return $this->format($dbo->getDateFormat(), $local, false);
+
+		if ($dbo === null)
+		{
+			$format = 'Y-m-d H:i:s';
+		}
+		else
+		{
+			$format = $dbo->getDateFormat();
+		}
+
+		return $this->format($format, $local, false);
 	}
 
 	/**


realityking:
I don't follow. JDatabase is still in the type hint for backwards compatibility reasons, the method used (getDateFormat()) isn't deprecated.

I also don't understand what the $format parameter is for. Why not use JDate::format() directly?

AmyStephen:
Let me explain why this code is a problem for me, specific to using JLoggerDatabase class.
- JLoggerDatabase::addEntry 

It calls toSQL -- if I changed that method (below) to the JDate::format - then I would have to assume the default date option would work for everyone. (And, if that were true, we could simply remove the toSQL method.)

$entry->date = $entry->date->toSql();

https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/log/loggers/database.php#L126
- JDate::toSQL 

https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/date/date.php#L437

JDate::toSQL expects one of two things:

a. JDatabase object - the hinting isn't the problem - it's accessing the referred method. I have an object for the specific driver -- this has changed a bit with the JDatabaseFactory approach.

b. JFactory

I have neither. I am not using JFactory (nor will I), and I am using my own factory class to create the Driver connection, etc. So, I use the JDatabase package, but not the JDatabase object.

This way, by adding format - those who are in the same situation as I am in, but don't default to the format I provided, can pass in their database format. Otherwise, the defaults will work for me and there is no backwards compatibility issues for current use. 

That helps to separate it better for those of us trying to use the platform - but aren't the CMS.

Did that help?

realityking:
I see the general goal your after - decoupling JDate from JFactory and the database package - but I still don't quite see where you'd use the $format argument. Do you plan on changing JLoggerDatabase after this is merged?

AmyStephen:
I added it for those who can't use the default format -- and aren't using JFactory. (In reality, the only database driver with it's own method for date format is Oracle, so if it makes more sense to just default the format, it won't bother me at all to remove the $format parameter - and simply do a default when the JFactory class isn't present). 

I do have a separate PR for the JLog package https://github.com/joomla/joomla-platform/pull/1153

There will be no changes to JLogDatabase for this, though. This $entry->date = $entry->date->toSql(); will stay the same.

AmyStephen:
@realityking - removed $format - this certainly is enough for me. Is this better?

realityking:
Looks fine - except for the the dozen or so style errors you just introduced :P

Maybe the $format part can be done in a different pull, I just don't see the use for it yet.

The code could be made a bit shorter (untested):

```
if ($dbo === null)
{
    if (class_exists('JFactory'))
    {
        $dbo = JFactory::getDbo();
    }
    $format = $dbo->getDateFormat();
}
else
{
    $format = 'Y-m-d H:i:s';
}
```

Actually not sure if that's better and I'm tired so take that with a grain of salt.

AmyStephen:
Where the heck did those come from? My IDE is showing External change. Good grief. 

No, @realityking - that won't work. If the $dbo is passed in, it should be used.

realityking:
Yeah I realized that and changed the code again, though I had managed to fix it before you look ;)

AmyStephen:
Yea, that'd do the trick, too. OK, I'll repull from the core repository and use that.

Thanks.

AmyStephen:
Actually, after testing that @realityking - it didn't work. What that code does is tries to run getDateFormat on a null $dbo object. So, I think the way it is - is probably the best approach (or, at least it works.)

eddieajau:
Amy, I see what you are trying to do, and I don't totally disagree with it in isolation, but on the point of consistency I don't know if I would want to standardise this across the board.  The Platform does assume that JFactory is inherently available.  You can choose not to use it, but I'm not sure I'd penalise the whole platform for your choice.  Do you follow what I'm saying?

AmyStephen:
Here's what's happening - I'm using JDatabaseFactory since JDatabase is deprecated and I am not using JFactory, I have my own application controller process. 

When you pass a JDatabaseFactory object into this JDate method, it gives you a "Catchable fatal error" indicating it must be an instance of JDatabase. 

So that it's clear, this call to JDate prevents me from using JLoggerDatabase. That's a bit ironic since the JDatabaseFactory object works just fine with the logger and it's the logger that is doing the heavy hitting with the database, not JDate. 

Would you settle for removing the JDatabase hint on the JDate call?

realityking:
Ah I think we're getting closer. Your not supposed to pass in the the JDatabaseFactory but your JDatabaseDriver object. That should work.

AmyStephen:
@realityking -The database object obtained from JDatabaseFactory, not JDatabaseFactory => https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/database/factory.php#L54 

Having said that, I am hacking all of the packages for namespacing and I did remove the JDatabase class completely since it's deprecated. That's probably the problem. 

I'm in the office right now, later, I'll add JDatabase back in and see if that fixes it. I probably made a bad assumption thinking that deprecated means you can remove it.

realityking:
Right you obtain a database object from JDatabaseFactory and that database object is what JDate expects.

As for the type hint, I think we can change it to JDatabaseDriver since you can actualy instantiate JDatabase on its own.

AmyStephen:
Changing the hint to JDatabaseDriver would solve this problem for me -- plus, it would get rid of the deprecated class. Is that a good solution?

=> EDITED. SORRY!

realityking:
That won't work. A JDatabaseFactory object doesn't have the getDateFormat() method JDate needs, that's in JDatabaseDriver.

AmyStephen:
@RealityKing - Let me try again. I am not passing in the JDatabaseFactory object. That would be silly. It's not a database object. I am passing in the object that I linked to above that is created using the JDatabaseFactory. It is a JDatabaseDriver object. So, if the JDate hint is changed to JDatabaseDriver from JDatabase, it will solve my problem.

The only reason I even mentioned JDatabaseFactory is that I wanted to point out to you guys (Andrew in particular) that I am trying to use the current, not deprecated approaches as one might expect other platform developers to do. 

Does that clear it up?

realityking:
Yes it does :)

I just opened #1158 I hope that gets you at least part of the way.

As for the problem with JLoggerDatabase, how about we change this line https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/log/loggers/database.php#L126 to:

```
$entry->date = $entry->date->toSql(false, $this->dbo);
```

Since JLoggerDatabase keeps track of its own DB object that'd would make sense to me. This would also cover the use case of using a different database type for logging than for the main application data. (Not sure anyone would actually do that but one could)

AmyStephen:
@realityking - yes, that's fine if the type hint is changed. (In fact, attempting that as a possible solution is exactly how I discovered the type hint failed.) So, combined, those two changes would work for developers who are using a JDatabaseDriver object with the JLog package, but not using JFactory or JDatabase. (Which I believe describes where the Platform team is heading.)

Now, that's not going to work for any database outside of JDatabaseDriver, but I feel like it's only fair to expect that using the JPlatform database package would be expected if you are trying to use another JPlatform package with database interactivity. To me, that's reasonable. If someone wants to use another database package, then they can create a custom Log class. That works too.

LouisLandry:
Since I just merged #1158 can we close this one, or are there still outstanding problems?  My week from hell is almost over and am just trying to get up to speed on what's changed.

AmyStephen:
No, unfortunately not.

We can either:
1. Do the PR I submitted (here).
2. Change the JDatabase Hint to be JDatabaseDriver and then make the change @realityking is recommending https://github.com/joomla/joomla-platform/pull/1155#issuecomment-5181405
3. (Or, I'll just hack JDate a bit - in addition to the namespacing hack I did.) 

Any one of those three will work (although I'd prefer to keep my hacks to the namespace).

LouisLandry:
2 Sounds like the path of least resistance to me and a needed change anyway.  You mind taking a crack at that?  I'm sitting in an airport terminal with really spotty connection on an iPad or I'd tackle it myself.

AmyStephen:
@LouisLandry - yup, glad to do it. I'll get it tonight after work. Thanks :)

AmyStephen:
Actually, now that I look at  #1158 - @realityking already made the database hint change for JDate. So, I'll close this issue and include the JLoggerDatabase SQL change in my next Log PR. Thanks guys.

PR #1156, opened 2012-04-17
Log Package changes

AmyStephen:
I intentionally created a number of commits, one for each of the proposed changes I am recommending for the Log package. That way, it's easier to review and provide feedback.

I have been working to integrate the log package into my application and ran into a few things.

Probably the most significant commit I would like feedback on relates to how loggers are selected. I believe there is a logic error, but upon further review of failed unit tests, I realize I do not know how the priority values work so I could be wrong. The reason I began that particular exploration is that a logger was not selected when specifying a single priority value and a single category. 

Anyway, I hope it's easier to review now and I will change it, as needed. There are three failed unit tests -- two are related to changes to the Exceptions (I don't know how to fix) and then an error due to the Priority / Priority and Category selection that might be errors, or might not, depends on the logic issue.

Thanks.


LouisLandry:
@AmyStephen, the use of the bitwise operator (http://php.net/manual/en/language.operators.bitwise.php) `&` instead of logical `&&` is on purpose.  It is so you can mix and match the different priorities a logger handles.  We have pretty good tests for that method working, can you maybe give me a couple of cases where findLoggers() isn't working for you and we can work through them to see where the failures are?  Most of this looks good though.

AmyStephen:
Thanks @LouisLandry - as I was reviewing the tests, I realized there was something fancy going on. Thanks for the link and explanation, I'll read up on that, get another test case added, and resubmit this later. 

Appreciate the feedback.

LouisLandry:
OK, I'm gonna close this one then.  If you want to re-submit the other stuff that's great, if you'd rather wait that's fine too.  The exceptions would be especially nice to get sorted out sooner than later.  Thanks for the effort.

AmyStephen:
I'll do it yet this week, maybe later this evening.

PR #1157, opened 2012-04-17
Removed JString usage in JStringNormalise.

robschley:
Using JString instead of normal string methods causes a significant performance hit. Removing the JString wrapper fixes this problem. As this class was designed for programatic values, this should not cause much of an issue.

### DIFF
--- a/libraries/joomla/string/normalise.php
+++ b/libraries/joomla/string/normalise.php
@@ -29,7 +29,7 @@ abstract class JStringNormalise
 	 */
 	public static function fromCamelCase($input)
 	{
-		return JString::trim(preg_replace('#([A-Z])#', ' $1', $input));
+		return trim(preg_replace('#([A-Z])#', ' $1', $input));
 	}
 
 	/**
@@ -45,8 +45,8 @@ abstract class JStringNormalise
 	{
 		// Convert words to uppercase and then remove spaces.
 		$input = self::toSpaceSeparated($input);
-		$input = JString::ucwords($input);
-		$input = JString::str_ireplace(' ', '', $input);
+		$input = ucwords($input);
+		$input = str_ireplace(' ', '', $input);
 
 		return $input;
 	}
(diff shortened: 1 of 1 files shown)


ianmacl:
Can you comment on how this deals with code deals with non ASCII character sets?

eddieajau:
@ianmacl it wasn't really designed with non-ascii support in mind.

robschley:
This class was not specifically designed to handle non-ASCII character sets. When originally submitted, I used the JString wrapper methods because I figured it would make the class more versatile. However, after more use and some actual profiling, I realized the cost of the JString wrappers makes the class unusable. 500 calls into the class added approximately 1 second of execution time.

realityking:
Could we add to the docblocks that this class works only with ASCII strings?

Also at least toKey() would use mb_strtolower() when mbstring is enabled. Is that still to slow? toVariable() would also mostly use mbstring functions.

robschley:
@realityking I was profiling on PHP 5.3 with mbstring installed. What would you like the docblocks to say? Do you want it at the class level or for each method?

elinw:
I always think it is better to do at the method level the generated docs that often divide those into separate pages. It's not like someone would normally use non ascii in this case but on the off chance that they are trying to run down an error it would be helpful.

eddieajau:
This gives you a couple of extra cpu cycles back as well:

<pre>   public static function toUnderscoreSeparated($input)
    {
        // Remove duplicate underscores.
        $input = preg_replace('#[ \-_]+#', '_', $input);

        return $input;
    }</pre>


Same principle could also be used in `toSpaceSeparated` and `toDashSeparated`.

chdemko:
I think the JString::splitCamelCase method ( https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/string/string.php#L83 ) function should move into this class too

eddieajau:
@chdemko does JStringNormalise::fromCamelCase do the job?

chdemko:
@eddieajau not exactly:

```
'ThisIsA_NASAAstronaut'
```

will generate

```
array('This', 'Is', 'A_NASA', 'Astronaut')
```

using JString::splitCamelCase

and

```
'This Is A_ N A S A Astronaut'
```

using JStringNormalise::fromCamelCase

eddieajau:
Ok. Probably best to upgrade the one in normalise.php and deprecate the JString one.

realityking:
@robschley 
Something like "The methods in this class only work with ASCII strings. Unicode strings are not supported and should not be used with this class" at the class level would be fine with me.

eddieajau:
Closing. Rebooted in #1162.

PR #1159, opened 2012-04-18
JLog Exception, Class Name capitalization, and DBO

AmyStephen:
Redid the PR that @LouisLandry reviewed with a fresh copy of core. Did _not_ include the logic error issue. I doubt it's a problem and I need to study it more closely before testing again. I will add a new test for the single priority, single category logger when I get this review done.

The other changes are in this PR and all tests pass -
- Options to pass in DBO object 
- For echo log, change /n to br tag for better display (no wrapping)
- Change WC3 to Wc3 for consistent Class naming
- Change FormattedText to Formattedtext for consistent Class names
- Change SysLog to Syslog for consistent Class Naming
- Move array into JLogger and remove from all child classes
- Change LogException to PHP SPL Exceptions

I also added the SQL change recommended by @realityking for the JDate/JDatabase object issue.

Appreciate everyone's help and feedback.  Please let me know if changes are needed.


libraries/joomla/log/log.php
```diff
@@ -372,7 +370,7 @@ protected function addLogEntry(JLogEntry $entry)
 				}
 				else
 				{
-					throw new LogException(JText::_('Unable to create a JLogger instance: '));
+					throw new InvalidArgumentException('Unable to create a JLogger instance: '. $class);
```
realityking: Since the loggers aren't an argument of this method I wonder if this is the correct exception.

libraries/joomla/log/loggers/database.php
```diff
@@ -134,7 +131,6 @@ public function addEntry(JLogEntry $entry)
 	 * @return  void
 	 *
 	 * @since   11.1
-	 * @throws  RuntimeException
```
realityking: I'd leave this one around. Since we don't catch the RuntimeException that could be thrown by JDatabaseDriver::getInstance() it would look from the outside like it throws a RuntimeException. The other alternative would be to catch & rethrow the exception.

libraries/joomla/log/loggers/echo.php
```diff
@@ -45,6 +32,6 @@ class JLoggerEcho extends JLogger
 	 */
 	public function addEntry(JLogEntry $entry)
 	{
-		echo $this->priorities[$entry->priority] . ': ' . $entry->message . (empty($entry->category) ? '' : ' [' . $entry->category . ']') . "\n";
+		echo $this->priorities[$entry->priority] . ': ' . $entry->message . (empty($entry->category) ? '' : ' [' . $entry->category . ']') . "<br />";
```
realityking: I wonder if this was originally done for the benefit of the CLI?

libraries/joomla/log/loggers/formattedtext.php
```diff
@@ -16,13 +16,13 @@
...
  * This class is designed to use as a base for building formatted text files for output. By
- * default it emulates the SysLog style format output. This is a disk based output format.
+ * default it emulates the Syslog style format output. This is a disk based output format.
  *
  * @package     Joomla.Platform
  * @subpackage  Log
  * @since       11.1
  */
-class JLoggerFormattedText extends JLogger
+class JLoggerFormattedtext extends JLogger
```
realityking, line 25: We should really try to make these autoloadable, good first step thou.

libraries/joomla/log/log.php
AmyStephen: RuntimeException?

libraries/joomla/log/loggers/database.php
AmyStephen: I wonder about that since we've not done that elsewhere. Not saying it's not a good idea but it should be consistent and part of the standard. Having a few scattered throughout the source code in methods that do not actually throw the exception seems confusing.

libraries/joomla/log/log.php
realityking: Makes sense to me.

libraries/joomla/log/loggers/echo.php
AmyStephen: That is likely (now that you mention it). It'd be nice to have an echo for both. Wonder if we could pass in a parameter for the create logger method for the break tag or the new line (defaulting to /n.)

libraries/joomla/log/loggers/formattedtext.php
AmyStephen, line 25: How would we do that?

AmyStephen:
@realityking - with the exception of the autoloading, I (believe I) have addressed your points. 

I added an option for using a break tag -- defaulting to current behavior so as to not change the output for existing CLI apps. Do you think this is an acceptable approach? 

Thanks for your feedback and direction.

libraries/joomla/log/loggers/syslog.php
```diff
@@ -23,21 +23,8 @@
...
  * @since       11.1
  */
-class JLoggerSysLog extends JLogger
+class JLoggerSyslog extends JLogger
 {
-	/**
-	 * @var    array  Translation array for JLogEntry priorities to SysLog priority names.
-	 * @since  11.1
-	 */
-	protected $priorities = array(
```
LouisLandry: This isn't OK.  If you look a little closer you will see that for the SysLog logger, there are different values for the priorities.  This is because they are used to derive constants which are expected by the SysLog function (http://php.net/manual/en/function.syslog.php).

_-- Sorry for not catching that earlier._

AmyStephen: I'll add it back for the SysLog class - so glad you spotted this, @LouisLandry Sorry for not being more careful.

AmyStephen: @LouisLandry Done - moved it back into the Syslog class https://github.com/AmyStephen/joomla-platform/commit/8eb5ad5b1b48ea29ae9a05f85ec3f58c962e86c0#diff-7

I assume it's still okay to abstract out the others? (I did not change back the other two.)

LouisLandry:
Looks good.  A couple of code style issues to clean up: http://developer.joomla.org/pulls/pulls/1159.html

Also, I think we can get rid of the Test class for LogException.  It doesn't do anything and adds no value.  There isn't anything to test.  As soon as the style issues are cleaned up I'll get it merged.  Thanks for the work Amy.

libraries/joomla/log/loggers/formattedtext.php
realityking, line 25: We'd have to rename them to something like JLogLoggerFormattedtext etc., but lets do that in a later pull.

realityking:
When you deal with the code style changes, I got one last request. Could you move LogException to legacy and register it in import.legacy.php? Thanks.

AmyStephen:
That was fun =) 

Think it's working now. Thanks guys.

LouisLandry:
Beautiful.  Thanks again.

PR #1164, opened 2012-04-21
Add JLogTest::testAdd()

AmyStephen:
I added a test to JLog for the new separator option -- and to test using the add method. (Existing tests send in the JLogEntry so this tests sending in a message, only.) 

Earlier this week, I thought I found a logic error for logs with one priority. The purpose of this testing was to try to see if such an error existed. I found a problem with my code, not JLog, but this test covers ground not already covered so I thought I'd share it if it is desired.


AmyStephen:
Hm. Closing this one - I think I missed the boat here.

PR #1165, opened 2012-04-21
Remove JText dependency on JStreamString for die Statement

AmyStephen:
The only occurrence of JText in this class is to translate the die Statement. Doesn't seem worth a dependency.

### DIFF
--- a/libraries/joomla/filesystem/streams/string.php
+++ b/libraries/joomla/filesystem/streams/string.php
@@ -295,4 +295,4 @@ class JStreamString
 	}
 }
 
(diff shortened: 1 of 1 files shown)


LouisLandry:
To be honest I'd rather see us throw an exception than die in a case like this, but for the sake of removing coupling I'll allow it this time ;-)

AmyStephen:
@LouisLandry - Good point. Listen - you are the one with the big picture in mind, so throw it back with direction or close it if it's not right. I appreciate the direction as it helps me understand what "uncoupled" means to the platform team and some of the finer points of what you are working towards in the framework. So, that type of feedback (should) help me be more useful. ;-)

PR #1168, opened 2012-04-22
Add Email Logger for JLog

AmyStephen:
I did a little reading on testing email capabilities and found recommendations to use a mocked class to simulate the email function. I noticed that JLoggerMessageQueueTest mocks up JApplication, so I used that approach as a guide. 

For the most part, the testing worked perfectly, except I am unable to retrieve the mock class properties from the test in a similar manner to how the application mock object is referenced. 

What I ended up doing (which I do not recommend be a permanent change), but I changed the Logger Mail class to be public so that I could do the comparisons using that approach.

Would appreciate feedback on what I should be doing to use the Mock class properties.

Also, these are essentially my first Unit tests, so any direction is appreciated.

### DIFF
--- a/docs/manual/en-US/chapters/packages/log.xml
+++ b/docs/manual/en-US/chapters/packages/log.xml
@@ -14,7 +14,7 @@
     formats.</para>
 
     <para>The classes included in the Log package are JLog, JLogEntry, LogException, JLogger as well as the classes
-    JLoggerDatabase, JLoggerEcho, JLoggerFormattedText, JLoggerMessageQueue, JLoggerSyslog and JLoggerW3C which support formatting
+    JLoggerDatabase, JLoggerEcho, JLoggerEmail, JLoggerFormattedtext, JLoggerMessagequeue, JLoggerSyslog and JLoggerW3c which support formatting
     and storage. Of all these classes, you will generally only use JLog in your projects.</para>
 
     <para>Logging is a two-step process.</para>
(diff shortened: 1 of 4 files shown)


LouisLandry:
Don't have time to do anything thorough on this with you right now, but instead of making it public you just need to use reflection.  We've added a helper to make that easy for you in the test classes.

``` php
<?php

// Invoke a non-public method.
$result = TestReflection::invoke($object, 'methodName', $arg1, $arg2);

// Get a non-public member.
$value = TestReflection::getValue($object, 'memberName');

// Set a non-public member.
TestReflection::setValue($object, 'memberName', $valueToSet);
```

That helper makes it much easier to test non-public aspects of your classes.  You are on the right track with mocking out the mailer, though I don't think you need to put that much effort into it.  You really just need a mock object that will validate the methods that are called, More along the lines of our `TestMock*` classes.  You'll find that stuff in https://github.com/joomla/joomla-platform/tree/staging/tests/core.

AmyStephen:
Perfect, very helpful. TestReflection::getValue was just what I needed.

My mock class is only those methods called. The only "fancy" thing I did was set an array of values but that was because I didn't want to create a test method for each value. Otherwise, that mock class is nothing fancy.

Appreciate your time. I think this ready for consideration.

AmyStephen:
Since it's been over two months, I'm going to close this.

LouisLandry:
Please open it again.  I think given the autoloader changes that Rouven made we may need to adjust a couple of things, but it is just as relevant ... just hasn't been on the top of my list to deal with recently.

PR #1178, opened 2012-05-01
JEditor unit tests

florianv:
Hello, 

I wrote unit tests for JEditor which cover everything besides _loadEditor().

JEditor doesn't pass two tests ending with a fatal error.

testGetContentWithoutData()
testSetContentWithoutData()

If for some reasons loadEditor fails : the plugin name passed in the construct of JEditor doesn't correspond to an existing plugin, getContent() and setContent() will call $this->_editor->update($args); on a non-object, because $this->_editor is still null.

The case is covered in the save() and display() methods but not in setContent() and getContent().


florianv:
Now it passes the tests

realityking:
3 code style errors: http://developer.joomla.org/pulls/pulls/1178.html

tests/suites/unit/joomla/html/JEditorInspector.php
```diff
@@ -0,0 +1,107 @@
...
+ * Inspector for the JEditor class.
+ *
+ * @package     Joomla.UnitTest
+ * @subpackage  HTML
+ * @since       12.2
+ */
+class JEditorInspector extends JEditor
+{
+	/**
+	 * Method for inspecting protected variables.
```
ianmacl, line 21: See http://developer.joomla.org/manual/chap-Joomla_Platform_Manual-Testing.html#idp6957800 for information on working with protected and private variables and methods.  New inspectors should not have methods to work with them as that is now provided by ReflectionHelper.

florianv:
Thanks. I'm closing it, and I will make a new pull request when I have some time for this

PR #1180, opened 2012-05-02
JSession still uses JRequest

florianv:
It causes a fatal error when instanciating JSession because _start() still uses JRequest. I don't know if you want to use this construction

### DIFF
--- a/libraries/joomla/session/session.php
+++ b/libraries/joomla/session/session.php
@@ -530,7 +530,9 @@ class JSession extends JObject
 			
 			if (is_null($cookie->get($session_name)))
 			{
-				if ($session_clean = $input->get($session_name, false, 'string'))
+				$session_clean = $input->get($session_name, false, 'string');
+				
+				if ($session_clean)
 				{
 					session_id($session_clean);
 					$cookie->set($session_name, '', time() - 3600);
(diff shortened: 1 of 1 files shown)


libraries/joomla/session/session.php
```diff
@@ -523,12 +521,19 @@ protected function _start()
...
+			// Get the JInput object
+			$input = JFactory::getApplication()->input;
+			
+			// Get the JInputCookie object
+			$cookie = $input->cookie;
+			
+			if (is_null($cookie->get($session_name)))
 			{
-				if (JRequest::getVar($session_name))
+				if ($session_clean = $input->get($session_name, false, 'string'))
```
realityking: Please don't make assignments inside conditionals. We want to move away from that kind of style since it's a source of bugs and confusion.

realityking:
Otherwise looks good, will test with the CMS after the adjustment is made.

florianv:
Thanks for the feedback. I didn't find an efficient way to test it

realityking:
Thanks for the fix. You now have a code style error on line 534 (whitespace at the end of a line)

florianv:
I think Eclipse adds it, found a plugin that should remove it, I hope it works

PR #1183, opened 2012-05-03
Remove deprecated methods in JLog + additional fixes

realityking:
Most significant changes:
-oldstatic is reintroduced as a module cache option
-the support for legacy paths for plugins is removed


eddieajau:
Merged

PR #1184, opened 2012-05-03
Make the log package fully autoloadable.

realityking:


realityking:
Couldn't test locally so I hope the unit test run trough ;)

simonharrer:
http://developer.joomla.org/pulls/pulls/1184.html states all tests run through! :)

+1 for this!

chdemko:
@realityking , could you please keep inheritance from old class names to new ones to keep B/C

chdemko:
I think this pull request does not maintain B/C: renaming JLogger to JLogLogger may lead to B/C issue. (if a 3rd party dev has developped a new logger, his code inherits from JLogger...)

JLogger must be defined (marked as deprecated) and inherits from JLogLogger

PR #1185, opened 2012-05-03
Assign less values explicitly by reference. Use more type hinting.

realityking:


simonharrer:
Why hasn't this been merged?

realityking:
There was a unit test error caused by this that spread to all other pull requests so I closed this. I'll reopen it when I get a chance to fix the error.

PR #1186, opened 2012-05-04
Small improvements to JViewHtml

eddieajau:
Added __toString to JViewHtml (proxies for the render method).
Added chaining support to setLayout and setPaths in JViewHtml.


realityking:
You got an unit test error: http://developer.joomla.org/pulls/pulls/1186.html

elinw:
That's so so so helpful.

eddieajau:
Odd, I wonder what's causing that error.

PR #728, opened 2012-01-09
Fixed JFormFieldMedia incorrect folder processing

piotr-cz:
**test instructions:**
com_media with configured 'image_path' as a 2+ level directory (ie. media/Pliki),
bug doesn't occur when image_path configured to 1 level directory (images)
(see Content > Media manager > Options > Path to images folder)

**details:**
JFormFieldMedia has a hardcoded method to compute $folder variable by stripping first directory in the image path value and a filename.
$folder variable is then passed in a query 'folder='.$folder when calling com_media component

patch will strip complete configured image_path in the $folder array

Closes #342

PR #1187, opened 2012-05-04
Add new params to media field: preview_width / preview_height

benjaminpick:
Improvement of #1105 : Allowing full-size preview.

### DIFF
--- a/libraries/legacy/form/field/media.php
+++ b/libraries/legacy/form/field/media.php
@@ -203,10 +203,8 @@ class JFormFieldMedia extends JFormField
 			$width = isset($this->element['preview_width']) ? (int) $this->element['preview_width'] : 300;
 			$height = isset($this->element['preview_height']) ? (int) $this->element['preview_height'] : 200;
 			$style = '';
-			if ($width > 0)
-				$style .= 'max-width:' . $width . 'px;';
-			if ($height > 0)
-				$style .= 'max-height:' . $height . 'px;';
+			$style .= ($width > 0) ? 'max-width:' . $width . 'px;' : '';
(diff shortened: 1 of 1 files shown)


timoline:
Hi, ok thx
I tested it, but directory="folder" isnt working correct for me anymore, 
I mean: when pressing select it not goes to the folder.........can you can confirm that?
if I change back the code to:(line 130)
array_shift($folder);

it works ok....but maybe it has someting todo with the getParams('com_media') setting

Also a new param "default" would be nice

default = the default image to use

benjaminpick:
Default Image: Do you mean, the image that is shown when no image is
selected (currently none)?

Or as default value, i.e. "when no image is selected then select this
one" - this kind of default you could easily implement in your extension
(something like $param->getString('image', 'default image')).

timoline:
Hi, I read it a few times, but I think the first one

Just show the image which is set by default, when the user select a new image then show the new image
The default value is like a placeholder

benjaminpick:
For the directory-Folder-Bug: Which exactly are your values in the XML as well as the getParams('com_media') ? It should be a regression of #728 .

timoline:
```
        <field
            name="logo_small"
            type="media"            
            label="COM_JOOMLEAGUE_ADMIN_CLUB_LOGO_SMALL"
            description="COM_JOOMLEAGUE_ADMIN_CLUB_LOGO_SMALL_DESC"
            class="inputbox"
            preview="yes"
            directory="/com_joomleague/database/clubs/small"/>  
```

echo JComponentHelper::getParams('com_media')->get('image_path', 'images');

images

benjaminpick:
From documentation: "This attribute should be relative to the top level /images/ folder." http://docs.joomla.org/Media_form_field_type

PR #1189, opened 2012-05-05
Added JDatabaseDriver::disconnect() to support disconnecting from the database cleanly.

robschley:
This is useful when building applications with JApplicationDaemon because the database resource that gets copied after forking is shared between parent and child and if either of the processes close the connect, it is closed for both. With this, you can use the onFork event to disconnect and reconnect the database to get a new, unshared resource.


mahagr:
There's a related feature I've really would like to see in Joomla platform: reconnect/retry if database query fails because of the connection was cut by the server timeout. The most common reason is JInstaller extracting a large zip and trying to insert component without checking if the connection is still there.

PR #1192, opened 2012-05-08
Added behavior.load to allow for a single go-to place to register/load javascript libraries.

rcorral:
With this I added a central place where javascript libraries can be registered along with their CSS and JS files. Which is something that other implementations don't account for.

An example of this would be doing this:
JHtmlBehavior::register('modal', array(
  'script' => array('system/modal.js' => array()),
  'stylesheet' => array('system/modal.css' => array(array(), true, false, true))
  )
);

If more CSS or JS files need to be included, they can be added to the 'script' or 'stylesheet' array. For each element in the array the key is the path to the file, this file will get parsed by JHtml::_includeRelativeFiles, and the value for the key is an array of arguments that you would liked passed to JHtml::script or JHtml::stylesheet respectively.

I modified JHtmlBehavior::modal as an example of how this new feature can be implemented. It makes a call to self::load('modal'); to add the library to the document.

Please review and comment.

Cheers,
Rafael

### DIFF
--- a/libraries/joomla/html/html/behavior.php
+++ b/libraries/joomla/html/html/behavior.php
@@ -24,6 +24,137 @@ abstract class JHtmlBehavior
 	protected static $loaded = array();
 
 	/**
+	 * @var   array   Array contains a list of all scripts and all the files that relate to it
+	 */
+	protected static $scripts = array();
+
+	/**
+	 * Method that initializes all known behaviors
+	 *
+	 * @return  void
+	 *
+	 * @since   11.4
+	 */
+	protected static function init()
+	{
+		// Only load once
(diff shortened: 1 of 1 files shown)


LouisLandry:
@rcorral this pull request is no longer mergeable.  While I am personally indifferent to the feature, I'm happy to merge it if others are positive towards it.  I would probably want to see a couple of unit tests so that we can avoid regressions and confusion about the feature down the road though.

For now I am going to close the pull request.  This isn't because I'm rejecting it, but because it really isn't in a reviewable state right now.  If you still want to pursue it please rebase the branch so that it is mergeable again and re-open the pull request.  If we can get some interest from other developers that will certainly help.

Thanks a bunch for the initiative and work!

PR #1194, opened 2012-05-09
Use JPATH_THEMES constant to set themes directory in JApplicationWeb::render()

mbabker:
The `JPATH_THEMES` constant is set in the CMS, unit tests, and in some of the example apps I've seen to define the path where themes (templates) are located.  Instead of checking for a hard-coded themes folder under `JPATH_BASE`, this allows the themes directory to be set to this constant (which, for example, allows the CMS to continue using the "templates" directory if so desired when the applications are refactored) and retains the current fallback of `__DIR__ . '/themes'`

### DIFF
--- a/libraries/joomla/application/web.php
+++ b/libraries/joomla/application/web.php
@@ -320,7 +320,7 @@ class JApplicationWeb extends JApplicationBase
 		// Fall back to constants.
 		else
 		{
-			$options['directory'] = (defined('JPATH_BASE') ? JPATH_BASE : __DIR__) . '/themes';
+			$options['directory'] = defined('JPATH_THEMES') ? JPATH_THEMES : (defined('JPATH_BASE') ? JPATH_BASE : __DIR__) . '/themes';
 		}
 
 		// Parse the document.


libraries/joomla/application/web.php
```diff
@@ -320,7 +320,7 @@ protected function render()
 		// Fall back to constants.
 		else
 		{
-			$options['directory'] = (defined('JPATH_BASE') ? JPATH_BASE : __DIR__) . '/themes';
+			$options['directory'] = defined('JPATH_THEMES') ? JPATH_THEMES : __DIR__ . '/themes';
 		}
```
chdemko: to keep B/C, we could use:

```
$options['directory'] = defined('JPATH_THEMES') ? JPATH_THEMES : (defined('JPATH_BASE') ? JPATH_BASE : __DIR__) . '/themes';
```

mbabker: Modified as suggested.

elinw:
What do you think about doing a config default path if that is going to be the fallback structure?

mbabker:
I think we're fine as is.  If the config item `themes.base` isn't set, then the fallback is to check the constants.  The CMS defines `JPATH_THEMES`, so there's no issue from that aspect.  For applications that don't define `JPATH_THEMES` or the config item, we then fall back onto option 3; checking for `JPATH_BASE` or `__DIR__` as the base folder, and the themes folder underneath that is used instead.

PR #1196, opened 2012-05-09
Autoload JMenu and JRouter child classes

mbabker:
Following the example in `JPathway::getInstance()`, I've modified JMenu and JRouter's getInstance methods to allow for autoloading the child classes.  If the class can't be found first, then it will look in the hardcoded `includes/<file>.php` location for the class.


realityking:
Nice one! I wonder if we should move JRouter to its own package as well.

PR #1205, opened 2012-05-13
Reverting 1201

eddieajau:
This pull is to revert the changes in 1201.  For some reason it has changed the way internal data persist (made some independent tests fail).  But looking at it, I actually think we should go further and not allow JInput to modify the source input (for example, it should get a copy of $_REQUEST, not a reference so that that input variable can make changes without affecting the rest of the application).

### DIFF
--- a/libraries/joomla/input/input.php
+++ b/libraries/joomla/input/input.php
@@ -68,12 +68,12 @@ class JInput implements Serializable
 	/**
 	 * Constructor.
 	 *
-	 * @param   array  &$source  Source data (Optional, default is $_REQUEST)
+	 * @param   array  $source   Source data (Optional, default is $_REQUEST)
 	 * @param   array  $options  Array of configuration parameters (Optional)
 	 *
 	 * @since   11.1
 	 */
-	public function __construct(&$source = null, array $options = array())
+	public function __construct($source = null, array $options = array())
 	{
 		if (isset($options['filter']))
 		{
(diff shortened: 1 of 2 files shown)


chdemko:
I'm ok for this (not modifying the source input).

LouisLandry:
I generally agree with this.  I'd rather let the input objects encapsulate the array... and in this sense that would be best served by copying it.  Keeping a reference to the array would only make sense if the expectation was that we wanted people using these super-global arrays and the fact is we do not.

realityking:
This actually creates a new set of problems for the CMS. We wanna keep supporting JRequest for 3.x because there's no other transition strategy in place. However this requires that both JInput and JRequest share the same data source. One way around this would be to make JRequest a wrapper for JInput but I'm kinda afraid we're creating a new set of bugs this way.

Thoughts?

piotr-cz:
@realityking 
When is JRequest supposed to be dropped? It's marked as deprecated in CMS since 1.7
I believe many developers learn from CMS codebase so in my opinion it should be swapped by JInput first in there but still stay as deprecated for 1 STS afterwards

realityking:
@piotr-cz 
That's why I wrote we wanna keep it around in 3.x. My plan is to drop using it in 3.0 (or 3.1 if we don't get done) and remove JRequest as a class not before 4.0. However that's not a final decision by anyone - just my plan.

elinw:
Certainly not before 4.0 imo, it is very widely used in many places and by extensions.

Deprecation notices are there to give people plenty of warning to change not to say changes are happening tomorrow. 

How can reverting to what was present before create new problems?

realityking:
Oh I was unclear )or misread the change not sure anymore), the change itself doesn't create a problem. But if we decide to go the route Louis suggested for the $_REQUEST object as well we'll get the problem I described.

Issue #1206, opened 2012-05-14
missing imports prevent command line installation of extensions


amh-mw:
Originally described in http://forum.joomla.org/viewtopic.php?f=624&t=716847

Steps to reproduce:
1. Download the english language extension from http://joomlacode.org/gf/project/jtranslation1_6/frs/?action=FrsReleaseBrowse&frs_package_id=5651 and unzip to /tmp
2. Modify the paths in the following script per your environment and run it
3. The installation will fail due to classes failing to autoload

``` php
#!/usr/bin/php
<?php
define('_JEXEC', 1);
define('JPATH_BASE', '/var/www/joomla');
require_once 'configuration.php';
require_once 'includes/defines.php';
require_once 'libraries/import.php';
JInstaller::getInstance()->install('/tmp/site_en-US');
?>
```

mbabker:
This should go to the CMS repo (https://github.com/joomla/joomla-cms).  With the current Platform version, these classes will be auto-loaded, making the jimport statements unnecessary.  CMS 2.5 is using Platform version 11.4, which would need these jimport statements.

amh-mw:
Roger that. Opening https://github.com/joomla/joomla-cms/issues/207

PR #1207, opened 2012-05-14
Patch #1206

amh-mw:
Add imports necessary for joomla-platform #1206


amh-mw:
Closing per comments on #1206

PR #1208, opened 2012-05-15
comment for database quote methods to avoid IDE error

florianv:


florianv:
Thanks, I didn't notice it

PR #1212, opened 2012-05-19
Fix typos in JInstaller.

realityking:


realityking:
Didn't mean to remove Sam's comment but I removed the change to JInstallerAdapterComponent because I'm not sure myself. The issue was discovered by @mbabker and I asked him to take another look. I'm leaving the JText change since that's a fatal error.

mbabker:
I reverted to `$db->insertid()` locally because I'm not quite sure where @gpongelli was going with this change, but it did cause installation to fail for me because there was no `$key` variable set.

PR #1216, opened 2012-05-21
Remove workaround for old WebKit versions.

realityking:
This will improve (perceived) performance a bit for users of current versions of WebKit (and lower performance for older versions). since the majority of WebKit users use Chrome and update quickly this should be a net gain.

### DIFF
--- a/libraries/legacy/application/application.php
+++ b/libraries/legacy/application/application.php
@@ -394,13 +394,6 @@ class JApplication extends JApplicationBase
 				echo '<html><head><meta http-equiv="content-type" content="text/html; charset=' . $document->getCharset() . '" />'
 					. '<script>document.location.href=\'' . htmlspecialchars($url) . '\';</script></head></html>';
 			}
-			elseif (!$moved && $navigator->isBrowser('konqueror'))
-			{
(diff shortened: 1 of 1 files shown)


libraries/legacy/application/application.php
```diff
@@ -394,13 +394,6 @@ public function redirect($url, $msg = '', $msgType = 'message', $moved = false)
 				echo '<html><head><meta http-equiv="content-type" content="text/html; charset=' . $document->getCharset() . '" />'
 					. '<script>document.location.href=\'' . htmlspecialchars($url) . '\';</script></head></html>';
 			}
-			elseif (!$moved && $navigator->isBrowser('konqueror'))
-			{
-				// WebKit browser (identified as konqueror by Joomla!) - Do not use 303, as it causes subresources
-				// reload (https://bugs.webkit.org/show_bug.cgi?id=38690)
```
chdemko, line 400: It has only been fixed one year ago. Are we sure all browser instances have been updated in the world?

realityking, line 400: They certainly haven't. But at least Chrome users (biggest group of WebKit users) update very fast and I don't think we should penalize the users of newer browsers.

PR #1217, opened 2012-05-22
Change JApplication to use JCrypt, not JSimpleCrypt. + some clean-up

realityking:


libraries/joomla/base/object.php
```diff
@@ -1,13 +0,0 @@
...
- *
- * @copyright   Copyright (C) 2005 - 2012 Open Source Matters, Inc. All rights reserved.
- * @license     GNU General Public License version 2 or later; see LICENSE
- */
-
-defined('JPATH_PLATFORM') or die;
-
-JLog::add('JObject has been moved to /object/object.php.', JLog::WARNING, 'deprecated');
-require_once JPATH_PLATFORM . '/joomla/object/object.php';
\ No newline at end of file
```
chdemko, line 13: Why do you remove this code?

realityking, line 13: I explained it in the commit message: "We don't need a proxy file for JObject since it has always been autoloaded."

PR #1222, opened 2012-05-25
Let JFactory::getXML() opt to receive a SimpleXMLElement.

realityking:
This adds a deprecated log message to the constructor of JXMLElement and to reduce the number of times this is actually output prevents loading it all the time. I'm not sure whether we shouldn't default to load SimpleXMLElement instead of JXMLElement. For now I opted for the safer option but I'd like to hear what others think.

### DIFF
--- a/libraries/joomla/access/access.php
+++ b/libraries/joomla/access/access.php
@@ -496,7 +496,7 @@ class JAccess
 		else
 		{
 			// Else return the actions from the xml.
-			return self::getActionsFromData(JFactory::getXML($file, true), $xpath);
+			return self::getActionsFromData(JFactory::getXML($file, true, false), $xpath);
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


ianmacl:
Is there really a need for this method these days?  If somebody wants a SimpleXmlElement they should probably call simplexml_load_file or simplexml_load_string.

realityking:
I thought about that too. The only advantage I can see is that the error handling is mostly done for you - the rest is just fluff. What do others think?

LouisLandry:
Ya, I tend to be with @ianmacl on that.  The XML based stuff in Joomla was written in a time when 4.3 was the minimum requirement and we didn't have really good XML parsing in PHP natively.  That really isn't the case anymore, and I would lean towards deprecating all of it for removal.

piotr-cz:
I decided not to use `JFactory::getXML()` in project because I need to catch exeptions on creating the object and ignore one of libxml parsing errors.

realityking:
Superseded by #1239.

PR #1224, opened 2012-05-27
Update mooRainbow to version 1.4.

realityking:


media/system/css/mooRainbow.css
```diff
@@ -83,7 +83,7 @@ label {
 	margin-top: 190px;
 	margin-left: 315px;
 }
-span.moor-ballino { /* Style hue ° (degree) !! */
+span.moor-ballino { /* Style hue Âº (degree) !! */
```
LouisLandry: Looks like you may have gotten a junk character in here.

infograf768:
file has to be encoded as UTF8

infograf768:
and
-label {
-   font-family: mono;
  -}

has to be taken off as it has unwanted display results in CMS back-end.

realityking:
I took of the first commit so we can just deal with the update.

PR #1228, opened 2012-05-30
Reduce dependency on JFolder

realityking:
The cache package still depends on filesystem, however JCache itself doesn't.
JLanguage only depends on the string package and JXMLElement after this.
The session package shouldn't depend on the filesystem package anymore.

Note: B/C issue created by removing JLanguage::parseXMLLanguageFiles(). It's simply not needed anymore due to the RecusiveDirectoryIterator in JLanguage::parseLanguageFiles()

### DIFF
--- a/libraries/joomla/cache/cache.php
+++ b/libraries/joomla/cache/cache.php
@@ -91,26 +91,44 @@ class JCache
 	 */
 	public static function getStores()
 	{
-		jimport('joomla.filesystem.folder');
-		$handlers = JFolder::files(__DIR__ . '/storage', '.php', false, false, array('helper.php'));
+		$handlers = array();
+
+		// Get an iterator and loop trough the driver classes.
+		$iterator = new DirectoryIterator(__DIR__ . '/storage');
 
 		$names = array();
-		foreach ($handlers as $handler)
+		foreach ($iterator as $file)
 		{
-			$name = substr($handler, 0, strrpos($handler, '.'));
-			$class = 'JCacheStorage' . ucfirst(trim($name));
+			$fileName = $file->getFilename();
 
-			if (class_exists($class))
+			// Only load for php files.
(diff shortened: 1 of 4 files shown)


libraries/joomla/cache/cache.php
```diff
@@ -91,26 +91,44 @@ public static function getInstance($type = 'output', $options = array())
...
 		{
-			$name = substr($handler, 0, strrpos($handler, '.'));
-			$class = 'JCacheStorage' . ucfirst(trim($name));
+			$fileName = $file->getFilename();
 
-			if (class_exists($class))
+			// Only load for php files.
+			// Note: DirectoryIterator::getExtension only available PHP >= 5.3.6
+			if (!$file->isFile()
+				|| substr($fileName, strrpos($fileName, '.') + 1) != 'php'
```
chdemko, line 107: you can use $file->getExtension() != 'php'

libraries/joomla/language/language.php
```diff
@@ -672,7 +672,7 @@ public static function exists($lang, $basePath = JPATH_BASE)
 		// Check if the language exists
 		jimport('joomla.filesystem.folder');
```
chdemko: you can remove this line

libraries/joomla/cache/cache.php
realityking, line 107: As noted in the comment above it DirectoryIterator::getExtension was only introduced in PHP 5.3.6. Since some LTS server distributions ship with an older version of PHP 5.3 I didn't feel comfortable just changing that in a random pull request. All the other places we also use the substr() workaround.

Personally I'd be fine with raising the limit further, but we should be aware of the impact.

libraries/joomla/language/language.php
realityking: Done.

PR #1232, opened 2012-05-30
Modified JFactory::getApplication to allow instanciating web or cli applications

florianv:
We could now use :

JFactory::getApplication('MyWebApp', 'web')
JFactory::getApplication('MyCliApp', 'cli')

to load the Application object in the Factory, instead of 
JFactory::$application = JApplicationWeb::getInstance('MyWebApp');

### DIFF
--- a/libraries/joomla/factory.php
+++ b/libraries/joomla/factory.php
@@ -82,6 +82,7 @@ abstract class JFactory
 	 * Returns the global {@link JApplication} object, only creating it if it doesn't already exist.
 	 *
 	 * @param   mixed   $id      A client identifier or name.
+	 * @param   string  $type    Application type : Web or Cli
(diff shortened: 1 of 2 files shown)


elinw:
If you add the new argument to the end it won't create backward compatibility issues.

florianv:
Thanks. The problem is that we have to pass the in-between default values which is not cool.

Eventually, the Application is always accessible by auto-loading so I could remove the $type param and do something like (for web Application) :

if (class_exists($id) && (is_subclass_of($id, 'JApplicationWeb'))) but the eventual problem is that this check is already done in JApplicationWeb::getInstance(), so on the first load it is done two times.

elinw:
This is showing up a not mergable, you probably need to update your repository to the current trunk.

PR #1236, opened 2012-06-01
Fix a fatal bug in JSession and lay the groundwork for lazy sessions.

realityking:


libraries/joomla/session/session.php
```diff
@@ -396,6 +407,20 @@ public function isNew()
...
+	 * @param   JInput  $input  JInput object for the session to use.
+	 *
+	 * @return  void.
+	 *
+	 * @since   12.2
+	 */
+	public function initialise(JInput $input)
+	{
+		$this->input = $input;
+	}
```
chdemko, line 411: Why not returning $this for chaining ?

libraries/joomla/session/session.php
```diff
@@ -517,6 +542,29 @@ public function clear($name, $namespace = 'default')
...
+		$this->_state = 'active';
+
+		// Initialise the session
+		$this->_setCounter();
+		$this->_setTimers();
+
+		// Perform security checks
+		$this->_validate();
+
+		return true;
```
chdemko, line 554: why not returning $this for chaining?

libraries/legacy/application/application.php
```diff
@@ -958,6 +958,8 @@ protected function _createSession($name)
 		}
 
 		$session = JFactory::getSession($options);
+		$session->initialise($this->input);
```
chdemko, line 961: if chaining is done this could be replaced by

```
$session->initialise($this->input)->start();
```

libraries/joomla/session/session.php
realityking, line 411: Since the rest of JSession is not chainable (and not easily made so) wouldn't that cause more confusion that utility?

PR #1238, opened 2012-06-02
Use the PHP 5.3 const syntax where possible.

realityking:
This is mainly for consistency with class constants. Note that this only works for constants that are not conditional and that they take only a simple value and not the result of an expression. Since they are compile time instead of runtime based they get cached a bit more efficient with an opcode cache (negligible improvement thou)

### DIFF
--- a/libraries/joomla/application/router.php
+++ b/libraries/joomla/application/router.php
@@ -12,8 +12,8 @@ defined('JPATH_PLATFORM') or die;
 /**
  * Set the available masks for the routing mode
  */
-define('JROUTER_MODE_RAW', 0);
-define('JROUTER_MODE_SEF', 1);
+const JROUTER_MODE_RAW = 0;
+const JROUTER_MODE_SEF = 1;
 
 /**
  * Class to create and parse routes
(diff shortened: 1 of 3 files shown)


libraries/joomla/application/router.php
```diff
@@ -12,8 +12,8 @@
 /**
  * Set the available masks for the routing mode
  */
-define('JROUTER_MODE_RAW', 0);
-define('JROUTER_MODE_SEF', 1);
+const JROUTER_MODE_RAW = 0;
```
chdemko, line 15: is there a B/C isssue if a 3rd party dev has used

```
if (defined('JROUTER_MODE_RAW'))
{
    ...
}
```

realityking, line 15: No, at runtime they behave exactly the same. As a matter of fact they are even stored the same way. The only different is if they're defined at runtime (define) or at compile time (const) the latter being a tiny bit faster with opcode cacheing. (I read that Drupal saw an overall performance increase of 1% but they have far more constants than we do).

the more important point is that this is more consistent with how we define class constants.

PR #1243, opened 2012-06-03
Fix a fatal bug in JAccess + docblock fix.

realityking:
Also fixes a docblock in the session package that causes a mismatch in JDepend.


chdemko:
@realityking could you please divide your pull request in 3 parts:
- the access issue
- the docblock issue
- the logger issues

the Changelog will be more readable

realityking:
I removed the JLog part since you covered everything that was important in you PHP 5.4 pull request. I'll leave this one for the two remaining issues since they're both one liner.

PR #1250, opened 2012-06-05
Fixes for Oracle driver, Oracle query and PDO driver

pasamio:
This pull request introduces various fixes for the Oracle driver, Oracle query class and PDO driver. 
# Summarised changes:
- Add double quote as the Oracle name quote to handle name quoting properly.
- Added support for supplying a query to execute to verify that a connection is active.
- Override default connection verification query for Oracle to support "dual" idiom.
- Update getTableColumns() to support rewriting a prefix on Oracle.
- Fix getTableColumns() to return results consistent with other drivers.
- Add replacePrefix() to Oracle driver to ignore double quoted items (it being the name quote character).
- Fix getEscaped() for PDO driver to not automatically quote it's results inline with other drivers.
- Store a copy of the error message in PDO execute() before trying to run keep-alive query which wipes out previous value.
- Suppress error message from lastInsertId() to avoid PDO warning you that a given database platform doesn't support it natively.
- Add a sudden death when trying to check if connected when already trying to check if connected to prevent infinite loop.
- Update Oracle query to only mangle the query for limiting and offset when applicable otherwise pass the query through unmodified.

### DIFF
--- a/libraries/joomla/database/driver/oracle.php
+++ b/libraries/joomla/database/driver/oracle.php
@@ -28,6 +28,17 @@ class JDatabaseDriverOracle extends JDatabaseDriverPdo
 	public $name = 'oracle';
 
 	/**
+	 * The character(s) used to quote SQL statement names such as table names or field names,
+	 * etc.  The child classes should define this as necessary.  If a single character string the
+	 * same character is used for both sides of the quoted name, else the first character will be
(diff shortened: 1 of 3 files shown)


realityking:
One CS error: http://developer.joomla.org/pulls/pulls/1250.html

PR #1253, opened 2012-06-05
Joomla PHPCS is excluding files and folders starting with js

stefanneculai:
Currently the rule: <exclude-pattern>*.js</exclude-pattern> is excluding the files and folders that start with js. That is because . is interpreted as any single character instead of the dot character.

Escaping dots from the exclude-pattern will fix that.


LouisLandry:
Would you mind sending this request against https://github.com/joomla/coding-standards instead?  When I get a moment I'm going to submodule out the phpcs standard in this repository to use that.

stefanneculai:
Done.

PR #1255, opened 2012-06-06
Write unit tests for the JArchive package.

aaronschmitz:
A few unit tests added... and some to go still.


realityking:
Cool. What did you change in logo.tar and logo.zip?

aaronschmitz:
I removed the __MACOSX folder from the zip file and some similar artifact from the tar file. The way the unit tests are written, these extra files don't get deleted after the extraction is performed.

realityking:
Thanks. The pull looks fine to me except that it seems like you're using spaces for indentation but we use tabs. Could you fix that? Thank you.

aaronschmitz:
The original was using spaces, and I switched it to tabs. That's why the deltas are so large.

realityking:
Gotcha. Thanks again!

PR #1257, opened 2012-06-06
Add "preset" attribute to form fields and adapt list/checkboxes multiple default/preset values

chdemko:
It's a solution for Joomla CMS [#28402] JFormFieldList (and derivatives) do not utilize multiple default values when multiple=true (http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28402)

It has been extended for supporting preset values in form fields
## in an single valued field, you can now use

```
<field name="text1" type="text" preset="Preset value" />
<field name="text2" type="text" preset="KEY_TO_PRESET_VALUE" translate_preset="true" />
```

This will fill the field the first time the form is edited
## in a list/checkboxes

```
<field name="list1" type="list" default="1">
    <option value="1">1</option>
    <option value="2">2</option>
    <option value="3">3</option>
</field>
```

This will fill the field the with 1 if the current value is empty

```
<field name="list2" type="list" preset="1">
    <option value="1">1</option>
    <option value="2">2</option>
    <option value="3">3</option>
</field>
```

This will fill the field the with 1 the first time the form is edited

```
<field name="list3" type="list">
    <option value="1" default="true">1</option>
    <option value="2">2</option>
    <option value="3" default="true">3</option>
</field>
```

This will fill the field the with 1,3 if the current value is empty

```
<field name="list4" type="list">
    <option value="1" preset="true">1</option>
    <option value="2">2</option>
    <option value="3" preset="true">3</option>
</field>
```

This will fill the field the with 1,3 the first time the form is edited

### DIFF
--- a/libraries/joomla/form/fields/list.php
+++ b/libraries/joomla/form/fields/list.php
@@ -59,16 +59,68 @@ class JFormFieldList extends JFormField
 		// Get the field options.
 		$options = (array) $this->getOptions();
 
+		// Get the value(s)
+		if ($this->multiple)
+		{
+			if (empty($this->value))
+			{
+				// Filter the option using the preset flag
+				$filter = array_filter(
+					$options,
+					function($option)
+					{
+						return $option->preset;
+					}
+				);
+
+				// If the data exists or there is no preset
+				if ($this->form->existValue($this->fieldname, $this->group) || empty($filter))
+				{
+					$filter = array_filter(
+						$options,
+						function($option)
+						{
+							return $option->default;
+						}
+					);
+				}
+
+				// Extract the values from the filter
+				$values = array_map(
+					function ($option)
+					{
+						return $option->value;
+					},
+					$filter
+				);
+			}
+			elseif (is_array($this->value))
+			{
+				$values = $this->value;
+			}
+			else
+			{
+				$values = array($this->value);
+			}
+		}
+		else
+		{
+			$values = array($this->value);
+		}
+
 		// Create a read-only list (no name) with a hidden input to store the value.
 		if ((string) $this->element['readonly'] == 'true')
 		{
-			$html[] = JHtml::_('select.genericlist', $options, '', trim($attr), 'value', 'text', $this->value, $this->id);
-			$html[] = '<input type="hidden" name="' . $this->name . '" value="' . $this->value . '"/>';
+			$html[] = JHtml::_('select.genericlist', $options, '', trim($attr), 'value', 'text', $values, $this->id);
+			foreach ($values as $value)
+			{
+				$html[] = '<input type="hidden" name="' . $this->name . '" value="' . $value . '"/>';
+			}
 		}
 		// Create a regular list.
 		else
 		{
-			$html[] = JHtml::_('select.genericlist', $options, $this->name, trim($attr), 'value', 'text', $this->value, $this->id);
+			$html[] = JHtml::_('select.genericlist', $options, $this->name, trim($attr), 'value', 'text', $values, $this->id);
 		}
 
 		return implode($html);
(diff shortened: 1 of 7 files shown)


libraries/joomla/form/fields/list.php
```diff
@@ -59,16 +59,48 @@ protected function getInput()
...
+
 		// Create a read-only list (no name) with a hidden input to store the value.
 		if ((string) $this->element['readonly'] == 'true')
 		{
-			$html[] = JHtml::_('select.genericlist', $options, '', trim($attr), 'value', 'text', $this->value, $this->id);
-			$html[] = '<input type="hidden" name="' . $this->name . '" value="' . $this->value . '"/>';
+			$html[] = JHtml::_('select.genericlist', $options, '', trim($attr), 'value', 'text', $values, $this->id);
+			foreach ($values as $value)
+			{
+				$html[] = '<input type="hidden" name="' . $this->name . '" value="' . $value . '"/>';
```
realityking, line 117: How does this work? If they all have the same name I'd think only one would get send to the server.

elinw:
We've had quite a big discussion on the platform list about how to store multiselect values, does this mean that the maintainers have reached consensus on this being the right approach?

@Roveun
I think in a list you would not have the same control name for all of the values, that's only for things like checkboxes and radio buttons.

I would think you need a test with this.

chdemko:
@realityking , for multiple field the name is 'myname[]' (note the []) this is why you can store multiple value

@elinw I don't see these posts. Can you post the urls here?

elinw:
https://groups.google.com/forum/?fromgroups#!searchin/joomla-dev-general/multiple$20field/joomla-dev-general/SVZxhtQN_II/7kI2u4w0Si8J  is one place. 

As far as I understand what Sam said, the platform wants to stay neutral in deciding whether multiple values are stored as normal strings, json strings or something else and basically it's up to developers to implement. So I'm not against making it possible to do this, I just want to be clear on what is being done here. This is just about the rendered form, not about what is saved.
 I do think that it would be good to allow us to use the native html to use the selected element for multiple options but i'm not sure why we should call it preset when it is called selected in the standard and also why we wouldn't just add selected so that it matches conventional use rather than make up our own term.
http://www.w3.org/wiki/HTML/Elements/select#Example_B
http://www.w3.org/TR/html401/interact/forms.html#h-17.6.1

chdemko:
No problem to change

preset="value|true" to selected="value|true"

but for text fields selected="something" is not accurate

elinw:
Check boxes and lists are extremely different

In text/textarea the idea of selected doesn't make sense since there is no select.
In a list you just should really put selected as an element of the option.  The problem with that is that you have lists that are generated from the database or something like the integer field which generates its own dropdown so that's why you really want to have it in the field definition or maybe let the form developer choose one way or the other.
I'm not sure what you mean by selected="true" it is a boolean so it is either present or not present if you are using it in that way. If you said selected = "true" it would preselect a value equal to true. 
What happens if the selected value(s) are not present in the list? Are they just ignored?

So the proper rendered html is:

```
<option value="1" selected>1</option>
<option value="2">2</option>
<option value="3" selected>3</option>
```

Checkboxes shoudl render 
checked="checked"  rather than selected.

http://www.w3.org/TR/html401/interact/forms.html
http://www.w3.org/WAI/UA/TS/html401/cp0101/0101-CHECKBOX.html

For check boxes we already have the checked element  .
https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/form/fields/checkboxes.php#L67

So to me the only real issue is in lists.

chdemko:
In text/textarea, the idea of preset="Enter your text here..." can be used

the preset attribute is only for fields edited the first time (text/list/checkboxes...)

For examples, look at the test code

If values are not present in the list, they will be ignored

chdemko:
Closing this one (need to play with it)

elinw:
I did a pull request to fix the preset (checked) issues in checkboxes specificially.  #1268
Also I fixed a problem with the checked in checkbox.  #1270

chdemko:
See #1280

PR #1260, opened 2012-06-07
BUGFIX: discover template install

rdeutz:
lang was used but not set so the template installation failed and the style for the template wasn't created


chdemko:
This was already set at the beginning of the function. What is the utility of this assignment?

rdeutz:
My fault, this is an bug in an older version, already fixed and I didn't check against the last version. But maybe it is better within the if as outside, because we need it only in the "then" path. What is process for such a situation, doing another pull request ?

chdemko:
I've removed the useless code

PR #1261, opened 2012-06-07
Added some method inside JDatabaseDriver to create database and alter character set

gpongelli:
Added two public method inside JDatabaseDriver:
- alterDbCharacterSet to change database's character set
- createDatabase to create a new database

AlterDbCharacterSet needs database name as parameter, createDatabase
needs a JObject, similar to that used inside CMS' "initialise" function,
and a bool to set UTF8 database's support.

They use two protected method to obtain correct query in database's
supported syntax:
- getAlterDbCharacterSet
- getCreateDatabaseQuery

Protected method inside JDatabaseDriver return query for MySQL and
MySQLi syntax, this commit contain also overridden version for
PostgreSQL database.

There are tests only for protected methods inside MySQL, MySQLi and
PostgreSQL test classes.

### DIFF
--- a/libraries/joomla/database/driver.php
+++ b/libraries/joomla/database/driver.php
@@ -378,6 +378,27 @@ abstract class JDatabaseDriver extends JDatabase implements JDatabaseInterface
 	}
 
 	/**
+	 * Alter database's character set, obtaining query string from protected member.
+	 *
+	 * @param   string  $dbName  The database name that will be altered
+	 *
+	 * @return  string  The query that alter the database query string
+	 *
+	 * @since   12.2
+	 * @throws  RuntimeException
+	 */
+	public function alterDbCharacterSet($dbName)
+	{
+		if (is_null($dbName))
+		{
+			throw new RuntimeException('Database name must not be null.');
+		}
+
+		$this->setQuery($this->getAlterDbCharacterSet($dbName));
+		return $this->execute();
+	}
+
+	/**
 	 * Connects to the database if needed.
 	 *
 	 * @return  void  Returns void if the database connected successfully.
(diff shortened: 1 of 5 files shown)


libraries/joomla/database/driver.php
```diff
@@ -397,6 +418,31 @@ public function __construct($options)
...
+		if (is_null($options))
+		{
+			throw new RuntimeException('$options object must not be null.');
+		}
+
+		$this->setQuery($this->getCreateDatabaseQuery($options, $utf));
+		return $this->execute();
+	}
+	
+
```
elinw: Extra blank line here. 443 and 444  also have white space.

libraries/joomla/database/driver.php
```diff
@@ -485,6 +531,48 @@ public function __construct($options)
...
+			$query = 'CREATE DATABASE ' . $this->quoteName($options->db_name) . ' CHARACTER SET `utf8`';
+		}
+		else
+		{
+			$query = 'CREATE DATABASE ' . $this->quoteName($options->db_name);
+		}
+
+		return $query;
+	}
+
```
elinw: This line 574 has white space.

gpongelli:
Added style fixes and checks for createDatabase method.

gpongelli:
Fixed.

Eng. Gabriele Pongelli

libraries/joomla/database/driver.php
```diff
@@ -397,6 +418,38 @@ public function __construct($options)
...
+	 * @since   12.2
+	 * @throws  RuntimeException
+	 */
+	public function createDatabase($options, $utf = true)
+	{
+		if (is_null($options))
+		{
+			throw new RuntimeException('$options object must not be null.');
+		}
+		elseif (!isset($options->db_name) || empty($options->db_name))
```
realityking: I think this can be simplified to just the empty() check.

libraries/joomla/database/driver.php
```diff
@@ -397,6 +418,38 @@ public function __construct($options)
...
+		{
+			throw new RuntimeException('$options object must have db_name set.');
+		}
+		elseif (!isset($options->db_user) || empty($options->db_user))
```
realityking: Same as above.

libraries/joomla/database/driver.php
```diff
@@ -397,6 +418,38 @@ public function __construct($options)
 	abstract public function connected();
 
 	/**
+	 * Create a new database using information from $options object, obtaining query string
+	 * from protected member.
+	 *
+	 * @param   JObject  $options  JObject coming from CMS' "initialise" function to pass user
```
realityking: We don't actually need a JObject here. A generic object will do. Also please don't base you comment based on CMS use but be generic.

libraries/joomla/database/driver.php
gpongelli: Changed, I'll commit.

gpongelli:
Committed changes.

Gabriele.

ianmacl:
So what's the point of public function alterDbCharacterSet($dbName) ?

gpongelli:
Hi Ian,
I added that method because it's used during CMS installation, as createDatabase one, to complete installation on Postgresql.
These methods were born after a private discussion with Andrew, who point me adding these methods here.

Gabriele.

elinw:
Can you update your branch please?

gpongelli:
updated [HERE](https://github.com/joomla/joomla-platform/pull/1625), I'll close this pull.

PR #1262, opened 2012-06-07
change yes/no to true/false as will be committed in joomla-cms

benjaminpick:
As decided in http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28176 , true/false should be used instead of yes/no for consistency.

### DIFF
--- a/libraries/legacy/form/field/media.php
+++ b/libraries/legacy/form/field/media.php
@@ -170,13 +170,17 @@ class JFormFieldMedia extends JFormField
 		$showAsTooltip = false;
 		switch ($preview)
 		{
+			case 'no': // Deprecated parameter value
 			case 'false':
 			case 'none':
 				$showPreview = false;
 				break;
+
+			case 'yes': // Deprecated parameter value
 			case 'true':
 			case 'show':
 				break;
+
 			case 'tooltip':
 			default:
 				$showAsTooltip = true;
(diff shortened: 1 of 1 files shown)


libraries/legacy/form/field/media.php
```diff
@@ -170,11 +170,11 @@ protected function getInput()
 		$showAsTooltip = false;
 		switch ($preview)
 		{
-			case 'no':
```
chdemko, line 173: you should keep this line to keep B/C

libraries/legacy/form/field/media.php
```diff
@@ -170,11 +170,11 @@ protected function getInput()
...
+			case 'false':
 			case 'none':
 				$showPreview = false;
 				break;
-			case 'yes':
```
chdemko, line 177: you should keep this line to keep B/C

infograf768:
@Christophe
These lines do not exist yet in CMS

chdemko:
@infograf768 yes, but they are in platform 12.1

infograf768:
so shall we add in CMS and this patch both
case 'yes'
and
case 'true'
?
(same for no and false)

chdemko:
It is not necessary

realityking:
If we keep the old ones around please deprecate them.

PR #1267, opened 2012-06-09
Enabling multiple color fields

infograf768:
only a single JFormFieldColor field can exist (functionally) within a form
See: 
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28426

The whole mooRainbow.css is included as it had a wrong encoding


realityking:
Thanks!

PR #1269, opened 2012-06-09
Reduce dependency on JObject.

realityking:


libraries/joomla/form/form.php
```diff
@@ -120,11 +120,6 @@ public function bind($data)
 				// Handle a JRegistry.
 				$data = $data->toArray();
 			}
-			elseif ($data instanceof JObject)
```
LouisLandry: This isn't really a dependency.  It simply looks to see if the incoming value is a JObject and acts accordingly.  You sure you want to get rid of it?

realityking: Fair point, I'll leave it out for now.

PR #1270, opened 2012-06-09
Add tests for JFormFieldCheckbox and fix an issue in the field.

elinw:


pasamio:
Doesn't this change the behaviour of the field to not be checked if the value does match? Shouldn't we validate both options?

elinw:
Yes we should :). So this is how it works (assuming I did it right)
If you have no value saved and no checked attribute you get an empty checkbox.
If you have no value saved and checked attribute you get a checked checkbox.
If you have a value saved  you get a checked checkbox.
If you have a value saved and a checked attribute you get a checked checkbox.  (redundant but it's the last combination).

PR #1280, opened 2012-06-13
Add "preset" attribute to form fields and adapt list/checkboxes multiple default/selected values

chdemko:
Currently, the "default" attribute in form fields is used in two way
- filling a field when the current the value is empty (editing a form)
- storing the field when filtering data (saving a form)

These two behaviors need to be distinct because we need to provide an initial value for some fields and allow to save an empty value.

This pull request introduces a new attribute for field: "preset". It is used for filling the attribute when the current value is empty (it falls back to the "default" attribute if it does not exist)

For multiple fields (list, checkboxes, and groupedlist), similar behavior has been introduced.

It's a solution for Joomla CMS [#28402] JFormFieldList (and derivatives) do not utilize multiple default values when multiple=true (http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28402)

It has been extended for supporting preset values in form fields
## in an single valued field, you can now use

```
<field name="text1" type="text" preset="Preset value" />
<field name="text2" type="text" preset="KEY_TO_PRESET_VALUE" translate_preset="true" />
```

This will fill the field the first time the form is edited
## in a list

```
<field name="list1" type="list" default="1">
    <option value="1">1</option>
    <option value="2">2</option>
    <option value="3">3</option>
</field>
```

This will fill the field the with 1 if the current value is empty

```
<field name="list2" type="list" preset="1">
    <option value="1">1</option>
    <option value="2">2</option>
    <option value="3">3</option>
</field>
```

This will fill the field the with 1 the first time the form is edited

```
<field name="list3" type="list">
    <option value="1" default="true">1</option>
    <option value="2">2</option>
    <option value="3" default="true">3</option>
</field>
```

This will fill the field the with 1,3 if the current value is empty

```
<field name="list4" type="list">
    <option value="1" selected="true">1</option>
    <option value="2">2</option>
    <option value="3" selected="true">3</option>
</field>
```

This will fill the field the with 1,3 the first time the form is edited

### DIFF
--- a/libraries/joomla/form/form.php
+++ b/libraries/joomla/form/form.php
@@ -214,31 +214,45 @@ class JForm
 		}
 
 		// Filter the fields.
-		foreach ($fields as $field)
+		foreach ($fields as $element)
 		{
-			// Initialise variables.
-			$name = (string) $field['name'];
+			// Get the field
+			$field = $this->loadField($element, $group);
 
-			// Get the field groups for the element.
-			$attrs = $field->xpath('ancestor::fields[@name]/@name');
-			$groups = array_map('strval', $attrs ? $attrs : array());
-			$group = implode('.', $groups);
-
-			// Get the field value from the data input.
-			if ($group)
+			if ($field)
 			{
-				// Filter the value if it exists.
-				if ($input->exists($group . '.' . $name))
+				// Initialise variables.
+				$name = (string) $element['name'];
+
+				// Get the field groups for the element.
+				$attrs = $element->xpath('ancestor::fields[@name]/@name');
+				$groups = array_map('strval', $attrs ? $attrs : array());
+				$group = implode('.', $groups);
+
+				// Get the field value from the data input.
+				if ($group)
 				{
-					$output->set($group . '.' . $name, $this->filterField($field, $input->get($group . '.' . $name, (string) $field['default'])));
+					// Filter the value if it exists.
+					if ($input->exists($group . '.' . $name))
+					{
+						$output->set($group . '.' . $name, $this->filterField($element, $input->get($group . '.' . $name, $field->default)));
+					}
+					else
+					{
+						$output->set($group . '.' . $name, $this->filterField($element, $field->default));
+					}
 				}
-			}
-			else
-			{
-				// Filter the value if it exists.
-				if ($input->exists($name))
+				else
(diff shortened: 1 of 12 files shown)


libraries/joomla/form/form.php
```diff
@@ -622,6 +636,29 @@ public function getValue($name, $group = null, $default = null)
 	}
 
 	/**
+	 * Method to test if a field has a value.
+	 *
+	 * @param   string  $name   The name of the field for which to get the value.
+	 * @param   string  $group  The optional dot-separated form group path on which to get the value.
+	 *
+	 * @return  bool  TRUE on success, FALSE otherwise.
```
elinw, line 644: spell out please

pasamio, line 644: p-l-e-a-s-e

chdemko, line 644: I don't understand what you mean

elinw, line 644: Spell out the word boolean.

piotr-cz, line 644: `@return  boolean  True on success.`

piotr-cz:
Why not `default="1,3"` or `default="1|3"`, This seems more natural to me?

elinw:
@piotr-cz  Preset/selsected/checked is not the same as default. Default is what will save if the field is blank although it also shows when a user opens the form. A preset is selected already when the user opens the form for the first time but the user can unselect all and unless there is a default value can save blank.

THat said,  I obviously agree more with the general approach @piotr-cz  more than with this approach since his approach is close was implemented in checkboxes (except not using the default attribute) which would now be undone, changing the behavior of the field.

piotr-cz:
@elinw Thanks for explaining. I'd be for same approach that has been implemented in [checkboxes.php](https://github.com/joomla/joomla-platform/commit/a6b59ba47e0b435f49a6cd89f29c98648c76933b#libraries/joomla/form/fields/checkboxes.php) because of simplicity (can't think of real life cases for `selected` attribute).

But I'm not sure which one would feel generic if I'd like to alter default field value programatically?

LouisLandry:
I'm going to close this since it can no longer be merged, and there appears to be consensus around a different approach.  Thanks a bunch!

PR #1282, opened 2012-06-13
Fix a parse error

elkuku:
So it happened that a parse error was committed and Mr. Jenkins let is pass silently ([almost](http://build.joomla.org/job/platform/847/console)) pushing it to the master branch.

It would be nice to implement a call to php lint to prevent those "accidents" in the future.


realityking:
Thanks!

PR #1283, opened 2012-06-14
Move more classes to legacy.

realityking:
Unit tests for legacy run fine with this,


tests/suites/legacy/table/JTableContentTest.php
```diff
@@ -59,7 +59,7 @@ protected function getDataSet()
 		$stubpath = JPATH_TESTS . '/suites/unit/joomla/table/stubs';
 
 		$dataSet->addTable('jos_assets', $stubpath . '/jos_assets.csv');
```
ianmacl, line 61: I'll merge but we should probably change the other $stubpath just to be consistent.

realityking, line 61: Not sure what you mean, the file is still in the joomla suit.

PR #1285, opened 2012-06-14
Fix issue #834: Bug in JString::parse_url

stefanneculai:


chdemko:
You have unit test and checkstyle errors 
http://developer.joomla.org/pulls
- Could you post the system used for testing (os version, php version). I want to be sure this will work on every os and on php >= 5.2

stefanneculai:
The comment of the parse_url was saying: "return  mixed  Associative array or false if badly formed URL." In fact, all methods that call parse_url expect to get an empty array. I have changed that and should be fine now.

I tested it on: Windows and Linux with php 5.3.4

libraries/joomla/string/string.php
```diff
@@ -954,35 +954,19 @@ public static function compliant($str)
...
-		// All other characters will be encoded
-		$encodedURL = str_replace($entities, $replacements, urlencode($url));
-
-		// Parse the encoded URL
-		$encodedParts = parse_url($encodedURL);
-
-		// Now, decode each value of the resulting array
-		if ($encodedParts)
+		$parsedURL = parse_url($url);
+		if ($parsedURL == false)
```
chdemko: use === test instead of ==

stefanneculai: Done.

elinw:
I have to say that on reflection I think it is a bug that an empty array rather than false is being returned. I think the intent is to act like the php functions and in that case it should return false. The question is, does fixing  that break anything?

stefanneculai:
Yes, making that fix would break 5 unit tests from other packages.

elinw:
And who knows what in the other half of the code.  So at least for now the doc block will match what actually happens, but this is something we probably should follow up on.

diananeculai:
I need to use pare_url in my project and this fix would really be helpful. Can you please merge it?

ianmacl:
If this method is becoming more or less a shell around the php function then it makes more sense to deprecate the method rather than altering it.  Also, there is nowhere that indicates what the actual bug is.

elinw:
@dianaprajescu 
Can't you use the php parse_url if it is really working with urf8?

diananeculai:
Yes, php parse_url works with UTF-8. I'm using it on a URL-encoded string according to RFC 3986, which performs character translation of textual identifiers to UTF-8.

diananeculai:
Could you please do something about this? I spent again a lot of time because of this problem, I changed the branch and forgot about it. It's not easy to notice this and I spent a lot of time debugging my code useless. I'm using curl to send a request which has in it's URL an URL-econded string, after setting the request URL: <code>$options[CURLOPT_URL] = (string) $uri;</code> the string is no longer encoded.

stefanneculai:
I have opened a pull request (#1382) that should fix this.

realityking:
I'm not sure about this. According to this PHP bug parse_urls behavior may be platform dependent: https://bugs.php.net/bug.php?id=52923

I'd like to a see a source or some testing across PHP versions and platforms before we merge this.

realityking:
Ha, I just realize that comment is from Mark Dexter, I'll ask him about it.

stefanneculai:
OK, but currently doing

<code>$uri = new JUri('http://mydomain.com?url=http%3A%2F%2Ffoo.com');
echo $uri->ToString();</code>

would return <b>http://mydomain.com?url=http://foo.com</b> instead of <b>http://mydomain.com?url=http%3A%2F%2Ffoo.com</b>. 

In my opinion it's more important to have it working well on such strings. As Diana said, she needs to make a CURL request to an URL similar to the one above, but because of the current parse_url this is impossible.

elinw:
Is it separating into segments correctly? The way I handled this issue in the url input filter is to get the separate parts of the array as strings, validate them as strings, then put them back together.  If I do this:
<?php
$x = parse_url('http://mydomain.com?url=http%3A%2F%2Ffoo.com');

var_dump($x);

I get

array(3) {
  ["scheme"]=>
  string(4) "http"
  ["host"]=>
  string(12) "mydomain.com"
  ["query"]=>
  string(24) "url=http%3A%2F%2Ffoo.com"
}

So maybe we could handle it somewhat similarly?

elinw:
What if we added an argument with the option not to change percent encoding or that specifically allowed them in the query?

elinw:
Stefan can we close this and the other one now?

stefanneculai:
Sure :-)

PR #1289, opened 2012-06-17
Standardize default Content-Type handling across all transports.

piotr-cz:
According to IETF (http://www.ietf.org/rfc/rfc2616.txt) and W3C (http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) are case-insensitive, but in both documentations used in camel case.

### DIFF
--- a/libraries/joomla/http/transport/socket.php
+++ b/libraries/joomla/http/transport/socket.php
@@ -94,8 +94,12 @@ class JHttpTransportSocket implements JHttpTransport
 				$data = http_build_query($data);
 			}
 
+			if (!isset($headers['Content-type']))
+			{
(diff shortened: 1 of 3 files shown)


libraries/joomla/http/transport/socket.php
```diff
@@ -94,8 +94,12 @@ public function request($method, JUri $uri, $data = null, array $headers = null,
 				$data = http_build_query($data);
 			}
 
+			if (!isset($headers['Content-type']))
```
chdemko: you mean Content-**T**ype ?

piotr-cz: definitely :)

PR #1290, opened 2012-06-18
Method in POST requests - router for RESTful web application

stefanneculai:
Most of the RESTful web applications allow to pass method as a parameter in the POST request, because not all clients support advanced methods such as PUT or DELETE.

I have added in a variable called <code>methodInPostRequest</code> in the <code>JApplicationWebRouterRest</code> which allows or not such requests.

There are tests for the new code too.


LouisLandry:
Very nice.  This was an oversight on my part.  Thanks!

Issue #1266, opened 2012-06-09
JMail addRecipient, addCC, addBCC and addReplyTo problems


diananeculai:
These methods were ignoring the $name parameter. This problem was solved for the addRecipient method, issue #1190, but it still has an untreated case: it uses array_combine(): <code>$combined = array_combine($recipient, $name);</code> and doesn't verify the returned value. If $recipient and $name accidentally don't have the same size then the returned value will be false and using it in <code>foreach ($combined as $recipientEmail=>$recipientName)</code> is not a good thing.

The same fix should be applied to addCC and addBCC. The addReplyTo method should be similar with these methods, but instead its parameters are different.

elinw:
Would you submit the code to fix this?

diananeculai:
Yes, I will, but after I finish what I have planed for this week for my project.

chdemko:
Fixed

PR #1292, opened 2012-06-18
Fix issue #1266

diananeculai:
I created the add function in order to avoid duplicated code since addRecipient, addCC, addBCC and addReplyTo are almost the same. #1266

### DIFF
--- a/libraries/joomla/mail/mail.php
+++ b/libraries/joomla/mail/mail.php
@@ -296,7 +296,7 @@ class JMail extends PHPMailer
 	 *
 	 * @return  JMail  Returns this object for chaining.
 	 *
-	 * @since   11.1
+	 * @since   12.2
 	 * @throws  InvalidArgumentException
 	 */
 	public function addAttachment($attachment, $name = '', $encoding = 'base64', $type = 'application/octet-stream')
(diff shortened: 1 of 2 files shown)


libraries/joomla/mail/mail.php
```diff
@@ -179,16 +179,17 @@ public function setBody($content)
...
 	 *
-	 * @param   mixed  $recipient  Either a string or array of strings [email address(es)]
-	 * @param   mixed  $name       Either a string or array of strings [name(s)]
+	 * @param   mixed   $recipient  Either a string or array of strings [email address(es)]
+	 * @param   mixed   $name       Either a string or array of strings [name(s)]
+	 * @param   string  $method     The parent method's name.
 	 *
 	 * @return  JMail  Returns this object for chaining.
 	 *
 	 * @since   11.1
```
mbabker, line 190: For the new add method, please change the since tag to 12.2

elinw:
Diana do you want to close 1266?

chdemko:
All add\* methods should return $this for chaining as stated in the docblock

diananeculai:
Done.

PR #1294, opened 2012-06-20
Force cached CSS/JavaScript reload after update

chdemko:
When css/js/images are updated on a site, they are not reloaded on browsers that use caching. This pull request add a timestamp to files loaded by the JHtml library to force the reload of cached files.


realityking:
Great idea to finally add this. But isn't filemtime rather slow?

chdemko:
I know that, but how to get a true timestamp on files by not using filemtime?

realityking:
Fair point. For Joomla itself we could use a hash of the version, but that wouldn't help extensions.

Maybe someone else has a bright idea?

chdemko:
I propose a new way: using the signature of the file folder to force update: if a file named MD5SUM exists in the file folder, append its content to the url. The content of the MD5SUM file must represent the signature of the folder, it must be generated by:
- the 3rd party developer
- the PLT for joomla core css/images/js folders

realityking:
Great idea! We probably have to generate that file on the server and not beforehand to prevent version disclosure but that's a CMS detail.

chdemko:
@realityking There are so many ways to detect Joomla! version (for example downloading a core js file and comparing this file)

libraries/joomla/html/html.php
```diff
@@ -340,13 +340,18 @@ protected static function includeRelativeFiles($folder, $file, $relative, $detec
...
+					/*
+					 * Loop on 1 or 2 files and break on first found.
+					 * Add the content of the MD5SUM file located in the same folder to url to ensure cache browser refresh
+					 * This MD5SUM file must represent the signature of the folder content
+					 */
 					foreach ($files as $file)
 					{
 						// If the file is in the template folder
-						if (file_exists(JPATH_THEMES . "/$template/$folder/$file"))
+						if (file_exists($resource = JPATH_THEMES . "/$template/$folder/$file"))
```
realityking: Small nitpick, could you move to the assignment out of the function call? I think that'd be more readable. There are more of these in other lines below.

realityking:
Fair warning: If no one has anything against this or a better solution by the end of the week I'll pull it.

betweenbrain:
I can't see anything technically against this pull request, and am happy to see this being implemented. +1

nikosdion:
Maybe there is an easier way: get the filesize and date/time, concatenate them, calculate the MD5 and use it as the unique hash for each file. Security-wise it's marginally better: you can not detect the exact version of an extension just by looking at the MD5 sum appended to each JS/CSS file. Furthermore it seems cheaper to do a single stat for each file than one fs query to check for the MD5SUM file and a very expensive read to read it. FYI, that's pretty much how Admin Tools' CSS and JS aggregator figure out if it has to regenerate the concatenated file if one of its components has changed.

realityking:
Well that won't work in the (admittedly rare) case where the file size doesn't change on updates. That would probably only happen when fixing typos thou.

Question is if that's a compromise we wanna make.

nikosdion:
Dude, no compromise! I hate compromises. Read again what I wrote. I wrote "get the filesize AND date/time". Think md5($filesize.'/'.$timestamp) where $filesize is size in bytes and $timestamp the last modified date of the file on disk. This is very accurate. Every time you install an extension update the file is overwritten and gets a new timestamp.

The only edge case would require you to have a FAT32 filesystem, install version N and within 4 seconds install version N+1, the file having changed but its size remain unchanged. OK, if you manage to do that under normal website operating conditions I will buy you a keg of the beer of your choice ;)

realityking:
Indeed I overlooked the date/time. Sorry.

Why even add the file size than?

filemtime is also quite slow (and not cached by PHP) so I'm not sure that's not worse than reading that file.

chdemko:
At the beginning I've used filemtime for each file but it's slow (no caching). Reading the file and computing its md5 could be slow on mootools files for instance.

Another solution is to look at a file called

myfile.(js|css|img).md5 (assuming it has been creating by the third party / code developer)

realityking:
Thanks!

PR #1295, opened 2012-06-20
Eval fix

ghost:
Checking the modules which are called under this, I can't see any reason for the eval call here as it simply duplicates the return function.


chdemko:
The eval function is used to count the number of modules. Removing it simply return the string, not the count

ghost:
I'm not following.  The code today is:
 $str = 'return ' . implode(' ', $words) . ';';
  return eval($str);

$words is an array of numbers, 1,2,3,4,5
So $str = 'return ' . implode(' ', $words) . ';'; results in the string: return 12345
eval($str) will be eval('return 12345;'), 
return 12345; will return 12345 to the caller
return is the caller, so it returns 12345

I'm not seeing the difference between:
return 12345;
and
return eval('return 12345;');

ghost:
oh wait...I see, $words is one part numbers and one part something else, and together they form php code... erg, gonna close this and take another look

PR #1296, opened 2012-06-20
Eval fix

ghost:
Removing the need for eval by putting in the case statements for the conditions.


libraries/joomla/document/html/html.php
```diff
@@ -493,9 +493,171 @@ public function countModules($condition)
 				: count(JModuleHelper::getModules($name));
 		}
```
realityking, line 495: I think we can cop out here if the count is only one since that means no expression has been used. Or am I missing a corner case here?

realityking:
That's a lot of code :(

How about we only use eval() when there's actually an expression used and just tell people that's a feature they can't have if their host disables eval?

realityking:
Also I wonder how this compares in performance? This code is possibly executed a number of times on each page load.

libraries/joomla/document/html/html.php
ghost, line 495: I don't see any reason why it can't exit there when it's only 1....  but if I was going to fix that, then I'd move it further up a bit and do  switch for 1, 2, and more words.   2 is an error condition of some sort[since you can't have only one argument if doing a comparison or arithmetic... ie countModules('leftNav and') is an incomplete condition.    As such, that would require going in and adding unit tests for the 3 options which I don't feel like doing right now.

realityking, line 495: 2 is an error, I agree. But what would you gain by moving the check for one further up? You still have to count the modules for that position.

ghost, line 495: Clarity... I just prefer switch statements to lots of if statements....  like I said, if _I_ do it then I do it the way I like to.  :-)

ghost:
An implementing this logic without the eval is to remove it altogether.

The atomic template, for example, uses:

```
        <?php if($this->countModules('atomic-topquote') or $this->countModules('position-15') ) : ?>
```

Instead of

```
        <?php if($this->countModules('atomic-topquote or position-15') ) : ?>
```

So if not many templates use it this way it could be removed.

chdemko:
You could use

```
$fn = create_function("", "$str" );
return $fn();
```

in place of

```
return eval($str);

```

ghost:
I'm leaving this pull request open for the time being, however I'd say hold off on using it as I also submitted a patch request to handle the cases of 1 or 2 parameters in order to fix a minor security hole.  If those are applied, then this patch will need to be rewritten to include that... and createFunction can be evaluated for use instead.

ghost:
Closing this for now in deference to a more limited functional change in https://github.com/joomla/joomla-platform/pull/1307 and getting unit tests in place for these switch statements

PR #1297, opened 2012-06-21
Fix an issue in bind() when load() is directly from database meaning articletext is "".

elinw:


libraries/legacy/table/content.php
```diff
@@ -115,7 +115,7 @@ protected function _getAssetParentId($table = null, $id = null)
 	public function bind($array, $ignore = '')
 	{
 		// Search for the {readmore} tag and split the text up accordingly.
-		if (isset($array['articletext']))
+		if (isset($array['articletext']) && !empty($array['articletext']))
```
chdemko: in that case, simply use

```
if (!empty($array['articletext']))
```

elinw:
Done.

PR #1298, opened 2012-06-22
Adding changes for behavior.jquery from Joomla CMS pull 218

jlleblanc:
Adds jQuery into core, loadable though behavior.jquery. Code from Joomla CMS pull 218: https://github.com/joomla/joomla-cms/pull/218


nikosdion:
This already looks very good. I have one suggestion, though. There are many Joomla! extensions out there which already use their own copy of jQuery with jQuery.noConflict(). These extensions would screw up Joomla!'s jQuery and any other extensions which depend on this pull request. So, instead, I would do jQuery.noConflict('joomlaJQ') which puts jQuery under the joomlaJQ object. Developer can then do something like (function($) {  ... })(joomlaJQ) to run their jQuery statements with the version of jQuery shipped with Joomla!.

jlleblanc:
@nikosdion while extensions using their own copy of jQuery would be a problem, I disagree that we should create yet another variable name in an attempt to accommodate them. While we are adding jQuery to support Bootstrap, another part of the appeal of adding jQuery to Joomla is to get everyone to use the same copy. That's not going to happen if we create a scenario where multiple copies of jQuery keep getting silently loaded with no problems.

nikosdion:
Point taken. You're right. It's best to weed out misbehaving extensions.

libraries/joomla/document/html/renderer/head.php
```diff
@@ -157,6 +157,22 @@ public function fetchHead($document)
...
+					$buffer .= ' defer="defer"';
+				}
+
+				if ($strAttr['async'])
+				{
+					$buffer .= ' async="async"';
+				}
+
+				$buffer .= '>' . $document->_preScripts[$strSrc] . '</script>' . $lnEnd;
+			}
```
realityking, line 174: "The defer and async attributes must not be specified if the src attribute is not present." (http://www.w3.org/TR/html5/the-script-element.html#the-script-element)

libraries/joomla/html/html/behavior.php
```diff
@@ -65,6 +65,47 @@ public static function framework($extras = false, $debug = null)
...
+	 * Method to load jQuery into the document head
+	 *
+	 * @param   string   $file  An optional path to use in place of the local path. (for CDNs)
+	 * @param   boolean  $debug  An optional argument toggling debug mode. $file takes precedence.
+	 *
+	 * @return  void
+	 *
+	 * @since   12.2
+	 */
+	public static function jquery($file = null, $debug = null)
```
realityking, line 77: How would the $file argument work? As far as I can tell the first caller wins. How did you see this being used?

realityking:
You also have a few code style errors: http://developer.joomla.org/pulls/pulls/1298.html

realityking:
You also have a few code style errors: http://developer.joomla.org/pulls/pulls/1298.html

libraries/joomla/document/html/renderer/head.php
jlleblanc, line 174: Thanks, I'll update this in a bit. I'm not as familiar with defer and async as I should be. If jQuery gets loaded with async, will my jQuery.noConflict() snippet afterwards wait until jQuery is loaded before executing?

libraries/joomla/html/html/behavior.php
jlleblanc, line 77: This is a way of handling CDNs. If someone wants to add a CDN plugin in the future and call it on one of the earlier system plugin events, that'll register it in the system before any other extension tries to. Otherwise, yes, first caller wins.

jlleblanc:
@realityking will look into the style errors momentarily, thanks!

libraries/joomla/document/html/renderer/head.php
realityking, line 174: Nope, inline scripts will be executed as soon as they are encountered. Async really shouldn't be used to load frameworks. An earlier draft of HTML5 actually allowed defer on inline scripts which made sense to me but was dropped for some reason.

libraries/joomla/html/html/behavior.php
realityking, line 77: Wouldn't that mean that activating such a plug-in causes jQuery to always load?

jlleblanc, line 77: Yes, activating such a plugin would cause jQuery to always load, unless you coded it in such a way that it only loaded on certain pages. For instance, you could have a plugin that works with a component to let you choose which pages load jQuery and which ones don't.

realityking, line 77: I don't think that's a behavior we should be encouraging. I think we can come up with a better way to support CDNs so I'd suggest leaving this out for now.

jlleblanc, line 77: Yes, it'd be best if we had a consistent way of managing all JavaScript libraries, which would provide a way for admins to specify CDNs of their choosing. I was trying to avoid addressing CDNs specifically, since I don't know whether that has a good enough shot of making it into core.

Perhaps now would be a good time to revisit the JavaScript library manager concept. I'll create another branch that implements that, as well as jQuery.

jlleblanc:
Closing this pull request in favor of #1359.

PR #1300, opened 2012-06-22
Reduce the use of JBrowser.

realityking:
Note that JHtml still depends on JBrowser, we'll need to fix that before we can move JBrowser to legacy.

Legacy unit tests run fine on my local machine.

### DIFF
--- a/libraries/joomla/application/web.php
+++ b/libraries/joomla/application/web.php
@@ -520,19 +520,6 @@ class JApplicationWeb extends JApplicationBase
 
 				echo $html;
 			}
-			/*
-			 * For WebKit based browsers do not send a 303, as it causes subresource reloading.  You can view the
-			 * bug report at: https://bugs.webkit.org/show_bug.cgi?id=38690
-			 */
-			elseif (!$moved && ($this->client->engine == JApplicationWebClient::WEBKIT))
-			{
-				$html = '<html><head>';
-				$html .= '<meta http-equiv="refresh" content="0; url=' . $url . '" />';
-				$html .= '<meta http-equiv="content-type" content="text/html; charset=' . $this->charSet . '" />';
-				$html .= '</head><body></body></html>';
-
-				echo $html;
-			}
 			else
 			{
 				// All other cases use the more efficient HTTP header for redirection.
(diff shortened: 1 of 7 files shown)


libraries/joomla/application/web.php
```diff
@@ -965,6 +952,18 @@ protected function header($string, $replace = true, $code = null)
 	}
 
 	/**
+	 * Determine if we are using a secure (SSL) connection.
+	 *
+	 * @return  boolean  True if using SSL, false if not.
+	 *
+	 * @since   12.2
+	 */
+	public function isSSLConnection()
```
chdemko, line 961: isn't it a static method?

libraries/legacy/application/application.php
```diff
@@ -1104,6 +1111,18 @@ public static function isWinOS()
 	}
 
 	/**
+	 * Determine if we are using a secure (SSL) connection.
+	 *
+	 * @return  boolean  True if using SSL, false if not.
+	 *
+	 * @since   12.2
+	 */
+	public function isSSLConnection()
```
chdemko, line 1127: static method ?

libraries/joomla/application/web.php
realityking, line 961: I didn't make it static in JApplication to facilitate us eventually moving to JApplicationWeb. and removing JApplication. So I'd be fine doing it here, however I think that's potentially confusing. Maybe we should make it static after dropping JApplication?

chdemko:
a static method could always be called using
$this->isSSLConnection

realityking:
True. But at least for JApplication I'd prefer people not using it since it will eventually disappear.

libraries/joomla/environment/browser.php
```diff
@@ -653,9 +651,12 @@ public function isMobile()
 	 * @return  boolean  True if using SSL, false if not.
 	 *
 	 * @since   11.1
+	 * @deprecated  13.3  Use the isSSLConnection method on the application object.
 	 */
 	public function isSSLConnection()
 	{
+		JLog::add('JBrowser::isSSLConnection() is deprecated. Use the isSSLConnection method on the application object instead.', JLog::WARNING, 'deprecated');
```
chdemko: The line is too long (checkstyle)

realityking: Thanks, fixed.

libraries/joomla/environment/browser.php
```diff
@@ -653,9 +651,13 @@ public function isMobile()
 	 * @return  boolean  True if using SSL, false if not.
 	 *
 	 * @since   11.1
+	 * @deprecated  13.3  Use the isSSLConnection method on the application object.
 	 */
 	public function isSSLConnection()
 	{
+		JLog::add('JBrowser::isSSLConnection() is deprecated. Use the isSSLConnection method on the application object instead.'
+			, JLog::WARNING, 'deprecated');
```
chdemko: checkstyle

realityking: Fixed.

PR #1305, opened 2012-06-22
Import plugins on authorise instead of loading

ShMaunder:
*Apologies if this has been done incorrectly - this is my first ever pull request.

I think the authorise method is suppose to importPlugin() - I can't see any point in using getPlugin() as this will return plug-in data (which isn't captured anyway) and doesn't import them for the dispatcher.

Also, when calling the authorise method, you have to import both plug-in groups manually which seems pointless:
JPluginHelper::importPlugin('user');
JPluginHelper::importPlugin('authentication');
$authorisations = JAuthentication::authorise($response, $options);

### DIFF
--- a/libraries/joomla/user/authentication.php
+++ b/libraries/joomla/user/authentication.php
@@ -326,9 +326,9 @@ class JAuthentication extends JObject
 	 */
 	public static function authorise($response, $options = array())
 	{
-		// Get plugins in case they haven't been loaded already
-		JPluginHelper::getPlugin('user');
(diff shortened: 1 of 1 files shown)


elinw:
https://github.com/joomla/joomla-platform/commit/003359dedadee117d8dcbd4d30b18514525b7c0c

I think the code style error on this one has been fixed the other day but it's still showing up in the pull test.

ShMaunder:
I will redo this as I've totally messed it up. In any case, it may need to be discussed if the import was originally intended or not.

PR #1308, opened 2012-06-22
Use JFactory::getXML() and JXMLElement less.

realityking:


tests/suites/unit/joomla/form/JFormTest.php
```diff
@@ -1360,7 +1360,7 @@ public function testLoad_BadInput()
 		$form = new JFormInspector('form1');
 
 		$this->assertThat(
-			$form->load(JFactory::getXml('<notform><test /></notform>', false)),
+			$form->load(simplexml_load_file('<notform><test /></notform>', false)),
```
chdemko: it should be

```
simplexml_load_string('<notform><test /></notform');
```

realityking: Thanks, fixed.

chdemko:
I think you are mixing two PR here ...

realityking:
Indeed. Thanks, I've fixed.

libraries/joomla/access/access.php
```diff
@@ -496,7 +496,8 @@ public static function getActionsFromFile($file, $xpath = "/access/section[@name
 		else
 		{
 			// Else return the actions from the xml.
-			return self::getActionsFromData(JFactory::getXML($file, true), $xpath);
+			$xml = simplexml_load_file($file);
```
chdemko, line 499: you could have an error here
simplexml_load_file produces an E_WARNING error message for each error found in the XML data.

libraries/joomla/installer/adapters/file.php
```diff
@@ -428,7 +428,7 @@ public function uninstall($id)
 			// @todo remove code: . '/files/' . $manifest->filename);
 			$this->parent->setPath('extension_root', JPATH_ROOT);
 
-			$xml = JFactory::getXML($manifestFile);
+			$xml = simplexml_load_file($manifestFile);
```
chdemko, line 431: check for E_WARNING

libraries/joomla/installer/adapters/library.php
```diff
@@ -290,7 +290,7 @@ public function uninstall($id)
 			// Set the library root path
 			$this->parent->setPath('extension_root', JPATH_PLATFORM . '/' . $manifest->libraryname);
 
-			$xml = JFactory::getXML($manifestFile);
+			$xml = simplexml_load_file($manifestFile);
```
chdemko, line 293: check for E_WARNING

chdemko:
You have to check for E_WARNING when you use simplexml_load_file

realityking:
That's actually intentional, otherwise we'd hide all the XML errors. Since of all these deal with local files shipped with the app or extension I don't think we need to handle them.

When accepting XML that's non local or user supplied that's a different matter.

The alternative would be adding this (from JFactory::getXML()) everywhere:

```
libxml_use_internal_errors(true);

$xml = simplexml_load_file($data);

if (empty($xml))
{
    JLog::add(JText::_('JLIB_UTIL_ERROR_XML_LOAD'), JLog::WARNING, 'jerror');

    JLog::add($data, JLog::WARNING, 'jerror');

    foreach (libxml_get_errors() as $error)
    {
        JLog::add($error->message, JLog::WARNING, 'jerror');
    }
}
```

or using error suppression. I don't think either is necessary.

PR #1310, opened 2012-06-23
Code Style in Jhtmltest plus fix places where files are created twice causing errors

elinw:
There were a bunch of tests that had corrupted unit test results because JHtmlTest was having problems in testStylesheet  when it was attempting to create a file with the same name as a file created earlier.


elinw:
P.S. This whole test needs to be refactored but this is really just phase one so it is possible to read the file easily and also to fix the situation where the same file being tested twice is causing tests to error out.

PR #1311, opened 2012-06-23
Unify the checking of the platform.

realityking:
The standardizes everything on the IS_WIN/IS_MAC/IS_UNIX constants. JApplication::isWinOS() is only deprecated for legacy purposes.

### DIFF
--- a/libraries/joomla/filesystem/path.php
+++ b/libraries/joomla/filesystem/path.php
@@ -9,12 +9,6 @@
 
 defined('JPATH_PLATFORM') or die;
 
-// Define a boolean constant as true if a Windows based host
-define('JPATH_ISWIN', (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN'));
-
-// Define a boolean constant as true if a Mac based host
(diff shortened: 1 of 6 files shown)


libraries/joomla/filesystem/path.php
```diff
@@ -9,12 +9,6 @@
 
 defined('JPATH_PLATFORM') or die;
 
-// Define a boolean constant as true if a Windows based host
-define('JPATH_ISWIN', (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN'));
-
-// Define a boolean constant as true if a Mac based host
-define('JPATH_ISMAC', (strtoupper(substr(PHP_OS, 0, 3)) === 'MAC'));
-
 if (!defined('JPATH_ROOT'))
```
chdemko, line 12: Could we have a B/C issue by removing these constants?

realityking, line 12: Yeah we do. We could move the to the import.legacy.php folder, make them aliases of IS_WIN and IS_MAC and deprecate them. What do you think?

chdemko:
it's a good solution

PR #1313, opened 2012-06-24
Joomla CMS [#28741] E_STRICT: JArrayHelper::pivot should be declared as static

chdemko:
See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28741

### DIFF
--- a/libraries/joomla/utilities/arrayhelper.php
+++ b/libraries/joomla/utilities/arrayhelper.php
@@ -16,7 +16,7 @@ defined('JPATH_PLATFORM') or die;
  * @subpackage  Utilities
  * @since       11.1
  */
-class JArrayHelper
+abstract class JArrayHelper
 {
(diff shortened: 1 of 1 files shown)


libraries/joomla/utilities/arrayhelper.php
```diff
@@ -16,7 +16,7 @@
  * @subpackage  Utilities
  * @since       11.1
  */
-class JArrayHelper
+abstract class JArrayHelper
```
realityking, line 19: I think either we shouldn't do this or do the same with JUtility to be consistent.

chdemko:
This is not a real class since we never do new JArrayHelper. I'm for doing the same on JUtility (maybe in another pull request?)

realityking:
Works for me.

PR #1315, opened 2012-06-26
Update documentation

elinw:


elinw:
https://github.com/joomla/joomla-platform/blob/staging/docs/coding-standards/en-US/chapters/comments.xml#L53

https://github.com/joomla/joomla-platform/blob/staging/docs/coding-standards/en-US/chapters/comments.xml#L94

Sigh.  And no comments about being pedantic please :).

PR #1316, opened 2012-06-26
Add a proxy for execute method

chdemko:
Since a controller is designed to be executed using the execute medhod, it could be a fancy and shorter way to call

```
$controller();
```

instead of

```
$controller->execute();
```

### DIFF
--- a/libraries/joomla/controller/base.php
+++ b/libraries/joomla/controller/base.php
@@ -50,6 +50,20 @@ abstract class JControllerBase implements JController
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


realityking:
I have to say I'm not a big fan of that syntax, but lets see what other are thinking.

beat:
I think an object should stay an object syntax-wise too ;-)
Or are we moving away from OOP ? :-O

LouisLandry:
I certainly think it's clever, but not really necessary.

PR #1318, opened 2012-06-27
Changing error messages into ErrorException

chdemko:
This will allow to not use the silent operator @ for php function that emit error message such as fopen, ...

i.e.

instead of doing

```
$handle = @fopen('/path/to/file', 'w');
```

we could do

```
try
{
    $handle = fopen('/path/to/file', 'w');
}
catch (ErrorException $e)
{
    // Deal with error
}
```

### DIFF
--- a/libraries/import.php
+++ b/libraries/import.php
@@ -6,6 +6,17 @@
  * @license    GNU General Public License version 2 or later; see LICENSE
  */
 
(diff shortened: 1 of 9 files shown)


chdemko:
Closing this one. I've to think to a better solution

Issue #1309, opened 2012-06-22
JInput object in JSession


diananeculai:
In JSession there is a variable defined as: <code>private $_input = null;</code>. In the rest of the code it is used <code>$this->input</code>.

I see 2 fixes for this issue: change the variable definition from $_input to $input or replace $this->input with $this->_input. Which would be better?

realityking:
Good catch.

$this->_input is the correct solution since our code style says to use the underscore prefix for private variables.

diananeculai:
Also, wouldn't it be a good idea to set the $_input variable in the _setOptions() method, like other private variables? Is this what the initialise() method should do? The method's comment is a bit confusing.

PR #1319, opened 2012-06-27
Fix issue #1309

diananeculai:
Replace $this->input with $this->_input.


chdemko:
@realityking Do we also have to use $this->_inout in JSession::checkToken?

chdemko:
ok, it's a static function, we can't use $this->_input

PR #1320, opened 2012-06-27
Check that $wpoption['modulemode'] == 1, not just that it is set.

okonomiyaki3000:
This is a fix for this bug:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28676

### DIFF
--- a/libraries/joomla/cache/controller/callback.php
+++ b/libraries/joomla/cache/controller/callback.php
@@ -144,7 +144,7 @@ class JCacheControllerCallback extends JCacheController
 			if (isset($woptions['modulemode']) && $woptions['modulemode'] == 1)
 			{
 				$document = JFactory::getDocument();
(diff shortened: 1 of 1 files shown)


libraries/joomla/cache/controller/callback.php
```diff
@@ -141,7 +141,7 @@ public function get($callback, $args = array(), $id = false, $wrkarounds = false
 				$locktest = $this->cache->lock($id);
 			}
 
-			if (isset($woptions['modulemode']))
+			if (isset($woptions['modulemode']) && $woptions['modulemode'] == 1)
 			{
 				$document = JFactory::getDocument();
 				$coptions['modulemode'] = $woptions['modulemode'];
```
chdemko: in that case you could replace this line by

```
$coptions['modulemode'] = 1;
```

PR #1322, opened 2012-06-27
Support applications without JMenu.

realityking:
Deprecate some code not needed with the autoloader.

### DIFF
--- a/libraries/legacy/menu/menu.php
+++ b/libraries/legacy/menu/menu.php
@@ -93,6 +93,8 @@ class JMenu extends JObject
 
 			if (!class_exists($classname))
 			{
+				JLog::add('Non-autoloadable JMenu subclasses are deprecated.', JLog::WARNING, 'deprecated');
+
 				// Load the menu object
 				$info = JApplicationHelper::getClientInfo($client, true);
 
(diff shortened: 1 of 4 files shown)


libraries/legacy/menu/menu.php
```diff
@@ -112,7 +114,8 @@ public static function getInstance($client, $options = array())
 			}
 			else
 			{
-				throw new Exception(JText::sprintf('JLIB_APPLICATION_ERROR_MENU_LOAD', $client), 500);
```
chdemko: why are you removing this line since you don't remove it router.php?

realityking: That's intentional. If you look in the old code for the 3 classes, JPathway and JRouter used to return a JError object but JMenu returned null. Without this the CMS installation actually fails to work.

We could deal with this in JApplication of course, I thought this was desirable for B/C but I wouldn't mind changing it.

realityking:
Added some fixes for things I overlooked the first time. We really should catch these exceptions in JApplication. Now we'd be able to deal with JMenu::getInstance() throwing exceptions as well, but I'm not sure of the B/C impact.

piotr-cz:
Yes!

PR #1323, opened 2012-06-28
Joomla doesn't cache modules for logged users

Denitz:
This is a big issue since 2.5.0, modules should be cached for logged users also.

### DIFF
--- a/libraries/legacy/module/helper.php
+++ b/libraries/legacy/module/helper.php
@@ -416,8 +416,8 @@ abstract class JModuleHelper
 		$cache = JFactory::getCache($cacheparams->cachegroup, 'callback');
 		$conf = JFactory::getConfig();
 
-		// Turn cache off for internal callers if parameters are set to off and for all logged in users
-		if ($moduleparams->get('owncache', null) === '0' || $conf->get('caching') == 0 || $user->get('id'))
+		// Turn cache off for internal callers if parameters are set to off
+		if ($moduleparams->get('owncache', null) === '0' || $conf->get('caching') == 0)
 		{
 			$cache->setCaching(false);
 		}


Denitz:
https://github.com/joomla/joomla-cms/pull/256
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28760

ianmacl:
How do you account for modules that have user specific data?

Denitz:
You can either disable caching in module settings, or implicitly disable caching by not using 'owncache' parameter.

Denitz:
Any news???

ianmacl:
This is in the legacy tree so it is really up to whether the CMS folks want it or not.  If they want to way in with a yay or a nay then we can make a decision on it.

elinw:
@Denitz I'd like to see you post to the CMS list and general list (to catch module developers) for feedback.  It's a major change in behavior especially if there is no option to use to the 2.5 (and all of CMS history) behavior globally. I think it would require quite a lot of work for webmasters with existing sites potentially (depending on what they are currently doing) to have to change settings on many modules.
  I think in general the CMS would introduce this in a manner that defaults to the existing behavior and maybe you could propose this in a way that deals with it comprehensively possibly adding a parameter to cache and possibly changing core and sample data (I haven't looked but if module instances are now going to have to pay different attention to the settings for this).

Denitz:
Elin, this is not a major change in behavior, before 2.5. modules were cached for logged users.
So in fact not caching is incorrect behavior! It seriously affect performance, even for core modules - try to load Related Articles with 50.000 articles.
It is up to module code to decide whether to cache it or not, and normal developer understand this and can use module cache parameters.
Having a big experience in caching with Joomla, introduced lots of caching features in previous versions, I can say that this issue is obvious.
No need to change core data or add additional parametes, current code is enought to allow developers the full control of modules caching.

elinw:
Can you show me the pull request/commit in the last two years that changed this?

mahagr:
I've never seen any module caching contents for registered users unless the developer has done the caching manually (which we had to do back in golden Joomla! 1.5 times -- and our code hasn't changed since).

Denitz:
Joomla 1.5 cached module correctly:
if ($mod_params->get('cache', 0) && $conf->getValue( 'config.caching' )) {
  ... cache
}

But 2.5.0 even in beta 1 stopped to cache module for logged user.

2 mahagr: I have seen tons of modules which do require caching for logged users.
We shouldn't implicitly decline the module caching but rely on module code, the code should decide how to cache itself depending on module params.

Otherwise Joomla module caching is not flexible and hard-coded - back in golden Joomla! 1.0 times

mahagr:
I agree that the module should decide when to cache and how. You cannot compare static content with a module that updates every few seconds.

elinw:
I really think that it has been this way at least since 1.6 was released.
Thinking about this, obviously this won't impact 2.5 since that is in LTS (and had  a 1.6, 1.7 and the run up to 2.5 for people to identify and request changes) and developers and users have a right to expect stable behavior from an LTS.  But it could be reasonable to change the behavior.   I still would like to see a general way to allow webmasters to enable/disable caching for logged in users. It's hard for experienced users to remember that using cache at all is something that webmasters approach extremely cautiously but at the same time, it's good for us to encourage (similarly to SEF).  If we ever were to move to have caching on by default, it would be for guests only I would guess. So it would make sense to even think about how to implement this in some detail including the user interface.

mahagr:
The issue with Joomla caching is that you are generally either caching the content for time x or you're not. If we want to add support for module caching, it needs to be done so that module can have more power on it.

For example module may want to update output only when something has changed (by making a simple query). That is useful for example if the module is displaying activity stream or forum posts. Module may also want to display different content based on viewlevels or it could be for that user only -- or the content could be identical to all users. I have also module that has small parts which are dynamic -- for example to mark if user has already visited the item. And the list goes on...

So very careful thinking is needed before adding caching for different users as there are many different requirements how the caching should be done.

ianmacl:
As this is in the legacy tree and really CMS specific, I'm going to close this issue until the CMS folks have discussed it and made a decision.  If they want to go ahead with it, that's fine.  It's really up to them.

PR #1324, opened 2012-06-28
Remove getter for non-existing "class" property

ghost:
Remove getter for non-existing "class" property (we're using class as part of the element).


elinw:
My question is why isn't class a property?

ghost:
Don't know who desgined that, but we're using "class" as part of "element". I don't like it too.

PR #1325, opened 2012-06-28
Don't use deprecated APIs.

realityking:


chdemko:
JXMLElement is now deprecated. Why do you it?

realityking:
Backwards compatibility. I was considering moving it to legacy and only using it when legacy is loaded. Do you think we should just not use it?

LouisLandry:
Ya, let's just not use it as much as is possible.

realityking:
Updated to remove JXMLElement from everything but JFactory::getXML().

PR #1327, opened 2012-06-28
Fix two parse errors

elkuku:
Yes i know that php lint takes a lot of time but...


realityking:
Guess that what you get for relying on the unit tests. Thanks!

PR #1329, opened 2012-06-30
JSession: Removed duplicated code, code style

piotr-cz:
Minor fixes not affecting functionality

### DIFF
--- a/libraries/joomla/session/session.php
+++ b/libraries/joomla/session/session.php
@@ -368,12 +368,6 @@ class JSession implements IteratorAggregate
 				continue;
 			}
 
-			// If the class doesn't exist we have nothing left to do but look at the next type. We did our best.
-			if (!class_exists($class))
-			{
-				continue;
-			}
-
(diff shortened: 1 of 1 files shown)


libraries/joomla/session/session.php
```diff
@@ -408,11 +402,7 @@ public function isActive()
 	public function isNew()
 	{
 		$counter = $this->get('session.counter');
-		if ($counter === 1)
-		{
-			return true;
-		}
-		return false;
+		return (bool) ($counter === 1)
```
realityking: You're missing the ; at the end of the line.

piotr-cz: Thanks, fixed.

I don't understand why there is a typecast to boolean in `JSession->isActive` and I repeated the style. Should I remove these?

realityking: However you want, it's not needed but personally I don't mind it.

piotr-cz: Lets keep it then

PR #1330, opened 2012-06-30
Joomla! throws error with valid xml string

bembelimen:
Test:

$xml = JFactory::getXML('&lt;foo /&gt;');

expected result:

No JError warning

actual result:

Joomla! calls a "JLIB_UTIL_ERROR_XML_LOAD" JError Warning

Solution:

Don't check for "empty()", use === false instead.

### DIFF
--- a/libraries/joomla/factory.php
+++ b/libraries/joomla/factory.php
@@ -407,7 +407,7 @@ abstract class JFactory
 			$xml = simplexml_load_string($data, 'JXMLElement');
 		}
 
-		if (empty($xml))
+		if ($xml === false)
 		{
 			JLog::add(JText::_('JLIB_UTIL_ERROR_XML_LOAD'), JLog::WARNING, 'jerror');
 


realityking:
Could you add some unit tests for this so we don't regress on this?

bembelimen:
I updated the comment, cause the &lt; wasn't visible.

realityking:
I think you misunderstood me. I meant to add a test for this behavior to https://github.com/joomla/joomla-platform/blob/staging/tests/suites/unit/joomla/JFactoryTest.php

bembelimen:
No, I didn't missunderstand you, we posted nearly at the same time ;) So my comment wasn't related to you, just an update notification what I changed.

I will look into the test stuff.

bembelimen:
Ok, I looked in the test cases and I wouldn't have a problem to write one, if the return value of the method is broken. But I have no idea how I catch a JError warning. Any suggestions?

realityking:
Not sure what you mean, JFactory::getXML() doesn't use JError.

bembelimen:
Ok, I try to explain it more:

Background:
If you load this (&lt;foo /&gt;) valid (XML) string with JFactory::getXML() you get an empty (J)SimpleXML Object. But this object is correct. That means empty($xml) says true but $xml !== false. The XML string is only invalid, if the simplexml loader functions return false.

To the problem:
Joomla! tests only if the object is empty and not if it is false. That means, that if we use &lt;foo /&gt; we will jump into the if clause (cause empty($xml) === true). And there JLog logs a JError message.

So my question ist:
How do I write a test which detects this misbehavior? I need something like:
Input: &lt;foo /&gt;
Detection of a JLog call => test false

Link: https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/factory.php#L412

realityking:
Ah sorry, now I understand the confusion. JError is also a class thats quite hard to unit tests, and this is indeed not used here.

I just looked at the code again and it looks like the actual return value for the method is correct. It's just the log messages you want to remove, correct? In that case I'll pull this as is.

bembelimen:
Yep, exactly. Thank you.

PR #1335, opened 2012-07-01
Make JToolbar more semantic.

realityking:


libraries/legacy/toolbar/button/help.php
```diff
@@ -42,11 +42,11 @@ public function fetchButton($type = 'Help', $ref = '', $com = false, $override =
 		$class = $this->fetchIconClass('help');
 		$doTask = $this->_getCommand($ref, $com, $override, $component);
 
-		$html = "<a href=\"javascript:void(0)\" onclick=\"$doTask\" rel=\"help\" class=\"toolbar\">\n";
```
ianmacl, line 45: What is this supposed to do?  It doesn't look like anything happens if you click on this button.  Did I miss something?

ianmacl:
https://tinker.io/f7b7e

Demo link for using button instead of href.  If people can try it in a couple of browsers it would be helpful.  You want to make sure that the alert appears and that the message is logged to the console.

elinw:
Works for me in Chrome on Ubuntu.

libraries/legacy/toolbar/button/help.php
realityking, line 45: You're right, I accidentally removed the onclick attribute. I'll add it back.

realityking, line 45: Fixed.

realityking:
Closing since JToolbar has been moved to the CMS.

PR #1341, opened 2012-07-02
Joomla CMS [#28633] JUserHelper getProfile method should be static

ShMaunder:
In no way a major issue; noticed while using JUserHelper::getProfile that it is
not declared as static:

public static function getProfile($userId = 0)

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28633

I don't believe there are backward compatibility issues.


realityking:
Thanks.

PR #1342, opened 2012-07-03
New Cache Package

LouisLandry:
This is **not** ready to be merged, but I wanted to start the conversation about how we dramatically simplify the caching package for Joomla.  The existing system is entirely too complex and doesn't honor the principle of separation of concerns.  Hopefully this can serve as a starting point for us to figure out what caching looks like for Joomla in the longer term.  As with everything these days I'm striving for simple, clear and concise.


piotr-cz:
I'm quite happy with JCache shipped with 12.1.

But few things I've been missing:
- not possible to overwrite some of the cache options (which are set based on CMS settings).
- I'd like to save in image in cache before start to manipulate it using JImage. Manipulated image should be stored in cache as well until User decides to save it in image format and not JCache format. I ended up doing it without JCache.

ghost:
Hm, I'd rather go in the opposite direction and allow for more complex settings such as Stash: https://github.com/tedivm/Stash

I'm also not thrilled with losing the group/pool/bin setting on the cache....  I realize that almost no one uses it though so that's likely just me.

With Stash, instead of having that special runtime property for caching in memory, it adds an "ephermeral" stash handler which keeps it treated by itself.

Moreover, their recent commits include  a configurable Multihandler cache storage handler which can be configured to use multiple cache backends intelligently...ie for incredibly large objects[say the JDocument object] the cache can be to Ephemeral + File, really small objects could go to Ephemeral + APC[with it's shared memory model], and everything else can go to Ephemeral + Memcache

The cool factor is on lookups, it will check in all configured storage areas in an optimized order, ie check Ephemeral, then APC, then Memcache - so lookups take a small hit equal to all the failures[variables from Ephemeral.

pasamio:
I'd suggest that for someone wanting a more complete cache library with more complicated caching strategies that they could then look into using Stash. Given you point out yourself that almost nobody uses it, it doesn't make sense to maintain a feature that is barely used and is better supported elsewhere in place of providing a simpler and more reliable caching mechanism to meet the majority of use cases.

mahagr:
+1 for the simplification -- except that there's really a need to clean up the whole cache. For example in Joomla! 1.5 there was no way to clean up APC cache, which in many cases meant that you had to disable caching and put it back on after few hours to make changes visible. Or you needed to log in the server and do it from command line. Fortunately it's possible in Joomla! 2.5. :)

There's also a need to free up disk space when using file cache, otherwise unused cache entries will slowly fill up the disk.

florianv:
+1 but I think it should be possible to inject a Memcached object in the constructor of JCacheMemcached, allowing a null value and calling _connect() as fallback.

LouisLandry:
That's an interesting idea @florianv ... I honestly hadn't thought of that.

@mahagr I agree that in general that is a need.  I'll take a look at it, but that wasn't really the exercise here.  Mostly I am trying to get a feel for how people see the simplification and straightforward approach.  I think we've always been too complex in how we deal with cache with all these controllers and handlers and such.

Thanks for the feedback guys!

florianv:
@LouisLandry I think about that because in the future we will probably have a dependency injection container holding the JRegistry options and it's likely we will also get drivers from him. So I think there should be a way to inject them via constructor or setter.

libraries/joomla/cache/cache.php
```diff
@@ -10,708 +10,215 @@
...
-	 * @since  11.1
+	 * @var    JRegistry  The options for the cache object.
+	 * @since  12.3
 	 */
-	public $_options;
+	protected $options;
 
 	/**
-	 * Constructor
+	 * Constructor.
```
elinw, line 34: Why?

LouisLandry, line 34: Why ask why?

libraries/joomla/cache/memcached.php
```diff
@@ -0,0 +1,203 @@
...
+	 * @var    Memcached  The memcached driver.
+	 * @since  12.3
+	 */
+	private $_driver;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param   JRegistry  $options  Caching options object.
+	 *
```
elinw, line 31: Why? Why all these periods?

eddieajau:
How are you thinking of instantiating the cache objects (or are we not there yet)?

libraries/joomla/cache/cache.php
elinw, line 34: Because I can.

libraries/joomla/cache/memcached.php
LouisLandry, line 31: Because they are the Best. Descriptions. Ever.

libraries/joomla/cache/cache.php
LouisLandry, line 34: That can be fixed. :wink:

LouisLandry:
I was thinking something like `$cache = new JCacheApc;`.  What do you think?

eddieajau:
I was just wondering if you wanted to include a `getInstance($driver)` convenience method.  I'm just thinking of the case where the driver type is coming out of a configuration setting.

ianmacl:
It would probably have to be getInstance($options) and $options would have to have a 'type' or 'driver' property.  I would be okay with that so long as it doesn't try and cache cache objects.  That isn't JCache's job.

eddieajau:
Or we add a factory class...

dongilbert:
@ianmacl what other properties would be present in the `$options` array that would warrant such an implementation, rather than just passing the driver string?

ianmacl:
Sorry Don,  didn't see the question before.  Different cache drivers would obviously have different options.  If you were using memcached, for example, you might have the address or IP of the memcache server(s) or other connection info.  For a file cache you might want to have a configuration option for the root path of the cache folder.  For a database cache, you might want to be able to inject in a database driver object.

dongilbert:
Makes sense - I didn't think of that before.

libraries/joomla/cache/memcached.php
javigomez, line 31: @elinw the period makes the code assistant of the IDE to stop showing more information. When you put a mouse over a function/method, the IDE usually shows the description of this method on a tooltip. If there is no dot/period, the code assistant continues reading showing something like: Constructor @param JRegsitry.... instead of only showing the description of the method: Constructor. I guess Louis did it for that.

florianv:
It's worth looking at https://github.com/fabpot/Pimple . It's a really simple dependency injection container that we could use to replace the factory. 

If we want a configuration file to define the dependencies, it becomes suddently more complicated because we need to check for circular references : A needs B, B needs A. The dependencies must be seen as a graph, and each param/service is a node.

eddieajau:
Closing this PR as it is incorporated into the new Joomla Framework.

Issue #1179, opened 2012-05-02
scriptfile not supported by language, library, template


HLeithner:
A scriptfile is not supported by these 3 installer types.

Looking at the php files shows no handling of the scriptfile tag.

elinw:
see #901 also.

eddieajau:
I think this is fixed now. If not, re-open the issue.

TorstenE:
Hey Eddieajau, Joomla 3.3.6 does not Support install script in templates

mbabker:
It was added to 3.4.0.

Issue #901, opened 2012-02-20
Manifest SQL entires are not run for modules and plugins on discovery install


elinw:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=24956

This issue has been in the CMS tracker for a year and I'm just moving it here anyway. Ian or Sam could just decide to close if they wish, but it needs to get resolved one way or another..

eddieajau:
I think this is fixed now. If not reopen or resubmit.

PR #1343, opened 2012-07-04
Handling msg object in adapter uninstall

mbabker:
1) Presently, a PHP Notice "ob_end_clean(): failed to delete buffer. No buffer to delete in /libraries/joomla/installer/adapters/plugin.php on line 638" is thrown during plugin uninstall because a `$msg` object is created before the `ob_start()` command is issued.  This instance is not necessary since the object is set at the correct time after the custom uninstall method.

2) In the file, module, and package adapters, the `$msg` object wasn't registered to extension_message, contradictory in behavior to the component and plugin adapters.  This is corrected.

### DIFF
--- a/libraries/joomla/installer/adapters/file.php
+++ b/libraries/joomla/installer/adapters/file.php
@@ -485,6 +485,11 @@ class JInstallerFile extends JAdapterInstance
 			$msg = ob_get_contents();
 			ob_end_clean();
 
+			if ($msg != '')
+			{
+				$this->parent->set('extension_message', $msg);
+			}
+
 			// Let's run the uninstall queries for the extension
 			$result = $this->parent->parseSQLFiles($this->manifest->uninstall->sql);
 
--- a/libraries/joomla/installer/adapters/module.php
+++ b/libraries/joomla/installer/adapters/module.php
@@ -798,6 +798,11 @@ class JInstallerModule extends JAdapterInstance
 		$msg = ob_get_contents();
 		ob_end_clean();
 
+		if ($msg != '')
+		{
+			$this->parent->set('extension_message', $msg);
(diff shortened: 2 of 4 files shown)


pasamio:
Actually 1 is caused because you deleted the plugin preflight uninstall method. If you put the uninstall preflight method back then you won't have this error. Do you mind putting it back?[1]

Additionally it's supposed to append at the later point as the comment alludes to however it doesn't, this is the real bug.

[1] https://github.com/mbabker/joomla-platform/commit/a3600e91df25f3511e3b39949a0faa4c8d5bc72d#L6L622

mbabker:
When I removed the preflight from uninstall, the plugin adapter was the only one which had this available.  Is it intended that all of the adapters have preflight available at uninstall, or was the plugin adapter a special case?

pasamio:
The idea was for library plugins (which are still reasonably popular) to be able to prevent uninstallation if their dependent extensions were installed. Obviously uninstalling the underlying library for an extension could cause problems for the extension and potentially break the site. Almost every other extension type doesn't have that behaviour except libraries (so if it wasn't there, that's an oversight) where uninstalling the plugin can take out the site.

mbabker:
Gotcha, I'll fix it up later on then.

elinw:
We have the same issue in (and issue reports for) all of the other adapters   (see #901 and #1179)  Sam can you give me your take on those two issues? They have been sitting a long while.

pasamio:
@elinw: Re #1179 I don't think it makes sense to have those three have script files similar to those items specifically geared at interacting with the user (component, module, template). I don't have any objection to it, I just don't feel it makes sense. Re #901 sounds like a minor oversight

PR #1345, opened 2012-07-06
Handle scriptfile for file extensions

mbabker:
At present, the scriptfile for file extension types isn't copied over during install, leaving it unavailable for uninstall routines.  This change sets it up so that it is installed in a manner similar to packages.

### DIFF
--- a/libraries/joomla/installer/adapters/file.php
+++ b/libraries/joomla/installer/adapters/file.php
@@ -99,7 +99,7 @@ class JInstallerFile extends JAdapterInstance
 			}
 		}
 		// Set the file root path
-		$this->parent->setPath('extension_root', JPATH_ROOT);
+		$this->parent->setPath('extension_root', JPATH_MANIFESTS . '/files/' . $this->get('element'));
 
 		/**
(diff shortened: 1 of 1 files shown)


pasamio:
@ErisRod good point, the changes on line 102, line 429/451, line 588 to 593 need to be reverted. This will mean that the additions on line 346 need to be reworked. I should have looked at this closer as the intent of it is to be able to put files anywhere in the file system and this breaks it. @mbabker do you mind looking into getting this resolved quickly?

pasamio:
Alternatively, @ErisRod if you could update and put in a pull request to fix things that'd be good as well.

mbabker:
Could someone point me to code that's broken?  I use a files extension to drop in alternate layouts for CMS templates and haven't seen an issue with install/uninstall.

pasamio:
I assume the reason is because you're changing extension_root. Though I could be wrong, @ErisRod care to expand on your issue?

HLeithner:
It seams that the problem was in my install xml, with the scriptfile tag using a wrong path. After changing my xml file it seams to work perfectly. Sorry for false error reporting.

PR #1346, opened 2012-07-07
Update staging

infograf768:
See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28535

This solves the display of the Modified date

### DIFF
--- a/libraries/joomla/html/html.php
+++ b/libraries/joomla/html/html.php
@@ -879,7 +879,7 @@ abstract class JHtml
 		else
 		{
 			return '<input type="text" title="' . (0 !== (int) $value ? self::_('date', $value, null, null) : '')
-				. '" value="' . (0 !== (int) $value ? self::_('date', $value, null, null) : '') . '" ' . $attribs
+				. '" value="' . (0 !== (int) $value ? self::_('date', $value, 'Y-m-d H:i:s', null) : '') . '" ' . $attribs
 				. ' /><input type="hidden" name="' . $name . '" id="' . $id . '" value="' . htmlspecialchars($value, ENT_COMPAT, 'UTF-8') . '" />';
 		}
 	}


realityking:
Is there a reason you don't do the same change in line 881 and 875?

Also in the future, please use a more descriptive title for pull requests. They are used the automatically generate change log.

ianmacl:
Since this is a regression I'd like to see a test for this behaviour so that it doesn't happen again.  Also it would be helpful to have at least some explanation as to what is actually happening.

infograf768:
What is happening is that the Modified date is displayed as full date, i.e. (
Wednesday, April 1st 2012
 instead of 'Y-m-d H:i:s', for example
2012-07-05 22:07:27

As, in some languages the "full" date is much longer than the 22 characters allowed for this field (We have some with 33), the "full" date is truncated.
We do display in 2.5.6 'Y-m-d H:i:s'.
The original patch was to solve the doubled time zone but overlooked this aspect. https://github.com/joomla/joomla-platform/pull/1293https://github.com/joomla/joomla-platform/pull/1293

The reason why lines 875 and 881 are not the same is that the Calendar field is automatically displayed as  'Y-m-d H:i:s' in 875.
See calendar.php: $this->value = $date->format('Y-m-d H:i:s', true, false);
On the Joomlacode tracker are some screenshots.

infograf768:
Another solution is to use;
. '" value="' . (0 !== (int) $value ? self::_('date', $value, $format, null) : '') . '" ' . $attribs

elinw:
Andrea proposed the patch. Here is her comment:

There really isn't enough room to increase the size when the screen is at the
minimum size. It sounded like the main reason the Platform changed the format
is they didn't want to make the call to the database class for the format. That
format is the same for everyone in all databases: Y-m-d H:i:s so we could just
pass the format instead. Then we don't have a b/c issue with the formatting.

I uploaded patch 4 with the change. If we do this we will also want to get the
platform to make the change as well.

infograf768:
Can this be merged please? We need to correct it in 2.5.x and master

realityking:
I'm merging this now but I'd really be good if someone submitted unit tests for this,

PR #1347, opened 2012-07-09
Move JXMLElement to legacy.

realityking:


LouisLandry:
Looks like you have an error executing the unit tests with this change.  Should be a simple fix when you get a moment.

realityking:
Thanks, should be fixed now.

PR #1349, opened 2012-07-10
Some additional preparations for lazy sessions.

realityking:
This is mostly small clean up, we need access to the store name to make some decisions based on whether the database is used as the store or not. This way we don't have to rely on the configuration for that.

The interesting part here is https://github.com/joomla/joomla-platform/pull/1349/files#L1R591. This code is kinda specific to the Joomla-Platform and used in several places, it does limit the ability of using our session system without our user system for example. The question is where else to put it. Any suggestions?

### DIFF
--- a/libraries/joomla/application/web.php
+++ b/libraries/joomla/application/web.php
@@ -1049,13 +1049,6 @@ class JApplicationWeb extends JApplicationBase
 			$session->restart();
 		}
 
-		// If the session is new, load the user and registry objects.
-		if ($session->isNew())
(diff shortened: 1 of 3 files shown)


realityking:
One alternative I originally had for the CMS is putting it in a event (the onAfterSessionStart event trigger will be added in the next pull request) but that's not really a solution for the platform.

realityking:
I discussed this with Louis and I think we came up with a better solution. I'll be back.

PR #1350, opened 2012-07-10
Add a more complex truncation method to JHtmlString (with tests)

elinw:
This method allows you to truncate a string of html to a fixed number of characters of text not counting html tags. It builds on the existing string.truncate method.

### DIFF
--- a/libraries/joomla/html/html/string.php
+++ b/libraries/joomla/html/html/string.php
@@ -182,8 +182,8 @@ abstract class JHtmlString
 		for ($maxLength; $maxLength < $baseLength;)
 		{
 			// We need to trim the ellipsis that truncate adds.
-
 			$ptString = rtrim($ptString,'.');
+
 			// Now get the truncated string if HTML is allowed.
 			$htmlString = JHtml::_('string.truncate', $html, $maxLength, $noSplit, $allowHtml = true);
 			$htmlString = rtrim($htmlString,'.');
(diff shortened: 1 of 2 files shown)


elinw:
This is something that we needed in the CMS because sometimes you have such a high ratio of html to text that when you truncate retaining html you end up with almost no text.  It seemed like it might be useful for other people rendering on the web so I'm sending it along.

elinw:
Okay fixed some bugs in Truncate and the truncate tests as well as getting all my tests to pass.

elinw:
I'm not sure about the += versus ++ code style issue ... it's not an increment but maybe I should write out the a = a + 1; ?

mahagr:
I find shorthands easier to read. Just be careful when using those inside for loops and inside functions as they may have different meaning than you expect.

$i++ and $i-- first calls the function and only after that increases the value. If you want to increase value before calling the function, you must use ++$i or --$i instead. 

Oh, almost forgot to say that $i+=1 works like ++$i...

elinw:
Did you see the line? I just changed it to write out the equation. It's just adding 1 to a  variable one time.

mahagr:
No, I didn't see that one. I just meant it as general note your comment as:

```
$i = 0;
echo $i++; // 0
echo ++$i; // 2
```

Alone it works like you expect it to:

```
$i = 0;
$i++;
echo $i; // 1
```

elinw:
Yes I know how increments work.

No that's not what I'm doing and that's why I am questioning Jenkins marking it as a CS error.
I am doing
$something = strpos(something);
 bla blah blabh

we now need to get $something+1 because we want to use it as a length not a position

$something += 1;

or 
$something = $something +1;

as you show $something++; is not a good idea.

elinw:
Replaced by #1413

Issue #1354, opened 2012-07-12
Incorrect static variable caching of instances causing infinite nesting in JForm


klas:
![Debugger screenshot](https://p.twimg.com/AxjPkwPCQAEM4da.jpg:large)

Storing instances in static variable accesible from the whole class is causing infinite nesting as each new instance will already contain this variable. Solution is to store them in static variable with  a method scope

realityking:
What app is that? Eclipse?

klas:
Yes, Eclipse.

klas:
After some thought - only proper solution is to remove all this inside class variable cachings and move them to a  factory class. In second phase it would be probably wise to also move all instances creation methods to separate factory classes. 

The easiest and short term solutin - add new generic / abstracted JFactory  method  that would take class name as a parameter and would take care about instances creation for all classes, for now just call classes existing getInstance methods and cache them in static variable array. Something like

```
class JFactory {
  static $instance = array();
 ...
 ...
  createInstance( $classname, $parameters)  {

    $instance[array($classname, parameters)] = $classname::getInstance($parameters);

  }
 ...
 ...
}
```

Final goal would be to have abstract JFactory and then implement separate factories for platform subsections (.e.g by current subfolders) with concrete methods for each class in folder, replacing current getInstance methods currently residing inside classes.

dongilbert:
Continue the conversation on the pull request here: #1355

PR #1355, opened 2012-07-12
Fix static variable caching of instances causing infinite nesting in JForm,

klas:
#1354

### DIFF
--- a/libraries/joomla/form/form.php
+++ b/libraries/joomla/form/form.php
@@ -62,12 +62,6 @@ class JForm
 	 */
 	protected $xml;
 
-	/**
-	 * Form instances.
-	 * @var    array
-	 * @since  11.1
-	 */
-	protected static $forms = array();
 
 	/**
 	 * Method to instantiate the form object.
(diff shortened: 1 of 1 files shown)


pasamio:
Sounds like a good case for a unit test as well to verify that we don't break things going forwards either.

LouisLandry:
@klas The reason we moved away from having method statics is because it is impossible to test with them because you cannot access the array in any way to modify it -- even from within the class or it's children.  Is it possible what you are seeing is a side-effect of late static binding or something?  The code definitely seems to work as it is now.  Can you put together a small POC gist or something that shows things not working we can look at?

klas:
Hi,

I don't think it is a sideeffect - it makes sense as class static variable has the same value across instances so when you create an instance and store it inside a class you start an indefinite loop.

I found this as my debugger (XDebug) kept on crashing - I presume debugger tries to trace all variables independently while PHP interpreter obviously has a way to break such loops as I haven't found any noticeable differences in memory usage or timings. 

As I see it, the way to properly solve it is to either use method scoped static variable or - this would be much better imho - for now use factory class to create and store instances for all platform methods and in the future add abstract factory class to avoid modifying factory each time new classes are addded.

eddieajau:
Thanks for looking into the problem. I'm going to close this PR for now since we don't have a mergable solution.

klas:
@dongilbert @mbabker @phproberto  this issue still exist (in all major classes, not just JForm). My proposal in the PR comment was to create a generic factory - just for creating and storing joomla class instances, something like code here  https://github.com/joomla/joomla-platform/issues/1354#issuecomment-9157110

Could any of you take another look at this?

Issue #1252, opened 2012-06-05
JUser is initialized when we try to store using JTableCategory


fastslack:
Im trying to insert a new category using the platform and JTableCategory class like:

```
$category = JTable::getInstance('Category', 'JTable', array('dbo' => $this->db_new));
$category->bind($object);
```

and i get this error:

```
Fatal error: Class 'JRequest' not found in /joomla-platform/libraries/joomla/session/session.php on line 526
```

Looking a bit into the code i see that into the store() method, the JUser class is called and it try to initialize a new session.

This is needed if we try to use the JUser table from platform/cli ?

fastslack:
Seems that this is fixed on master and staging branches. Im using the 12.1 tag, im wrong to using it?

LouisLandry:
You aren't wrong to use 12.1 ... it is just a bug in 12.1 that slipped through the cracks and was fixed more recently.  You should be safe using master I believe.

fastslack:
Nope, im getting "Application Instantiation Error" because the store() method of the JTableCategory class is calling to JFactory to get the JUser instance, but im on CLI so no user/session is initialised.

libraries/joomla/table/category.php -> Line 202 

```
    public function store($updateNulls = false)
    {
        $date = JFactory::getDate();
        $user = JFactory::getUser();

        if ($this->id)
        {
            // Existing category
            $this->modified_time = $date->toSql();
            $this->modified_user_id = $user->get('id');
        }
        else
        {
            // New category
            $this->created_time = $date->toSql();
            $this->created_user_id = $user->get('id');
        }

        // Verify that the alias is unique
        $table = JTable::getInstance('Category', 'JTable', array('dbo' => $this->getDbo()));
        if ($table->load(array('alias' => $this->alias, 'parent_id' => $this->parent_id, 'extension' => $this->extension))
            && ($table->id != $this->id || $this->id == 0))
        {

            $this->setError(JText::_('JLIB_DATABASE_ERROR_CATEGORY_UNIQUE_ALIAS'));
            return false;
        }
        return parent::store($updateNulls);
    }
```

Im using the master branch now.

elinw:
Is that the same issue as # 267?

Does your cli set a user for the created_by_user value?  If there is no session and value set from the application should it be null?

oops I meant #276

fastslack:
yes is related with #276

I have not user set on my CLI script and i dont need it neither, so that causes the error. The "Application Instantiation Error" is from the JFactory::getUser()

[...]

Issue #276, opened 2011-08-22
The Platform needs to decouple the JUser class from the JTable class


cozimek:
Specifically, the following lines in /libraries/joomla/database/table/content.php are problematic (according to Louis Landry): 247, 253, 266.  Thanks to Louis for tracking this down at Joomla!Day South Africa 2011 with me.

Cheerio,
Ryan

AmyStephen:
Interesting. What is the problem?

https://github.com/joomla/joomla-platform/blob/master/libraries/joomla/database/table/content.php#L247

elinw:
Has this been cleaned up in the meantime?

LouisLandry:
The issue is that the `JTable` bone is connected to the `JUser` bone ... and the `JUser` bone is connected to the `JSession` bone... and in a CLI application where you may want to edit/store content but don't have a session this causes problems.  We need a way to inject a `JUser` object into `JTable` classes  to be used instead of calling `JFactory::getUser()`.  Perhaps something like `JTable::setUser(JUser $user)`.  It can still default to asking the factory, but there should be an option to inject the user dependency.

elinw:
oooo I missed he session part.

elinw:
Louis,

Should the user just be set to null  if there is no user specifically available (either JUser or set explicitly in the application  or a default user?  Should there be any cases besides the user table where you have to have a userId and get an exception if you don't?  I was looking in JContent and it uses the idea of a default user as a fall back but what if there is no default?  I guess I mean to what extent is/should the platform (as opposed to application) setting required fields?

fastslack:
Possible fix b2ecaa89f036376a4a43b4d892d401dd24b5fda4

eddieajau:
I'm sorry but we aren't accepting issues for this repository any more. Please see http://developer.joomla.org/cms/report-an-issue.html for how to report and issue for the CMS Platform.

Issue #227, opened 2011-08-13
possible fatal error in JUser::load()


Hoffi1:
Because JTable::getInstance() could return false (e.g. if an extension unfortunately registers a user.php file for it's own table object) the line "if (!$table->load($id))" could cause a fatal error.

To solve it (at this point) change the line to "if (!$table or !$table->load($id))"

But, of course, best solution would be if JTable::getInstance() could handle different classes stored in files with same name (in different folders).

Thanks,
Hoffi

(also submitted on http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26628)

eddieajau:
Can you try something for me.  If you are having this problem in the backend, can you find /administrator/includes/framework.php and after the `jimport('joomla.user.user')` put:

```
JLoader::register('JTableUser', JPATH_PLATFORM.'/joomla/database/table/user.php');
```

Also do this in the frontend `/includes/framework.php` file.  I think that may solve the problem.

Hoffi1:
Hi Andrew,

thanks for your suggestion. I have tried this and I think I must correct my report. The fatal error I had seen on an Joomla 1.6.4 with Discussions 1.3. There your suggestion doesn't help. Then I have tried this on a fresh Joomla 1.7.0 and I can't reproduce the error. I don't know how but 1.7 seems to be able to handle multiple files named user.php to load the correct JTable derived class.
I had renamed Discussions' files on our site to solve the problem completely before we upgraded to 1.7. Because I had not found any relevant differences in JTable, JPath, and JUser I was sure the problem is also present on 1.7 - which seems to be a wrong mind.
But one (design) question is left: Is it generally allowed / a good idea having files like user.php in a folder registered via JTable::addIncludePath() or should all these files have unique names to prevent conflicts? JPath returns only one path containing a file with matching name. And JTable::getInstance() returns false if this file doesn't contain the class requested regardless if another path contains the right file. So I see the potential for the fatal error I originally described although I can't see it any longer in "real world".

Many thanks and sorry for my wrong conclusion,
Hoffi

chdemko:
See https://github.com/joomla/joomla-platform/pull/397

elinw:
The issue Christophe referenced has been closed ... what does that mean for this issue?

LouisLandry:
This has been fixed along the way.  Closing it.  If we have a regression feel free to re-open.

PR #1356, opened 2012-07-12
JFactory::getUser(0) returns null if session doesn't have JUser object

mahagr:
We have had some bug reports like this:

```
Fatal Error: Call to a member function authorisedLevels() on a non-object
```

after running following code:

```
$user = JFactory::getUser($userid);
$accesslevels = (array) $user->authorisedLevels();
```

More details from the bug can be found from here:
http://www.kunena.org/forum/K-2-0-Support/123824-500-Internal-Server-Error

After some inspection I found out that the function indeed returns NULL when $id = 0 and $instance = NULL. Please see my comments in the code below:

``` php
    public static function getUser($id = null)          // assume $id = 0
    {
        $instance = self::getSession()->get('user');    // assume $instance = null

        if (is_null($id))                               // (is_null(0)) === false
        {
            if (!($instance instanceof JUser))
            {
            }
        }
        elseif ($instance->id != $id)                   // Notice: Trying to get property of non-object
                                                        // (NULL != 0) === false
        {
            $instance = JUser::getInstance($id);
        }

        return $instance;                               // $instance === null
    }
```

I know that this can happen if session was broken, but can it happen in core if the session is valid, for example when user logs out?

### DIFF
--- a/libraries/joomla/factory.php
+++ b/libraries/joomla/factory.php
@@ -220,7 +220,7 @@ abstract class JFactory
 				$instance = JUser::getInstance();
 			}
 		}
-		elseif ($instance->id != $id)
+		elseif (!($instance instanceof JUser) || $instance->id != $id)
 		{
 			$instance = JUser::getInstance($id);
 		}


mahagr:
Looks like the bug doesn't exist in Joomla! 1.5, but it does affect both J! 2.5 and Platform.

See also:
https://github.com/Kunena/Kunena-2.0/issues/976

pasamio:
What about changing line 216 to be empty instead of is_null? Would that solve the problem?

mahagr:
It solves the problem, yes.

But I do see a 'problem' with that change: doesn't user object contain some extra ACL information, even for guests? If there is anything that gets loaded from the DB or calculated during the request, your proposal would mean extra load to the server.

mahagr:
I can answer to my own question: I should have read the code before answering as I remembered it wrong.

I'm afraid that empty() doesn't work because of then $id = 0 returns current user instead of creating new object.

mahagr:
My original change combined both if and else into one slightly more complicated if(), but it also required a small change in JUser::getInstance(null) handling as the code currently tries to find username.

Here is the proposal for JFactory::getUser():

```
if (!($instance instanceof JUser) || (!is_null($id) && $instance->id != $id))
{
    $instance = JUser::getInstance($id);
}
```

It's slightly harder to understand and as I said, it requires a change to JUser as well. On the other hand current handling of NULL in JUser::getInstance() doesn't make any sense to me... Users without name are guests and do not need extra DB check.

elinw:
#227 Also has to do with this I think. 

Two related recent things. One is what if the "user" is a cli app?  
Also we had this discussion on the jbs list
https://groups.google.com/forum/?fromgroups#!searchin/joomlabugsquad/session/joomlabugsquad/8mX-Zfe0hMI/lN72kxzAgGkJ

Also don't forget #740

mahagr:
@elinw: 

Yes, there are at least 3 related issues, but to me they seem to be separate bugs and not symptoms from a one bug. I've worked around #740 by being sure that certain functions are not being called during login and logout. In there the issue was different -- user object was fine except for some cached values in it. But your message in bug squad list says that variation of the same bug can be seen in the backend.

This issue is slightly different as JUser doesn't exist in the session or has been changed into something else. I have no idea what causes this -- but it seems to happen sometimes for some users when user has been logged off, is logging off or if user suddenly becomes a guest (session corruption?). I've never been able to reproduce the issue in my tests or in any site I have.

elinw:
Also #276 and #1252.
I think these are not literally the same bug but the whole set of issues (and also possibly the problem of changing session values)   that would be worth thinking about coherently. 

Rouven fixed a bug the other day were two sessions were potentially starting #1353 and there is an old issue about "When using database session driver and msqli driver for database sessions aren't closed properly and write is never performed"

LouisLandry:
@mahagr This change makes sense to me.  Can you quickly let me know, does it actually solve your problem?  There is a bit of a confusing back and forth.  You seem to indicate that some other change needs to happen as well, is that the case?

I'd just like a little clarity before I merge it.  Thanks!

mahagr:
@LouisLandry Current code works and solves my problem (we currently have workaround for it).

Later I was simply trying to say that if JUser::getInstance() would have written in a slightly different way, I could have simplified if/else structure in JFactory::getUser() to just two lines of code. This is because of JUser::getInstance(null) currently makes a query to find a user who has empty username, which makes no sense to me.

LouisLandry:
@mahagr Thanks!

mahagr:
@elinw Found my fix, but it's not included in Joomla CMS. Do I create a new pull for this?

mbabker:
Do a pull against the CMS with it.  Still working on syncing up some of the Platform code to the CMS.

PR #1359, opened 2012-07-13
An implementation of the JavaScript library loader and behavior.jquery

jlleblanc:
Started with @realityking's JavaScript library loader concept, made fixes, and added support for pre/post JavaScript library initialization. This should take care of the major concerns raised by developers I've heard from, while still making it possible for the vast majority to call `JHTML::_('behavior.jquery');`.

Closes pull request #1298.


pasamio:
This appears to have style and unit test problems. Can you verify that there are not UT or CS problems with this pull request?

libraries/joomla/document/html/html.php
```diff
@@ -430,6 +438,55 @@ public function setBuffer($content, $options = array())
...
+	public function registerLibrary($name, $version, $url,
+		$mime = "text/javascript", $defer = false, $async = false,
+		$pre = null, $post = null)
+	{
+		$name = strtolower($name);
+
+		// Check if that library has already been added and if that's the case whether this one is newer
+		if (isset($this->libraries[$name]))
+		{
+			$key = array_keys($this->libraries[$name]);
```
tompap, line 464: I am not sure to understand this part of the code. 
In my understanding is correct the goal is to only load the latest version of a lib.
1) if someone is trying to load a lower version => ignore
2) if someone is trying to load a higher version => replace

$key = array_keys($this->libraries[$name]);
$currVersion = $key[0];

$currVersion will be equal to the word "version" not to X.X.X ?

And I was more expecting a code like

if (isset($this->libraries[$name]) {
    if (version_compare($version,$this->libraries[$name]['version'],'>')) {
        unset($this->libraries[$name]);
    } else {
        return $this;
    }
} 

But also should the code merged the pre and post script ?

beat:
One more vote for this pull request :-)
Following discussion in https://github.com/joomla/joomla-platform/pull/736 's conclusions. :-)

jlleblanc:
Actually, I'm going to close this PR for now. As @pasamio and @tompap pointed out, there are some issues with the code. My summer got away from me and I unfortunately won't have time to work on this for a while.

PR #1360, opened 2012-07-13
Route Query String Fix and Raw Route

pasamio:
This pull request fixes an issue where if a query string contained a colon in it (e.g. ?filters=content_id:20) that PHP's parse_url function would get confused. This meant that it would fail to properly process the path resulting in the first path segment being truncated erroneously and no match found for the route. This pull request preprocesses the route to remove the query string before parse_url function so that it will work reliably.

The second change is pushing into the input the raw route for the request. This is to permit direct access to the route segments in the controller should it wish to make further decisions based on the route segments. A use case is multiple hard coded mappings where the last segment is not variable however may have meaning to the individual controller (e.g. there are limited options available that can be reasonably elaborated).

### DIFF
--- a/libraries/joomla/application/web/router/base.php
+++ b/libraries/joomla/application/web/router/base.php
@@ -132,7 +132,7 @@ class JApplicationWebRouterBase extends JApplicationWebRouter
 		$controller = false;
 
 		// Trim the query string off.
-		$route = preg_replace('/([^?]*).*/', '\1', $route);
+		$route = preg_replace('/([^?]*).*/u', '\1', $route);
 
 		// Sanitize and explode the route.
 		$route = trim(parse_url($route, PHP_URL_PATH), ' /');
--- a/libraries/joomla/application/web/router/base.php
+++ b/libraries/joomla/application/web/router/base.php
@@ -131,6 +131,9 @@ class JApplicationWebRouterBase extends JApplicationWebRouter
 		// Initialize variables.
 		$controller = false;
 
+		// Trim the query string off.
+		$route = preg_replace('/([^?]*).*/', '\1', $route);
+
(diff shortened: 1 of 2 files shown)


elinw:
@pasamio 

Can you clarify if that is whole usage is utf-8 safe? Because there is quite a bit of conflicting information about the php parse_url function and whether it works with utf-8 segments. The php manual even with comments this spring is full with advice about how to solve lack of utf-8 support and we have jString:parse_url fo that reason, yet recent testing by a number of us seems to show that it is not a problem.  If it is a problem we should be using the uft8 friendly implementation since in the past in theory utf-8 strings have been treated as malformed urls without it. Alternatively, we were wondering if the JString Parse_url could be deprecated.

http://php.net/manual/en/function.parse-url.php
http://www.php.net/manual/en/function.parse-url.php#108787
http://3v4l.org/kPADL
https://github.com/joomla/joomla-platform/pull/1285

pasamio:
I have no idea if it's UTF-8 safe or not, all I know is there is a bug when you pass in what looks like a port in the query string which confuses PHP's parse_url (to be fair, we're parsing a segment). I'd guess like most of PHP it's far from safe for most definitions of the word.

libraries/joomla/application/web/router/base.php
```diff
@@ -131,6 +131,9 @@ protected function parseRoute($route)
 		// Initialize variables.
 		$controller = false;
 
+		// Trim the query string off.
+		$route = preg_replace('/([^?]*).*/', '\1', $route);
```
LouisLandry: Would you add the `u` modifier to the regex so that it becomes UTF-8 safe (just in case) and I'll merge this.

PR #1362, opened 2012-07-13
JUpdate load form XML - fixed empty response data

juliopontes:
Hi,

the joomla update are getting a empty response from Http because use $response->data instead of $response->body.


elinw:
This is the kind of thing where a test would be really helpful.

PR #1363, opened 2012-07-13
[CMS #28823] Re-add isError into the JError legacy class

ShMaunder:
Not sure the reason for the removal of isError() from the JError legacy class in J3.0 Alpha 1 - https://github.com/joomla/joomla-platform/pull/802 ?

I understand its old and "instanceof Exception" is the better way... but it just seems like this removal might break any extensions that use it (including my old code).

If we have a legacy class, we may as well keep it all in tact.

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28823

### DIFF
--- a/libraries/legacy/error/error.php
+++ b/libraries/legacy/error/error.php
@@ -62,23 +62,19 @@ abstract class JError
 	protected static $stack = array();
 
 	/**
-	 * Method to determine if a value is an exception object.  This check supports
-	 * both JException and PHP5 Exception objects
+	 * Method to determine if a value is an exception object.
 	 *
-	 * @param   mixed  &$object  Object to check
+	 * @param   mixed  $object  Object to check.
 	 *
 	 * @return  boolean  True if argument is an exception, false otherwise.
 	 *
-	 * @since   11.1
-	 *
 	 * @deprecated  12.1
+	 * @since   11.1
 	 */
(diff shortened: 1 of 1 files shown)


libraries/legacy/error/error.php
```diff
@@ -62,6 +62,27 @@
...
+	 *
+	 * @param   mixed  &$object  Object to check
+	 *
+	 * @return  boolean  True if argument is an exception, false otherwise.
+	 *
+	 * @since   11.1
+	 *
+	 * @deprecated  12.1
+	 */
+	public static function isError(& $object)
```
realityking: This doesn't need the by reference. Please remove it.

libraries/legacy/error/error.php
```diff
@@ -62,6 +62,27 @@
...
+	{
+		// Deprecation warning.
```
realityking: Useless comments should be removed as well. Thanks.

realityking:
I'd be fine with this, what do others think?

ShMaunder:
^ Done.

Its not a big deal if this doesn't get committed. I feel we should keep this legacy class in one piece for backwards compatibility.

LouisLandry:
I agree, should put it back to make things as smooth as can be.

PR #1364, opened 2012-07-14
Have the session fire an event after it is started and use this to populate the session with some default objects.

realityking:
This is an alternative approach to #1349 that doesn't cause JSession to depend on JUser and JRegistry.

### DIFF
--- a/libraries/joomla/application/web.php
+++ b/libraries/joomla/application/web.php
@@ -1038,25 +1038,41 @@ class JApplicationWeb extends JApplicationBase
 			'force_ssl' => $this->get('force_ssl')
 		);
 
+		$this->registerEvent('onAfterSessionStart', array($this, 'afterSessionStart'));
+
 		// Instantiate the session object.
 		$session = JSession::getInstance($handler, $options);
-		$session->initialise($this->input);
+		$session->initialise($this->input, $this->dispatcher);
 		if ($session->getState() == 'expired')
 		{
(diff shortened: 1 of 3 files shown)


elinw:
I don't know if you have been following the discussions in the other pull requests and on lists but there is really some serious issue with the dependency on JUser. Is what you are going here going to address that?
Most of the links are here #1356 it would be helpful if you would comment there.

P.S. it would be nice to have some tests in general in here http://developer.joomla.org/coverage/joomla_session.html so as not to go from broken one way to broken another way. :)

realityking:
This doesn't address the issue form #1356, it maintains the current way while allowing lazy sessions.

As for unit tests, yes that'd be nice but it's pretty hard because obviously we can't start a session when running the unit tests. (the CLI SAPI might not even have the session methods). We could mock all the PHP methods out and make sure we cll them with he right arguments but that's probably it.

LouisLandry:
I'm honestly not real excited about having an event triggered there.  There _are_ ways we can get some test coverage done for JSession, but not without doing some real thinking about the way it's structured and create wrapper functions around the PHP functions so that they can be replaced with a mock.

realityking:
@LouisLandry
Do you have an alternate suggestion for the event? I was planning to use the same event to move the who's online stuff from JApplication to a plug-in.

realityking:
Updated with some suggestions from Louis.

PR #1369, opened 2012-07-15
JModuleHelper - JPath and addIncludePath

juliopontes:
Hi,

I've changed some include files to use more platform code improved with use of JPath to find module files. Added a feature that will be used just by devs that wish to override module code.
I've implement new method addIncludePath to find modules in this new path.
these changes will not affect actual working code of cms.

Cheers,

Julio Pontes


eddieajau:
Julio, this pull request is really only of use to the CMS, so I feel it's better added to the CMS codebase.

PR #1371, opened 2012-07-15
Let mysqli be the base class for mysql instead of the other way around.

realityking:
This this is a necessary first step to eventually get rid of mysql.


elinw:
+1

PR #1372, opened 2012-07-15
Fixes for extension update adapter

mbabker:
The update adapter is failing in the CMS master repo currently.  These changes get it working again.

### DIFF
--- a/libraries/joomla/updater/adapters/extension.php
+++ b/libraries/joomla/updater/adapters/extension.php
@@ -186,7 +186,7 @@ class JUpdaterExtension extends JUpdateAdapter
 		xml_set_element_handler($this->xmlParser, '_startElement', '_endElement');
 		xml_set_character_data_handler($this->xmlParser, '_characterData');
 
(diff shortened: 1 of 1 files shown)


libraries/joomla/updater/adapters/extension.php
```diff
@@ -186,7 +186,7 @@ public function findUpdate($options)
 		xml_set_element_handler($this->xmlParser, '_startElement', '_endElement');
 		xml_set_character_data_handler($this->xmlParser, '_characterData');
 
-		if (!xml_parse($this->xmlParser, $response->data))
+		if (!xml_parse($this->xmlParser, $response->body))
```
realityking, line 189: Where are the coming from? #1362 had a change just like that.

mbabker, line 189: Any changes to JHttpResponse that came about when JHttpFactory was added?

realityking, line 189: It has been called body from the start. Maybe they were missed by the switch to JHttp.

PR #1373, opened 2012-07-16
Required and validate elements errors messages are not translatable (Regression).

infograf768:
Required and validate elements error messages are not translatable (Regression). + errors in front-end wrong display.
 See http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28827

### DIFF
--- a/libraries/joomla/form/form.php
+++ b/libraries/joomla/form/form.php
@@ -1839,7 +1839,7 @@ class JForm
 				// Does the field have a defined error message?
 				if ($element['message'])
 				{
-					$message = $element['message'];
+					$message = JText::_($element['message']);
 				}
 				else
 				{
@@ -1884,10 +1884,11 @@ class JForm
 		if ($valid === false)
 		{
 			// Does the field have a defined error message?
-			$message = (string) $element['message'];
+			$message = (string) ($element['message']);
 
 			if ($message)
 			{
+				$message = JText::_($element['message']);
 				return new UnexpectedValueException($message);
 			}
 			else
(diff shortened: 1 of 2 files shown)


elinw:
Further, I don't think this should be throwing an exception, are we really going to log every time a user makes a typo in their password?

To make this work in other places to at least use the string when the exception is thrown you need to also add
                $message = JText::_($element['label']);
after 1890 
also I guess after 1842 similarly

infograf768:
$message = JText::_($element['label']);
is present where it should

What has also to be solved is $message = $element['message'];
1. It should be translatable
2. We should separate message_required and message_validate as a field may contain both elements "required" and "validate"
Example: when the validate = equals for confirm password, the error should not be 
Invalid field: Confirm password
but
JLIB_USER_ERROR_PASSWORD_NOT_MATCH="Passwords do not match. Please re-enter password."
While keeping for the required message
JLIB_FORM_VALIDATE_FIELD_REQUIRED="Field required: %s"
i.e
Field required: Confirm Password

Example in site.xml
        <field name="admin_password2" type="password" id="admin_password2" class="inputbox"
            label="INSTL_ADMIN_PASSWORD2_LABEL"
            required="true"
            validate="equals"
            field="admin_password"
            message_validate="JLIB_USER_ERROR_PASSWORD_NOT_MATCH"
            message_required="JLIB_USER_ERROR_CONFIRM_PASSWORD_REQUIRED"
        />

I thought this could be done in another patch.

PR #1374, opened 2012-07-16
[IMP] JImage->toFile() returns boolean

phproberto:
Add a return value to the method toFile in the JImage library.


LouisLandry:
You need to change the doc block if you want to do this.

phproberto:
You are true thanks!

realityking:
Thanks!

PR #1376, opened 2012-07-17
Fix a doc block

elinw:


elinw:
There are other odd things but that docblock is just a copy of the previous one (addNode)

libraries/joomla/form/form.php
```diff
@@ -2026,7 +2026,7 @@ protected static function addNode(SimpleXMLElement $source, SimpleXMLElement $ne
 	}
 
 	/**
-	 * Adds a new child SimpleXMLElement node to the source.
+	 * Update the attributes of the a child node
```
eddieajau: Would this be better as "Merges a new element into a source <field> element."?  Just going by the description in the next method (mergeNodes).

elinw: Yeah it's a bit confusing but I think it really is identical to the first code block in the mergeNodes and what I put  is based on the comment for the first code block. THat's why I said there were other oddities, because the code in mergeNodes is identical to that block.  
Also it is specifically an update not an add and the method only deals with attributes of an element not  the element. The addNode method adds the element itself.

I think that whole block of methods could use some love but at minimum the docblock shouldn't be a copy and paste from another method.

PR #1379, opened 2012-07-17
Remove more form fields that belong to the CMS.

realityking:
Removed are the following classes:
JFormFieldContentLanguage
JFormFieldEditor
JFormFieldMedia
JFormFieldMenuitem


eddieajau:
Is it worth updating the description of this pull request to list the fields that are moved (for the sake of the change log)?

realityking:
Good point. Done.

LouisLandry:
Great stuff as always.

PR #1380, opened 2012-07-17
Remove our custom session ID generating code if favor of session_regenerate_id().

realityking:
Note there's one non-optional use of session_id() left on line 624. I'm not really sure what it does so I didn't touch it for now.

I confirmed the CMS works fine with this change (tested admin login/logout). The session id changes on logout. (It doesn't on login because apparently we don't fork the session on login)

### DIFF
--- a/libraries/joomla/session/session.php
+++ b/libraries/joomla/session/session.php
@@ -606,7 +606,7 @@ class JSession implements IteratorAggregate
 		// Start session if not started
 		if ($this->_state === 'restart')
 		{
-			session_id($this->_createId());
+			session_regenerate_id(true);
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


eddieajau:
Couple of thoughts:
1. Should we use session_regenerate_id(true)?
2. Should we provide some opt-in backward compatibility to use the "old" way (but deprecate it) just in case we find some edge case where sites break?

realityking:
Indeed, it should be session_regenerate_id(true). I had that in an earlier version but apparently it got lost. Fixed.

As for B/C, could we wait if something breaks?

eddieajau:
Yeah, that's fine. Just asking the question ;)

LouisLandry:
Love the direction all of this is going. :-)

PR #1382, opened 2012-07-18
Solve parse_url issues

stefanneculai:
Because the parse_url method from JString has issues and the parse_url from PHP has no longer the UTF8 problem, I have replaced all JString::parse_url with parse_url from PHP and marked the parse_url from JString as deprecated.


realityking:
I'm not sure about this, see my response in #1285.

I suggest we keep the discussion there and than decide which approach we wanna take.

elinw:
So it seems like PHP's parse_url is not really UTF-8 safe on all platforms. This being the case, I think what we should do is fix the problems in our parse_url but avoid making any b/c breaks.

stefanneculai:
Fixed by pull request #1458

PR #1383, opened 2012-07-18
Remove use of intval() and floatval() in favor of casting.

realityking:
We mainly use the cast syntax but sometime we use intval() and floatval(), the pull makes things more consistent.


eddieajau:
I believe it's a tad faster as well. Merged.

realityking:
Yeah I found that info too but the absolute time is still insignificant.

PR #1385, opened 2012-07-18
Make JImageFilter child classes autoloadable

### LINKED ISSUE #1384: JImage::filter - The Grayscale image filter is not available.
I'm building joomla cms gallery but there is problem with JImage filters. Filter class isn't loaded. 

Here is sample code:

``` php
$image = new JImage($src);
jimport('joomla.image.filters.grayscale');
$image->filter('grayscale');
```

When I run it, this error is thrown" "The Grayscale image filter is not available."

It works when I import class file manually using require_once.

mbabker:
This should fix #1384


LouisLandry:
Thanks Michael.

PR #1389, opened 2012-07-19
JFactory::createDocument always creates HTML documents.

Bakual:
Fixing the "format" URL parameter. Bug was introduced when the depreceated "no_html" was removed.
Now it is no longer possible to create other document types using "&format=raw".
This PR brings back this function but doesn't reintroduce the "no_html" parameter.

### DIFF
--- a/libraries/joomla/factory.php
+++ b/libraries/joomla/factory.php
@@ -695,8 +695,8 @@ abstract class JFactory
 	{
 		$lang = self::getLanguage();
 
-		$jinput = JFactory::getApplication()->input;
-		$type = $jinput->get('format', 'html' , 'word');
+		$input = self::getApplication()->input;
+		$type = $input->get('format', 'html' , 'word');
 
 		$attributes = array('charset' => 'utf-8', 'lineend' => 'unix', 'tab' => '  ', 'language' => $lang->getTag(),
 			'direction' => $lang->isRTL() ? 'rtl' : 'ltr');
--- a/libraries/joomla/factory.php
+++ b/libraries/joomla/factory.php
@@ -695,10 +695,13 @@ abstract class JFactory
 	{
 		$lang = self::getLanguage();
 
+		$jinput = JFactory::getApplication()->input;
+		$type = $jinput->get('format', 'html' , 'word');
+
 		$attributes = array('charset' => 'utf-8', 'lineend' => 'unix', 'tab' => '  ', 'language' => $lang->getTag(),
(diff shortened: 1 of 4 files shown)


libraries/joomla/factory.php
```diff
@@ -695,10 +695,13 @@ protected static function createDocument()
 	{
 		$lang = self::getLanguage();
 
+		$jinput = JFactory::getApplication()->input;
```
realityking: You'll have to replace `JFactory` with `self`, or you'll get a code style error. Also the convention right now is to use $input, that isn't really important thou.

realityking:
Thanks a lot! Got a small comment and have to wait on unit tests to run but all looks good :)

Bakual:
I'm all for following conventions as it makes it easier to read the code.

I was following the example given in http://docs.joomla.org/Retrieving_request_data_using_JInput. Maybe someone should change it there.

mbabker:
If the method is in the same class, you use self instead of the class name.  Otherwise, that reference is correct.

Bakual:
Ah, I meant to change the "$input" convention in the example, not the "JFactory" vs "self" one. That was just an oversight I should have known.

realityking:
+1

Just waiting for the pull tester to run.

libraries/joomla/factory.php
```diff
@@ -695,10 +695,13 @@ protected static function createDocument()
 	{
 		$lang = self::getLanguage();
 
+		$input = self::getApplication()->input;
+		$type = $input->get('format', 'html' , 'word');
```
realityking: You got an extra space before the second comma.

realityking:
You also got a bunch of unit test errors. This commit should fix it: https://github.com/realityking/joomla-platform/commit/55a352512733c7ec216df6b1fd58a97257de1f7a

Could you cherry pick it into your branch? Thanks.

I'd really like to merge this tomorrow so it can make its way to the CMS.

Bakual:
I'm not sure what you mean with the commit you linked. It seems to be all related to the testing files, and I have to admit I have no clue how those work.

However I have fixed the codestyle error.

realityking:
Your patch is breaking the unit tests. The commit I linked is the fix, you just have to cherry-pick that commit into your branch and we should be good to go.

Bakual:
Included your fix into my fork. I didn't find out how I could include your commit otherwise. Still learning Git :-)

realityking:
Thanks. Merged.

PR #1390, opened 2012-07-20
Discourage use of the silence operator.

realityking:


pasamio:
Not sure I entirely agree, PHP's obsession with returning PHP warnings/errors with the actual error message being in the warning (only partially following the C convention here). The only way to work around this and get useful information is to use $php_errormsg however the PHP example includes the use of the silence operate in it's canonical example:
http://us2.php.net/manual/en/reserved.variables.phperrormsg.php

realityking:
I agree we'll never be able to completely get rid of them (at least not with PHP's current error capabilities). But there are a number of places where we wouldn't need them if we'd use a isset() or other checks like that. That's why this would only add a warning, not an error.

libraries/joomla/string/string.php
```diff
@@ -15,9 +15,9 @@
 if (extension_loaded('mbstring'))
 {
 	// Make sure to suppress the output in case ini_set is disabled
-	@ini_set('mbstring.internal_encoding', 'UTF-8');
```
LouisLandry, line 18: These I'm not sure I like.  The others I'm fine with.

LouisLandry:
This is no longer mergeable.  Additionally I have reservations on the silence operator being used on the mbstring settings.  If errors, notices, or warnings are suppressed for that we don't have any sort of backup to know what went wrong in using the mbstring functions.

I'm going to close this one for now until you can get it rebased @realityking.  Do that and we can continue the discussion.  Remove those supression operators on the `ini_set()` calls for mbstring and I'll just merge it.  Reopen when you are ready, thanks!

PR #1391, opened 2012-07-20
[IMP] function to create thumbnails from the current image

phproberto:
Finally after the previous JImage required modifications I submit the optimized createThumbs function to quickly create thumbnails from the current image.

Example. Create 2 thumbnails with cropping: 
jimport('joomla.image.image');
$imagen = new JImage(JPATH_SITE . DIRECTORY_SEPARATOR . 'images' . DIRECTORY_SEPARATOR . 'powered_by.png');
$thumbSizes = array('50x50', '250x100');
$imagen->createThumbs($thumbSizes, 4);


realityking:
Not sure we really want this in the platform, it's more application logic to me. Also the would make JImage depend on JFolder and JFile which I'm pretty sure we don't want.

phproberto:
Can make it JFile and JFolder independent but no problem.

realityking:
Right, that's doable. The question is if we really want this in the platform, I'd be interested what other think about that.

phproberto:
And what's the problem in using JFile & JFolder? If both are already disposable on the platform. Want fully independant libraries to avoid dependencies ? Is that a platform rule?

realityking:
We wanna start limiting the dependencies between packages to those that are really required. This should make it easier to just use certain packages from the platform. We also made the decision not to expand the FTP support to new functionality so there's less need for JFolder/JFile.

phproberto:
I see no sense in a platform that can't make things easier to developers because it's own rules.

Obviously I'm nobody to questionate platform rules so I'll take this as a FAIL! :)

Thanks for your time reviewing this.

realityking:
I didn't say it was a rule. It's my opinion. Note I didn't close this since I wanted to let other people weight in ;)

elinw:
This would be such a great thing to have.

phproberto:
So do you think it's better modify it to be non JFile & JFolder dependant or just wait?

piotr-cz:
It makes sense to me to have such functionality in the platform. There is more than 1 native extension that could use thumbnails (com_articles, com_media, com_contact, com_banners, com_users, com_templates) and really any extension that works with images.

Question is if the default place for thumbnails is the /thumb folder where these could be overwritten or /cache/[extension].

phproberto:
The default place can be cache/thumbs for example and developers can override it. 
It can't be /cache/[extension] because the function doesn't know from where is called.

I used thumbs folder because if the user has uploaded an image probably that folder has the required permissions and is more comfortable to have that images there. For example: images/photos and then images/photos/thumbs

phproberto:
Lol closed this accidentally :D

piotr-cz:
@phproberto 
You are right, the developer should decide on naming strategy to manage thumbnail files (maybe `md5($this->getPath());`).

But how to apply sharpen filter before thumbnail is being saved? Perhaps the method should return a new image object that can be further manipulated. Then it would not make sense to add new JImage method but handle it other way.

Let's see what other think

e-motiv:
A lot, but i mean "A LOT" of third party extensions use their own thumb creation method.  I myself have already seen at least five popular ones.  Others and I have also let this know in a years' old post in some development discussion group of joomla (but it got overcrowded with a discussion of whether to put JImage outside or inside another class, which i suppose is solved now).
So, if some people have opionions against putting this in the platform, why not put it in the joomla non-platform library?
I must admit, I don't know exactly in which folders platform and non-platform library code are put nor where it should be discussed, but I think this thumb function should finally get it through somehow and beg someone to push this in a joomla 2.5 update, platform or no platform. (In the latter case, don't forget to post a link to the relevant discussion space.)

Also a note on caching: shouldn't be there an "if exact thumb image already exist in thumbsfolder, return that one" be included in this class (and not outside of it) in stead of recreating it every time?

drmmr763:
This is a great feature that I myself considered writing for the platform. Would love to see this in core. JImage is a great base for so many things, but to have an image library that doesn't even have a nice thumbnail feature is kind of pointless. Let's get some thumbnail creation in the platform to reduce the amount of code _every_ _single_ image slider / thumb view extension has to write to do this very thing.

betweenbrain:
+1

phproberto:
I'm going to clean it (JFile, JFolder & DIRECTORY_SEPARATOR) and try again to get this merged.

nternetinspired:
One of the key reasons for moving to Bootstrap was re-usability, to provide extension developers with the tools they need to do the job in order to cut out senseless duplication. The sheer number of extensions that are using their own (often horribly flawed) method for creating thumbs is proof that there is a need for such functionality. 

Let's have image thumb creation done _once_, done _properly_ and usable _by all_.

Well done Roberto!

Skullbock:
I totally agree with @nternetinspired here. Having a JImage class and not having a thumnail creation method seems pointless to me. Moreover, nearly every component needs this, so why not include it into the platform?

Chris-Jones-Gill:
+1

phproberto:
I have removed the JFile & JFolder dependencies and replaced DIRECTORY_SEPARATOR with slashes.

Fast example:

```
    $image      = new JImage(JPATH_SITE . '/images/banner1.jpg');
    $thumbSizes = array("150x100", "200x100");
    $thumbs     = $image->createThumbs($thumbSizes, 4);
```

Will return an array of 2 JImage objects with the path set for further thumb manipulation:

```
Array
(
    [0] => JImage Object
    (
        [handle:protected] => Resource id #116
        [path:protected] => /var/www/joomla256/images/thumbs/banner1_150x100.jpg
    )

    [1] => JImage Object
    (
        [handle:protected] => Resource id #118
        [path:protected] => /var/www/joomla256/images/thumbs/banner1_200x100.jpg
    )

)
```

realityking:
Thanks for that. You're on the right track. I'll look trough it a bit more later.

piotr-cz:
I've been very happy to see this method as you guys.
I use to sharpen thumbnail images in the image editor as mostly after resampling these lose some sharpness. To use this method I'd have to resave each thumbnail image and so double compress.

I'd like to have noSave parameter, so I can use the returned array to further manipulate each resource and save using the path variable.

libraries/joomla/image/image.php
```diff
@@ -141,6 +141,86 @@ public static function getImageFileProperties($path)
...
+			{
+				$thumbSizes = array($thumbSizes);
+			}
+			// No thumbFolder set -> we will create a thumbs folder in the image folder
+			if (is_null($thumbsFolder))
+			{
+				$thumbsFolder = dirname($this->getPath()) . '/thumbs';
+			}
+
+			if (is_dir($thumbsFolder) || @mkdir($thumbsFolder))
```
realityking: I'd remove the mkdir part. I'd let the calling method deal with creating the folder.

libraries/joomla/image/image.php
```diff
@@ -141,6 +141,86 @@ public static function getImageFileProperties($path)
...
+		}
+		if (!empty($thumbSizes))
+		{
+			// Accept a single thumbsize string as parameter
+			if (!is_array($thumbSizes))
+			{
+				$thumbSizes = array($thumbSizes);
+			}
+			// No thumbFolder set -> we will create a thumbs folder in the image folder
+			if (is_null($thumbsFolder))
```
realityking: Shouldn't we just require a folder to be specified?

realityking:
One bigger question left as well: I see you're both storing the thumbs on the disk and returning the JImage instances for them. I think that's too much for one method. If both functionalities are desired I suggest splitting it into two methods, one to generate them and one that generates and stores them depending on the first. This would adress @piotr-cz's use case.

libraries/joomla/image/image.php
AmyStephen: Probably not a big deal, but I disagree, @realityking I think it's almost a defacto standard, at this point, that the image resizing process also create the subfolder, if that location does not exist.  

If the calling routine wants to create it first, nothing is stopping it, but causing an error when it doesn't exist seems like an unfriendly way to deal when a folder can simply be created and the application continue on it's merry way.

realityking: Ok. It's certainly not something I very insensitive about. My main concern (besides doing too much in a method) is that this will fall flat when there's more than one directory to be created (i.e. the path passed is /var/www/images/collection/thumbs but the folder /var/www/images/collection doesn't exist)

AmyStephen: Agree on raising an exception if the parent folder doesn't already exist.

phproberto:
@piotr-cz I think that's fine and make this cover more developer needs. A noSave parameter false by default. The method will remain simple to use as most users will not use it.

@realityking (re: split method). Do you mean a method that stores an array of JImage objects to disk? It will be just a foreach with toFile. In @piotr-cz case that method will be useful but we are saving him about 5 lines of code? It worths create a static saveImages method?

libraries/joomla/image/image.php
phproberto: Why complicate it? If the user wants to use a specific folder still can do it. I think this should remain as simple as possible. What kind of issues are you thinking on?

realityking:
Not quote I was thinking more along the lines of a generatThumbs() methods that does all the generating but doesn't bother with the filesystem and a createThumbs() method that calls generatThumbs() but stores the result to the disk.

libraries/joomla/image/image.php
phproberto: I allways keep in mind to ease developers work. Why don't save us the folder creation when we need it? If we cover the noSave requirement the method will remain really simple to use but flexible to any needs.

phproberto:
@realityking I understand it now. Thank you. That's great and seems better than add a parameter to the function. That way user can identify each function easilly.

I'll commit the changes to create both methods and will also try to solve the mkdir issue.

Thanks all for the feedback / ideas!

piotr-cz:
@phproberto Thanks for considering my comment. @realityking got good idea about splitting method into two, this seems better than noSave param.

I suggest that generateThumbs would return an array of thumbs file paths, while createThumbs an array of resources

phproberto:
@piotr-cz as the generateThumbs method will not handle paths it must return an array of JImage objects for further processing without paths. The createThumbs method that stores the result into disk will return the array of JImages with path set.

phproberto:
Created both generate and create methods. Added an extra check/exception when parent folder doesn't exist or current folder cannot be created.

Also added @since 12.2. Not sure if I should do it.

libraries/joomla/image/image.php
```diff
@@ -141,6 +141,134 @@ public static function getImageFileProperties($path)
 	}
 
 	/**
+	 * Method to generate thumbnails from the current image. It allows creation by resizing
+	 * or croppping the original image.
+	 *
+	 * @param   mixed    $thumbSizes      string or array of strings. Example: $thumbSizes = array('150x75','250x150');
+	 * @param   integer  $creationMethod  1-3 resize $scaleMethod | 4 create croppping
```
realityking, line 154: Please add a constant for cropping (like there are for the scale methods) and change the default values to the corresponding constant.

realityking:
Somewhere along the way you changed the permissions of the file:
libraries/joomla/image/image.php 100644 → 100755

Please undo that.

phproberto:
Fixed permissions and used constants as default.

libraries/joomla/image/image.php
```diff
@@ -141,6 +147,134 @@ public static function getImageFileProperties($path)
...
+	 * @param   string   $thumbsFolder    destination thumbs folder. null generates a thumbs folder in the image folder
+	 *
+	 * @return array
+	 *
+	 * @throws  LogicException
+	 * @throws  InvalidArgumentException
+	 *
+	 * @since 12.2
+	 */
+	public function createThumbs($thumbSizes, $creationMethod = 2, $thumbsFolder = null)
```
realityking: Same here.

phproberto: Oops! Done.

PR #1393, opened 2012-07-21
Speed improvement to JRoute::_()

realityking:
JRoute::_() may be executed several hundred times on a complex site so even a small optimization can be measureable.


mahagr:
Thanks! I've been playing with some other optimizations as well, but they still need some live testing and discussion of what's the best way to implement those.

PR #1396, opened 2012-07-21
Simplify JInstaller a tiny bit.

realityking:


realityking:
@pasamio @mbabker Could one of you review this? Thanks.

LouisLandry:
@realityking I'm really feeling like we need to move the installer and updater libraries to legacy en route to the CMS tree.  Given that I'm not inclined to really merge this change, but would rather suggest you run with it in the CMS tree.  I'm happy to be overridden by @pasamio if he wants to run with it here, but that's my feeling.

Incidentally either way it is no longer mergeable and needs to be rebased if we were to merge it.  I'm closing it for now until it is mergeable again.  If you guys want to get it into the platform I'll leave it up to you.

pasamio:
I'd much rather this shift to CMS sooner rather than later and we can build a new mechanism built around JController tasks that we join together to build coherent pieces.

PR #1398, opened 2012-07-21
Optimize images to reduce their filesize.

realityking:


LouisLandry:
_sigh_ .... can we please get these things into the CMS tree and out of the platform tree. :-)

realityking:
At the very least we should move the stuff not actually used by the platform :)

PR #1399, opened 2012-07-22
Avoid using deprecated APIs and remove more deprecated code in JModuleHelper.

realityking:


realityking:
The attempt to keep supporting test() didn't work correctly and would rarely be required so we might as well drop it.

LouisLandry:
Don't understand how the title of this PR lines up with the changes.

realityking:
Yeah the class name in the title was nonsense.

PR #1401, opened 2012-07-22
Remove deprecated code in JModuleHelper.

realityking:
Fix the deprecation warning in JHtmlSelect.


realityking:
Close and code added to #1399.

PR #1404, opened 2012-07-24
Correct confusion between required and validate messages

infograf768:
When a field contains both a required and a validate element, we get a confusion for the error messages.
This patch takes element['message'] out of the required conditional as, in this case, the label or the name can be used for the error.
(Patch also ready for the CMS with fields and strings changes)


realityking:
Thanks. Fixed.

infograf768:
Can you commit to your php53 branch, as well as the CMS-specific patch I sent you?

PR #1405, opened 2012-07-24
Add basic PHARchive support.

LouisLandry:
Add basic package definition files for using https://github.com/LouisLandry/packager to package either a platform PHAR or a platform + test libraries PHAR.  Additionally there are a few small fixes here and there to allow the platform to work as a PHAR.


realityking:
I don't know much about PHAR but do we really need to duplicate the import file in a 3rd and 4th location? This is just young to get out of sync.

LouisLandry:
The short answer is yes, we really do.

PR #1406, opened 2012-07-24
Fix problem with JLoader prefixes

rvsjoen:
When registering a custom prefix that has more than a single uppercase character, JLoader searches the wrong path. For example, registering

``` php
JLoader::registerPrefix('Jcms', JPATH_PLATFORM . '/cms');
```

will cause it to look for, say, JcmsViewHtml in libraries/cms/cms/view/html.php because the "cms" part of the prefix is mistakenly interpreted as a path segment. 

This patch will make sure the whole first segment of a CamelCasedWord is used as the loader prefix.

### DIFF
--- a/libraries/loader.php
+++ b/libraries/loader.php
@@ -296,7 +296,8 @@ abstract class JLoader
 	{
 		foreach (self::$prefixes as $prefix => $lookup)
 		{
-			if (strpos($class, $prefix) === 0)
+			$chr = strlen($prefix) < strlen($class) ? $class[strlen($prefix)] : 0;
+			if (strpos($class, $prefix) === 0 && ($chr === strtoupper($chr)))
 			{
 				return self::_load(substr($class, strlen($prefix)), $lookup);
 			}


realityking:
Hmm, is there a fix without perfuming a regex?

In some testing I did the time spend in _autoload started to get significant. (It's called so often) Not sure it's smart to make it more complex.

rvsjoen:
This updated version does not use regular expressions and seems to me to work correctly. The basic idea is to check whether or not the next character after the prefix being searched for is an uppercase character (To denote a word boundary in a CamelCasedWord).

rvsjoen:
There seems to be an issue with the unit tests here, looking into that

http://developer.joomla.org/pulls/pulls/1406.html

rvsjoen:
The unit test are running ok now, also got rid of a notice if the prefix we're testing against happens to be the same length or longer than the class name.

pasamio:
In your description you say "more than a single uppercase character" but all of the examples are "Jcms" with a single upper case character. Am I missing something?

elinw:
I think he is saying it expects exactly one character and that the character will be upper case.  More than a single upper case character.

rvsjoen:
After some digging I discovered the root of this behavior, and it is not that the loader doesn't handle prefixes that are longer than a single character. 

One example is that if the prefix 'J' is registered (like it is) then if I register a "Jcms" prefix it won't care, because it finds the J prefix first, and then assumes that the "cms" is part of the path. so checking whether or not a class name starts with the prefix currently being looked for is not enough, we also have to check if the next character is an uppercase character.

The patch takes are of this by making sure the "whole" prefix is used  (meaning the first word in a CamelCasedString) even if a smaller part of the string matches another prefix.

rvsjoen:
So that makes the original description a bit wrong in the way that it describes only the way the issue manifested itself when I tried to register a Jcms prefix in addition to the core 'J' prefix.

PR #1408, opened 2012-07-24
[IMP] Plugin get overridable layout

phproberto:
Helper function to ease the plugin layout overrides based on the JModuleHelper getLayoutPath function.

For example the plugin Content - Pagenavigation will search his overrides in:

template/YOUR_TEMPLATE/html/plg_content_pagenavigation

And the default layout would be:

plugins/content/pagenavigation/tmpl


mbabker:
I'd be supportive of something like this.  I currently use some internal code in a plugin that allows users to override the output, and I know of a few other plugins doing similar workarounds.

piotr-cz:
I guess plugins were meant to not output any HTML, so there was not override functionality built in (this has been discussed few times already), but there are exceptional cases.
Most JoomlaWorks plugins have this functionality built in since Joomla 1.5 and it's great.

I'm for this to be in core.

@phproberto: Consider replacing `DIRECTORY_SEPARATOR` with `/`. I think this is the convention in the platform as PHP deals perfectly fine with both types of slashes in paths and this makes code more readable.

phproberto:
You are true @piotr-cz

I forgot to remove the DIRECTORY_SEPARATOR's. I'll modify it ASAP.

Thanks for the advice and support !

phproberto:
I've fixed the DIRECTORY_SEPARATOR with slashes. 

Thank you all for your feedback.

phproberto:
Any news about this? Will be great to see it approved and submitted to joomla-cms to be in Joomla! 3.0

AmyStephen:
This makes good sense, to me. It's just one more step towards getting rendering more flexible. Probably dozens of ways to do this but this looks suitable to me, inline with how other output rendering is handled. Even if a new approach is used with the new Integrated Content system, having this for 3.N series would be good.

Thanks Roberto!

piotr-cz:
Just hope it's not too late to include it in the 3.0 trunk. Probably most forces are on Bootstrap integration.

libraries/joomla/plugin/helper.php
```diff
@@ -27,6 +27,49 @@
 	protected static $plugins = null;
 
 	/**
+	 * Get the path to a layout from a Plugin
+	 *
+	 * @param   string  $type    Plugin type
+	 * @param   string  $name    Plugin name
+	 * @param   string  $layout  Layout name
+	 *
+	 * @return  string  Layout path
```
pasamio, line 36: Can you add a `@since 12.2`

LouisLandry:
Why wouldn't you just use JViewHtml to support this functionality?  You have layouts and overrides just built into that.  It's very lightweight as well.  It then keeps the clear separation of presentation responsibilities independent of the event handling system.

libraries/joomla/plugin/helper.php
phproberto, line 36: Done ;)

phproberto:
@LouisLandry This is based on the same method already existent for module layouts. I've been watching the JViewHtml class and you are right, it's better prepared for layouts but that will probably be harder than create a simple method.

AmyStephen:
Core code needs to be inline with architectural directions.

phproberto:
Any news about this? :(

PR #1409, opened 2012-07-25
[FIX] replace slashes with DIRECTORY_SEPARATOR

phproberto:


jonnsl:
why? slashes works fine on windows and linux.

okonomiyaki3000:
It's true, / works. Also, I must have missed something but has the use of DS been deprecated?

phproberto:
/ works most of the times and in this case all the times. Only fails when you try to manipulate a path obtained from the system. For example with a str_replace or an explode. We should use allways DIRECTORY_SEPARATOR as convention.

DS is considered deprecated from Joomla 1.6 but backward compatibility will be broken in 3.0.

http://forum.joomla.org/viewtopic.php?p=2720756#p2742713

realityking:
No we shouldn't. The whole point of deprecating and removing DS is to force people to just use slashes. We saw a measurable performance increase in the CMS when we dropped using DS because we require far less string concatenations.

If we need to compare paths - the only real issue with slashes - than there's JPath::clean().

If there's an actual issue on Windows please report it or submit a patch, but we won't use DIRECTORY_SEPARATOR generally.

realityking:
@okonomiyaki3000 
Missed your post. Yes DS was deprecated and it is removed in Platform 12.1 and Joomla 3.0. It's fairly trivial to add to your extension if you need it. I'd recommend investing the time thou.

phproberto:
As I like to play I did some tests.

50000 string concatenations => time elapsed: 0.029999 seconds

That's really innapreciable.

But the time elapsed by developers in write DIRECTORY_SEPARATOR instead of just / and the improvement in code readability make a winner.

I'll start using / ;)

Thanks for the info to all and sorry!

PR #1411, opened 2012-07-25
Autoloader cleanup

ianmacl:
Moving JHtml files so that they can be autoloaded.  This allows for more readable code because you can now use, f.e., JHtmlSelect::radiolist($params) instead of using _.


LouisLandry:
Have we also adjusted JHtml to not "include" the previous path as a location to look for helper classes?

bembelimen:
I think, this is a bad idea.

Let's say a component uses behavior.caption

## old handling:

Component calls JHtml::_('behavior.caption');
I create a plugin which can make any time: JHtml::register('behavior.caption', function() { return null; }); to deactivate it.

## new handling:

Component calls JHtmlBehavior::caption();
I have to create an Plugin with an JHtmlBehavior class which is loaded before the original Behavior class is included. In this class I have to implement all methods (framework, modal, etc.) which I plan to use and also add a dummy caption method to deactivate the caption itself. And with every update I have to check if the class had any security issues and have to update it.

So I see no sence in autoload the JHtml stuff. The _ method works very well, if you work the correct way and JHtml::_ have some advantages like the register mentioned above.

realityking:
I agree we should stick to JHtml::_() for now but it still makes sense to make the classes autoloadable since it simplifies a few things. The legacy JHtml\* classes are already autoloadable and so are those the CMS keeps around, so far we haven't seen any problems with it.

LouisLandry:
@bembelimen this doesn't change that behavior at all.  All it does is make the initial call to JHtml::addIncludePath() unnecessary.  @ianmacl you need to rebase so that it is mergeable.  Do that and I'll merge it this weekend.

ianmacl:
This should be ready to go now.

PR #1412, opened 2012-07-25
Moving rules and fields for autoloading

ianmacl:
Moving JForm fields and rules to new folders so they can be autoloaded.


LouisLandry:
Have we also removed the "include" paths here so that it isn't trying to add an include a folder full of things that doesn't exist.

elinw:
@ianmacl  This can be closed now I think.

LouisLandry:
Doesn't look like it can be closed to me.  This is still an issue it seems -- unless i'm missing something.

@ianmacl can you rebase this and double check on the include paths bit please?

elinw:
Oh for some reason I thought Rouven had done field(s) and rule(s) but it was just rules to rule.

LouisLandry:
@ianmacl when you get this fixed up re-open please.  Marked for 12.3 release.

PR #1413, opened 2012-07-26
Add a more complex truncate method

elinw:
This method allows specification of a given length of plain text embedded in HTML of unlimited length when truncating. This offers a more complex option than the truncate method which allows specification of total length (number of character) of either plain text or html+plaintext. 

It also corrects some minor problems in the string.truncate method.

### DIFF
--- a/libraries/joomla/html/string.php
+++ b/libraries/joomla/html/string.php
@@ -216,7 +216,7 @@ abstract class JHtmlString
 
 		// Now deal with more complex truncation.
 		$diffLength = 0;
-		for (; $maxLength <= $baseLength;)
+		while ($maxLength <= $baseLength)
 		{
 			// Get the truncated string assuming HTML is allowed.
 			$htmlString = JHtml::_('string.truncate', $html, $maxLength, $noSplit, $allowHtml = true);
--- a/libraries/joomla/html/string.php
+++ b/libraries/joomla/html/string.php
@@ -35,6 +35,11 @@ abstract class JHtmlString
 	 */
 	public static function truncate($text, $length = 0, $noSplit = true, $allowHtml = true)
 	{
+		// Assume a lone open tag is invalid HTML.
+		if ($length == 1 && substr($text, 0, 1) == '<')
+		{
+			return '...';
+		}
 		// Check if HTML tags are allowed.
 		if (!$allowHtml)
 		{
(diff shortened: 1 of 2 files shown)


libraries/joomla/html/html/string.php
```diff
@@ -35,6 +35,11 @@
 	 */
 	public static function truncate($text, $length = 0, $noSplit = true, $allowHtml = true)
 	{
+		// Assume a lone open tag is invalid HTML.
+		if ($length == 1 && substr($text, 0, 1) == '<')
```
pasamio: If $length == 1 then surely you can just do $text == '<' without the substr?

elinw:
Length is the maximum length allowed  and then we want to know if the first character of the  string (of unknown length) to be truncated to one character  is an open tag because if it is that causes some difficulty.

libraries/joomla/html/html/string.php
elinw: Basically if you are truncating to one character when you have a string longer than one character you have some problems no matter what because we requre a ... to indicate the truncation, but in this case we are just saying look you have html here and you only want one character, we're just going to call that truncated totally and be done.   The real problem is if you just truncate to a <.  You could make an argument that if someone starts of with an  and they truncate to one character (which becomes ...) that we should include all of the html. We could do it but it seems extremely edge case.

elinw: Actually ...
If you look at the test at line 364 of the test changes you will see that in truncateComplex if there is only html, such as an image tag, we will return that even with a length of 1 whereas in truncate normal you should get either nothing in the mode that does not allow html which should then be turned to ...  or ... in the mode that does allow html because the $length doesn't give you room to close the tag (you really must have at least 5 characters to do something with valid html and that's pushing it since I don't know that there is a one letter self closing tag  <? /> ).  So what was happening was when truncate was given this scenario it was doing the truncation (at line 55/60) and then seeing there was no > in the truncated string so it was returning the < as a character. I'm proposing here that we make the assumption the < in the first position of a 1 character string is an opening tag and handle it as such.

I was really not intending to do a rewrite of truncate itself but I found there are some issues with short lengths. Also we're going to have to look at its assumptions again in light of some of the new tags from html5.

elinw:
Actually technically the ellipsis is one character even though we are creating it with 3.

PR #1415, opened 2012-07-27
JRegistryFormatINI::stringToObject() preg_match key validation

phproberto:
Fix an actually non-relevant but potential bug reported by eaxs in joomla-cms.

// Validate the key.
if (preg_match('/[^A-Z0-9_]/i', $key))
{
    // Maybe throw exception?
    continue;
}

Issue 267 reported by eaxs in joomla-cms
https://github.com/joomla/joomla-cms/issues/267

### DIFF
--- a/libraries/joomla/registry/format/ini.php
+++ b/libraries/joomla/registry/format/ini.php
@@ -137,7 +137,7 @@ class JRegistryFormatINI extends JRegistryFormat
 			list ($key, $value) = explode('=', $line, 2);
 
 			// Validate the key.
-			if (preg_match('/[^A-Z0-9_]/i', $key))
+			if (!preg_match('/[^A-Z0-9_]/i', $key))
 			{
 				// Maybe throw exception?
 				continue;


elinw:
Would you mind changing this title to something actually describing the change?

phproberto:
Sorry. Edited the title.

pasamio:
Do you mind putting in a unit test that shows this bug in action and can demonstrate it working properly?

phproberto:
As it's not a real bug can't be tested and doesn't worth to lose any time. Only have to see the if statement. Otherwise free to close the pull request.

pasamio:
I'm confused though, if it isn't a real bug what are we fixing? It sounds like it's a bug, all I'm asking for is a unit test to feed data to the function and replicate the error/prove it's fixed and ensure it doesn't get broken into the future.

phproberto:
If you see the if statement you can see that it doesn't do anything with or without the error.  It's only a logical structure to complete in the future. Unit test has no sense as the if only contains a continue. It takes 2 min. to identify the problem.

We better forget this. If someday the logical structure is used the bug will appear and someone will fix it.

pasamio:
Then it should only take two minutes to build a unit test to verify it and ensure it works into the future.

phproberto:
Sure. Thank you for your time.

PR #1417, opened 2012-07-28
Allowing override of the xx-XX.localise.php 

infograf768:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28877

This patch let's override the settings in the localise.php file without touching at the original file in the core language folders by letting add an xx-XX.localise.php file in the language/overrides/ folders.
For example the maximum number of characters allowed to search.

### DIFF
--- a/libraries/joomla/language/language.php
+++ b/libraries/joomla/language/language.php
@@ -199,9 +199,9 @@ class JLanguage
 			$paths[3] = JPATH_ADMINISTRATOR . "/language/$lang/$lang.localise.php";
 		}
 
-		ksort($paths); 
+		ksort($paths);
 		$path = reset($paths);
-		
+
 		while (!class_exists($class) && $path)
 		{
 			if (file_exists($path))
--- a/libraries/joomla/language/language.php
+++ b/libraries/joomla/language/language.php
@@ -186,6 +186,26 @@ class JLanguage
 		$class = str_replace('-', '_', $lang . 'Localise');
 		if (!class_exists($class) && defined('JPATH_SITE'))
 		{
+			// Class does not exist. Try to find it in the Site Language Overrides Folder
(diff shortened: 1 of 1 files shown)


pasamio:
Do you mind refactoring this to use an array and a loop for each of them. Then you can optimise the if (defined(JPATH_SITE) to wrap around something that will add twice to a $paths array lets say. Then something like this could work:

```
        $paths = array();
        if (defined('JPATH_SITE'))
        {
            // Note: Manual indexing to enforce load order.
            $paths[0] = JPATH_SITE . "/language/overrides/$lang.localise.php";
            $paths[2] = JPATH_SITE . "/language/$lang/$lang.localise.php";
        }

        if (defined('JPATH_ADMINISTRATOR'))
        {
            // Note: Manual indexing to enforce load order.
            $paths[1] = JPATH_ADMINISTRATOR . "/language/overrides/$lang.localise.php";
            $paths[3] = JPATH_ADMINISTRATOR . "/language/$lang/$lang.localise.php";
        }
        ksort($paths); // not sure if this is necessary but it can't hurt. it'll ensure the array is properly ordered.
        $paths = array_values($paths); // this will remove the indexes. also should not be necessary.

        $path = reset($paths);
        while (!class_exists($class) && $path)
        {
            if (file_exists($localise))
            {
                require_once $localise;
            }
            $path = next($path);
        }
```

The `ksort` and `array_values` may not be necessary however it never hurts to check. They're there for if say `JPATH_ADMINISTRATOR` isn't defined and the array only has two entries but I don't think it'd make a difference however it doesn't hurt to try it out and verify. If it does make a difference then the code is already there.

infograf768:
looks that
$paths = array_values($paths);
was not necessary

PR #1419, opened 2012-07-29
Add active flag to JPaginationObject, add coverage-legacy to build/.gitignore

mbabker:
This adds a flag to JPaginationObject so that items can be reported as the active page in the pagination.

Also, I added the legacy unit report coverage to the .gitignore since it wasn't there.

### DIFF
--- a/libraries/joomla/pagination/pagination.php
+++ b/libraries/joomla/pagination/pagination.php
@@ -721,6 +721,10 @@ class JPagination
 				$data->pages[$i]->base = $offset;
 				$data->pages[$i]->link = JRoute::_($params . '&' . $this->prefix . 'limitstart=' . $offset);
 			}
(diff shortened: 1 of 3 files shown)


libraries/joomla/pagination/pagination.php
```diff
@@ -721,6 +721,10 @@ protected function _buildDataObject()
 				$data->pages[$i]->base = $offset;
 				$data->pages[$i]->link = JRoute::_($params . '&' . $this->prefix . 'limitstart=' . $offset);
 			}
+			elseif ($i = $this->get('pages.current'))
```
realityking: Use $this-> pagesCurrent instead. The get() method is deprecated.

mbabker: Done

PR #1422, opened 2012-08-01
OAuth 1.0 and 1.0a client

diananeculai:


realityking:
This looks pretty great already. You got a couple of uni test errs thou: http://developer.joomla.org/pulls/pulls/1422.html

diananeculai:
Thanks! 

Yes, I know and I am trying to fix them. The fun fact is that locally the tests pass - the problem is at setting session the right way in tests.

diananeculai:
Solved it, no errors now :)

libraries/joomla/oauth/1aClient.php
```diff
@@ -0,0 +1,588 @@
...
+
+		if (empty($verifier))
+		{
+			// Generate a request token.
+			$this->_generateRequestToken();
+
+			// Authenticate the user and authorise the app.
+			$this->_authorise();
+		}
+		// Callback
```
elinw: This should really have a blank line above the comment.

pasamio:
@dianaprajescu Can you rename it to just oauth/client.php and we'll have oauth2/client.php. Unless someone is planning on implementing a 1.0 client (not 1.0a), this will be fine.

elinw:
Yes we do have someone working on OAuth1. It's commonly used when data are public (e.g Open Street Map and large govt databases) and also for widgets that phone home

pasamio:
Does it share a high level of code sharing with 1a or is it completely different like 2 appears to be with no ability to share? If it can share a significant chunk of code I think integrating it into the constructor and setting the behaviour that way and perhaps defaulting to 1a if that's the more predominant value then going forwards that way.

elinw:
It's extending 1a pretty extensively since really 1a is 1 plus. That sounds like a reasonable approach to me.   I'd be happy to see what you're suggesting if  Diana thinks it is workable.

pasamio:
Well the way I see it is we have two options:
1) If OAuth1 and OAuth1a don't share any code either we split them into their own packages. In a sense at the moment that's probably the path of least resistance.
2) If they do share some amount of code that we can re-use between the two, expanding through a constructor param that selects the appropriate version (1 or 1a) and mutates the object behaviour that way.

elinw:
In terms of the protocols they are almost the same but 1a adds to 1 to deal with the session fixation issue which is not relevant for 1 users. http://oauth.net/core/1.0/  I recently read also that Dropbox uses OAuth1 not  1a.
Open Street Map apparently allows 1a to fall back to 1 and maybe others do too, which to me indicates that just letting it be a param makes sense. 

http://wiki.oauth.net/w/page/12238555/Signed%20Callback%20URLs

I'm pretty sure the red summarizes the differences.
http://wiki.oauth.net/w/page/12238535/OAuth%20Session%20Fixation%20Advisory

You an see from the above that the real reason people use 1 is that there are devices with no web browsers and and no keyboards or other input mechanisms so they need to use 1.

diananeculai:
There are only few differences between OAuth 1.0 and 1.0a so I have included OAuth1.0 in this client, using a constructor param, like @pasamio said.

There are some problems with the unit test and I would appreciate some help. Locally the tests are passing with no problems, but in the pull tests results I currently get Segmentation fault. This is after my last commit, before it I was getting: 
<code>JSessionTest::testGetExpire
Session expire should be 20
Failed asserting that 900 matches expected 20.
.../tests/suites/unit/joomla/session/JSessionTest.php:131</code>
I can't see how this is related to my code.

eddieajau:
@dianaprajescu do you have any developer documentation for this PR?

diananeculai:
@eddieajau no, I don't have any documentation.

@LouisLandry I have opened this PR <b>4 months</b> ago, isn't that quite a long time? You should've gave me some feedback and collaborate on this if you were already working on something similar. It seems that I've wasted time for nothing. Both JTwitter and JLinkedin use this client and I was waiting for you guys to give me feedback and finally merge this PR in order to make PRs for JTwitter and JLinkedin.

LouisLandry:
@dianaprajescu the pull request I have opened does not conflict with this one.  It is only implementing the server side of OAuth1.  What you may see as a client class in that package is actually representing a "client application" from a server perspective.  It is true that I would generally like us to take a slightly different approach on the OAuth clients.  I spoke with Aaron about those ideas briefly at the Joomla World Conference.  In short, I'd like the OAuth authentication piece be injected into a `JHttp` class instead of wrapping a `JHttp` class to allow OAuth authentication for requests.  I would like for us to support this type of authentication in the `JHttp` class similar to how we support basic auth today.  I'm very sorry that I haven't had/made the time to give more constructive feedback on this, but c'est la vie.

diananeculai:
@LouisLandry  Yes, the client class confused me, I haven't looked over it's content but the file is named oauth1/client.php, which is the same with my client. It makes sense to include this in a JHttp class, but for the moment wouldn't it be a good idea to have this OAuth client merged, just like Aaron's was until me or someone else has time to make the change?

eddieajau:
@dianaprajescu could you consider writing some documentation about how to use this package please. For those of us that are new to oAuth (like me), it would be helpful to see how you intend developers to use it.  A markdown file in /docs/manual/en-US/chapters/packages would be ideal.  I certainly appreciate the work you have already put into this but for me I don't know how to review this because there is no documentation of any kind (you probably just need to recycle some of your blog posts from GSOC).  

Also, a brief introduction in the description of this pull request would be helpful as the changelog is automatically generated from this information.

For what it's worth, Louis needs to do the same in his pull request ;)

Thanks in advance.

diananeculai:
@eddieajau I wrote some documentation for the oAuth client. Let me know what you think.

eddieajau:
Thanks @dianaprajescu. Do you do any examples of connecting to real services during GSOC (like an example of connecting to Twitter)?

diananeculai:
@eddieajau <a href="https://gist.github.com/3258852">here</a> is the app I was using for testing the JTwitter package. <a href="https://github.com/dianaprajescu/joomla-platform/blob/JTwitter/libraries/joomla/twitter/oauth.php">This</a> is the class that extends the JOAuth1Client. And <a href="https://github.com/dianaprajescu/joomla-platform/blob/JTwitter/libraries/joomla/twitter/statuses.php#L307">here</a> is an example of an authorised request.

diananeculai:
I would like to bring this PR into your attention too. The JTwitter and JLinkedIn packages that are ready highly depend on the OAuth client. Moreover, the PR #1786 depends on it too.

elinw:
+1  and it opens up a lot of other possibilities.

Buddhima:
+1 too. If you look at the "List of OAuth service providers"  at http://en.wikipedia.org/wiki/OAuth you can see that most of  OAuth providers are using OAuth 1.0 or 1.0a
Therefore adding OAuth 1.0 client would make developers life much easy and enable to develop packages for Joomla with OAuth 1.0 in future.

AmyStephen:
@dianaprajescu - first, major kudo's for hanging in there over six months and keeping this PR current. You appear to be someone who does not discourage easily. Tenacity comes in handy with open source. =)

Also, I've learned in many years of sometimes being confused as to why something is or isn't happening to pick up on subtle hints which, more often than not, are usually shared once. As I look back through this history, I can't help but wonder if you have you had a chance yet to address @LouisLandry 's concerns? If not, that might be helpful. It might also be useful to review Louis's code on the other PR (which was closed). Maybe there are ideas on approach you might borrow?

Thank you for your work, Diana, and for hanging in there.

diananeculai:
Thanks, @AmyStephen ! I'm aware of what @LouisLandry said, but I also suggested to get it merged like this for the moment, until I have time to change it and nobody told me this can't be done. Right now I don't have the necessary time to work on this, so we should probably close this PR and I'll reopen it later.

AmyStephen:
@dianaprajescu Please don't do so on my account. Maybe a maintainer can comment on what they might be looking. I am just guessing.

AmyStephen:
@dianaprajescu Another idea is to post on the platform list and ask what might be needed to get this code in. If you don't have time, we could try to find someone to help. Getting some exposure on list could help get all the pieces moving right. If you do, I'll look for your post and see if there is something I might be able to do to help or if I can find something.

betweenbrain:
I'd hate to see this PR closed. I see a lot of support for it an not much opposition. What can we do to help get it merged?

elinw:
Just follow Louis's instructions and send a PR to @dianaprajescu

AmyStephen:
Might be worth joining the discussion Diana started on list a couple of days ago. 

https://groups.google.com/d/topic/joomla-dev-platform/xddM8195hpo/discussion

PR #1423, opened 2012-08-02
JOAuth 2.0 Client

aaronschmitz:
This is an OAuth 2.0 client for requesting access tokens and sending signed requests. I created it as a part of my JGoogle GSoC project (http://goo.gl/5dj50).


aaronschmitz:
See also Diana's 1.0a client: https://github.com/joomla/joomla-platform/pull/1422

libraries/joomla/oauth/oauth2client.php
```diff
@@ -0,0 +1,344 @@
...
+ * @package     Joomla.Platform
+ * @subpackage  Oauth
+ *
+ * @copyright   Copyright (C) 2005 - 2012 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE
+ */
+
+defined('JPATH_PLATFORM') or die;
+jimport('joomla.environment.response');
+jimport('joomla.environment.uri');
```
realityking: The jimport for JUri isn't necessary anymore.

libraries/joomla/oauth/oauth2client.php
```diff
@@ -0,0 +1,343 @@
...
+jimport('joomla.environment.response');
+
+/**
+ * Joomla Platform class for interacting with an OAuth 2.0 server.
+ *
+ * @package     Joomla.Platform
+ * @subpackage  Oauth
+ * @since       12.2
+ */
+class JOauthOauth2client
```
stefanneculai: I find it strange to name the class J<b>OauthOauth</b>2client. Wouldn't it be better to change the name to JOauth2client (it sounds more natural)?

elinw:
Then wouldn't the autoloader be looking for a file named 2client?

stefanneculai:
Would be a problem to name it 2client.php? 
Maybe make a folder called v2 and put the client.php in it. This way it would be JOAuthV2client.

aaronschmitz:
I suppose it wouldn't be a problem.

elinw:
I think I'd like to hear from the maintainers on this one.

Just FYI Aaron needs to know about the patch() issue in order to know how to fix the unit test that has an error.

aaronschmitz:
Ok. I switch to extending JHttp like JGithub does. If the maintainers accept @dianaprajescu's pull for the patch method I'm 95% of the way there, and if they don't this is a superior alternative to using JHttpTransport.

For now, I renamed the class to JOauth2client and the file to 2client, but I'm flexible on what others prefer.

pasamio:
@aaronschmitz do you mind renaming it from oauth/v2client.php to oauth2/client.php? Given it seems to be an independent implementation with no sharing of code between the two implementations.

elinw:
@pasamio can you comment on whether the class name should be JOAuth or  JOauth (my understanding is that the loader looks for non consecutive upper case letters which would make JOAuth okay.  Codestyle originally called for uppercase where appropriate (HTML ZXML) but we veered away from that with implementation of the autoloader.  We probably should aim for some consistency.

aaronschmitz:
Closing in favor of https://github.com/joomla/joomla-platform/pull/1480.

PR #1424, opened 2012-08-03
J form field combo wrong element type

okonomiyaki3000:
JFormFieldCombo::getInput() is generating a text input and a hidden list. The combobox javascript actually expects a select. So, combo field actually just shows up as a text field with none of the features of a combobox. 

Anyway, isn't it weird that this is part of the platform while the javascript that it requires is part of the cms?

### DIFF
--- a/build/.gitignore
+++ b/build/.gitignore
@@ -1,4 +1,5 @@
 /coverage
+/coverage-legacy
 /logs
 /pdepend
 /code-browser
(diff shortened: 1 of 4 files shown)


okonomiyaki3000:
Hmmm? what just happened?

realityking:
Looks like you rebased or something.

As for the JavaScript, that's part of the platform as well. Take a look at the media/system folder.

PR #1425, opened 2012-08-03
JFormFieldCombo::getInput()

okonomiyaki3000:
JFormFieldCombo::getInput() should return a fully formed 'select' element with a class of 'combobox' as that is what is expected by the combobox javascript functions.

Isn't it a bit odd that this is part of 'platform' but the required javascript is part of 'cms'?

### DIFF
--- a/libraries/joomla/form/fields/combo.php
+++ b/libraries/joomla/form/fields/combo.php
@@ -44,20 +44,11 @@ class JFormFieldCombo extends JFormFieldList
 
 		// Initialize some field attributes.
 		$attr['class'] = $this->element['class'] ? 'combobox ' . $this->element['class'] : 'combobox';
-		if ((string) $this->element['readonly'] == 'true')
-		{
-			$attr['readonly'] = 'readonly';
-		}
-		if ((string) $this->element['disabled'] == 'true')
-		{
-			$attr['disabled'] = 'disabled';
-		}
+		((string) $this->element['readonly'] == 'true') ? $attr['readonly'] = 'readonly' : null;
+		((string) $this->element['disabled'] == 'true') ? $attr['disabled'] = 'disabled' : null;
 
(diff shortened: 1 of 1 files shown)


libraries/joomla/form/fields/combo.php
```diff
@@ -40,16 +40,24 @@ protected function getInput()
...
 		// Initialize some field attributes.
-		$attr .= $this->element['class'] ? ' class="combobox ' . (string) $this->element['class'] . '"' : ' class="combobox"';
-		$attr .= ((string) $this->element['readonly'] == 'true') ? ' readonly="readonly"' : '';
-		$attr .= ((string) $this->element['disabled'] == 'true') ? ' disabled="disabled"' : '';
-		$attr .= $this->element['size'] ? ' size="' . (int) $this->element['size'] . '"' : '';
+		$attr['class'] = $this->element['class'] ? 'combobox ' . $this->element['class'] : 'combobox';
+		if ((string) $this->element['readonly'] == 'true')
+		{
+			$attr['readonly'] = 'readonly';
+		}
```
realityking: Please keep using the tenary operator. It's much compacter.

realityking:
Does this keep working with behavior.combobox?

okonomiyaki3000:
No, it STARTS working with behavior.combobox which is currently broken.

elinw:
Combo box has been broken for a long, long time.

okonomiyaki3000:
It looks like the JavaScript was changed at some point but the HTML never was. So it should start to work again now.

okonomiyaki3000:
Since the Isis theme in 3.0 uses Chosen to make very nice selects, it further breaks the combo-box functionality. Should an effort be made to fix it or should combo just be dropped from the platform? It makes no sense to keep a broken field type.

PR #1426, opened 2012-08-03
Fix improper detection of UTF-8 encoding in strcasecmp and strcmp.

elinw:
The test of whether locale is UTF-8 is incorrect because it is missing the 'utf8' string which is commonly what is in the string returned when locale0 is obtained.


infograf768:
Makes sense. We have both depending on OSes.

elinw:
See https://github.com/joomla/joomla-cms/issues/179.

PR #1431, opened 2012-08-05
Handling in HttpTransport when the requested uri does not exist

elinw:
The transports are currently not handling it well when the domain for a requested uri completely does not exist. At this point they are somewhat inconsistent in what they do (e.g. stream silences the message and then throws an exception similarly to what I propose here for socket). Rather than use a generic exception message this uses php and curl errors to give more detailed information about what went wrong.

### DIFF
--- a/libraries/joomla/http/transport/socket.php
+++ b/libraries/joomla/http/transport/socket.php
@@ -159,6 +159,10 @@ class JHttpTransportSocket implements JHttpTransport
 	{
 		// Create the response object.
 		$return = new JHttpResponse;
+		if (empty($content))
+		{
+			throw new UnexpectedValueException('No content in response.');
+		}
 
 		// Split the response into headers and body.
 		$response = explode("\r\n\r\n", $content, 2);
@@ -167,7 +171,7 @@ class JHttpTransportSocket implements JHttpTransport
 		$headers = explode("\r\n", $response[0]);
 
 		// Set the body for the response.
-		$return->body = $response[1];
+		$return->body = empty($response[1]) ? '' : $response[1];
 
 		// Get the response code from the first offset of the response headers.
 		preg_match('/[0-9]{3}/', array_shift($headers), $matches);
@@ -177,6 +181,7 @@ class JHttpTransportSocket implements JHttpTransport
 		{
 			$return->code = (int) $code;
 		}
+
 		// No valid response code was detected.
 		else
 		{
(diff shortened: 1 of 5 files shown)


elinw:
These came up when testing bad urls. The challenge was that fsockopen() sends a warning when the uri is not there ... the sending of the warning was making it very difficult to handle problem so this was the solution I finally came to after going around in circles for a while.

piotr-cz:
Some time ago I've send a PR #1022 that was addressing this issue across all transports ( fix for Curl has been merged yesterday in #1438).

I think it's important to keep same convention across all transport and throw an exception in the same place (probably `request` function).

elinw:
Well .. I agree consistent is good and as you see I also came to the conclusion that silencing was the only option for fsocketopen().  The no conent condition is different than the bad url condition. I also think it's worth considering that there may only be headers and no body (or body and no headers) to cover all scenarios.

elinw:
There was a reason for doing it twice but I can't remember what it was (although I will test again with your version and see if it copes with all of my cases).  I guess that it really doesn't matter that there are two different exceptions happening, one where the host is not foundand the other where it can't be opened for another reason.

piotr-cz:
@elinw 
More descriptive exceptions are always welcome.
But there are some cases (authentication to external REST service) when there are multiple requests exchanged and every saved request counts.

Besides in my opinion in your previous code both assertions for `fsockopen` had same result and passing first would mean passing second as well.

Thanks for taking time at it.

elinw:
Yes I was trying to remember why I did that and I think I just was avoiding touching the existing code for no good reason and really more focused on the empty $content problem.

elinw:
Wow that was a not easy exercise in rebasing

libraries/joomla/http/transport/socket.php
```diff
@@ -158,23 +158,33 @@ protected function getResponse($content)
...
+		}
 		// Split the response into headers and body.
 		$response = explode("\r\n\r\n", $content, 2);
 
 		// Get the response headers as an array.
 		$headers = explode("\r\n", $response[0]);
 
+		if (empty($response[1]))
+		{
+			throw new UnexpectedValueException('No [1] offset in response.');
```
pasamio: If we got headers and an empty body, I'm not sure this is necessarily an exception.

pasamio: Went and looked at the spec, a HEAD request by definition won't return a body:

<blockquote>
The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response.
</blockquote>


So we really shouldn't return an exception here.

PR #1432, opened 2012-08-05
Specify additional options for cURL transfers.

diananeculai:
I have added a new parameter to the request method through which to specify additional options for the cURL transfer.

Currently the HTTP client will automatically follow redirects. I have removed this because (see RFC [0]):
- it is not always necessary to follow redirects
- following redirects automatically may lead to an infinite loop
- HTTP redirection codes may need to be handled differently

[0] http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html - 10.3 Redirection 3xx

### DIFF
--- a/libraries/joomla/http/http.php
+++ b/libraries/joomla/http/http.php
@@ -111,15 +111,16 @@ class JHttp
 	/**
 	 * Method to send the HEAD command to the server.
 	 *
-	 * @param   string   $url      Path to the resource.
-	 * @param   array    $headers  An array of name-value pairs to include in the header of the request.
-	 * @param   integer  $timeout  Read timeout in seconds.
+	 * @param   string   $url          Path to the resource.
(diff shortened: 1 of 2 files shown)


pasamio:
Calling it $curlOptions and jamming it into JHttp feels wrong. JHttp is supposed to be transport agnostic and putting something named as one of the specific options doesn't feel right. 

Why can't these be included with the $options that are created when the transport is constructed? It's obviously cURL specific, so if you want to set specific cURL options then you can instantiate a cURL transport and set the parameters appropriately?

diananeculai:
Yes, I haven't noticed that, it works by passing the cURL options through the $options parameter. This means that the option to follow redirects should also be passed this way and <code>$options[CURLOPT_FOLLOWLOCATION] = true;</code> should be removed.

I will close this pull request and make a new one with this change if that's okay.

PR #1434, opened 2012-08-06
JhtmlBehavior - improve  _getJSObject

okonomiyaki3000:
This function has several flaws which I think I've solved. I think that with a little more work I can eliminate it entirely as json_encode() should be used whenever possible.

### DIFF
--- a/libraries/joomla/html/html/behavior.php
+++ b/libraries/joomla/html/html/behavior.php
@@ -808,7 +808,7 @@ abstract class JHtmlBehavior
 
 	/**
 	 * Internal method to get a JavaScript object notation string from an array
-	 * 
+	 *
 	 * This function differs from json_encode() in two important ways:
 	 *	1) If the array contains a key 'fullScreen', the resulting object will have a property 'size' which will be an object with properties 'x' and 'y'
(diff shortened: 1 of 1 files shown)


okonomiyaki3000:
I changed my mind about removing it entirely. There are too many cases of things being passed with the \ prefix. That practice would need to be deprecated first. It could, perhaps, be purged from all Joomla code but there's no way to know if 3rd parties are making use of it. 

In any case, it does not affect this patch as the \ behavior is still supported here.

realityking:
Why would we remove the use of \? The whole reason that method exists is to support data types beyond those that are handled by JSON.

okonomiyaki3000:
In my opinion, there are better ways to support types such as functions while letting json_encode take care of whatever it possibly can. But I suppose the \ solution is one that has worked well and maybe shouldn't be changed.

LouisLandry:
Good cleanup.  If you can rebase it so that it merges cleanly I'll get it merged in this weekend.  Thanks a bunch Elijah.

okonomiyaki3000:
Sorry, I'm on Japan time. It'll have to wait till Monday.

PR #1436, opened 2012-08-06
JLoader : files located at the root path, where the prefix is registered...

florianv:
Files located at the root path, where the prefix is registered, cannot be autoloaded.

Ex : Prefix `PN` registered in `/path/to/pn`.
A class `PNVariable` located in `path/to/pn/variable.php` won't be autoloaded, because the auto loader try to find it in `path/to/pn/variable/variable.php`

### DIFF
--- a/libraries/loader.php
+++ b/libraries/loader.php
@@ -321,7 +321,7 @@ abstract class JLoader
 		$fileName = false;
 
 		// If there is only one part.
-		if (count($parts == 1))
+		if (count($parts) == 1)
 		{
 			// Keep the possible file name.
 			$fileName = $parts[0];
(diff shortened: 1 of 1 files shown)


pasamio:
Do you mind rebasing the commits to remove the extra two commits that don't really add much value. 

I'm not sure I entirely agree with the direction as this would encourage people to place files in the root and I don't feel that this is generally what we'd be after. I can see a few use cases (my own included) where this would be useful however. I'll wait for feedback from others.

florianv:
Hello, I rebased it here : https://github.com/joomla/joomla-platform/pull/1445.

For Joomla itself, it is not so useful but it can allow to autoload the factory.

For Platform applications, if a separate library come with it, you might want to have some files in the root path of the library, or the front controller of the application located in the root path of the application, for example.

pasamio:
For future reference if you rebase and do a force push to the original branch, the original pull request will be updated with the rebase'd commits.

florianv:
Ok, thanks for the tip.

PR #1438, opened 2012-08-06
Fixing warnings triggered by cURL when no response is received.

pasamio:
This pull request resolves a situation where when no response from a server is received a PHP warning is thrown by the cURL transport. Instead the code triggering the warning is mitigated and additionally a check to see if there was any response data is added to validate a response was actually received or throw an exception.

### DIFF
--- a/libraries/joomla/http/transport/curl.php
+++ b/libraries/joomla/http/transport/curl.php
@@ -169,6 +169,12 @@ class JHttpTransportCurl implements JHttpTransport
 		// Create the response object.
 		$return = new JHttpResponse;
 
+		// Check if the content is actually a string.
+		if (!is_string($content))
+		{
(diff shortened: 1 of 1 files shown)


piotr-cz:
Please take a look at my PR #1022.
This is addressing same issue across all JHttpTransports

elinw:
Actually there's a difference between getting empty content and what you get from a bad url where in fact you are not going to get a normal repsonse code since there is no server. 
 $return = new JHttpResponse

I haven't tested it yet but I bet it isn't pretty if the $uri doesn't exist, it probably needs an exception and then JTransportCurl needs to handle that.  And yes the error issue is a problem because php is going to send a notice before any exception gets thrown.

Needs to catch that

PR #1443, opened 2012-08-09
Regex filter fits

Radek-Suski:
Modified the regular expression in language debugger to include dot sign as a valid KEY character and allow including an escaped double quotes inside the value (right-hand)


pasamio:
Wasn't there a bug with escaped double quotes on PHP 5.2? Isn't that why we did the whole _QQ_ thing?

realityking:
It wasn't supported in 5.2 at all AFAIK. But we don't support PHP 5.2 anymore ;)

realityking:
Works just fine in my testing.

Radek-Suski:
Thanks a lot :)

pasamio:
Cool, I couldn't remember if it was PHP 5.2 or 5.3 that had it or some version of PHP 5.3. Yay for workarounds!

PR #1444, opened 2012-08-09
Fix cache error due to incorrect commit 83eb8fae 

dextercowley:
I goofed with the prior pull request. This should fix it. Thanks.


realityking:
Thanks Mark.

PR #1447, opened 2012-08-12
Add stripIframe method

elinw:
In doing some work with feeds I noticed that we have stripImage available, but we don't have a way to strip iFrames which pose similar or more serious issues than images. A separate method makes sense because sometimes it is very reasonable to take the images from a feed but not to want the risks associated with iFrames.


realityking:
Could you change both @ since tags to 12.2? Thanks.

elinw:
Done

Issue #1437, opened 2012-08-06
Bug in JArchiveZip::extract() - missing return


jms2win:
In Joomla 3.0 alpha1, we have identified an issue in the JArchiveZip class.

The JArchive::extract() function performs some test and depending on the file type, call an adapter to perform the action.

public static function extract($archivename, $extractdir)
{
. . . .
. . . .
. . . .
   $result = $adapter->extract($archivename, $extractdir);
. . . .
. . . .
        if (!$result || $result instanceof Exception)
        {
            return false;
        }
        return true;
    }

The issue is that the JArchiveZip::extract() function does NOT return any value and therefore the value received by JArchive is NULL.
This result by a JArchive::extract() that ALWAYS return false.

The fix consists in adding return inside the JArchiveZip::extract routine.

Here it is the code as we have modified it.

```
public function extract($archive, $destination, array $options = array())
{
    if (!is_file($archive))
    {
        if (class_exists('JError'))
        {
            return JError::raiseWarning(100, 'Archive does not exist');
        }
        else
        {
            throw new RuntimeException('Archive does not exist');
        }
    }

    if ($this->hasNativeSupport())
    {
        return $this->_extractNative($archive, $destination, $options);
    }
    else
    {
        return $this->_extract($archive, $destination, $options);
    }
}
```

You will see in the last section if ($this->hasNativeSupport()) that we have added return to the call on _extractNative() and _extract() functions.

elinw:
Will you just make a pull request for this?

jms2win:
Sorry but in the past, I already tried to do a pull request and didn't know how to do on the correct version of the file.
If you want , I can go in the bug tracker and attache the file that was the one in Joomla 3.0 alpha 1 that I have modified.

I prefer to let "bug squad" to update the correct source.

elinw:
Just go here and clic edit to modify the file
https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/archive/zip.php#L120

jms2win:
Done

elinw:
And merged ... will you close this please? Thanks for doing this work.

realityking:
Closing since this is fixed. Thanks.

PR #1451, opened 2012-08-14
Add missing return in the "extract()"

jms2win:
This apply the fix reported in the issue
https://github.com/joomla/joomla-platform/issues/1437#issuecomment-7539014


elinw:
See I told you it wasn't hard!

PR #1454, opened 2012-08-17
Prevent unit tests from fatally failing if database is not available.

aaronschmitz:
Without these changes, all unit test crash when phpunit attempts to call a method on a null object.

### DIFF
--- a/tests/core/case/database.php
+++ b/tests/core/case/database.php
@@ -313,7 +313,14 @@ abstract class TestCaseDatabase extends PHPUnit_Extensions_Database_TestCase
 	 */
 	protected function getConnection()
 	{
-		return $this->createDefaultDBConnection(self::$driver->getConnection(), ':memory:');
+		if (self::$driver)
+		{
+			return $this->createDefaultDBConnection(self::$driver->getConnection(), ':memory:');
+		}
+		else
+		{
+			return null;
+		}
 	}
 
 	/**
--- a/tests/core/case/database.php
+++ b/tests/core/case/database.php
@@ -313,7 +313,7 @@ abstract class TestCaseDatabase extends PHPUnit_Extensions_Database_TestCase
 	 */
 	protected function getConnection()
 	{
-		if (self::$driver)
+		if (!is_null(self::$driver))
 		{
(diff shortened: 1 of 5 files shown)


stefanneculai:
Take a look at the following line:
https://github.com/aaronschmitz/joomla-platform/blob/50c7794826a7e763604f37e6c07412564f961188/tests/core/case/database.php#L491

I think this line should mark the test as incomplete if <code>driver=null</code> and this way <code>getConnection</code> is never called. I would say that there is a problem with using empty which has a strange behavior depending of PHP version. Try <code>is_null(static::$driver)</code> instead of <code>empty(static::$driver)</code> - this may fix the issue.

This is only a guess. Sorry if I am wrong.

aaronschmitz:
I don't follow how that change will prevent getConnection from being executed - please explain.

stefanneculai:
It doesn't look like <code>PHPUnit_Extensions_Database_TestCase</code> uses getConnection before setUp so I supposed that it is called after that. If driver is null then the test should stop and be marked as incomplete. This way getConnection is never called if driver is null.  Maybe I'm wrong.

aaronschmitz:
I hope this is along the lines of what you were thinking, @stefanneculai.

stefanneculai:
I was suggesting that the fix to this issue would be just to change <code>if (empty(static::$driver))</code> from http://goo.gl/p3spv to <code>if (is_null(static::$driver))</code> or <code>if (static::$driver === null)</code>. I think this could replace all the changes that you did in this pull request. Maybe you can give it a try.

aaronschmitz:
According to the PHP documentation empty(null) == true - http://php.net/manual/en/function.empty.php. The problem is that marking the unit test as skipped doesn't immediately terminate the current test.

stefanneculai:
I know that but on some versions of PHP empty(null) returns false. :-)

Also, marking a test as skipped terminates the current test immediately (I have just tried that).

aaronschmitz:
Well, I tried making the change you suggested on 491 just to be sure, and it didn't work. If you have or anyone else have a simpler solution that solves the problem, I'm all ears... but I haven't found one yet.

LouisLandry:
This looks good enough.  Thanks.

PR #1456, opened 2012-08-17
Fix JInput loading in JApplicationCli

kukudroid:
The autoloader is quite case sensitive ;)


elkuku:
err... see &rArr; #1457

PR #1457, opened 2012-08-17
Fix JInput loading in JApplicationCli

elkuku:
The autoloader is quite case sensitive ;)


realityking:
I wonder if we should just remove that check. Anyway, merged.

PR #1458, opened 2012-08-17
Fix bug in JString::parse_url that causes urlencoded strings to be decoded

aaronschmitz:
Switch JString::parse_url to return false on failure as specified in it's doc block.
JString::parse_url is now identical to parse_url for ascii strings.

### DIFF
--- a/libraries/joomla/form/rules/url.php
+++ b/libraries/joomla/form/rules/url.php
@@ -62,7 +62,7 @@ class JFormRuleUrl extends JFormRule
 		 * accurately without a scheme.
 		 * @see http://php.net/manual/en/function.parse-url.php
 		 */
-		if (!array_key_exists('scheme', $urlParts))
+		if ($urlParts && !array_key_exists('scheme', $urlParts))
 		{
 			return false;
 		}
--- a/libraries/joomla/string/string.php
+++ b/libraries/joomla/string/string.php
@@ -961,12 +961,11 @@ abstract class JString
 	 */
 	public static function parse_url($url)
 	{
-		$result = array();
+		$result = false;
 
 		// Build arrays of values we need to decode before parsing
-		$entities = array('%21', '%2A', '%27', '%28', '%29', '%3B', '%3A', '%40', '%26', '%3D', '%24', '%2C', '%2F', '%3F', '%25', '%23', '%5B',
-			'%5D');
-		$replacements = array('!', '*', "'", "(", ")", ";", ":", "@", "&", "=", "$", ",", "/", "?", "%", "#", "[", "]");
+		$entities = array('%21', '%2A', '%27', '%28', '%29', '%3B', '%3A', '%40', '%26', '%3D', '%24', '%2C', '%2F', '%3F', '%23', '%5B', '%5D');
+		$replacements = array('!', '*', "'", "(", ")", ";", ":", "@", "&", "=", "$", ",", "/", "?", "#", "[", "]");
 
 		// Create encoded URL with special URL characters decoded so it can be parsed
 		// All other characters will be encoded
@@ -980,7 +979,7 @@ abstract class JString
 		{
 			foreach ($encodedParts as $key => $value)
 			{
-				$result[$key] = urldecode($value);
+				$result[$key] = urldecode(str_replace($replacements, $entities, $value));
 			}
 		}
 		return $result;
--- a/tests/suites/unit/joomla/string/JStringTest.php
+++ b/tests/suites/unit/joomla/string/JStringTest.php
@@ -597,6 +597,13 @@ class JStringTest extends PHPUnit_Framework_TestCase
 		$actual = JString::parse_url($url);
 		$this->assertEquals($expected, $actual, 'Line: ' . __LINE__ . ' Results should be equal');
 
+		// Test all parts of query
+		$url = 'https://john:doe@www.google.com:80/folder/page.html#id?var=kay&var2=key&true';
+		$expected = parse_url($url);
+		$actual = JString::parse_url($url);
+		$this->assertEquals($expected, $actual, 'Line: ' . __LINE__ . ' Results should be equal');
+
+		// Test special characters in URL
 		$url = 'http://joomla.org/mytestpath/È';
 		$expected = parse_url($url);
 		// Fix up path for UTF-8 characters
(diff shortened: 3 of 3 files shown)


stefanneculai:
Looks good.

libraries/joomla/form/rules/url.php
```diff
@@ -62,7 +62,7 @@ public function test(SimpleXMLElement $element, $value, $group = null, JRegistry
 		 * accurately without a scheme.
 		 * @see http://php.net/manual/en/function.parse-url.php
 		 */
-		if (!array_key_exists('scheme', $urlParts))
+		if ($urlParts && !array_key_exists('scheme', $urlParts))
```
realityking, line 65: Wouldn't it make more sense to error out right after the call to parse_url if it returns false? Or am I missing a usecase?

libraries/joomla/string/string.php
```diff
@@ -961,12 +961,11 @@ public static function compliant($str)
 	 */
 	public static function parse_url($url)
 	{
-		$result = array();
+		$result = false;
```
realityking, line 964: That's an API change, why do we wanna do that?

libraries/joomla/string/string.php
```diff
@@ -961,12 +961,11 @@ public static function compliant($str)
...
 
 		// Build arrays of values we need to decode before parsing
-		$entities = array('%21', '%2A', '%27', '%28', '%29', '%3B', '%3A', '%40', '%26', '%3D', '%24', '%2C', '%2F', '%3F', '%25', '%23', '%5B',
-			'%5D');
-		$replacements = array('!', '*', "'", "(", ")", ";", ":", "@", "&", "=", "$", ",", "/", "?", "%", "#", "[", "]");
+		$entities = array('%21', '%2A', '%27', '%28', '%29', '%3B', '%3A', '%40', '%26', '%3D', '%24', '%2C', '%2F', '%3F', '%23', '%5B', '%5D');
+		$replacements = array('!', '*', "'", "(", ")", ";", ":", "@", "&", "=", "$", ",", "/", "?", "#", "[", "]");
```
realityking, line 968: The only change here is removing the unescaping for "%", correct?

libraries/joomla/string/string.php
```diff
@@ -980,7 +979,7 @@ public static function parse_url($url)
 		{
 			foreach ($encodedParts as $key => $value)
 			{
-				$result[$key] = urldecode($value);
+				$result[$key] = urldecode(str_replace($replacements, $entities, $value));
```
realityking, line 982: Why is this needed?

realityking:
Sorry for the many questions, I'm a bit afraid we break something.

stefanneculai:
@realityking I have pulled this out and tried all kind of tests and looks good. I think @aaronschmitz can explain all those strange str_replace and urldecode. I have tried to tackle this issue several times but Aaron's solution doesn't fail any test - or at least I could not find one that fails.

libraries/joomla/string/string.php
diananeculai, line 964: Probably because this is written in the doc block: <code>@return  mixed  Associative array or false if badly formed URL.</code>

libraries/joomla/string/string.php
diananeculai, line 982: Because urlencoded values will be decoded otherwise. For example if you have <code>$url = 'www.example.com?keywords=foo%20bar'</code> <code>$result['query']</code> should be 'keywords=foo%20bar', instead you'll get 'keywords=foo bar'

libraries/joomla/string/string.php
aaronschmitz, line 964: 1. As @dianaprajescu said, it's now consistent with the doc block (I would hence argue that we're not changine the API, but fixing a bug in our implementation of the API).
2. The whole point of this function is to mimic parse_url with added unicode support. Parse_url returns false in such an instance, so that's our target functionality.

libraries/joomla/string/string.php
aaronschmitz, line 968: That's correct. Escaping percent causes really bad things to happen since we're adding a bunch of other %'s to the string in the process.

libraries/joomla/string/string.php
aaronschmitz, line 982: Again, as @dianaprajescu said, this prevents it from decoding URL's. Both tests that I added on lines 620-630 would fail in the current version of this function. This causes _significant_ problems because this function is used in JURI to help explode URL's for things like HttpTransports.

In my use case, JHttp::get('https://www.googleapis.com/calendar/v3/calendars/me%40aaronschmitz.com/events?singleEvents=true&orderBy=startTime&timeMax=2012-09-01T00%3A00%3A00%2B02%3A00&timeMin=2012-08-01T00%3A00%3A00%2B02%3A00') turned into a request for https://www.googleapis.com/calendar/v3/calendars/me%40aaronschmitz.com/events?singleEvents=true&orderBy=startTime&timeMax=2012-09-01T00:00:00 02:00&timeMin=2012-08-01T00:00:00 02:00 which perhaps unsurprisingly failed.

libraries/joomla/form/rules/url.php
aaronschmitz, line 65: You would know better than I do. This simply maintains current functionality. If we want to investigate changing that functionality, perhaps that could be considered in another pull request.

aaronschmitz:
@realityking, the original code is incredibly confusing at first glance, and I only make it worse - I'm guessing my pseudo-code explanation will shed some light on the black magic.
1. Urlencode the url
   - This converts unicode to escape codes so it doesn't get filtered
2. Decode "important" characters
   - If the characters in the array on line 968 are left encoded parseurl will not be able to identify any or the parts of the url)
3. Run the string through parseurl.
4. Re-encode the "important" characters in each chunk of the array
   - This puts the strings back in a consistent state where all characters are urlencoded
5. Urldecode each chunk

diananeculai:
@aaronschmitz's solution solves my problem too and it also seems to be OK with @stefanneculai. See pull requests #1285 and #1382.

libraries/joomla/form/rules/url.php
elinw, line 65: Basically missing scheme is what causes parse_url in  php in general not to work correctly ( because it will mistake the host for a string if there is no scheme) but that doesn't necessarily mean that there is nothing to parse or worth parsing.

LouisLandry:
This is great work @aaronschmitz ... really great.

infograf768:
This has broken multilang in the CMS when using unicode aliases

aaronschmitz:
I'm happy to try to resolve the issue if you can give me the url that causes the trouble.

mbabker:
Using JM's multi-lingual data set, the URL http://localhost/joomla-cms/es/inicio/19-categorías-en-español.html is
supposed to load the category with ID 19 but is loading the article with ID 19.  Reverting the changes to the JString method makes this load as expected.

EDIT: Likewise, http://localhost/joomla-cms/es/federico-garcía-lorca.html should go to article ID 57 but is attempting to load category ID 57.

infograf768:
@aaronschmitz 
I can send you the db dump if it can help

aaronschmitz:
Bug Acknowledged. I got a test database from @mbabker.

The issue is that the browser doesn't actually send http://localhost/joomla-cms/es/inicio/19-categorías-en-español.html. Instead it sends http://localhost/joomla-cms/es/inicio/19-categor%C3%ADas-en-espa%C3%B1ol.html. Before, JString::parse_url kindly converted this back to unicode.

While this was convenient, it's not really proper behavior and it broke things because JString is also used by JHTTP where we definitely don't want URL's getting converted to unicode. Likewise, we could fix this in JUri::getInstance where the URI is retrieved from $_SERVER, but to me that also seems like an incorrect solution because that's not the actual URI (true unicode URI's are illegal - hence the browser's punycode conversion).

The problem manifests itself in the router. I'm looking at how to fix it now.

elinw:
I'm seeing more general set  of routing issues in some places  in the cms trunk ... and michael what you are mentioning sounds like the  id clashing we had a long time ago in which case it's a cms issue or at least a different issue.

aaronschmitz:
Ok, @infograf768. I've put together two distinct solutions for the issue.
1. Decode URL before sending it to the router
   - https://github.com/aaronschmitz/joomla-cms/commits/routing_hack
   - I don't really like this solution, but it's easy and it essentially undoes the problems from the changes to JString.
   - No backwards-compatibility issues
   - Not 100% RFC compliant
2. Encode the URL's within the router that comparisons are performed against
   - https://github.com/aaronschmitz/joomla-cms/commits/fix_urlencoding
   - This solution is more complicated and prone to causing new bugs
   - Internationalized routers in 3rd-party components will need to make changes
   - Compliant with RFC 3986

[RFC 3986](http://tools.ietf.org/html/rfc3986) - the present standard on Uniform Resource Identifiers clearly defines the allowable characters in a URI ([Page 49](http://tools.ietf.org/html/rfc3986#page-49)). Clearly, non-ASCII characters are prohibited yet the Joomla currently outputs whatever is in the slug/alias.

Solution 1 does not resolve this problem. Solution 2 resolves this problem by replacing unicode characters with their percent encoded equivalents. Most browsers still display these encodings as the unicode character, but the URL's themselves are now RFC compliant.

Please look over these two solutions and let me know which is more desirable. If you think it needs more discussion, I can easily post it to the mailing list as well.

infograf768:
The simplest the best IMHO (routing_hack), even if we do not exactly follow exactly RFC 3986, as it is fully B/C.
The fact that ASCII unreserved characters should be presented as ASCII instead of percent encodings should be solved indeed globally. Another time I guess.

realityking:
Another vote for option No. 1.

aaronschmitz:
Here you go. As requested I moved it into the router itself.
https://github.com/joomla/joomla-cms/pull/347

PR #1460, opened 2012-08-17
Improvement for PostgreSQL database.

gpongelli:
These changes come from [2.5 CMS' Postgresql pull request](https://github.com/joomla/joomla-cms/pull/166) to let CMS works on this database.

### DIFF
--- a/libraries/joomla/session/storage/database.php
+++ b/libraries/joomla/session/storage/database.php
@@ -43,7 +43,12 @@ class JSessionStorageDatabase extends JSessionStorage
 
 			$db->setQuery($query);
 
-			return (string) $db->loadResult();
+			$result = (string) $db->loadResult();
+
+			$result = str_replace('\0\0\0', chr(0).'*'.chr(0), $result);
+			
+			return $result;
+
 		}
 		catch (Exception $e)
 		{
@@ -65,6 +70,8 @@ class JSessionStorageDatabase extends JSessionStorage
 	{
 		// Get the database connection object and verify its connected.
 		$db = JFactory::getDbo();
+
+		$data = str_replace(chr(0).'*'.chr(0), '\0\0\0', $data);
 
 		try
 		{
--- a/libraries/joomla/session/storage/database.php
+++ b/libraries/joomla/session/storage/database.php
@@ -46,9 +46,8 @@ class JSessionStorageDatabase extends JSessionStorage
 			$result = (string) $db->loadResult();
 
 			$result = str_replace('\0\0\0', chr(0).'*'.chr(0), $result);
-			
-			return $result;
 
+			return $result;
 		}
 		catch (Exception $e)
 		{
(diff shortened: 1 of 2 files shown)


libraries/joomla/session/storage/database.php
```diff
@@ -43,7 +43,12 @@ public function read($id)
 
 			$db->setQuery($query);
 
-			return (string) $db->loadResult();
+			$result = (string) $db->loadResult();
+
+			$result = str_replace('\0\0\0', chr(0).'*'.chr(0), $result);
+			
+			return $result;
+
```
realityking, line 49: Nitpick: Useless new line. Can you remove it? Thanks.

realityking:
Could you explain a bit why these changes are needed for postgres?

gpongelli:
Line removed.

About str_replace, these changes come from [here](https://github.com/gpongelli/joomla-cms/commit/9d2160e229f2c4b8df035946bc886c59f777d7e5).

realityking:
Thanks.

Maybe I should be a bit more specific, the question I have is what's different about postgres that we need those changes. I'd just like to understand so we can avoid them in the future.

realityking:
Code style error: http://developer.joomla.org/pulls/pulls/1460.html

gpongelli:
Sincerely I don't remember which was the error, try asking to @elkuku .
The only thing I remember is that without that code I wasn't able to correctly use joomla on Postgresql.

Committed code style changes.

elkuku:
Hum, let me try - first I should state that I am **no database expert** at all ;)

There are two issues here. They exist both on Postgresql and on SQLite, a driver I am playing with..

1)
When inserting new records the PK should be "NULL" not "0".

The db engine considers 0 a perfect valid integer, so it will insert a record with the primary key set to 0 instead of auto incrementing.

Mysql does not care and auto increments both on 0 and NULL
M$Sql - not tested.

Not sure how to hande this on a db specific level, so I opted for doing it in JTable.

2)
(_scratching my head_.. what was that about.....)

I believe it was about a strange NULL character PHP generates when serializing stuff. The dabase engine can not handle them and will simply truncate the strings when those characters appear.

I also was not able to porperly escape those in the db driver so I opted for str_replaceIng back and forth in JSessionStorageDatabase.

Hope that makes some sense,
Nikolai

pasamio:
Re 1:
That's a good addition, I've done something similar in the past on an SQLite project where the same behaviour exists. I think MySQL is probably the exception in letting zero values for auto-incrementing columns work like a null. MSSQL identity fields I think will have an issue with either zero or null so perhaps it'd be better to actually not include the field at all if it's empty in the insertObject (e.g. unset $k) which should work fine for all four of those platforms since it'll be assumed to be null if not specified. But we can look at that in another pull request.

Re 2:
From the PHP web site there's this:

<blockquote>
Note:
Object's private members have the class name prepended to the member name; protected members have a '*' prepended to the member name. These prepended values have null bytes on either side.
</blockquote>


Which appears to cause PostgreSQL some issues as noted in the comments as well:
http://www.php.net/manual/en/function.serialize.php#96504

I don't actually think this is enough since if the null character is the problem, serialising an object with private members will also break. The change will only work for protected members however as soon as an object with private members is put into the session it will break again. I assume that single quotes were used to prevent PHP from turning the `\0` into a literal null character by design.

eddieajau:
I'm not seeing a clear decision here. If there are still fixes to be made please issue a pull request against the CMS repo. If it proves stable, we can incorporate it into the Framework.

Thanks.

gpongelli:
As you can see, these changes are related to platform's files, so why have I to make a pull request to CMS repository ?

dongilbert:
The CMS is taking the current platform code back into it's repo to ease maintenance of the legacy platform. The platform has been upgraded to the Joomla Framework; a package based, composer integrated platform. And, since there wasn't a clear decision on what to do here, we felt it best for the CMS to handle the change as they see fit, since it will effect them more than the framework. Does that help?

PR #1463, opened 2012-08-17
Remove the weird exception handling from string.php.

aaronschmitz:
This handling can cause problems in JArrayHelper::sortObjects due to https://bugs.php.net/bug.php?id=50688.

Iconv throws a notice if there is an illegal character regardless of whether ignore is set so this handling is pointless (and in fact broken since it will always fall back to IGNORE not TRANSLIT). Also, //TRANSLIT,IGNORE not //TRANSLIT//IGNORE will properly both transliterate and ignore.

See also: https://github.com/joomla/joomla-cms/issues/179#issuecomment-7835266

### DIFF
--- a/libraries/joomla/string/string.php
+++ b/libraries/joomla/string/string.php
@@ -726,8 +726,14 @@ abstract class JString
 	{
 		if (is_string($source))
 		{
-			$iconv = @iconv($from_encoding, $to_encoding . '//TRANSLIT,IGNORE', $source);
-			return $iconv;
+			switch (ICONV_IMPL)
+			{
+				case 'glibc':
+				return @iconv($from_encoding, $to_encoding . '//TRANSLIT,IGNORE', $source);
+				case 'libiconv':
+				default:
+				return iconv($from_encoding, $to_encoding . '//IGNORE//TRANSLIT', $source);
+			}
 		}
 
 		return null;
--- a/libraries/joomla/string/string.php
+++ b/libraries/joomla/string/string.php
@@ -710,23 +710,6 @@ abstract class JString
 	}
 
 	/**
-	 * Catch an error and throw an exception.
-	 *
-	 * @param   integer  $number   Error level
-	 * @param   string   $message  Error message
-	 *
-	 * @return  void
-	 *
-	 * @link    https://bugs.php.net/bug.php?id=48147
-	 *
-	 * @throw   ErrorException
-	 */
-	private static function _iconvErrorHandler($number, $message)
-	{
-		throw new ErrorException($message, 0, $number);
-	}
-
-	/**
 	 * Transcode a string.
 	 *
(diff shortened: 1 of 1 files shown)


realityking:
Are you sure about `//TRANSLIT,IGNORE`? I can find no reference for that online.

Also according to the PHP documentation with `//IGNORE` added no error should be thrown. Is that a PP bug? If so what version are affected?

aaronschmitz:
https://bugs.php.net/bug.php?id=48147 has reports of ignore not working since at least 5.2.6. I can confirm the problem on the latest version of PHP on my system.

With regards to the TRANSLIT,IGNORE: I saw it mentioned as a solution elsewhere and it seemed to work. My test is at https://gist.github.com/2a40d736da8cf1749c60. I encourage you to run it and compare your results.

realityking:
Thanks for the link to the test script. I had to add the correct header for UTF-8 for it to work but here are my results:

//TRANSLIT,IGNORE: no output. Error message: Notice: iconv(): Wrong charset, conversion from `UTF-8' to`ASCII//TRANSLIT,IGNORE' is not allowed
//IGNORE,TRANSLIT: no output. Error message: Notice: iconv(): Wrong charset, conversion from `UTF-8' to`ASCII//TRANSLIT,IGNORE' is not allowed

//TRANSLIT//IGNORE: Correct output, no error
//IGNORE//TRANSLIT: Correct output, no error

//TRANSLIT: no output. Error message: Notice: iconv(): Detected an illegal character in input string
//IGNORE: Output is with both UTF-8 charters gone so that's correct
Plain: no output. Error message: Notice: iconv(): Detected an illegal character in input string

PHP version: 5.4.4 on OS X

Conclusion: //TRANSLIT,IGNORE and //IGNORE,TRANSLIT seem to not work at all and are not recognized at all. All others behave as expected. If the option contains //IGNORE there's no error message generated.

My hunch:

As so often the PHP documentation probably already has the information we need:

`Note that the iconv function on some systems may not work as you expect.
In such case, it'd be a good idea to install the » GNU libiconv library.
It will most likely end up with more consistent results.`

aaronschmitz:
Thanks for looking at this. I'm still concerned about this issue (my platform - Ubuntu 12.04 x64 - isn't exactly an obscure OS), but I'm going to close this pull request while I try to find another solution to this problem that provides consistent behavior across platforms. If you have any advice, I'd love to hear it.

realityking:
Apparently the main issue is libiconv vs glibc's implementation of iconv. While it's highly discouraged by PHP to create implementation specific code using that information I'm afraid we'll end up with that.

aaronschmitz:
This seems to work for both versions, but I don't like it at all.

pasamio:
Unfortunately we can't always have elegant code but so long as it works then it is a step in the right direction.

PR #1464, opened 2012-08-17
If the locale isn't available, skip the unit tests so they don't fail.

aaronschmitz:


elinw:
It would be good to add a comment about why this needed to be done or in a year we'll be thinking that turning it off means there is a problem instead of being a solution to the problem.

eddieajau:
Ditto. The description is shown in the changle log we automatically generate - just helps people be aware of what has changed.

LouisLandry:
Thanks a bunch @aaronschmitz.

PR #1472, opened 2012-08-19
Add password hashing interface and basic implementation.

ianmacl:
This will provide basic password hashing capabilities using a straightforward and easy to use interface.

Use is easy:
$hasher = new JCryptPasswordSimple();

$hash = $hasher->create('MyPasswordHere', JCryptPassword::BLOWFISH);

---

$isPasswordCorrect = $hasher->verify('MyPasswordHere', $hash);

The basic implementation supports BLOWFISH, MD5 and the legacy Joomla password format.  For more complex requirements an alternative implementation of the interface can be supplied and used.

### DIFF
--- a/libraries/joomla/crypt/password/simple.php
+++ b/libraries/joomla/crypt/password/simple.php
@@ -28,7 +28,7 @@ class JCryptPasswordSimple implements JCryptPassword
 	 * Creates a password hash
 	 *
 	 * @param   string  $password  The password to hash.
-	 * @param   string  $prefix    The prefix of the hashing function.
+	 * @param   string  $type      The hash type.
 	 *
 	 * @return  string  The hashed password.
 	 *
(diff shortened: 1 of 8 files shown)


libraries/joomla/crypt/password/simple.php
```diff
@@ -0,0 +1,151 @@
...
+				}
+				else
+				{
+					$prefix = '$2a$';
+				}
+
+				$salt = $prefix . str_pad($this->cost, 2, '0', STR_PAD_LEFT) . '$' . $this->getSalt(22);
+
+				return crypt($password, $salt);
+				break;
```
stefanneculai: I don't think this break is needed here.

libraries/joomla/crypt/password/simple.php
```diff
@@ -0,0 +1,148 @@
...
+	 * @param   string  $password  The password to hash.
+	 * @param   string  $type      The hash type.
+	 *
+	 * @return  string  The hashed password.
+	 *
+	 * @since   12.2
+	 */
+	public function create($password, $type = JCryptPassword::BLOWFISH)
+	{
+		switch ($type)
```
realityking, line 39: Why did you solve this using a switch a not trough inheritance?

realityking:
Also it'd be great if you could take a look at refactoring JUser::getCryptedPassword to use the new classes.

ianmacl:
I used a switch rather than inheritance because I wanted to keep it really simple.  This isn't meant to be a full featured do everything library - the interface allows other libraries to be used.

I plan to use this in JUser as a next step.

PR #1473, opened 2012-08-19
Fix a grammar mistake in a comment

elinw:


elinw:
It's kind of amusing to call curl stupid and make a grammar mistake while doing it.

eddieajau:
Merged. Thanks.

Issue #1427, opened 2012-08-03
JUser load 


stefanneculai:
When successfully loading a user, shouldn't guest set to false? Right now if you do <code>new JUser(id);</code> guest will be false and if you do <code>$user = new JUser; $user->load(id); </code> guest will be true.

elinw:
Well it should be false unless the id is 0.  But yes, do you want to make a pull request to fix load()?
That should really be part of the test if you want to update that. That test should be doing  more detailed checking of expected properties.

elinw:
@stefanneculai  Will you close this one please?

stefanneculai:
I thought that would close by itself when the PR is merged.

Fixed by pull request #1474.

realityking:
It does that only when the commit message is formatted in a certain way so you have to be careful if you want that behavior.

PR #1474, opened 2012-08-19
Set guest properly after load in JUser

stefanneculai:
This is a fix for issue #1427

### DIFF
--- a/libraries/joomla/user/user.php
+++ b/libraries/joomla/user/user.php
@@ -828,7 +828,7 @@ class JUser extends JObject
 
 		$this->_params->loadString($table->params);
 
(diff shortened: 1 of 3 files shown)


libraries/joomla/user/user.php
```diff
@@ -828,6 +831,16 @@ public function load($id)
 		// Assuming all is well at this point lets bind the data
```
elinw: Would you make this "let's" not "lets" please while you're updating tis file.

stefanneculai: Done.

elinw:
Also would you change the title to be descriptive?

stefanneculai:
Any thoughts about this?

PR #1481, opened 2012-08-21
Making the Joomla platform loadable as a dependency from Composer

eirslett:
Adding the file "composer.json" will allow Composer (http://getcomposer.org/) to import the Joomla Platform as a dependency.
It is now temporarily defined as a package in Packagist.org: http://packagist.org/packages/joomla/joomla-platform, pointing to my fork, which can be changed if this pull request is accepted.

### DIFF
--- a/libraries/joomla/http/transport/stream.php
+++ b/libraries/joomla/http/transport/stream.php
@@ -123,7 +123,13 @@ class JHttpTransportStream implements JHttpTransport
 		$context = stream_context_create(array('http' => $options));
 
 		// Open the stream for reading.
-		$stream = fopen((string) $uri, 'r', false, $context);
+		$stream = @fopen((string) $uri, 'r', false, $context);
+
+		// Check if the stream is open.
+		if (!$stream)
+		{
+			throw new RuntimeException(sprintf('Could not connect to resource: %s', $uri));
+		}
 
 		// Get the metadata for the stream, including response headers.
 		$metadata = stream_get_meta_data($stream);
@@ -181,9 +187,9 @@ class JHttpTransportStream implements JHttpTransport
 
 	/**
 	 * method to check if http transport stream available for using
-	 * 
+	 *
 	 * @return bool true if available else false
-	 * 
+	 *
 	 * @since   12.1
 	 */
 	static public function isSupported()


LouisLandry:
Cool.  You should change the PHP requirement to 5.3.1+, and also would you mind squashing those two commits into one?

AmyStephen:
Really great to see this. Wondering if it makes sense that the Joomla project be the entity that manages the composer package -- and do so as a normal step of the official release process? IIRC, the composer instructions state it is preferred (if not required?) that project maintainers submit/maintain these packages.   

Part of the benefit to this could be that as the separation process is more complete, the project might see start packaging individual components and perhaps even comply with PSR-0, at least for the individual packages. Right now, I do those steps manually so that I can use Joomla platform components, along with other software, and comply with those standards. 

Glad to see this - it's a good way to share some great code with the broader PHP community.

eirslett:
Ok, I changed the PHP version. I tried to squash the commits, but ended up ruining the commit history, sorry. :-(
@AmyStephen: Of course, the Joomla project will manage the packagist package eventually, when the pull request is merged. I submitted the fork to packagist as a proof-of-concept. But there's not really much to manage - when the Joomla Platform git repository is tagged with a new version, packagist interprets that as a release of the package - and that's it.

realityking:
Are you a cli user? Try git rebase -i upstream/staging and set all the commits but the first one to fix.

eirslett:
Oops, I pressed the wrong buttons I guess :S I'll have to try this once more!

AmyStephen:
@eirslett Don't get me wrong, I am happy to see this but I do believe it is best if this should is set up on the packagist.org website by someone acting in an official role for the project who then has control of the account. I understand that to be strongly encouraged by packagist, if not required. 

Part of the problem is, I don't believe there is any way to change the package name once it has started being distributed (so the Joomla project would not be able to use "joomla-platform" to handle this after your account first did so) and to the best of my knowledge - transfering ownership of a package is not possible. 

Agree to get it going, but the right way, which would be following the same approach used by any web service used by the project - that it be established by someone acting in an official role, much like how the github resource is managed, and for that matter, the Twitter and FB accounts, etc.

Thanks for getting this started! I know it sounds silly but best to do it right from the start than to lose options later.

eirslett:
Changing ownership of a package is no problem! If Joomla has a packagist account, I add that account as maintainer for the package and then remove myself. And if not, I can change my Packagist username (it's my only uploaded package) and e-mail address to yours.

AmyStephen:
Perfect. Thanks @eirslett. I think this means of distribution will prove to
be very useful.

PR #1483, opened 2012-08-21
Making the Joomla platform loadable as a dependency from Composer (2)

eirslett:
See this discussion:
https://github.com/joomla/joomla-platform/pull/1481#issuecomment-7914023

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,11 @@
+{
+    "name": "joomla/joomla-platform",
+    "type": "library",
+    "description": "The Joomla Platform is a platform for writing Web and command line applications in PHP.",
+    "keywords": ["joomla","platform"],
+    "homepage": "https://github.com/joomla/joomla-platform",
+    "license": "GNU Public License",
+    "require": {
+        "php": ">=5.3.1"
+    }
+}


ianmacl:
Please rebase and squash and then we'll get this merged in.

eirslett:
Ok, they're squashed now.

dongilbert:
When loading in the Joomla Platform via composer, what would get loaded? Just the libraries/joomla directory? or would the whole repo including the import.php, legacy, etc be loaded as well and all placed into the vendors/joomla-platform folder?

eirslett:
What composer does is clone the repository that is defined by the package at packagist (packagist does not store any actual code, it just stores metadata about the package - i.e. the location of the git repository). So when you require "joomla/joomla-platform" in composer.json, composer will clone the whole repository and put the root in "vendor/joomla/joomla-platform". It will even include the .git folder. That is happening to all composer packages, not just this joomla one. If we don't want that, we can do 2 things:
1) Make a separate git repository (a mirror) for the composer package, where the git root is in library/joomla. Or one could even divide each separate library into its own package, e.g. package names "joomla/database", "joomla/filesystem" etc., which is the approach taken by the Symfony framework, but that's a lot more complex.
2) Getting involved with the development of composer/packagist, extending their functionality so only a subdirectory of a repository is checked out

A third option: Just let it be like that for now. Eventually the Joomla platform will follow the psr-0 standard for autoloading (I guess), in which case that can be stated in composer.json. Then, the developer includes "vendor/autoload.php", which is auto-generated by composer, and the contents of vendor/joomla/joomla-platform will be unimportant. Except for the size of the repository of course (30 MB), but realistically that shouldn't be a problem for an average web server or shared hosting account.

Actually, having the whole repository in vendor/ might be a good thing. For example if you are using Joomla's database classes, you might find the mocks in tests/core/mock useful?

PR #1486, opened 2012-08-23
Validate that the key parameter is a string

pasamio:
This pull request validates that the `$key` parameter provided to the function is actually a string before trying to use this. I came across this as I was working to modify JTable to be able to support multiple primary keys in a somewhat transparent way. By default JTable passes through it's keys onto the database `insertObject` and `updateObject` function which works well for single keys but not properly for `insertObject` when the `$key` is an array.

### DIFF
--- a/libraries/joomla/database/driver.php
+++ b/libraries/joomla/database/driver.php
@@ -851,7 +851,7 @@ abstract class JDatabaseDriver extends JDatabase implements JDatabaseInterface
 
 		// Update the primary key if it exists.
 		$id = $this->insertid();
-		if ($key && $id)
+		if ($key && $id && is_string($key))
 		{
 			$object->$key = $id;
 		}


libraries/joomla/database/driver.php
```diff
@@ -851,7 +851,7 @@ public function insertObject($table, &$object, $key = null)
 
 		// Update the primary key if it exists.
 		$id = $this->insertid();
-		if ($key && $id)
+		if ($key && $id && is_string($key))
```
stefanneculai, line 854: I think that it would be enough to do  <code>if (is_string($key) && $id)</code>. If the condition fails for $key it will fail for is_string($key) too.

HermanPeeren, line 854: But if $key would not exist you would then get a **Notice: Undefined variable: key**
That's why $key is tested first and cannot be left.

stefanneculai, line 854: Yes, but $key is by default null. I can't see any case when $key is not defined in this context.

HermanPeeren, line 854: You are right (I was wrong).

PR #1488, opened 2012-08-24
onGetGroupsByUser; request for new event trigger

mdobrinic:
Request to create a new event trigger, to be able to write a plugin that can insert groups from a custom source.

### DIFF
--- a/libraries/joomla/access/access.php
+++ b/libraries/joomla/access/access.php
@@ -347,12 +347,12 @@ class JAccess
 				// Add trigger to allow for external group assignment
 				// nesting prevention:
 				if (!self::$loaded) {
-					self::$loaded=true;
+					self::$loaded = true;
 					JPluginHelper::importPlugin('user');
 				}
-				$dispatcher = JDispatcher::getInstance(); 
+				$dispatcher = JEventDispatcher::getInstance();
 				$r = $dispatcher->trigger('onGetGroupsByUser', array($userId, $recursive));
-        
+
 				if (! empty($r)) {
 					// Collect result from all fired plugins
 					foreach ($r as $plugin_r) {
--- a/libraries/joomla/access/access.php
+++ b/libraries/joomla/access/access.php
@@ -346,7 +346,8 @@ class JAccess
         
 				// Add trigger to allow for external group assignment
 				// nesting prevention:
-				if (!self::$loaded) {
+				if (!self::$loaded) 
+        {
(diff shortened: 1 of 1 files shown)


libraries/joomla/access/access.php
```diff
@@ -333,7 +340,27 @@ public static function getGroupsByUser($userId, $recursive = true)
...
 				$result = $db->loadColumn();
 
-				// Clean up any NULL or duplicate values, just in case
+        // Execute the query and load the rules from the result.
+        $db->setQuery($query);
+        $result = $db->loadColumn();
+        
+				// Add trigger to allow for external group assignment
+				// nesting prevention:
+				if (!self::$loaded) {
```
elinw: We put opening curly brackets on new lines, can you please fix this here and elsewhere?

elinw:
http://developer.joomla.org/pulls/pulls/1488.html 
You've got a failing unit test and some code style errors (don't worry the first 129 are not yours).

libraries/joomla/access/access.php
```diff
@@ -333,7 +340,27 @@ public static function getGroupsByUser($userId, $recursive = true)
...
+					self::$loaded=true;
+					JPluginHelper::importPlugin('user');
+				}
+				$dispatcher = JDispatcher::getInstance();
```
elinw: JEventDispatcher

elinw:
Would you close ths one please since it s the same as the other?

LouisLandry:
This is the same as what other one?

elinw:
#1494 but that's closed instead.

Isn't this repeated? 
https://github.com/mdobrinic/joomla-platform/blob/0ed0fd2442d66a2e5e110114fdba29c754605bbd/libraries/joomla/access/access.php#L340

I really like the idea of being able to manipulate what groups a user is part of more easily and based on context.  For example that would have let us possibly handle guest differently or let us dynamically create a single-user group for each user.  But I'm not sure that I would do it this way because I'd want to know that I'm not going to end up with group id conflicts without a plan for handling them (ignore or replace? throw an exception?). I'm also not sure if that should be happening in JAcccess or JUser.
Maybe there should be a method for adding a group that could manage some of these things.

eddieajau:
Apart from a few code-style issues, I don't like the way the trigger is done.  I'm going to close this PR and suggest you raise the issue on the mailing list to get a consensus on how to solve your problem.

PR #1490, opened 2012-08-25
Fix IS_MAC operating system check

dongilbert:
On my OS X 10.8 machine, when I run the code `strtoupper(substr(PHP_OS, 0, 3))`, I get the result of 'DAR', which is short for Darwin. Under the previous revision, the IS_UNIX would be defined as true, and IS_MAC is defined as false, even though I am running the check on a Mac. 

In the platform docs, it says IS_UNIX is only set to true if it is a flavor of UNIX or Linux, but not Mac. This is not true, as Darwin IS Mac, but is labeled as IS_UNIX using the existing code. This fixes that.

### DIFF
--- a/libraries/import.php
+++ b/libraries/import.php
@@ -20,11 +20,11 @@ if (!defined('IS_WIN'))
 }
 if (!defined('IS_MAC'))
 {
-	define('IS_MAC', ($os === 'MAC') ? true : false);
+	define('IS_MAC', ($os === 'MAC' || $os === 'DAR') ? true : false);
 }
 if (!defined('IS_UNIX'))
 {
-	define('IS_UNIX', (($os !== 'MAC') && ($os !== 'WIN')) ? true : false);
+	define('IS_UNIX', (($os !== 'MAC' || $os !== 'DAR') && ($os !== 'WIN')) ? true : false);
 }
 
 // Import the platform version library if necessary.


libraries/import.php
```diff
@@ -20,11 +20,11 @@
 }
 if (!defined('IS_MAC'))
 {
-	define('IS_MAC', ($os === 'MAC') ? true : false);
+	define('IS_MAC', ($os === 'MAC' || $os === 'DAR') ? true : false);
 }
 if (!defined('IS_UNIX'))
 {
-	define('IS_UNIX', (($os !== 'MAC') && ($os !== 'WIN')) ? true : false);
+	define('IS_UNIX', (($os !== 'MAC' || $os !== 'DAR') && ($os !== 'WIN')) ? true : false);
```
pasamio, line 27: Why not an && here?

dongilbert, line 27: True. However, after thinking about it some more, we don't need to run the checks on $os again, we should just compare against IS_MAC or IS_WIN.

Something like:

define('IS_UNIX', (IS_MAC === false && IS_WIN === false));

mbabker:
Would you mind updating import.legacy.php as well, since it has the same check?

dongilbert:
New pull request here - https://github.com/joomla/joomla-platform/pull/1491

PR #1491, opened 2012-08-26
Fix IS_MAC operating system check

dongilbert:
This patch fixes the IS_MAC os check in the import.php and the import.legacy.php file.

It also improves the IS_UNIX check. Instead of checking the $os variable again, we just compare the the IS_WIN and IS_MAC constants.

### DIFF
--- a/libraries/import.legacy.php
+++ b/libraries/import.legacy.php
@@ -18,13 +18,17 @@ if (!defined('IS_WIN'))
 {
 	define('IS_WIN', ($os === 'WIN') ? true : false);
 }
-if (!defined('IS_MAC'))
-{
-	define('IS_MAC', ($os === 'MAC') ? true : false);
-}
 if (!defined('IS_UNIX'))
 {
-	define('IS_UNIX', (($os !== 'MAC') && ($os !== 'WIN')) ? true : false);
+	define('IS_UNIX', (IS_WIN === false) ? true : false);
+}
+
+/**
+ * @deprecated	13.3	Use IS_UNIX instead
+ */
+if (!defined('IS_MAC'))
+{
+	define('IS_MAC', (IS_UNIX === true && ($os === 'DAR' || $os === 'MAC')) ? true : false);
 }
 
 // Import the platform version library if necessary.
--- a/libraries/import.legacy.php
+++ b/libraries/import.legacy.php
@@ -20,11 +20,11 @@ if (!defined('IS_WIN'))
 }
 if (!defined('IS_MAC'))
 {
-	define('IS_MAC', ($os === 'MAC') ? true : false);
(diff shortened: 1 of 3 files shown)


ianmacl:
Shouldn't IS_UNIX be true if IS_MAC is true?

dongilbert:
See the docs here - http://developer.joomla.org/manual/ch01s02.html

"IS_UNIX - A bollean value, true is the platform is some flavor or Unix, Linux or similar (but not Mac)."

ianmacl:
That's...  odd.  Considering OS X is pretty much the only certified UNIX that people run Joomla on.

realityking:
Picking up @ianmacl's point. For what do we even need the IS_MAC constant? Can we remove it and just use IS_UNIX on OS X?

We only use it in JClientFTP to decide for the line ending, since OS X also uses LF we probably do it wrong over there anyways.

dongilbert:
Just goes to show you never know the extent of a single pull request. I would say drop the IS_MAC constant in favor of IS_UNIX altogether, since Unix and OS X both use LF for line endings.

Do you think anyone would be using IS_MAC in their extensions that this would still need to be set though?

realityking:
We should probably keep it around in the import.legacy.php marked with an @deprecated for backwards compatability. (and set both IS_UNIX and IS_MAC to true on OS X).

Would you do these changes?

dongilbert:
We should be good to go then with these added commits. I've removed IS_MAC altogether from import.php, and deprecated it in import.legacy.php.

realityking:
Not quite, you need to remove the use of IS_MAC in JClientFtp.

libraries/import.legacy.php
```diff
@@ -18,13 +18,17 @@
...
-if (!defined('IS_MAC'))
+if (!defined('IS_UNIX'))
 {
-	define('IS_MAC', ($os === 'MAC') ? true : false);
+	define('IS_UNIX', (IS_WIN === false) ? true : false);
 }
-if (!defined('IS_UNIX'))
+
+/**
+ * @deprecated	12.3	Use IS_UNIX instead
```
realityking: Please make this 13.3.

realityking:
Could you squash all these commits into one as well? Thanks.

dongilbert:
OK - commits squashed. Had some issues rebasing and then pushing, since I had already pushed the previous commits, but we're all good now.

realityking:
Uh not quite. The old commits are still there. Try a forced push (git push -f on the CLI) after squashing.

dongilbert:
Git not cooperating, or operator error - I don't know. But I created a new branch, remade the changes, and created a new pull request from my ismacfix branch. Much cleaner diff's

https://github.com/joomla/joomla-platform/pull/1498

PR #1492, opened 2012-08-26
Populate the session table with the minimum of data if the record doesn't yet exist.

realityking:
This should make small applications a bit easier since they don't have to create the session record themselves.

We should look into using the merge/replace into syntax on supported database drivers to save the count query. If someone could do the database abstraction for this that'd be great.


realityking:
Closing because there's some bug in here.

PR #1494, opened 2012-08-27
Add onGetGroupsByUser event trigger

mdobrinic:
code style updated; JEventDispatcher used instead of JDispatcher

### DIFF
--- a/libraries/joomla/access/access.php
+++ b/libraries/joomla/access/access.php
@@ -347,12 +347,12 @@ class JAccess
 				// Add trigger to allow for external group assignment
 				// nesting prevention:
 				if (!self::$loaded) {
-					self::$loaded=true;
+					self::$loaded = true;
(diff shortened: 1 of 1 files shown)


elinw:
http://developer.joomla.org/pulls/pulls/1494.html

You have a number of code style errors and a unit test failure. 
I'm not sure what the UT failure is about but it is probably the messy intertwining of user and session.
We use tabs and not space except inside docblocks. 

Wait are these two the same?

elinw:
If this one and the other one are the same will you close whichever one is not the one you want?

eddieajau:
I'm going to close this one mainly because I have no idea what the problem is you are trying to solve.

mdobrinic:
I'll look into replying before the end of the week;

I remember opening the ticket as a feature request, that I used to
establish group memberships from an external source.

Cheers!

Mark

PR #1502, opened 2012-08-29
copy language and media folder on discover

AndOne1974:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=29018


elinw:
Would you please give this a meaningful title and description?

AndOne1974:
done ;)

elinw:
Somehow you have made this pull request against master rather than staging. You can redirect it by editing the pull request target I believe.

Also, in discover install the files are supposed to be in the right place already so you wouldn't need to copy. For example this happens when we update the core an install a new extension.  I think in your use case you might want to use install from directory instead.

AndOne1974:
thx elinw, i will find another way (xml install with all packages with the file tag) of installing my component, modules and plugin with one step. pls close the pull request.

pasamio:
Discover was primarily intended for developers and development where all of the files would be in place and to just update the database to match what the file system had. This way you could extract out an extension into the right places for development and then have the database automatically populated instead of before where you would have to install the extension to create everything in the database and then re-write everything on the filesystem to fit into a developer flow.

PR #1503, opened 2012-08-29
Bring some sanity to JComponentHelper::isEnabled

realityking:
JComponentHelper::isEnabled is a rather confusing piece of API, this should let it behave as expected.

Please don't pull this until we have buy-in from the CMS since this a B/C break and a CMS specific piece of API!

### DIFF
--- a/libraries/legacy/component/helper.php
+++ b/libraries/legacy/component/helper.php
@@ -62,18 +62,17 @@ class JComponentHelper
 	/**
 	 * Checks if the component is enabled
 	 *
-	 * @param   string   $option  The component option.
-	 * @param   boolean  $strict  If set and the component does not exist, false will be returned.
+	 * @param   string  $option  The component option.
 	 *
 	 * @return  boolean
 	 *
 	 * @since   11.1
 	 */
-	public static function isEnabled($option, $strict = false)
+	public static function isEnabled($option)
 	{
-		$result = self::getComponent($option, $strict);
+		$result = self::getComponent($option, true);
 
-		return ($result->enabled | JFactory::getApplication()->isAdmin());
+		return $result->enabled;
 	}
 
 	/**


dongilbert:
You left an extra closing parenthesis on line 75

realityking:
Thanks, fixed.

libraries/legacy/component/helper.php
```diff
@@ -62,18 +62,17 @@ public static function getComponent($option, $strict = false)
...
-	 * @param   string   $option  The component option.
-	 * @param   boolean  $strict  If set and the component does not exist, false will be returned.
+	 * @param   string  $option  The component option.
 	 *
 	 * @return  boolean
 	 *
 	 * @since   11.1
 	 */
-	public static function isEnabled($option, $strict = false)
+	public static function isEnabled($option,)
```
LouisLandry: `,`

realityking: Thanks, fixed.

libraries/legacy/component/helper.php
```diff
@@ -62,18 +62,17 @@ public static function getComponent($option, $strict = false)
...
 	 *
 	 * @return  boolean
 	 *
 	 * @since   11.1
 	 */
-	public static function isEnabled($option, $strict = false)
+	public static function isEnabled($option)
 	{
-		$result = self::getComponent($option, $strict);
+		$result = self::getComponent($option, true);
```
mbabker, line 73: Another ,

realityking, line 73: I'm missing something. What's wrong?

mbabker, line 73: Oh, my bad.  It looked like everything was deleted after the comma on this line the first time I looked at this.  I really should get more than 4-5 hours of sleep at night...

realityking:
Response from the CMS wasn't very large but positive so this can go ahead.

PR #1509, opened 2012-09-01
fix JViewLegacy::getName model name discovery

dongilbert:
The current regex implementation for model name auto-discovery in the JViewLegacy prevents a developer from the use of the substr of "view" in the classname without consequences. Hence the reason for the custom error created just for that case. (see line 466 of the diff: https://github.com/joomla/joomla-platform/pull/1509/files#L0L466). This pull request fixes that.

Note: This fix breaks Backwards Compatibility (potentially) as the previous regex implementation used the /i modifier and searched for case insensitive strings. Here we use strpos (case-sensitive). Be sure your classnames are properly CamelCased (http://docs.joomla.org/Coding_style_and_standards#Classes)

This is how I tested - https://gist.github.com/3563618

### DIFF
--- a/libraries/legacy/view/legacy.php
+++ b/libraries/legacy/view/legacy.php
@@ -456,16 +456,15 @@ class JViewLegacy extends JObject
 	{
 		if (empty($this->_name))
 		{
-			$r = null;
-			if (!preg_match('/View((view)*(.*(view)?.*))$/i', get_class($this), $r))
+			$classname = get_class($this);
+			$viewpos = strpos($classname, 'View');
+			
+			if ($viewpos === false)
 			{
 				throw new Exception(JText::_('JLIB_APPLICATION_ERROR_VIEW_GET_NAME'), 500);
 			}
-			if (strpos($r[3], "view"))
-			{
-				JLog::add(JText::_('JLIB_APPLICATION_ERROR_VIEW_GET_NAME_SUBSTRING'), JLog::WARNING, 'jerror');
-			}
-			$this->_name = strtolower($r[3]);
+			
+			$this->_name = strtolower(substr($classname, $viewpos+4));
 		}
 
 		return $this->_name;
--- a/libraries/legacy/view/legacy.php
+++ b/libraries/legacy/view/legacy.php
@@ -456,16 +456,15 @@ class JViewLegacy extends JObject
 	{
 		if (empty($this->_name))
 		{
-			$r = null;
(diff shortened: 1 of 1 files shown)


elinw:
Thanks!

dongilbert:
No problem. Were you able to test? I implemented the change into a test site and saw no abnormalities. Everything continued to function as normal.

elinw:
I didn't test it directly yet but the pull tester says you have a few code style errors but it's passing all the unit tests.
http://developer.joomla.org/pulls/pulls/1509.html

pasamio:
For compatibility it should probably be stripos as you will note the earlier regexp has a /i modifier clause.

dongilbert:
@pasamio I had noticed that and was going to use stripos instead, but then I thought, what if a developer built a previews component (or something else that had view in the component name). PreviewsViewVideo would set the $_name property to 'sviewvideo', and that's not what would be expected (obviously). So I made it strpos instead. With the same view name, it would correctly set $_name to 'video' instead. What do you think? 

I think it would be easier to ensure that end-users correctly StudlyCapped the classnames then to limit the use of certain lowercase letter sequences in the component name. I've been running this on a local install, and haven't had any issues with the default components. (I know, it doesn't mean there won't be ANY issues) but I felt this would probably be the way to go.

pasamio:
The current code doesn't let them call it PreviewsViewVideo already (or Review) so we're fine there for the moment. We just have to note down that this could potentially cause BC issues though again it also shouldn't be a problem.

pasamio:
BTW can you look through the style issues that Elin linked to and work those out and it should be fine to merge.

dongilbert:
Ya, I checked the style issues. It was 4 things directly related to my code. Two lines with white space, and missing spaces around the $viewpos+4 the other 100+ were to do with variable naming and such. I have those fixed, I was just waiting to see the outcome of this discussion before pushing. 

So, should I change it to stripos instead then to avoid potential bc issues?

Sent from my iPhone

On Sep 2, 2012, at 6:13 PM, Sam Moffatt notifications@github.com wrote:

BTW can you look through the style issues that Elin linked to and work those out and it should be fine to merge.

pasamio:
Nah, leave it as strpos, we'll put it in the release notes that it's BC incompatible.

eddieajau:
@dongilbert Can you edit the pull description and make a note of the BC incompatibility (automagically gets put in the changelog). Thanks.

dongilbert:
description updated and style issues fixed

eddieajau:
Merged. Thanks.

PR #1510, opened 2012-09-01
Correct docblocks in mysqli and pdo.

stefanneculai:
As PHP documentations says mysqli and PDO do not return integers all the time.
- <code>mysqli::$insert_id</code>: The value of the AUTO_INCREMENT field that was updated by the previous query. Returns zero if there was no previous query on the connection or if the query did not update an AUTO_INCREMENT value. <b>Note: If the number is greater than maximal int value, mysqli_insert_id() will return a string.</b>
- <code>PDO::lastInsertId</code>: If a sequence name was not specified for the name parameter, PDO::lastInsertId() returns a string representing the row ID of the last row that was inserted into the database. If a sequence name was specified for the name parameter, PDO::lastInsertId() returns a string representing the last value retrieved from the specified sequence object.

I have checked and insertid from the other database drivers and they return integers.

### DIFF
--- a/libraries/joomla/database/driver.php
+++ b/libraries/joomla/database/driver.php
@@ -793,7 +793,7 @@ abstract class JDatabaseDriver extends JDatabase implements JDatabaseInterface
 	/**
 	 * Method to get the auto-incremented value from the last INSERT statement.
 	 *
-	 * @return  integer  The value of the auto-increment field from the last inserted row.
+	 * @return  mixed  The value of the auto-increment field from the last inserted row.
 	 *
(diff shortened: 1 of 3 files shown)


stefanneculai:
I really wonder how much it takes to comment or merge such a simple pull request - one month so far.

elinw:
Patience is a good thing :). 
But I think the title makes it sound more confusing than it needs to be. Can you make it "Correct docblocks in mysqli and pdo."

LouisLandry:
It takes two clicks to merge it ... not too much.

PR #1511, opened 2012-09-02
Improved debug check, added file handling to else portion of statement

mbabker:


mbabker:
Thinking a couple things out on this one...

PR #1516, opened 2012-09-04
Have JViewLegacy extend JView.

realityking:
This contains a small backwards compatibility break because the `display` method looses it `$tpl` argument.

### DIFF
--- a/libraries/legacy/view/legacy.php
+++ b/libraries/legacy/view/legacy.php
@@ -18,7 +18,7 @@ defined('JPATH_PLATFORM') or die;
  * @subpackage  View
  * @since       12.2
  */
-class JViewLegacy extends JObject
+class JViewLegacy extends JObject implements JView
 {
 	/**
 	 * The name of the view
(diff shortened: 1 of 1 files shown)


LouisLandry:
I think it's actually very important that we keep the `$tpl` argument to the display method.  It is used quite a bit by several extensions that I have worked with over the years, and I feel that if this change is going to be made it needs to be communicated and people given some time to adjust.

realityking:
For reference here's the discussion on the CMS list: https://groups.google.com/forum/?fromgroups=#!topic/joomla-dev-cms/D8FiWlqataA

I was probably a bit quick in submitting this. I'd be curios how you have used it? From a quick grep it looks like we don't use it at all in the CMS.

HermanPeeren:
If I leave $tpl=null as argument I get no errors. It looks like a dirty (but working) way of overloading the display method's arguments. Also then calling display() with a value for $tpl worked.

And another dirty implementation to trick the display-method into an optional argument, although it is not in the interface:

function display (){
        $a = func_get_args();
        if (count($a))
        {
            $tpl=$a[0];
        }

mahagr:
Generally in OO it's perfectly legal to add new parameters to the inherited functions -- as long as the function can be used in the same way as its parent. So I would leave the parameter as well as I do know that it's used by a few components.

realityking:
Of course we can lave the parameter - but what's the point when we don't actually use it internally. If anything we'd have to add the parameter to the render method as well.

elinw:
Huge amounts of use of the Joomla CMS do not involve the core application but rather extensions to the application. There are many many things that are not used internally (why ship JImage?) but they are there for people extending the CMS to use. We strongly encourage extension developers to use the API rather than roll their own whenever possible.  These involve not just retail extensions but also tens of thousands of custom uses.

PR #1517, opened 2012-09-06
Add error handling for __get and __call methods.

realityking:
Adding a __get or __call method may mask errors about unexisting method or properties if we don't do our own error handling. This pull should provide appropriate errors for all of these.

Note: The test results will show one error in JTableNested that I suspect is an actual bug but I'm not sure how to fix.

### DIFF
--- a/libraries/joomla/application/web/client.php
+++ b/libraries/joomla/application/web/client.php
@@ -224,6 +224,15 @@ class JApplicationWebClient
 					$this->detectRobot($this->userAgent);
 				}
 				break;
+
+			default:
+				// In dynamic methods we have to do the error handling ourself.
+				$trace = debug_backtrace();
+				trigger_error(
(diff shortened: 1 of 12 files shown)


pasamio:
I'm guessing it should be $this->_unlock() instead since that is a valid method of JTable.

I found where it was added to the file here:
https://github.com/joomla/joomla-platform/commit/017f8f78629ec386b02194ac9838c81bb4cf1856#L26L367

Even at that commit, JDatabaseQuery doesn't support lock and for some reason it's having a database object fed into it as well which seems a bit strange. Given almost every other occurrence in the file has been removed I think swapping this back to $this->_unlock() is probably the correct change.

pasamio:
As an aside, I like this change. I've been burnt a few times by calling the wrong method somewhere and getting a null back when I expected it to be something else.

LouisLandry:
@realityking love the direction on this.  Can you get it cleaned up and rebased so we can merge it?  Do we still have an issue figuring out the problem with the unit test failure?  Please re-open it when it is cleaned up.  Thanks!

PR #1521, opened 2012-09-09
Correct misleading docblock and log message

elinw:


LouisLandry:
_shrug_

elinw:
What can I say, it came up in the tracker :P

PR #1529, opened 2012-09-14
Notice error due undefined property

Kubik-Rubik:
Avoid error notice which is described here:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29168


Kubik-Rubik:
See this pull request: https://github.com/joomla/joomla-platform/pull/1530

PR #1531, opened 2012-09-15
Fix potentially broken code in JPagination

mahagr:
Potential endless loop by setting $i=0 inside for statement. It's though much more likely that the code has no effect at all (I think it's always setting $i = $i), but even then it's probably a typo and will cause issues if the if statement gets changed at some point.

See also:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29197

### DIFF
--- a/libraries/joomla/pagination/pagination.php
+++ b/libraries/joomla/pagination/pagination.php
@@ -707,9 +707,6 @@ class JPagination
 		{
 			$offset = ($i - 1) * $this->limit;
 
-			// Set the empty for removal from route
-			// @todo remove code: $offset = $offset == 0 ? '' : $offset;
-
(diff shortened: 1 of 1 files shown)


mahagr:
Oh, forgot to mention that (else)if ($i == $this->pagesCurrent) can be removed without changing meaning of the code:

```
if ($i != $this->pagesCurrent || $this->viewall) { ... }
elseif ($i == $this->pagesCurrent) { ... }
```

mahagr:
Removed TODO comment to remove &limitstart=0 from URL. Almost a year ago I had the same idea in my own JPagination override, but I quickly learned that it's not going to work.

Reason is that the backend keeps current page in the state and URLs without &limitstart=0 have different meaning -- they are pointing to the last page visited by the user and not the first page.

PR #1532, opened 2012-09-15
Fix performance issues in JModelLegacy::_getListCount()

mahagr:
Current code in JModelLegacy::_getListCount() forces database engine to fetch all items instead of just counting them. With hundreds of thousands or millions of rows the query gets very inefficient even if no results are actually fetched by the function.

Cloning and modifying JDatabaseQuery object into COUNT(*) allows database engine to do optimizations that will considerably speed up the code even with very simple queries like:

```
SELECT * FROM jos_users
```

if there are hundreds of thousands of users.


libraries/legacy/model/legacy.php
```diff
@@ -313,10 +313,25 @@ protected function _getList($query, $limitstart = 0, $limit = 0)
...
 	{
-		$this->_db->setQuery($query);
-		$this->_db->execute();
+		if ($query instanceof JDatabaseQuery)
+		{
+			// Create COUNT(*) query to allow database engine to optimize the query.
+			$query = clone $query;
+			$query->clear('select')->clear('order')->select('COUNT(*)');
+			$this->_db->setQuery($query);
+
```
pasamio: Can you remove the white space on this line?

mahagr:
Sorry, project settings were wrong. Now it should be better.

pasamio:
Sweet! Nice addition!

PR #1534, opened 2012-09-16
[IMP] Ensure that publish doesnt receive empty ids array

phproberto:
Trying to disable the default item in Joomla CMS finishes in a JModelAdmin::publish with an empty ids array.

http://awesomescreenshot.com/09ag6vbe3

I'm going also to fix CMS to ensure that we do not call this method with an empty array. 

Joomlacode tracker issue:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28990


pjwiseman:
Tested successfully by applying 1534 to the CMS which fixes the issue in 28990.

PR #1538, opened 2012-09-18
JMediawiki

nprasath002:
JMediawiki package for joomla platform


pasamio:
First of all this pull request can't be merged so needs to be rebased.
Second can you need to rebase this and squash some commits. It doesn't add value to have more than 10 commits in a row that are "codesniff cleanup".

Why is this changing GitHub related functionality for a MediaWiki change? There seems to be a lot of changes in here unrelated to MediaWiki.

pasamio:
This might be helpful:
http://git-scm.com/book/en/Git-Tools-Rewriting-History

I suggest if you are new creating a new copy of the branch you're working on and trying the rebase there first or alternatively create a new copy of the branch and push it up to GitHub then rebase this branch. That way you still have all of your original commits in the order they were created safely.

PR #1549, opened 2012-09-26
MySQLi, JHTML::tabs and JHTML::getJSObject fixes

obsidev:
This pull request is link to two issues

Javascript Error with JHTML tabs

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=29344

MySQLi does not display SQL errors messages anymore.

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=29345

Regards,
Jerome @ Obsidev

PR #1551, opened 2012-09-27
MySQLi display error message fix

obsidev:
Store error message and error number before calling mysqli_ping.


obsidev:
The separate pull request from https://github.com/joomla/joomla-platform/pull/1549

pjwiseman:
This was replaced by pull #1554.

PR #1553, opened 2012-09-27
JHTML Tabs and JHTML getJSObject fixes

obsidev:
With JHTML Tabs, the function getJSObject does not trim quote properly.
This fix would remove and strip single quotes.

Original issue :
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=29344

First pull request (with other different issues)
https://github.com/joomla/joomla-platform/pull/1549

(sorry, there is the other patch for mysql, but I don't find where exclude this file from my pull request in GitHub)

### DIFF
--- a/libraries/joomla/database/driver/mysqli.php
+++ b/libraries/joomla/database/driver/mysqli.php
@@ -503,6 +503,9 @@ class JDatabaseDriverMysqli extends JDatabaseDriver
 		// If an error occurred handle it.
 		if (!$this->cursor)
 		{
+			$this->errorNum = (int) mysqli_errno($this->connection);
+			$this->errorMsg = (string) mysqli_error($this->connection) . ' SQL=' . $sql;
+
 			// Check if the server was disconnected.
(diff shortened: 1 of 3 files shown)


elinw:
The tabs and html parts have been added in the CMS  version of the platform ... could you take the mysqli work out and move to a separate pull request?

What you need to do is to open a separate branch for the mysqli.   Since it is all in one file you could even just use the web UI for it -- go to the file, click edit and then github will automatically make a new branch for you to save to and then prompt you for a pull request.

obsidev:
Hi,
I did a separate pull request with just mysqli here : https://github.com/joomla/joomla-platform/pull/1551
I didn't found how create a branch directly in github, I have to install a git client on my current computer to create a branch and make a clean separate pull request for the tabs & jhtml.

pjwiseman:
This was replaced by pull 1555.

PR #1554, opened 2012-09-27
Fixing mysqli error message

obsidev:
Store error message and error number before calling mysqli_ping (which would clean errors)


pjwiseman:
CMS tracker: 
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=29345

PR #1556, opened 2012-09-27
Add JArrayHelper::invert

pasamio:
This adds a new method to JArrayHelper called 'invert'. The 'invert' method provides the ability to provide an associative array of arrays and then make the values in those arrays the keys with the key from the input associative array as the value for those entries.

So if we use the following code:

``` php
$input = array(
    'New' => array('1000', '1500', '1750'),
    'Used' => array('3000', '4000', '5000', '6000')
);
$output = JArrayHelper::invert($input);
```

The result for output would be equal to:

``` php
$output = array(
    '1000' => 'New',
    '1500' => 'New',
    '1750' => 'New',
    '3000' => 'Used',
    '4000' => 'Used',
    '5000' => 'Used',
    '6000' => 'Used'
);
```

This provides a quick means of a slightly more powerful 'array_flip' method where multiple values are concerned.

### DIFF
--- a/libraries/joomla/utilities/arrayhelper.php
+++ b/libraries/joomla/utilities/arrayhelper.php
@@ -331,6 +331,55 @@ abstract class JArrayHelper
 	}
 
 	/**
+	 * Takes an associative array of arrays and inverts the array keys to values using the array values as keys.
+	 *
+	 * Example:
+	 * $input = array(
+	 *     'New' => array('1000', '1500', '1750'),
+	 *     'Used' => array('3000', '4000', '5000', '6000')
+	 * );
+	 * $output = JArrayHelper::invert($input);
+	 *
+	 * Output would be equal to:
+	 * $output = array(
+	 *     '1000' => 'New',
+	 *     '1500' => 'New',
+	 *     '1750' => 'New',
+	 *     '3000' => 'Used',
+	 *     '4000' => 'Used',
+	 *     '5000' => 'Used',
+	 *     '6000' => 'Used'
+	 * );
+	 *
+	 * @param   array  $array  The source array.
+	 *
+	 * @return  array  The inverted array.
+	 *
+	 * @since   12.3
+	 */
+	public static function invert($array)
+	{
+		$return = array();
+		foreach ($array as $base => $values)
+		{
+			if (!is_array($values))
+			{
+				continue;
+			}
+
+			foreach ($values as $key)
+			{
+				// If the key isn't scalar then ignore it.
+				if (is_scalar($key))
+				{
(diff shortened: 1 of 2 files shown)


elinw:
Could you not use :: in the name of the branch since it makes pull confused.

elinw:
I think it would be good to document the ways in which this is similar to/different from array_flip specifically

Different: array_flip only will work with strings and integers. This uses is_scalar instead and casts booleans to integers and floats to integers (by truncating not rounding).

Same: If there are multiple instances of a value only the last one is retained, the others are lost.  (We've had an issue previously where this came up and we had to change the return on a method not to use the values in order to not lose the elements with the same values.)

If you wanted you could add these to the data provider

```
        // Two values that are the same
        'Case 3' => array(
            // Input
            array(
                'New' => array('3000', '1500', '1750'),
                'Used' => array('3000', '4000', '5000', '6000')
            ),
            // Expected
            array(
                '1500' => 'New',
                '1750' => 'New',
                '3000' => 'Used',
                '4000' => 'Used',
                '5000' => 'Used',
                '6000' => 'Used'
            )
        ),
        // Float as value
        'Case 4' => array(
            // Input
            array(
                'New' => array(1.1111, '1500', '1750'),
                'Used' => array('3000', '4000', '5000', '6000')
            ),
            // Expected
            array(
                '1' => 'New',
                '1500' => 'New',
                '1750' => 'New',
                '3000' => 'Used',
                '4000' => 'Used',
                '5000' => 'Used',
                '6000' => 'Used'
            )
        ),
        // Boolean as value
        'Case 5' => array(
            // Input
            array(
                'New' => array(false),
                'Used' => array('3000', '4000', '5000', '6000')
            ),
            // Expected
            array(
                '0' => 'New',
                '3000' => 'Used',
                '4000' => 'Used',
                '5000' => 'Used',
                '6000' => 'Used'
            )
        ),
    );
```

pasamio:
1) :: wasn't used in the branch name, it's used in the pull title and is the correct description of the item as it's a static method of the given class. The branch name is jarrayhelper_invert.
2) Being consistent here seems reasonable, I'll update that to check if it's a string or int. I was mainly trying to avoid objects or other arrays as array keys more than anything which triggers errors. The truncation behaviour is PHP's conversion to valid array types.
3) Yes, it's consistent there. I had considered throwing an exception or something like that but opted instead to just let it go through.

elinw:
ooops misread the name :)

I actually think it's okay either way but either way it should be documented in the tests to prevent accidental changes in behavior in the future.

See #948  where we had this come up before.

PR #1559, opened 2012-09-29
add support for redirect on installer helper

oc666:
This is fix of bug #29374

### DIFF
--- a/libraries/joomla/installer/helper.php
+++ b/libraries/joomla/installer/helper.php
@@ -46,13 +46,17 @@ abstract class JInstallerHelper
 
 		$http = JHttpFactory::getHttp();
 		$response = $http->get($url);
-		if (200 != $response->code)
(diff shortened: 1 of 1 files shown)


pasamio:
This has some style errors:

<pre>
Expected "if (...)\n...{...}\n...else\n"; found "if (...)\n...{...}\n...else ".
 .../libraries/joomla/installer/helper.php : 49
Usage of ELSE IF is not allowed; use ELSEIF instead.
 .../libraries/joomla/installer/helper.php : 53
</pre>

oc666:
@pasamio, I've fixed this error. 
Is there any guide for how to install code styling tool? (Preferred for Netbeans)

elkuku:
Here you go: http://docs.joomla.org/Joomla_CodeSniffer#Netbeans

PR #1560, opened 2012-09-29
Add set method to JInputFiles + tests

florianv:


florianv:
This is not so useful, but can be used for unit tests.
The "problem" is that we expect the set method to set something.

libraries/joomla/input/files.php
```diff
@@ -110,10 +110,30 @@ protected function decodeData(array $data)
...
+			'size' => 0
+		);
+
+		// We just consider the needed keys.
+		$intersection = array_intersect_key((array) $value, $pattern);
+
+		// If there are less keys (or value is not an array) we throw an exception.
+		if (count($intersection) !== 5)
+		{
+			throw new InvalidArgumentException('The files value is not formatted correctly.');
```
elinw: I'm pretty certain this should be "file's" not "files/"

florianv:
I am closing this, since it cannot be merged.

PR #1561, opened 2012-10-01
Test

pasamio:
This is to test the pull tester to work out why it all of a sudden is finding all sorts of errors. Do not merge.

### DIFF
--- a/test.txt
+++ b/test.txt
@@ -1,0 +1,1 @@
+Test File


mbabker:
My bet would be when Rouven updated the sniffs to work with the latest version of PHPCS.  He and I were having similar issues running the tool with the CMS ruleset and upgrading solved it and exposed some warnings that weren't being reported.  I guess the upgrade on this side didn't go as smoothly.

pasamio:
Ahh yes, I see it now:
https://github.com/joomla/joomla-platform/commit/77ccd48a5941c11b30c2ad176821358f99626bf7

Fun

pasamio:
Perhaps, perhaps not! So this with a base of bd396400e5cddfb5170e2c41f64b1827c63417f4 passed without any CS errors. I've shifted on forwards to 7b23e48756efa58a3edf2def15d8615b678b737f to see if this causes an issues with it.

elinw:
When we first started we had to do something to prevent the external stuff from being tested, which is why there is a probably out of date note from me about how not to test phpmailer and simple pie in the instructions for testing locally.

mbabker:
We have the external files and some other ones to not test in the exclude patterns, but those aren't being read right now.  Neither is the ignore in the build.xml file.  And, FWIW, I can't run any `ant command` statements on my MacBook because of 7b23e48756efa58a3edf2def15d8615b678b737f (getting a "Cannot run program "phpcs${script-suffix}"" message).  Guess that should be a separate report/fix...

PR #1563, opened 2012-10-01
Add deleted & required _unix2DOSTime method

phproberto:
Method _unix2DOSTime was in Platform 11.1 but does not exist actually. 

Is required by the method _addToZIPFile:

```
    // Get the hex time.
    $dtime = dechex($this->_unix2DosTime($ftime)); 
```

Was removed intentionally? If so what is the replacement? 

We have to re-add the method or replace its call.

### DIFF
--- a/libraries/joomla/archive/zip.php
+++ b/libraries/joomla/archive/zip.php
@@ -539,6 +539,33 @@ class JArchiveZip implements JArchiveExtractable
 	}
 
 	/**
+	 * Converts a UNIX timestamp to a 4-byte DOS date and time format
+	 * (date in high 2-bytes, time in low 2-bytes allowing magnitude
+	 * comparison).
+	 *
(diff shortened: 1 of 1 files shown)


eddieajau:
Wouldn't you use the native 5.3 support in favour of JArchiveZip?  See http://php.net/manual/en/book.zip.php

phproberto:
@eddieajau It isn't my code. I was trying to fix this issue reported on joomla tracker:

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29411

If the user has to use native PHP functions why don't fully delete this class? If the library is available it has to work even if it's there only for backward compatibility.

There is a method in this class that is calling another one that has been deleted. We should remove it or fix it.

LouisLandry:
I do think we should be looking into going "native" where possible, but this appears to simply fix a regression.  Thanks a bunch Roberto!

PR #1564, opened 2012-10-02
sendMail automatically adds sender as replyTo

phproberto:
Reference Tracker ID:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29417

Please notice me if accepted to apply the same fix to CMS.

### DIFF
--- a/libraries/joomla/mail/mail.php
+++ b/libraries/joomla/mail/mail.php
@@ -441,7 +441,6 @@ class JMail extends PHPMailer
 	public function sendMail($from, $fromName, $recipient, $subject, $body, $mode = false, $cc = null, $bcc = null, $attachment = null,
 		$replyTo = null, $replyToName = null)
 	{
-		$this->setSender(array($from, $fromName));
 		$this->setSubject($subject);
 		$this->setBody($body);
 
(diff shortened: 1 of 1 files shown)


drmmr763:
+1 This is a simple change that will fix tons of issues. Many servers (like mine) require an at-domain sender email, but this causes problems when you want to _reply_ to a site sender from an emailed form. Being able to set the header here is perfect, you get the proper send address, and a useful reply to address.

pasamio:
So I dislike adding code that relies upon the internal structure of PHPMailer when we could add a simple `$autoReplyTo = false` on roughly line 457 and then add an else statement with `$autoReplyTo = true`. Alternatively we can flip the boolean here and put the `$autoReplyTo = false` inside the existing two if clauses.

fastnetwebdesign:
+1 Nice work Roberto... causes some of my clients huge issues.

phproberto:
@pasamio I thought about the same searching how to fix it. 

I finally used this because JMail already extends PHPMailer and this approach allows you to use a $mailer->replyTo() and then a $mailer->sendMail() without the replyTo info. The mail will use the desired replyTo inserted because it checks its existence globally .

If we set the autoReply only based on the arguments received by sendMail that will stop working. So you will be creating another bug.

pasamio:
Fair enough, merged.

PR #1566, opened 2012-10-05
Improvments to getJSObject

okonomiyaki3000:
Resources are ignored
Object keys and strings are encoded with json_encode
If the input array contains (php) objects, their values are used for recursion rather than the objects themselves (the function takes an array, not an object)
Less string concatenation, more array implosion


LouisLandry:
Nice.  Thanks!

PR #1567, opened 2012-10-05
JHTMLBehavior::_calendartranslation

okonomiyaki3000:
rewritten for improved readability


LouisLandry:
There's a unit test error that we need to take care of before this can be merged.  http://developer.joomla.org/pulls/pulls/1567.html

PR #1568, opened 2012-10-05
Keychain package

eddieajau:
#### Overview

The keychain provides a way to securely store sensitive information such as access credentials or any other data.  This pull request contains a new class called `JKeychain` that extends `JRegistry` and supports encrypting and decrypting data through the use of public and private keys.  It also provides the necessary command-line utility (a Joomla Platform application itself) to generate the key files.
#### Documentation

Documentation is provided in the pull request.  A readable version is available at:

https://github.com/eBaySF/joomla-platform/wiki/Keychain
#### Business case

At eBay, we needed a way to be able to store sensitive credentials (server/database logins, etc) in our SCM repository in an encrypted form without the hassle of sharing those credentials over email or other electronic means.  The Keychain package was developed so we could store encrypted data in the source code repository (the security issues related to storing clear-text credentials in a semi-public, albeit internal, source code repository should be obvious).  Engineers and deployment servers only need obtain the the 'shared key' manually, but only once.  Should the credentials change, the encrypted data file in the repository simply needs to be updated and all engineers and servers have access to the new information on the next build.


elinw:
Cool. My only question is kind of a general one about the command line management ... should that be part of the repository?

eddieajau:
The command line app is required to build the test files for the unit tests; that's the main reason it's included in the pull request, but there's an obvious value in having it there so people can actually create data to use the package out of the box.  There's also additional value there because it's yet-another-platform app that we can refer people to.  To be honest, I'd like to see @LouisLandry's phar builder in the /bin/ folder as well :)

elinw:
Oh that make sense.  The phar builder would make sense too. I don't think you would want to ship apps that are not used in the development process but both of those would be.

Issue #1544, opened 2012-09-21
undefined constant JPATH_SITE : in web application


bishnoi:
Notice: Use of undefined constant JPATH_SITE - assumed 'JPATH_SITE' in .....\htdocs\jplateform\libraries\joomla\filesystem\path.php on line 15

emerges when testing for web application.

Don't know how put it up, but here is correction:

define('JPATH_ROOT', JPath::clean('JPATH_SITE'));

elinw:
Would you turn that into a pull request?

elinw:
@bishnoi  You can just browse to the file, click edit and make the proposed change. When you save it will give you the option to make a pull request.

bishnoi:
ok Thanks

IberoMedia:
Hello,

I am not sure how this system works, but I want to notify you that this notice is still coming up. I downloaded Joomla Platform 12.1 yesterday, from http://api.joomla.org/ and I am getting "Notice: Use of undefined constant JPATH_SITE - assumed 'JPATH_SITE' in ROOT\libraries\joomla\filesystem\path.php on line 12." After I added the quotes to the JPATH_SITE constant the warning dissappeared.

dongilbert:
@IberoMedia - This is not an "issue" per se. When building with the Joomla Platform, you need to defined certain constants in your application bootstrap file. See this file here - https://github.com/JTracker/jissues/blob/master/includes/defines.php

and here - https://github.com/JTracker/jissues/blob/master/includes/boostrap.php

JTracker is an app built using the Joomla Platform. This a good example of how to bootstrap your applicaiton

dongilbert:
IberoMedia, I responded to your comment on the issue,
https://github.com/joomla/joomla-platform/issues/1544

PR #1569, opened 2012-10-05
Use of undefined constant JPATH_SITE

bishnoi:
Notice: Use of undefined constant JPATH_SITE - assumed 'JPATH_SITE' in .....\htdocs\jplateform\libraries\joomla\filesystem\path.php on line 15

emerges when testing for web application

### DIFF
--- a/libraries/joomla/filesystem/path.php
+++ b/libraries/joomla/filesystem/path.php
@@ -12,7 +12,7 @@ defined('JPATH_PLATFORM') or die;
 if (!defined('JPATH_ROOT'))
 {
 	// Define a string constant for the root directory of the file system in native format
-	define('JPATH_ROOT', JPath::clean(JPATH_SITE));
+	define('JPATH_ROOT', JPath::clean('JPATH_SITE'));
 }
 
 /**


elinw:
See #1544.  Thanks!

LouisLandry:
Actually this won't work.  I mean, yes it removes the notice, but it doesn't solve the problem.  In this particular code block it is looking for the constant `JPATH_SITE` to be equivalent to the root path of the Joomla CMS instance.  Now obviously the whole shebang needs a rethink to avoid that problem, but simply setting it to be the string literal `'JPATH_SITE'` is problematic because there are very likely no files at a local filesystem path `JPATH_SITE`.

elinw:
From the platform point of view is the application path  a  reasonable fall back assumption about where the root path is?  I don't think that location is the location where I would expect that to be in most applications. Because it seems to me that most applications will be in folders not at root. So if it were an emergency and you had no other information at all I would normally look at least one level up.  It has always seemed really strange to me that we don't have the cms  frontend code in a site folder so I don't think we should standardize on that structure.

LouisLandry:
So I'm going to close this.  I think we need a more complete look at this particular bit of code to find a better answer.  Please feel free to re-open it if you get a more comprehensive fix put together.  Thanks!

PR #1572, opened 2012-10-07
Replace use of undefined $error in JComponentHelper::_load

mbabker:


ianmacl:
You should be logging a message if component[option] is not set instead of leaving the second string empty.

PR #1573, opened 2012-10-07
Set param for file to load with render method

mbabker:
This change to JApplicationWeb::render() allows downstream users to set a file to load using the same methods as the template and params values are set with.  For example, in the CMS, we would have to override this method to use the tmpl=component URL param.  Now, it's just another param to set.


pasamio:
Why not default it to index.php instead of requiring each implementation to define a default? Theme path I can understand as being needed to be configured, but index.php is a pretty safe bet for a file name.

mbabker:
I forgot get allows a default value.  I'll fix it in a bit.

mbabker:
OK, that's changed.

PR #1585, opened 2012-10-09
JHTMLBehavior::_calendartranslation rewritten for improved readability

okonomiyaki3000:
Changed test to reflect that '/' is now escaped by json_encode

### DIFF
--- a/libraries/joomla/html/behavior.php
+++ b/libraries/joomla/html/behavior.php
@@ -838,60 +838,83 @@ abstract class JHtmlBehavior
 	{
 		static $jsscript = 0;
 
-		if ($jsscript == 0)
-		{
-			$return = 'Calendar._DN = new Array ("' . JText::_('SUNDAY', true) . '", "' . JText::_('MONDAY', true) . '", "'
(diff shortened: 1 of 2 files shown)


pasamio:
You realise that if you push up to the same branch (even with a rebase) the original pull request will be updated. You don't need to keep opening new pull requests.

okonomiyaki3000:
Indeed, but I couldn't push because of fast-forward yadda-yadda so instead of figuring out what to do about it, I just deleted the branch and uploaded again.

LouisLandry:
This looks great @okonomiyaki3000.  Thanks for sticking with it.  One day we'll get you sorted out on rebasing and forced pushing branches.

pasamio:
In that particular case you need to force push to ignore the losing commits warning (git push -f repo branch [note: repo branch not required but when I'm force pushing stuff I'd like to be explicit where it's going]). Alternatively deleting and pushing again is still a valid solution and I think I've had GitHub pick up the branch again but I might be wrong. Rebasing triggers Git's internal safety mechanisms to kick into action to make sure you really meant to essentially delete commits and wipe them out - which while annoying is usually a good thing.

PR #1586, opened 2012-10-09
Loggers can be added which exclude specific categories

okonomiyaki3000:
Suppose you want to separate your logs by category. Until now, it was only possible if you knew the names of all possible categories (which you don't). So now, let's say you want your deprecation logs in one place, your query logs in another, and all others someplace else. Then you just need to do this:

```
// Normal log files
JLog::addLogger(array('text_file' => 'deprecated'), JLog::ALL, 'deprecated');
JLog::addLogger(array('text_file' => 'databasequery'), JLog::ALL, 'databasequery');
// Excluding categories
JLog::addLogger(array('text_file' => 'others'), JLog::ALL, array('deprecated', 'databasequery'), true);
```

### DIFF
--- a/libraries/joomla/log/log.php
+++ b/libraries/joomla/log/log.php
@@ -159,12 +159,13 @@ class JLog
 	 * @param   array    $options     The object configuration array.
 	 * @param   integer  $priorities  Message priority
 	 * @param   array    $categories  Types of entry
+	 * @param   boolean  $exclude     If true, all categories will be logged except those in the $categories array
 	 *
 	 * @return  void
 	 *
 	 * @since   11.1
 	 */
-	public static function addLogger(array $options, $priorities = self::ALL, $categories = array())
+	public static function addLogger(array $options, $priorities = self::ALL, $categories = array(), $exclude = false)
 	{
 		// Automatically instantiate the singleton object if not already done.
 		if (empty(self::$instance))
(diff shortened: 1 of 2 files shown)


eddieajau:
I like the idea, but what if exclude was implied by prefixing the category name with bang (!), for example:

``` php
$logger->addLogger(array(), JLogger::ALL, array('included', '!excluded'));
```

That way you could mix white and black lists?

Also, we'd want to see the unit tests adjusted for this change (whatever its final form takes) and also I'd like to see the description of the pull request filled in with some details about the change and how a developer could use it.

okonomiyaki3000:
At first, I thought it might be useful to have both an included and an excluded list for each logger but after thinking about it a bit, I realized it's not necessary. If you say you want to log categories a, b, and c, it implies that you would not like to log categories d, e, and f. You don't need to exclude them.

So why is it useful at all? Because you don't have a finite list of all possible categories but you might want to have logs of almost everything. Maybe there are one or two specific categories you don't want to log or would like to log separately, this allows you to do that.

eddieajau:
You are absolutely right.  Ok, if you can attend to the unit test adjustments and the PR description we can keep moving this forward.

I'm closing this pull request for now until the requested changes are uploaded. Please re-open the pull request at that time so we can review it again. Thanks.

okonomiyaki3000:
I will add a new function to the tests, could you tell me what is the correct number to use for 'since' (or, better yet, where I can find this)?

eddieajau:
You'll use `@since 12.3`.

PR #1587, opened 2012-10-09
Loggers can be added which exclude specific categories

okonomiyaki3000:
Also, unit test for adding loggers to exclude specific categories

Suppose you want to separate your logs by category. Until now, it was only possible if you knew the names of all possible categories (which you don't). So now, let's say you want your deprecation logs in one place, your query logs in another, and all others someplace else. Then you just need to do this:

``` php
// Normal log files
JLog::addLogger(array('text_file' => 'deprecated'), JLog::ALL, 'deprecated');
JLog::addLogger(array('text_file' => 'databasequery'), JLog::ALL, 'databasequery');
// Excluding categories
JLog::addLogger(array('text_file' => 'others'), JLog::ALL, array('deprecated', 'databasequery'), true);
```

### DIFF
--- a/libraries/joomla/log/log.php
+++ b/libraries/joomla/log/log.php
@@ -159,12 +159,13 @@ class JLog
 	 * @param   array    $options     The object configuration array.
 	 * @param   integer  $priorities  Message priority
 	 * @param   array    $categories  Types of entry
+	 * @param   boolean  $exclude     If true, all categories will be logged except those in the $categories array
 	 *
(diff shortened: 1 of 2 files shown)


LouisLandry:
This is fantastic!  Thanks for putting it together.  I'm marking it accepted for 12.3 and will give it a day or so for other comments.  All we have to do now is get you comfortable with rebasing and such so you don't have to open new pull requests every time you make a change ;-)

PR #1588, opened 2012-10-10
Completing tests for JModelLegacy

dongilbert:
Here are the completed tests for JModelLegacy.

### DIFF
--- a/tests/suites/legacy/model/JModelLegacyTest.php
+++ b/tests/suites/legacy/model/JModelLegacyTest.php
@@ -3,110 +3,220 @@
  * @package     Joomla.UnitTest
  * @subpackage  Model
  *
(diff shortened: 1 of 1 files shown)


LouisLandry:
Great stuff.  You can improve the bit where you are testing the exception a little by looking over: http://www.phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.exceptions

Also, for extra points you can fix up all the code styling and docblocks, etc to take after: https://github.com/joomla/joomla-platform/blob/staging/tests/suites/unit/joomla/model/JModelDatabaseTest.php

dongilbert:
Will do. Also, I'm going to refactor to use fixtures and the setUp() and tearDown() functions, instead of calling JModelLegacy::getInstance() everywhere.

ianmacl:
Merged.  Thanks for your contribution!

PR #1592, opened 2012-10-10
Consistent use of XML values for fields 'readonly' and 'disabled'

benjaminpick:
As I said in #1188, this should be implemented in all field implementations (including joomla-cms).

### DIFF
--- a/libraries/joomla/form/field.php
+++ b/libraries/joomla/form/field.php
@@ -141,6 +141,24 @@ abstract class JFormField
 	protected $required = false;
 
 	/**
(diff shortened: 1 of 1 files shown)


LouisLandry:
Thanks a bunch @benjaminpick.  I'm going to leave it open for a day or so in case anyone wants to comment and we'll get it merged in.  Is there a chance that later on when you get more time you could go through and clean up the field implementations you mention?

benjaminpick:
Not really, sorry.

PR #1604, opened 2012-10-15
New function: JDatabaseQuery:format() 

okonomiyaki3000:
A sprintf-like function for building queries or query fragments.

Usage:

``` php
$query->format('SELECT %1$n FROM %2$n WHERE %3$n = %4$a', 'foo', '#__foo', 'bar', 1);
```

Returns: 

``` php
'SELECT `foo` FROM `#__foo` WHERE `bar` = 1'
```

Maybe a more common use would be to make something like this:

``` php
$query = $db->getQuery(true);
$query->select($query->qn('extension_id'))
    ->from($query->qn('#__extensions'))
    ->where($query->qn('type') . ' = ' . $query->q('file'))
    ->where($query->qn('element') . ' = ' . $query->q($element));
```

Into something like this:

``` php
$query = $db->getQuery(true);
$query->select($query->qn('extension_id'))
    ->from($query->qn('#__extensions'))
    ->where($query->format('%1$n = %2$q', 'type', 'file'))
    ->where($query->format('%1$n = %2$q', 'element', $element));
```

### DIFF
--- a/libraries/joomla/database/query.php
+++ b/libraries/joomla/database/query.php
@@ -1565,4 +1565,191 @@ abstract class JDatabaseQuery
 		// Apply the distinct flag to the union.
 		return $this->union($query, $distinct, $glue);
 	}
+
+	/**
+	 * Find and replace sprintf-like tokens in a format string.
+	 * Each token takes one of the following forms:
+	 *     %%       - A literal percent character.
+	 *     %[t]     - Where [t] is a type specifier.
+	 *     %[n]$[x] - Where [n] is an argument specifier and [t] is a type specifier.
+	 *
+	 * Types:
+	 * a - Numeric: Replacement text is coerced to a numeric type but not quoted or escaped.
+	 * e - Escape: Replacement text is passed to $this->escape().
+	 * E - Escape (extra): Replacement text is passed to $this->escape() with true as the second argument.
+	 * n - Name Quote: Replacement text is passed to $this->quoteName().
+	 * q - Quote: Replacement text is passed to $this->quote().
+	 * Q - Quote (no escape): Replacement text is passed to $this->quote() with false as the second argument.
+	 * r - Raw: Replacement text is used as-is. (Be careful)
+	 *
+	 * Date Types:
+	 * - Replacement text automatically quoted (use uppercase for Name Quote).
+	 * - Replacement text should be a string in date format or name of a date column.
+	 * y/Y - Year
+	 * m/M - Month
+	 * d/D - Day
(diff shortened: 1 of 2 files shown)


pasamio:
Why wouldn't we improve support for parameter binding? What does this add over sprintf?

okonomiyaki3000:
This is quite different from sprintf. It works in a similar way but it does different stuff. This function is able to apply functions such as quote, quoteName, escape, etc. to the replacement strings. I think it makes it a little easier to write proper queries. In particular, I think the 'name quote' is often omitted because it is generally optional and it's a bit of a hassle to write `$query->qn('column_name')`. 

It can be used to format a full query as in the example above but maybe a more common use would be to make something like this:

``` php
$query = $db->getQuery(true);
$query->select($query->qn('extension_id'))
    ->from($query->qn('#__extensions'))
    ->where($query->qn('type') . ' = ' . $query->q('file'))
    ->where($query->qn('element') . ' = ' . $query->q($element));
```

Into something like this:

``` php
$query = $db->getQuery(true);
$query->select($query->qn('extension_id'))
    ->from($query->qn('#__extensions'))
    ->where($query->format('%1$n = %2$q', 'type', 'file'))
    ->where($query->format('%1$n = %2$q', 'element', $element));
```

Not a huge difference in this case but I find it a bit more readable.

pasamio:
Can you edit the pull request description to include the second more extensive example so that we have that for the automatic changelog and then I'm merge this pull request (not sure if GitHub will let you change it after the fact). Once you've updated, put in a comment so that I get an email notification to merge.

okonomiyaki3000:
Updated

pasamio:
Thanks, merged :)

elinw:
@okonomiyaki3000 Elijah, could you provide some code for the cms to test this in practice (just a query and a dump out of the query that can be put in a file somewhere for testing purposes is fine). You can just take the test data and use that if you want.  The cms issue for this is here: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=30003

okonomiyaki3000:
@elinw Do you mean that you want some sample format strings and values and the expected result of the function? The result should be slightly different depending on which database is used. For example, MySQL uses backtick for name quotes but some others may use brackets, etc.

elinw:
THat's fine, just say "this assumes you are using mysql"  ... we just need to have a couple of people do integration  tests  in the context of the application. It should not interfere with anything else from my reading of the code so I don't see a problem with doing it.

PR #1605, opened 2012-10-16
JFacebook

diananeculai:
This package is part of my GSoC project.


tests/suites/unit/joomla/facebook/JFacebookCommentTest.php
```diff
@@ -0,0 +1,500 @@
+<?php
+/**
+ * @package     Joomla.UnitTest
+ * @subpackage  Facebook
+ *
+ * @copyright   Copyright (C) 2005 - 2012 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE
+ */
+
+require_once JPATH_PLATFORM . '/joomla/facebook/facebook.php';
```
ianmacl: You should rely on the autoloader where possible.

libraries/joomla/facebook/album.php
```diff
@@ -0,0 +1,208 @@
...
+	 * @param   integer         $limit   The number of objects per page.
+	 * @param   integer         $offset  The object's number on the page.
+	 * @param   string          $until   A unix timestamp or any date accepted by strtotime.
+	 * @param   string          $since   A unix timestamp or any date accepted by strtotime.
+	 *
+	 * @return  mixed   The decoded JSON response or false if the client is not authenticated.
+	 *
+	 * @since   12.3
+	 */
+	public function getPhotos($album, $oauth, $limit = 0, $offset = 0, $until = null, $since = null)
```
ianmacl: I may be missing something but it seems you are passing $oauth into almost every single method.  It would seem to make sense to put that into one of your base classes so that it doesn't have to be passed every time.  Similar to the way that JGithubObject stores the credentials and then automatically uses them when sending requests.

LouisLandry:
@dianaprajescu I'm excited about this addition to the platform.  As such I've marked it for inclusion to 12.3.  I haven't had an opportunity to really go over the contents, but will try to do that very soon.  I do want us to re-think how we are using the OAuth client classes.  It's something I'd like to revisit in the Google package as well actually.  That being said, at first glance this looks very well put together.  Hopefully others will chime in as well.

libraries/joomla/facebook/album.php
diananeculai: Thanks for pointing this out, I've thought about it during the summer but haven't had time to change it then and after that I've forgotten. I started the package this way because most Facebook GET requests for public data don't require authentication. Anyway I think it's better to have the OAuth client in the base class. I'll commit the change soon.

tests/suites/unit/joomla/facebook/JFacebookCommentTest.php
diananeculai: Done.

elinw:
You might want to squash some commits.

LouisLandry:
@dianaprajescu, quite simply I suck.  I'm sorry I haven't found the time to dig into this, I promise I haven't forgotten about it.  I've been beyond busy with work, moving house, and preparing for the Joomla World Conference.  I really appreciate your patience, and the work you put in ... I just need a little more time before I can give you meaningful feedback.  Hopefully we can get some others to look things over and give their thoughts.

diananeculai:
No worries @LouisLandry . I'm also going through a very busy period: university duties, diploma project, personal projects, some collaborations and conferences, so I'm not sure I would've had time to carefully go through the feedback. @elinw that's why I haven't squashed commits, I've tried but there are too many of them and it's quite hard to keep up while I'm already tired and under time pressure.

dongilbert:
If this is still a go for 12.3, it probably needs some further review.

diananeculai:
I have just updated this branch and solved the conflicts so that it's mergeable again. I'll also squash some commits soon and change the documentation file from <code>facebook.xml</code> to <code>facebook.md</code>, it probably needs some updating too as I made some changes since I've wrote it..

I would really appreciate some <b>feedback</b> on this package! I might have some time every now and then to refine it and maybe we can get it merged soon.

dongilbert:
I would like to see this merged as well. :) A lot of work here.

If I have time in the coming days, I'll try to take it for a spin.

diananeculai:
Thanks very much. I really look forward to.

elkuku:
This looks very nice and will be usefull for a lot of applications (including some of mine)

Hopefully this will be merged soon ;)

AmyStephen:
@dianaprajescu This does look great, much appreciated. 

A thought occurs to me as some of these new packages start to appear, it might be a good opportunity to start thinking about packages that are offered as JPlatform options. The same could be done with the github library. Eventually, as the subsystems are uncoupled, just about every subsystem could be packaged separately. 

If we were to do so now with new packages, it could help create discussion/agreement as to what coupling is expected between packages, and, even think about how to package and distribute. 

Eventually, it would help clarify the core JPlatform distribution, encourage more packages, and avoid bloating for those options that maybe aren't needed for everyone.

dongilbert:
@AmyStephen - I wouldn't be totally opposed to the idea, but I think for
now we should keep everything status quo with integrating packages into the
base platform. :) You know, not try to put the cart before the horse with
offering packages. We don't have PSR-0 autoloader yet. :P

AmyStephen:
/me worried one of those "baby steps" lectures is next, but realizing @dongilbert is right, concedes the point. =)

diananeculai:
I just squashed some commits. Please let me know if there is anything else to do in order to get this merged.

dongilbert:
Does this rely on your oauth1 client?

diananeculai:
No, this is using OAuth2 client which already is in the platform.

eddieajau:
Diana, let's have a talk someday about getting these package into the Framework too.

diananeculai:
I looked over JGoogle and it doesn't seem much to change, but I'm not sure of this. I'm going through a busy period since I'm about to finish the University and I'm also working on a start-up. I can't promise to get this done very soon, but I talked with @elinw and I think she wants to do it.

eddieajau:
There's no hurry Diana. In terms of time frame I'm thinking sometime before the end of the year. If you want, put an issue on the tracker and we can schedule if for Q2, Q3 or Q4.

See https://github.com/joomla/joomla-framework/issues/milestones

diananeculai:
Just opened an issue: https://github.com/joomla/joomla-framework/issues/83

Schedule it for Q3 just in case, and I'll let you know if I finish it until Q2.

PR #1615, opened 2012-10-19
Add JGithubAccount object

mbabker:
This adds an account object to the GitHub package with methods to interact with the OAuth Authorizations API (http://developer.github.com/v3/oauth/) and the rate_limit end point (http://developer.github.com/v3/#rate-limiting).

### DIFF
--- a/libraries/joomla/github/account.php
+++ b/libraries/joomla/github/account.php
@@ -1,0 +1,233 @@
+<?php
+/**
+ * @package     Joomla.Platform
+ * @subpackage  GitHub
+ *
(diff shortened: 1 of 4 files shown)


LouisLandry:
Fantastic work as always Michael.  Any chance I could interest you in adding support for the hooks api? http://developer.github.com/v3/repos/hooks/

mbabker:
Can definitely look at it over the weekend.  Truth be told, I was only planning on the rate limit endpoint here since I hit it yesterday.  OAuth Authorizations was an added bonus :laughing:

PR #1616, opened 2012-10-19
Cross-Database GUID function 

okonomiyaki3000:
generates 8-4-4-4-12 hexadecimal unique id

### DIFF
--- a/libraries/joomla/database/query/sqlazure.php
+++ b/libraries/joomla/database/query/sqlazure.php
@@ -29,4 +29,19 @@ class JDatabaseQuerySqlazure extends JDatabaseQuerySqlsrv
 	 * @since  11.1
 	 */
 	protected $name_quotes = '';
+
+	/**
+	 * Generates a Globally Unique Identifier (32 hexadecimal digits separated by hyphens as 8-4-4-4-12).
+	 *
+	 * Usage:
+	 * $query->set('guid = ' . $query->GUID());
+	 *
+	 * @return  string
+	 *
+	 * @since   12.3
+	 */
+	public function GUID()
+	{
+		return 'NEWID()';
+	}
 }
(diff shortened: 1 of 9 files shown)


LouisLandry:
Do you have a solve for SQLite ?

okonomiyaki3000:
SQLLite doesn't have a function like this. It should be possible to duplicate the functionality with something like `hex(randomblob(32))` or some such thing but I don't exactly know the best approach at the moment as I have no SQLLite experience at all. 

I did notice that the SQLLite Query class seems incomplete as it is. The base query class has functions like year, month, etc. but, based on my quick check of SQLLite available functions, these do not seem to exist and the SQLLite query class doesn't provide alternatives.

libraries/joomla/database/query/sqlazure.php
```diff
@@ -29,4 +29,19 @@ class JDatabaseQuerySqlazure extends JDatabaseQuerySqlsrv
 	 * @since  11.1
 	 */
 	protected $name_quotes = '';
+
+	/**
+	 * Generates a Globally Unique Identifier (32 hexadecimal digits separated by hyphens as 8-4-4-4-12).
+	 *
+	 * Usage:
+	 * $query->set('guid = '.$query->GUID());
```
pasamio: You should put spaces between this like we would normally have in properly formatted code.

okonomiyaki3000:
Fixed the formatting and rebased as a single commit.

okonomiyaki3000:
Ah, there are a couple of test classes that extend `JDatabaseQuery` so, if it has an abstract method, they will need to implement it. This is all well and good but doesn't actually make much sense. Are we going to need test classes for each subclass of `JDatabaseQuery`? They each have to implement that function in a different way so...

okonomiyaki3000:
SQLLite implementation based on this: http://comments.gmane.org/gmane.comp.db.sqlite.general/62514

eddieajau:
Closing because this PR cannot be merged.

PR #1619, opened 2012-10-19
Move the installer and updater packages to legacy.

LouisLandry:


mahagr:
Just curious: Are you (at some point) going to propose a better way to install extensions into platform applications? Because of I think that even platform applications could benefit from easy updates and possibility to extend them by extensions.

mbabker:
The plan is to create a better installer library.  The current code is very closely tied in with the CMS, error prone, and because of how it's written, hard to test or troubleshoot.  A cleaner package will hopefully solve a lot of these problems.  I guess I'm kinda curious if anyone's started working on it yet.

mahagr:
Yes, I'm aware of the plans, but I'm curious as well. I'm also interested if there's a plan to get the new installer class into the platform.

pasamio:
There is a plan, I don't think anyone has any code yet.

PR #1621, opened 2012-10-20
Implementing a new event system

florianv:
This is not ready to be merged and probably won't never be.

I wanted to implement a new feature for the event package, but for some reasons I'm not 100% fan of the actual system.

For example, JEvent might be called JEventListener/Observer, and the difference between JDispatcher::attach and register is not totally clear.
Also, I'm not sure if the Dispatcher should instanciate the listener, I think the listener object must be passed to the dispatcher instead.

For these reasons, I ended up rewriting a new package.
So I am asking you what do you think about a new event package that can look like this.

This implementation allow to specify a priority for a given listener, so if you have multiple listeners for a given event, you can manage the order in which they are notified during the registration.

An event is a particular object having a name and holding arguments.
Each listener must take a JEvent object as parameter and can manipulate its arguments.
Additionnaly, they can stop the event propagation, so the next listeners are not notified.

Moreover, this Dispatcher allows to use closures as listeners.

Like the actual one it is possible to use any object as event listener.

Some examples : 

Using a closure as listener and a class, stopping the event propagation.

``` php
<?php

class MyListener
{
    public function onBeforeSomething(JEvent $event)
    {
        echo 'It is enough for me';

        // Stopping the event propagation.
        $event->stopPropagation();
    }
}

$listener = function (JEvent $event)
{
    echo 'Please pass me this event!';
};

$dispatcher = JEventDispatcher::getInstance();

$dispatcher->registerListener(array('onBeforeSomething'), new MyListener);
$dispatcher->registerListener(array('onBeforeSomething'), $listener);

$dispatcher->triggerEvent('onBeforeSomething');
```

Registering a listener with a given priority.
The default priority is 0.
If some listeners have the same priority, they will be notified in the order they were added.

``` php
<?php
class Listener1
{
    public function onBeforeSomething(JEvent $event)
    {
        echo 'Im the first';
    }
}

class Listener2
{
    public function onBeforeSomething(JEvent $event)
    {
        echo 'Im the second';
    }
}

$dispatcher = JEventDispatcher::getInstance();

// Registering the Listener1 with a priority = 2 
// for the onBeforeSomething event
$dispatcher->registerListener(
    array('onBeforeSomething'), 
    new Listener1,
    array('onBeforeSomething' => 2)
);

// Registering the Listener2 with a priority = 1 
// for the onBeforeSomething event
$dispatcher->registerListener(
    array('onBeforeSomething'), 
    new Listener2,
    array('onBeforeSomething' => 1)
);

$dispatcher->triggerEvent('onBeforeSomething');
```

Injecting you custom events.
There are two ways to override the default event object that is passed to all listeners.

``` php
<?php
// Create the onBeforeSomething event with special arguments.
$onBeforeSomething = new JEvent('onBeforeSomething');
$onBeforeSomething->setArgument('foo', 'bar');

$dispatcher = JEventDispatcher::getInstance();

// First method : registering a custom event.
$dispatcher->registerEvent($onBeforeSomething);

// Second method : when triggering the event.
$dispatcher->triggerEvent($onBeforeSomething);
```


LouisLandry:
@florianv I'm certainly interested in where you are going with this, but obviously changing something this low-level in the system will be delicate will have to be considered carefully.  I'm just back from a week long holiday so my head really isn't in the right place to go through things bit by bit.  I'd love to hear what any others have to say on your proposal and hopefully next week I can spend a little time playing with it.  In the meantime, can you rebase this and get it mergeable again so we have something we can actually tinker with?

PR #1622, opened 2012-10-20
New Feed Package.

LouisLandry:
This is not yet ready to merge, but I wanted to start the discussion about it.


elinw:
I like it; the older version of this is working great in the cms and this  rather nicely takes care of the issues we had to fix.

eddieajau:
I'd like to see some docs on how it's used.

elinw:
@eddieajau  Is that the main issue from your perspective?

eddieajau:
Yes. I've no issue with the code being in good condition. But without seeing how it's used, and especially as Louis will have no interest in doing any further work on this, we need the documentation.

elinw:
Ok so since we've used it in the CMS already I can write up some docs based
on that.

elinw:
Do you think that this will be merged? As I've mentioned I'm happy to update the since tags etc, Otherwise I will  make a copy of this update the package in the cms library.  We've been using it since 3.0 with no problem and this version fixes some issues and would if we have a little time let us look at taking simple pie out of the factory.

dongilbert:
Not sure why the docs were merged and this wasn't. Doesn't make sense to have docs with no code. Merged.

elinw:
Thanks Don!

PR #1625, opened 2012-10-21
Added some method inside JDatabaseDriver to create database and alter character set [v2]

gpongelli:
Added two public method inside JDatabaseDriver:
- alterDbCharacterSet to change database's character set
- createDatabase to create a new database

AlterDbCharacterSet needs database name as parameter, createDatabase needs a JObject, similar to that used inside CMS' "initialise" function, and a bool to set UTF8 database's support.

They use two protected method to obtain correct query in database's
supported syntax:
- getAlterDbCharacterSet
- getCreateDatabaseQuery

Protected method inside JDatabaseDriver return query for MySQL and MySQLi syntax, this commit contain also overridden version for PostgreSQL database.

There are tests only for protected methods inside MySQL, MySQLi and PostgreSQL test classes.

[Old Pull request](https://github.com/joomla/joomla-platform/pull/1261)

### DIFF
--- a/libraries/joomla/database/driver/postgresql.php
+++ b/libraries/joomla/database/driver/postgresql.php
@@ -1117,7 +1117,7 @@ class JDatabaseDriverPostgresql extends JDatabaseDriver
 	 *
 	 * @return  string  The query that creates database, owned by $options['user']
 	 *
-	 * @since   12.2
+	 * @since   12.3
 	 */
 	protected function getCreateDatabaseQuery($options, $utf)
 	{
--- a/libraries/joomla/database/driver/postgresql.php
+++ b/libraries/joomla/database/driver/postgresql.php
@@ -789,7 +789,7 @@ class JDatabaseDriverPostgresql extends JDatabaseDriver
 	/**
 	 * This function return a field value as a prepared string to be used in a SQL statement.
 	 *
-	 * @param   array   $columns      Array of table's column returned by ::getTableColumns.
+	 * @param   array   &$columns     Array of table's column returned by JDatabasePostgreSQL::getTableColumns.
 	 * @param   string  $field_name   The table field's name.
 	 * @param   string  $field_value  The variable value to quote and return.
 	 *
@@ -797,7 +797,7 @@ class JDatabaseDriverPostgresql extends JDatabaseDriver
 	 *
 	 * @since   11.3
 	 */
-	public function sqlValue($columns, $field_name, $field_value)
+	protected function sqlValue(&$columns, $field_name, $field_value)
 	{
 		switch ($columns[$field_name])
 		{
@@ -816,7 +816,6 @@ class JDatabaseDriverPostgresql extends JDatabaseDriver
 			case 'bigserial':
 			case 'integer':
 			case 'money':
-			case 'numeric':
 			case 'real':
 			case 'smallint':
 			case 'serial':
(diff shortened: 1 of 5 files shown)


elinw:
Gabriele you have a unit test failure but I'm not sure why.

gpongelli:
Could you write the error?
Thanks.

elkuku:
You may see it here: http://developer.joomla.org/pulls/pulls/1625.html

gpongelli:
I see it's related to JOAuth2ClientTest, I've not developed that test.
Check if this error is also present on master branch, I forked from it yesterday to create this pull request.

libraries/joomla/database/driver/postgresql.php
```diff
@@ -1110,18 +1109,17 @@ public function getAlterDbCharacterSet( $dbName )
...
-	 * 									not used in PostgreSQL "CREATE DATABASE" query.
+	 * @param   stdClass  $options  Object used to pass user and database name to database driver.
+	 * 									This object must have "db_name" and "db_user" set.
+	 * @param   boolean   $utf      True if the database supports the UTF-8 character set.
 	 *
-	 * @return  string	The query that creates database, owned by $options['user']
+	 * @return  string  The query that creates database, owned by $options['user']
 	 *
-	 * @since   12.1
+	 * @since   12.2
```
elinw: 12.3

tests/suites/database/driver/mysql/JDatabaseMySQLTest.php
```diff
@@ -17,6 +17,22 @@
 class JDatabaseMysqlTest extends TestCaseDatabaseMysql
 {
 	/**
+	 * Data for the getCreateDbQuery test.
+	 *
+	 * @return  array
+	 *
+	 * @since   11.3
```
elinw: 12.3

tests/suites/database/driver/mysqli/JDatabaseMySQLiTest.php
```diff
@@ -17,6 +17,22 @@
 class JDatabaseMysqliTest extends TestCaseDatabaseMysqli
 {
 	/**
+	 * Data for the getCreateDbQuery test.
+	 *
+	 * @return  array
+	 *
+	 * @since   11.3
+	 */
```
elinw, line 25: 12.3

gpongelli, line 25: Changed.

elinw:
So in some indirect way this change is impacting OAuth2 the issue to figure out is how.

ianmacl:
The error is an intermittent error related to comparing timestamps and has been fixed in master so don't worry about it (though if you rebase it will go away).

My main question regarding this pull request is that it would seem better to use the PHP methods to do this so that the client is aware of what it is supposed to be sending rather than using direct queries.

eddieajau:
The JDatabaseImporter classes are already handling schema manipulation.  My personal preference would be to continue to do this outside of the driver itself, maybe even starting some sort of ETL layer (which I had hoped to do someday myself) or schema helper, and moving stuff like dropTable in there as well.

ianmacl:
IMO those sorts of manipulation methods belong in the JDatabaseDriver, not the JDatabaseImporter class, IMO.  The importer methods will certainly use them, but shouldn't own them.  In some cases you can't create, remove databases and tables without access to the underlying connection object which shouldn't be something that JDatabaseImporter has access to.

gpongelli:
I added these methods to let installation phase works.
I did these changes for J2.5 , then I haven't followed subsequent joomla's changes.

I agree with Ian, the driver has to own all kinds of database's manipulation (tables, schema, etc); the other classes use  the driver to perform their task without owning this code.
In my humble opinion this is a better solution for a logical point of view (who owns what) and for code clearness (piece of code exists at only one point).

Eng. Gabriele Pongelli.

elinw:
It would be so helpful to have create database.  The current importer is extremely limited and there is no sqlsrv support so just as a side consideration the idea of being able to create and actually potentially do a complete installation of an application including creating the database in the abstraction layer is very exciting.

eddieajau:
I'm going to close this because I really want to see database manipulation handled in another class.

gpongelli:
Ok.
I did these changes to let CMS installs correctly on PostgreSQL.

PR #1633, opened 2012-10-24
Adding JMail::IsHTML and have it return $this for chaining. Includes tests.

dongilbert:
With the current implementation of JMail, you can't do something like this:

```
JFactory::getMailer()
    ->addRecipient($to)
    ->setSubject($subject)
    ->setBody($body)
    ->addAttachment($attachment)
    ->IsHTML(true)
    ->Send();
```

because `PHPMailer::IsHTML` doesn't support method chaining on the object. This PR fixes that.

### DIFF
--- a/libraries/joomla/mail/mail.php
+++ b/libraries/joomla/mail/mail.php
@@ -359,7 +359,7 @@ class JMail extends PHPMailer
 	 *
 	 * @since   12.3
 	 */
-	public function IsHTML($ishtml = true)
+	public function isHtml($ishtml = true)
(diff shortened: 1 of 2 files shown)


ianmacl:
If we're going to explicitly add this then it should be isHTML not IsHTML.

dongilbert:
Should it be `isHtml`? I know casing on functions and method names is irrelevant, but we do want to be consistent.

eddieajau:
Thanks Don. Merged.

PR #1640, opened 2012-10-25
Handle nested transactions

okonomiyaki3000:
Transactions need to support 'nesting' (actually save points) so that mistakes like this don't happen:

``` php
$db->transactionStart();

// do stuff

SomeFunction();

// oops, need to rollback for some reason
$db->transactionRollback();
// Oh no! It was already commited because:

/****** Elsewhere... ********/
function SomeFunction()
{
    // I don't know that we already started a transaction so I'll try to start one.
    $db->transactionStart();  
    // Hmm. Seems fine. Do stuff.

    // All good, let's commit. I think I'm just committing my stuff because I don't know anything!
    $db->transactionCommit(); 
}
```

I'm sure this situation never happens in Joomla but let's make sure it never can happen. 

Like this:

``` php
$db->transactionStart(true); // In case there is currently no transaction, this works as it always has

// do stuff

SomeFunction();

// oops, need to rollback for some reason
$db->transactionRollback(true);
// No problem, because:

/****** Elsewhere... ********/
function SomeFunction()
{
    // I don't know if we already started a transaction so I'll try to start one with the savpoint flag just in case.
    $db->transactionStart(true);  
    // Hmm. Seems fine. Do stuff.

    // All good, let's commit. I'll commit with the savepoint flag so as to not step on any transactions that may be 'enclosing' this one.
    $db->transactionCommit(true); 
}
```


pasamio:
I'm not sure I like the concept of the database driver being deceptive about what it's doing. At this point you're putting magic into the driver that the application developer can't avoid. The situation now will be when I call commit is it really committed? There is no way of knowing as a caller of the application if it's actually done what you wanted it to do or if it's silently ignored and the data wasn't actually committed. In a sense the database driver will lie to you that the commit ever happened which means you have bad faith that the data committed when what you thought was committed could be rolled back on the next statement.

okonomiyaki3000:
It is correct that you might call `transactionCommit` but not actually commit anything. This will happen if your code is being executed within the scope of another transaction. If that 'parent' transaction is committed, yours will also be. If it is not, yours shouldn't be anyway. 

The alternative (the current situation) is that you might think you are committing your transaction but, in fact, you are committing the 'parent' which might actually need to be rolled back at some point but now can't be. Or, you might try to rollback your transaction but you actually rollback all the way to the parent. Both of these could be very bad. 

So I believe I'm choosing the lesser of two evils. Perhaps this solution would be more palatable if the current transaction level was exposed.

dongilbert:
Would it be possible to do named or grouped transactions with a var passed through to the `transactionStart()` method? I know it's more a weakness of MySQL than the code, since it only supports a single transaction per db connection.

So in the transaction process, if a named transaction is started, it opens a new connection to the database, with it's own transaction history and ability to commit / rollback as needed. When you call `transactionCommit` or `transactionRollback`, you pass in the same name, and the db driver knows which connection to use to perform the task. Think if it as db connection instances.

okonomiyaki3000:
@dongilbert This is an interesting idea and can actually work mostly as you describe without any modification to existing database driver code. A developer who needs to do some work in a transaction should just get a new DBO instance specifically for that transaction. There are probably some cases where this is exactly what you would want to do. I don't think it's right for the general case though. 

For one thing, any changes that you've made in one connection's transaction will not be visible to the other connection. This is especially confusing when multiple transactions are happening (which is the issue I'm trying to address in the first place). They're also committed or rolled back independently of one another. In nearly all cases, I don't think this is what you want. 

If I start a transaction and later roll it back, I want all changes rolled back. That means even if I called some other functions which ran their own transactions without problems and committed them, I want those rolled back too. This is not possible if they were done on a different connection and this could lead to some serious problems. What you really want and what you should expect is that your transactions will be nested. Savepoints aren't exactly the same as nesting as we usually think of it but they are effectively the same.

pasamio:
I'd support putting in a flag for `transactionStart` and `transactionCommit` to opt-in to supporting being in a nested transaction but I dislike the idea of me saying commit this and having something else decide it thinks it knows better. We'd need to figure out some nuance around what pre-conditions are valid and what happens if someone starts a transaction inside another transaction (nested or otherwise) and how that all works.

More broadly consider if you come from another framework and you see our database abstraction layer. You may expect commit to behave like it does normally on every other platform. In MySQL, when you attempt to start a transaction more than once it lets you do that. However when you send commit it commits everything. I don't see a good reason to second guess the behaviour here of MySQL, which is our primary database engine. Oracle is by default always in a transaction and you either commit or rollback. Obviously Oracle supports savepoints which allow you to control how far you rollback but when you do a commit, it's committed. I don't have enough MSSQL experience to comment there but I'd expect it to behave similarly.

I don't think that by default implementing a behaviour that is completely distinct from the standard expected behaviour is a good idea. It leads to mistakes. Right now for sure we have a potential that something is committed that shouldn't be but this is the expected behaviour for most of these systems. As you point out yourself, databases don't have the concept of nested transactions. I don't think it is our position in the database layer to "transparently" support a concept that is not the norm or even a major issue. It isn't a bug that when you do a commit that it commits everything. It might not be intentional but that is the expected behaviour of all database engines I've encountered.

mahagr:
Maybe there should be warning if someone is trying nested transaction (flag idea) saying that it's not supported in MySQL? That warning should display/be logged even in other engines...

okonomiyaki3000:
Ok, I think I can work out a compromise that everyone will be satisfied with. I should be able to do it tomorrow.

pasamio:
I'm not aware of a database that supports nested tranactions, save points support rolling back part of a transaction but don't permit partial committing of the last segment. I don't have any objections to improving the API to add support for save points as an explicit API call.

okonomiyaki3000:
@pasamio Whether nested or using save points, I don't think committing the last segment but rolling back others would ever make any logical sense. 

Suppose nested transactions were a reality. Suppose you committed one or more nested transactions within an enclosing 'parent' and then finally rolled back the parent. The expected behavior would always be to roll back everything to the start of the parent transactions. 

When you talk about a last segment, it seems that what you really want is non-nested transactions in sequence. There is, of course, no issue with this as it is. 

Any time you are dealing with nestable structures, whether they are conditionals, loops, tags in some markup language, etc., whoever initiates the structure is responsible for closing it. Transactions should work the same way. 

Working with savepoints explicitly is not the answer as it requires the developer to know things about the context in which the code he's writing will be executed which he should not (maybe cannot) know. It also relies on the developer to do things in the 'right' way or risk breaking someone else's code in a way that is extremely difficult to debug.

pasamio:
Actually what I want is a database abstraction API that mirrors what the underlying system is doing, not something that through some sort of magic (imperceptible to the external user) corrupts the "commit" statement departing from decades of established standards (ANSI) and implementation (Oracle, MSSQL, Sybase, MySQL) to create a concept that doesn't exist and isn't really ideal for this layer.

What I want is when I say commit, the data hits the database and is committed. That way I can handle the situation where the commit fails for what ever reason and recover and not have my database API lie about what it's doing so that the code then moves onto the point where I can't reasonably recover the situation. Perhaps I'm old fashioned about my data integrity.

What I've stated I'm happy to compromise on is a flag that says that it's ok for the transaction not to actually commit and state that I don't care that my data doesn't get committed (at which point I wonder why we're using a transaction here). I'm happy to even say that someone can start a transaction without the flag but as soon as someone implements the flag for a nested transaction, if within that someone attempts to start a transaction that doesn't have the nested flag that this should raise some form of exception or error case (again, expected behaviour versus magic behaviour). I've also stated I'm supportive of adding support for adding save points for those who wish to use them.

The simple reality transactions aren't nested and work on a different level. If you feel transactions should work differently then you are most welcome to take that up with the folks ratifying the ANSI SQL standard and the various database vendors. Until such time as that changes, I feel the database abstraction layer needs to reflect the reality of the situation not what you feel should be the situation.

okonomiyaki3000:
Maybe I haven't properly explained the problem scenario but the thing I'm trying to avoid is an unrecoverable situation. One in which a commit was made prematurely and now can't be rolled back. 

Anyway, the idea of using a flag seems suitable. It allows all existing code to continue to function in exactly the same way while providing support for savepoints when needed.

libraries/joomla/database/driver/postgresql.php
```diff
@@ -849,55 +849,92 @@ public function sqlValue($columns, $field_name, $field_value)
 	 *
 	 * @return  void
 	 *
-	 * @since   12.1
+	 * @param   boolean  $toSavepoint  If true, commit to the last savepoint.
+	 *
+	 * @since   12.2
```
pasamio: Any reason you're changing the since tag here?

libraries/joomla/database/driver/postgresql.php
```diff
@@ -849,55 +849,92 @@ public function sqlValue($columns, $field_name, $field_value)
...
 	 * Method to roll back a transaction.
 	 *
-	 * @param   string  $toSavepoint  If present rollback transaction to this savepoint
+	 * @param   boolean  $toSavepoint  If true, rollback to the last savepoint.
 	 *
 	 * @return  void
 	 *
-	 * @since   12.1
+	 * @since   12.2
 	 * @throws  RuntimeException
```
pasamio: Same question as above, why is the since tag changing?

libraries/joomla/database/driver/postgresql.php
```diff
@@ -849,55 +849,92 @@ public function sqlValue($columns, $field_name, $field_value)
...
 
 	/**
 	 * Method to initialize a transaction.
 	 *
+	 * @param   boolean  $asSavepoint  If true and a transaction is already active, a savepoint will be created.
+	 *
 	 * @return  void
 	 *
-	 * @since   12.1
+	 * @since   12.2
```
pasamio: Ditto on since tag change.

libraries/joomla/database/driver/sqlsrv.php
```diff
@@ -784,49 +784,94 @@ public function setUTF()
 	/**
 	 * Method to commit a transaction.
 	 *
+	 * @param   boolean  $toSavepoint  If true, commit to the last savepoint.
+	 *
 	 * @return  void
 	 *
-	 * @since   12.1
+	 * @since   12.2
```
pasamio: More since tag changes?

libraries/joomla/database/driver/sqlsrv.php
```diff
@@ -784,49 +784,94 @@ public function setUTF()
...
 
 	/**
 	 * Method to roll back a transaction.
 	 *
+	 * @param   boolean  $toSavepoint  If true, rollback to the last savepoint.
+	 *
 	 * @return  void
 	 *
-	 * @since   12.1
+	 * @since   12.2
```
pasamio: More since tag changes?

libraries/joomla/database/driver/sqlsrv.php
```diff
@@ -784,49 +784,94 @@ public function setUTF()
...
 
 	/**
 	 * Method to initialize a transaction.
 	 *
+	 * @param   boolean  $asSavepoint  If true and a transaction is already active, a savepoint will be created.
+	 *
 	 * @return  void
 	 *
-	 * @since   12.1
+	 * @since   12.2
```
pasamio: More since tag changes?

pasamio:
I've made comments about the since tag changing, not sure why you're changing them...were those changes not in 12.1?

This pull request isn't mergeable at the moment, can you rebase it to get it mergeable, update the pull request description to match the new behaviour and we'll get this merged.

Once you're ready drop a note saying you want to reopen the pull request and we'll have a look again.

okonomiyaki3000:
Looks like those tag changes were probably copy/paste errors. I've fixed them and rebased with the latest changes to staging.

okonomiyaki3000:
@pasamio We can maybe look at this again.

pasamio:
If you're ready to reopen it, you should be able to say "reopen" in a comment and it should reopen the pull request. I did see Ian manually open one the other day so I'm partially curious to check if it's still working properly.

okonomiyaki3000:
reopen

okonomiyaki3000:
Oh, that's nice. I wonder if "open sesame" would have worked...

dongilbert:
"reopen sesame" would have worked. heh heh :)

pasamio:
It's coded for reopen and re-open. It doesn't look like you've added any extra commits. Have you made changes and pushed back to the same branch?

pasamio:
Let me rephrase that, GitHub is marking that it can now be merged however the since tags are still changing to 12.2 in some places. Did you go through and validate all of the since tag changes are accurate?

okonomiyaki3000:
The only place 12.2 appears in this diff is in the mysqli driver but this is unchanged from the official repo. Where I have made additions, I've tagged them with 12.3.

pasamio:
Perhaps GitHub was lagged, it was showing in the diff still when I made the comment but it looks all good now.

pasamio:
Can you fix up these unit test failures? Should have double checked again before I merged. 

http://developer.joomla.org/pulls/pulls/1640.html

okonomiyaki3000:
I'll have a look. When did we get a nosql driver?

okonomiyaki3000:
reopen sesame

okonomiyaki3000:
reopen

okonomiyaki3000:
Jenkins didn't show up. Do I need a new pull request?

pasamio:
You will need a new pull request. Build.j.org didn't report any problem though. Do you have a fix?

okonomiyaki3000:
The fix is here: https://github.com/joomla/joomla-platform/pull/1704

pasamio:
Also no idea when we got the nosql driver either.

okonomiyaki3000:
I don't imagine that Joomla will ever run MongoDB.

pasamio:
I was working on a MongoDB driver the other week.

PR #1647, opened 2012-10-27
Remove patch method from JGithubHttp

mbabker:
The patch method was added in to the parent JHttp class, making the method in the child JGithubHttp class no longer necessary.  Additionally, since the methods have different declarations, this causes a E_STRICT error.

Since the method is only used in the GitHub package, I've opted to totally remove the method instead of deprecating it.  Essentially, this leaves no need for the child class at this time, but I've left it for the time being.

### DIFF
--- a/libraries/joomla/github/http.php
+++ b/libraries/joomla/github/http.php
@@ -55,20 +55,4 @@ class JGithubHttp extends JHttp
 		// Set the default timeout to 120 seconds.
 		$this->options->def('timeout', 120);
 	}
-
-	/**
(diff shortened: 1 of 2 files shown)


ianmacl:
Since there are now no tests left in JGithubHttpTest.php the file should be removed (it causes a test failure).  Alternatively, you could write a simple test for the constructor to make sure the options get set.

mbabker:
I've added a test on the constructor.  Reopen please.

ianmacl:
Merged.

PR #1652, opened 2012-10-29
Cleaning in the test tree

elinw:


dongilbert:
Gettin' it DONE!!! Good job Elin. :+1:

ianmacl:
https://github.com/elinw/joomla-platform/blob/codecleaning/tests/suites/unit/joomla/object/JObjectTest.php#L42

That method needs to be removed because we don't want empty tearDown methods and because it is a parse error (you have no method body).

ianmacl:
Closing until parse errors are fixed.

elinw:
reopen please (let's see what else is wrong :P)

ianmacl:
Closing until code is fixed.

PR #1654, opened 2012-10-30
Adding $form parameter to JHtmlGrid::sort()

Bakual:
The JS function "Joomla.tableOrdering" allows 4 parameters (order, dir, task and form), but JHtmlGrid::sort() only uses three of them. This commit adds the form parameter which allows a developer to choose the name of the form (instead of "adminForm").
This can be handy in case there is more than one form in an output or if the form needs another name for some reason.

### DIFF
--- a/libraries/joomla/html/grid.php
+++ b/libraries/joomla/html/grid.php
@@ -71,7 +71,7 @@ abstract class JHtmlGrid
 	 *
 	 * @since   11.1
 	 */
-	public static function sort($title, $order, $direction = 'asc', $selected = 0, $task = null, $new_direction = 'asc')
+	public static function sort($title, $order, $direction = 'asc', $selected = 0, $task = null, $new_direction = 'asc', $form = null)
 	{
 		$direction = strtolower($direction);
 		$images = array('sort_asc.png', 'sort_desc.png');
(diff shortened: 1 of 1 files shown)


ianmacl:
Can somebody please test this and make sure nothing breaks if $form is not supplied?  I am concerned that the check at https://github.com/joomla/joomla-platform/blob/staging/media/system/js/core-uncompressed.js#L19 will fail and we'll introduce a bug with this change.

mbabker:
It shouldn't fail.  That same check is made in the tableOrdering method at https://github.com/joomla/joomla-platform/blob/staging/media/system/js/core-uncompressed.js#L210, before the submitform function is called and the check you've highlighted being executed.  Maybe to be safe, we should default the param in the sort method to 'adminForm', but that's the extent that we could do to feasibly stop a bug from creeping in.

ianmacl:
Except that now instead of not passing the variable at all we will be passing an empty string.  See http://jsfiddle.net/Syfbg/1/ for an illustration of the difference that this will result in.  I highly suspect that is will cause issues, but if somebody can do some actual testing I would be happy to be proven wrong.

ianmacl:
I'm closing this until more research and testing has been done.

mbabker:
OK, looking at it some more, the JS methods both expect the form object, not the string name of the form.  So, for this change to be accepted, the JS would have to be changed so that the form param is a string type versus object and the JS would then load the form element.  This fiddle further demonstrates that - http://jsfiddle.net/ZCqEB/1/

Bakual:
I see the problem and I learned something about Javascript today :-)
I have this fixed in my testing enviroment with the following code for the core-uncompressed.js:

``` javascript
Joomla.tableOrdering = function(order, dir, task, form) {
    if (typeof(form) === 'undefined') {
        form = document.getElementById('adminForm');
    } else if (typeof(form) === 'string') {
        if (!form) {
            form = 'adminForm';
        }
        form = document.getElementById(form);
    }
    form.filter_order.value = order;
    form.filter_order_Dir.value = dir;
    Joomla.submitform(task, form);
}
```

This works without problems as far as I see it. However I'm not sure if it's good codestyle and I don't know yet how to do the compressed variant.

PR #1656, opened 2012-11-01
Fix code style line lengths in user agent strings

dongilbert:
While continuing my code style cleanup in the unit test tree, I came across this rather large chunk of UA strings for the WebClient tests. They all (obviously) failed code style checks based on the line length, so I went ahead and broke these up as logically as I could, without extending the file length too much. If this is an acceptable approach, I'll continue doing it this way when line length code style errors arise - but I want to get feedback first to be sure.


elinw:
awesome

PR #1657, opened 2012-11-01
Code style work in the testing tree

elinw:
Cleaning up various things in the test tree including working on basic code sniff compliance,  only using setUp and tearDown where needed and some work on docblocks and code quality.


elinw:
See #1652 this came as a new pr instead of updating the old one.

PR #1658, opened 2012-11-02
More code cleaning in the test tree

elinw:


dongilbert:
@elinw - You and I should coordinate efforts here. I had waited until your last code style PR was merged, and then pulled that into my branch where I'm doing my cs fixes, but I failed to look at the PR queue before going through and running the interactive phpcs command and systematically fixes the errors. If you take a look at #1659, you'll see I duplicated some of your work. :(

No big deal though - we just gotta decide what to do. :)

PR #1660, opened 2012-11-02
Use include_once

ssv445:
I have found a bug recently, we were trying to autoload JHtmlGrid.
- We tried to autoload class JHtmlGrid
- Joomla autloader loaded libraries/joomla/html/grid.php (class in that file is JGrid)
- then JHtml tried to autoload the same class
- again the same file was included (as class expected in file was not there, so class_exists check failed)

So I propose here that for autoloading purpose, one file should be loaded once.
Added the same change here.

### DIFF
--- a/libraries/loader.php
+++ b/libraries/loader.php
@@ -332,7 +332,7 @@ abstract class JLoader
 			// Load the file if it exists.
 			if (file_exists($path))
 			{
-				return include $path;
+				return include_once $path;
 			}
 		}
 	}


pasamio:
https://github.com/joomla/joomla-platform/blob/master/libraries/joomla/html/grid.php

The class in that file is JHtmlGrid. Perhaps I'm missing something here.

As an aside, I don't think the fix for an incorrectly named class is to impose a significant performance hit on all users of the system.

ianmacl:
http://stackoverflow.com/questions/4326435/php-include-vs-include-once-speed

The include vs include_once is intentional for performance reasons.  This is one of the few places that we really work hard to optimize because it gets called so often.

I'm going to close this request and suggest you work on solving the actual bug you are experiencing (i.e. why is the file getting included by the autoloader if it has already been included somewhere else (i.e. the class should be defined).

mbabker:
FWIW, this is probably happening on a CMS 2.5 instance since when JGrid was originally added to the Platform, the file was in the location the autoloader expected to find the class JHtmlGrid (see #1317).

pasamio:
So CMS 3.0 will be fine, CMS 2.5 needs to work out what it wants to do with the problem then.

elinw:
@mbabker Do you want to backport #1317?

ssv445:
As per thread - http://stackoverflow.com/questions/186338/why-is-require-once-so-bad-to-use
I would like to suggest that _once form does not have a very big performance problem.

The real bug is inside CMS 2.5, but as a component developer I cannot fix it. 
I think a bad code or convention might break the system easily, these types of code makes joomla environment very complex (as many 3pd developers are there, and who knows, how many vendor's code is there). 

So I would request to re-check the issue and should provide some basic mechanism for such a unexpected faults.

Backporting should be done. I will look into that. But it does not guarantee the same problem come through a different path in future.

Thanks
Shyam

pasamio:
If the code is fixed then there won't be any performance issues of any sort. Masking the issue of an incorrectly named or positioned class is not a solution: it's a bad workaround to bandage a mistake that needs to be rectified. Instead of masking errors, the attempt should be to fix the original problem. And if you read that thread closely you will see there are performance impacts on using include_once over include. By definition it has to have an extra check to internally validate if a file has been included versus the majority use case of this function which is that a given file hasn't been included because the class doesn't exist already. The assumption is that people have correctly located their files with appropriately named classes. If this isn't the case then it should be fixed and not masked.

This is the platform project and this particular issue has been fixed in the platform. It is up to the CMS to fix the issue in 2.5 however the place for discussing this is not here. Please move your discussion to the CMS project either the mailing list or issue tracker.

elinw:
@ssv445  WHy don't you just make a pull request to the CMS 2.5.x branch to backport the fix?
I''m not sure why "as a component developer I can't fix" what does that mean? I would think that component developers would be the biggest source of bug fixes and new code in the CMS.

PR #1662, opened 2012-11-02
Complying with new 'WhiteSpaceBefore' Standard

dongilbert:
This _finally_ finishes code style fixing on the legacy tree, as long as more standards aren't added later. :p

### DIFF
--- a/tests/suites/legacy/request/JFilterInput-mock-general.php
+++ b/tests/suites/legacy/request/JFilterInput-mock-general.php
@@ -37,11 +37,13 @@ class JFilterInputJRequest
 	public function clean($source, $type = 'string')
 	{
 		$hash = md5($source . '|' . strtoupper($type));
+
 		if (!isset($this->_expectations[$hash]))
 		{
 			$this->_expectations[$hash] = array('source' => $source, 'type' => $type, 'result' => null, 'count' => 0);
 		}
 		--$this->_expectations[$hash]['count'];
+
 		return $this->_expectations[$hash]['result'];
 	}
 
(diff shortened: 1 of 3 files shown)


tests/suites/legacy/request/JFilterInput-mock-general.php
```diff
@@ -37,11 +37,13 @@ class JFilterInputJRequest
 	public function clean($source, $type = 'string')
 	{
 		$hash = md5($source . '|' . strtoupper($type));
+
 		if (!isset($this->_expectations[$hash]))
 		{
 			$this->_expectations[$hash] = array('source' => $source, 'type' => $type, 'result' => null, 'count' => 0);
 		}
 		--$this->_expectations[$hash]['count'];
```
mbabker, line 45: I know this line has nothing to do with the pull request, but why would the line be prefaced like this?  Seems like there's an error in the tests, unless it's a syntax I've never seen.

elinw, line 45: That's very odd.

dongilbert, line 45: I thought the same thing. But then realized its just decreasing the value by one. It looks odd because your expecting the single arrow before $this, instead of the de-increment operator. For clarity, this _should_ go after if it's not for a return value.

Blame @ianmacl - https://github.com/dongilbert/joomla-platform/blame/8f1a5ced1c4f39f77798fa641dd997c6697c840c/tests/suites/legacy/request/JFilterInput-mock-general.php#L45

dongilbert, line 45: Explanation: http://php.net/manual/en/language.operators.increment.php

PR #1664, opened 2012-11-04
Adding _e method to compliment commonly echo'd class underscore methods

dongilbert:
The static underscore methods of `JHtml`, `JRoute` and `JText` are (IMHO) preceded by the `echo` command more often than not.

This PR adds some syntatic sugar to those underscore methods, making it nicer for developers. If implemented, you would replace calls like this:

```
echo JHtml::_('tabs.start', 'tab_id', $options);
// would become
JHtml::_e('tabs.start', 'tab_id', $options);

echo JRoute::_('index.php?option=com_content');
// would become
JRoute::_e('index.php?option=com_content');

echo JText::_('JDEFAULT');
// would become
JText::_e('JDEFAULT');
```

This isn't a huge change, but I think it's a nice addition, and something that I would use extensively.

### DIFF
--- a/libraries/joomla/application/route.php
+++ b/libraries/joomla/application/route.php
@@ -105,4 +105,24 @@ class JRoute
 
 		return $url;
 	}
+
+	/**
+	 * Translates an internal Joomla URL to a humanly readible URL, and echo's it back.
+	 *
+	 * @param   string   $url    Absolute or Relative URI to Joomla resource.
+	 * @param   boolean  $xhtml  Replace & by &amp; for XML compilance.
+	 * @param   integer  $ssl    Secure state for the resolved URI.
+	 *                             1: Make URI secure using global secure site URI.
+	 *                             0: Leave URI in the same secure state as it was passed to the function.
+	 *                            -1: Make URI unsecure using the global unsecure site URI.
+	 *
+	 * @return  The translated humanly readible URL.
+	 *
+	 * @see     JRoute::_()
+	 * @since   12.3
+	 */
+	public static function _e($url, $xhtml = true, $ssl = null)
+	{
+		echo self::_($url, $xhtml, $ssl);
+	}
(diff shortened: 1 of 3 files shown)


ianmacl:
I'm not a big fan of this.  It obfuscates what is going on at a casual glance and makes code less readable.

elkuku:
I have to confess, that I'm in love with PHP's short open tags which have recently being undeprecated...
So

``` php
<?php echo JRoute::_('index.php?option=com_content'); ?>
```

would become

``` php
<?= JRoute::_('index.php?option=com_content'); ?>
```

without any added code ;)

dongilbert:
Yes, i would agree with that. But I don't think the Joomla coding standard allows for short tags.

ianmacl:
Yeah, we can't allow short tags until we move our requirement to PHP 5.4, which won't be quite yet.

elinw:
Too many hosts do not support short tags.

ianmacl:
<?= ?> cannot be disabled on PHP 5.4.

elkuku:
Well, this only makes sense when mixing PHP with HTML code. e.g. a view template. Since there are no view templates in the platform the question would be if a downstream user (a CMS maybe) would like to implement those methods.

What is possible, in therms of short tags, is to check if they are disabled and, with a small speed penalty, pre-parse the template and replace the short tags with "long" ones.

I would prefer this instead of implementing a new method. Hopefully it will not be long until we have PHP 5.4 available for playing ;)

Implementing the usage of short open tags in the Joomla! CMS would be a nice sophisticated idea - IMHO.

dongilbert:
Well, if there's not broad support for this, I will close the PR and patiently await the deprecation of PHP 5.3 support.

dongilbert:
Closing until more support is shown, if ever. :P

PR #1666, opened 2012-11-04
Implementing a new event system

florianv:
This is https://github.com/joomla/joomla-platform/pull/1621.

I added unit tests and changed some things.
## Resume : 

All event listeners must take a `JEvent` object as sole parameter.

``` php
<?php

class Listener
{
    public function onBeforeSomething(JEvent $event)
    {

    }

    public function onAfterSomething(JEvent $event)
    {

    }
}

$dispatcher = JEventDispatcher::getInstance();

// The listener is registered to events
// corresponding to it's method names.
$dispatcher->registerListener(new Listener);

// The listener can be registered only
// to some specific events.
$dispatcher->registerListener(
    new Listener, 
    array('onBeforeSomething')
);

/** 
 *  The listener can be registered to
 *  some events with a given priority.
 *  The default priority is 0.
 *  If some listeners have the same priority,
 *  they will be called in the order they were
 *  registered.
 **/
$dispatcher->registerListener(
    new Listener, 
    array()
    array('onBeforeSomething' => 4)
);
```

Closures

``` php
<?php

$listener = function(JEvent $event) {
};

$dispatcher = JEventDispatcher::getInstance();

// Closure listeners must be explicitly
// registered to a given event (or more)
$dispatcher->registerListener(
    $listener, 
    array('onBeforeSomething')
);

// Specifying a priority.
$dispatcher->registerListener(
    $listener, 
    array('onBeforeSomething')
    array('onBeforeSomething' => 4)
);
```

Closures and 'normal' objects can be used together in the dispatcher.
Additionnaly, the event propagation can be stopped by a listener.

``` php
<?php

class ListenerOne
{
    public function onBeforeSomething(JEvent $event)
    {
        echo 'It is enough for me';

        // Stopping the event propagation.
        $event->stopPropagation();
    }
}

$listenerTwo = function (JEvent $event)
{
    echo 'Please pass me this event!';
};

$dispatcher = JEventDispatcher::getInstance();

$dispatcher->registerListener(new ListenerOne);
$dispatcher->registerListener(
    $listenerTwo,
    array('onBeforeSomething')
);

$dispatcher->triggerEvent('onBeforeSomething');
```

By default a JEvent object is created and passed to the listeners
when triggering the event.
You can inject your custom event object using two methods :

``` php
<?php

// Create the onBeforeSomething event with special arguments.
$onBeforeSomething = new JEvent('onBeforeSomething');
$onBeforeSomething->setArgument('foo', 'bar');

$dispatcher = JEventDispatcher::getInstance();

// First method : registering a custom event.
$dispatcher->registerEvent($onBeforeSomething);

// Second method : when triggering the event.
$dispatcher->triggerEvent($onBeforeSomething);
```


pasamio:
This still isn't mergeable. Also you don't need to create a new pull request if you push up new or rebased commits back to the same branch, it'll update the pull request once it's reopened.

florianv:
I did a mistake with the previous one concerning the actual event package.
I prefered to open a new one and put the package in legacy, so it will be fine from now (I rebased it a few times already).

Btw, I don't get why it is not mergeable, could it be a problem of file mode or something ? I created the branch from the joomla staging branch yesterday.

pasamio:
A few things:
1) Make sure you have a remote for the platform project (`git remote add joomla git://github.com/joomla/joomla-platform.git`)
2) Make sure you do a `git fetch --all` to get all of the changes
3) When you do a rebase use the remote branch not your local branch: `git rebase joomla/master`

The staging or master branch in your fork won't automatically update, you have to manually update it. You need to checkout the branch and update it from the main platform branch.

I personally prefer to checkout new branches with the following: `git checkout -b [newbranchname] joomla/master`

Doing this (with a `git fetch --all` prior) ensures that I'm always up to date with the platform when creating a new branch. 

If you look at your commit in GitHub, you get this page:
https://github.com/florianv/joomla-platform/commit/e05dba591bf56464f7896421ef4ed96206f5ddf0

If you click on the commit ID next to the text "parent" you land on the following page:
https://github.com/florianv/joomla-platform/commit/6d6c3310cba3c71270cb3b182a6089ee27b86d78

Which is over a month old and appears to be the head of your own staging brach:
https://github.com/florianv/joomla-platform

florianv:
Thanks @pasamio. I think it is correct now

pasamio:
Yeah, this is now mergeable.

libraries/joomla/event/dispatcher.php
```diff
@@ -74,205 +63,361 @@ public static function getInstance()
...
-	 * @return  array  An array of results from each function call.
+	 * @return  JEventDispatcher  This method is chainable.
 	 *
-	 * @since   11.1
+	 * @throws  InvalidArgumentException
+	 *
+	 * @since   12.3
 	 */
-	public function trigger($event, $args = array())
+	public function registerListener($listener, array $events = array(), array $priorities = array())
```
eddieajau, line 150: Have you looked at using an SplPriorityQueue for the $priorities?  Would that work?

libraries/joomla/event/event.php
```diff
@@ -10,69 +10,191 @@
...
+ * The event can contain arguments and its listeners can manipulate them.
+ * Additionnaly, its propagation can be stopped.
  *
  * @package     Joomla.Platform
  * @subpackage  Event
- * @since       11.1
+ * @since       12.3
  */
-abstract class JEvent extends JObject
+class JEvent implements Serializable, Countable
```
eddieajau, line 21: Ironically, I think your new JEvent would benefit from my proposal to change JObject.  See https://github.com/joomla/joomla-platform/pull/1673/files#diff-3 (you can use the "properties" of JObject as your "arguments").  We could also add the Countable interface to the new JObject since that is related to IteratorAggregate.  That would trim this class down a bit.

eddieajau:
Obviously a bit of work to do, and some backward compatibility issues to think about, but I like where this is heading.

florianv:
Yes, it is possible to use SplPriorityQueue instead of an array.

Also, I put three parameters because I didn't want to force people to specify a priority per event, but finally I'm not sure it would be a big problem.

In this case it would become :

``` php
registerListener($listener, SplPriorityQueue $events = null)
```

For the closures many frameworks support them as event listeners so I did it.
I think it can be interesting as 'little' event listeners.
But on the other side, it could encourage people to do non OOP stuff..so I don't know what you think about it.

In the case we don't support them, I was thinking about having a tagging interface for event listeners

``` php

// Tagging interface for event listeners
// All event listeners implement this interface
interface JEventListener
{
}
```

which could be useful for type hinting and to quickly see if a class is designed to listen to some events, for example JApplicationWeb.

So, in this case we would have : 

``` php
registerListener(JEventListener $listener, SplPriorityQueue $events = null)
```

which looks better but we drop the closures.

And for JEvent, yes, it should extend the new JObject.

ghost:
A priorityQueue is not an array, it is a heap - so many of your methods that deal with intersecting and diffing arrays will not work for a queue.  Also keep in mind: splPriorityQueue is an object, this means it is passed by reference - any changes to it will change the base object.   Reading a heap is a destructive action, ie
foreach($priorities as $p)

Will read the queue in priority order and do something, when done the queue will be empty - so if you have an 'array' of events you just deleted them all if you didn't clone the queue before processing.

florianv:
@eddieajau I fixed an issue with SplPriorityQueue dequeing elements with the same priority and added more tests. I think it is good for further review.

ronnikc:
A new event system sure is on my list of things that id like to see for the benefit of us all :)

florianv:
I think we can benefit having `JEventInterface` and `JEventDispatcherInterface` but didn't know which name to use for them.

AmyStephen:
@redcomponent Are there capabilities you are missing from Joomla's event system? Interested in understanding that, esp. from an ecommerce POV.

@florianv - few questions.

I think events are a particular challenge for uncoupling. What approaches would you advise to prevent a dependency on JEvent and yet still ensure an object is able to schedule events or register plugins? (Can dependency injection help here?)

Also, interested in hearing how do you see Interfaces as potentially helping. Do you see Interfaces as potential "filters" (like tags) that could be used in addition to registration for a specific event? And, in regards to the coupling, what impact does type hinting have on coupling?

Also (as long as I am here!) - if we are going to modify the Event system, should we add criteria to the dispatcher where an event could be scheduled "until first failure" or "until first success?" That would be very useful for stepping up the game of JEvent to help out as a workflow engine.

Your examples are very helpful. Takes me time to think it all through, and my head is not completely around this yet, but do like the potential I am seeing.

Also,

florianv:
Interfaces are useful because people can implement them how they want to.

If someone wants to make his own `Dispatcher` implementation instead of using `JEventDispatcher`, it can still be compatible with `JEvent`.
If someone wants to make his own Event implementation he can still use `JEventDispatcher` to dispatch them.

An object that needs to notify listeners depends on the dispatcher.

So it could implement : 

``` php
<?php

interface DispatcherAwareInterface
{
   public function setDispatcher(DispatcherInterface $dispatcher);
}

class Something implements DispatcherAwareInterface
{
  protected $dispatcher;
  protected $onAfterSomething;
  protected $onBeforeSomething;

 public function __construct(DispatcherInterface $dispatcher)
 {
    $this->dispatcher = $dispatcher;
 }

 public function setOnBeforeSomethingEvent(OnBeforeSomethingEvent $event)
 {
    $this->onBeforeSomething = $event;
    return $this;
 }

 public function setOnAfterSomethingEvent(OnAfterSomethingEvent $event)
 {
   $this->onAfterSomething = $event;
   return $this;
 }

 public function registerOnBeforeSomethingListener($listener, $priority = 0)
 {
   $this->dispatcher->registerListener(
      $listener, 
      array('OnBeforeSomething'),       
      array('OnBeforeSomething' => $priority)
    );
 }

 public function registerOnAfterSomethingListener($listener, $priority = 0)
 {
   $this->dispatcher->registerListener(
      $listener, 
      array('OnAfterSomething'),       
      array('OnAfterSomething' => $priority)
    );
 }

 public function doSomething()
 {
   $this->dispatcher->trigger($this->onBeforeSomething);

   // Something

   $this->dispatcher->trigger($this->onAfterSomething);
 }

 public function setDispatcher(DispatcherInterface $dispatcher)
 {
    $this->dispatcher = $dispatcher;
    return $this;
 }
}

$dispatcher = new Dispatcher;
$something = new Something($dispatcher);

$onBeforeSomethingEvent = new OnBeforeSomethingEvent;

$onBeforeSomethingListener = function (OnBeforeSomethingEvent $event) {
    echo 'onbeforesomething';
};

$something->setOnBeforeSomethingEvent($onBeforeSomethingEvent);
$something->registerOnBeforeSomethingListener($onBeforeSomethingListener);

$something->doSomething();
```

Something is easy to test because the dispatcher and events can be mocked.
The listeners are easy to test because the event can be mocked.

There are some cases where the listeners need an instance of the object they are listening to.

``` php
class SessionEvent extends JEvent
{
   public function setSession(SessionInterface $session)
   {
     $this->setArgument('session', $session);
     return $this;
   }

   public function getSession($default = null)
   {
     return $this->getArgument('session', $default);
   }
}

class Session implements DispatcherAwareInterface, SessionInterface
{
   protected $dispatcher;
   protected $sessionEvent;

   public function setDispatcher(DispatcherInterface $dispatcher)
   {
     $this->dispatcher = $dispatcher;
   }

   public function setSessionEvent(SessionEvent $event)
   { 
     $this->sessionEvent = $event;
     $event->setSession($this);
   }
}

class SessionEventListener
{
   public function onSessionStart(SessionEvent $event)
   {
     $session = $event->getSession();
     // Verify the session token is matching..
    // Otherwise stop the event propagation and destroy the session
   }
}
```

AmyStephen:
Do you have any unit tests for that code? (Man, can he code quickly!)
- It appears the above is your response to " interested in hearing how do you see Interfaces as potentially helping", and your answer is essentially that it is very wide open, people can use interfaces as they wish, even to the point of building a Dispatcher independent of JDispatcher but compatible with JEvent. 

So, is that basically what you are saying?
- Regarding this comment you made: "An object that needs to notify listeners depends on the dispatcher."

Exactly, so, how can we better control which object is dependent and prevent JModel, for example, from dependence on JEvent -- but still get those Events scheduled?

Would the use of closures/anonymous functions, for example, on a JModel method parameter where the callback was outside of the class -- be an alternative to directly naming JEventDispatcher in the code? 

Any ideas of that nature how to keep from coupling those together but still keep the capability?
- What do you think of adding these capabilities to the JEvent?  "until first failure" or "until first success?"

ronnikc:
An event system in a multitier non-excluding system should not need to specify first error or success - but n error or n success or any other, in my opinion, given sub-event or "tag" to really be a home for a multipurpose event system that can be used not only for interaction between different elements or entities (be it core or 3rd party) but also as a base of workflows.

So parts of this is to allow for events to register and be used by all other entities but part is also making it completely independant so it could be used stand alone.

In terms of an ecommerce point of view - or specifically a point of view for 3rd parties in general the waste possibilities in a mature structure that does not exclude nor presets all events (as in Drupal) would greately make it possible for different elements of a Joomla site or application to utilize subparts from others in a more smooth and global approach instead of doing it staticly like today.

As for tagging or other sub  or 2nd tier capabilities like failures, success, passed test or what ever can be imagined i think its quite crucial in terms of allowing for an organic non-static structure to form.

Which form the multi-tier takes is not so important for me - whats important is that is multi tier so there is a way to couple action and type of action and also to register, read and utilize from any source that is connected to the event system.

I hope it didnt get too arbitrary :)

florianv:
1. yes

I think the model must do it model's work. If you want to add event capabilities to a model then it is an other layer (it is for the CMS) or you must implement it.

Ideally all packages must be standalone and it's the CMS that does the glue.

eddieajau:
@florianv, I'm going to close this PR but I'd like you to keep working on this package for the new Framework. Thanks!

dongilbert:
:+1: florian, I'd like to see this in the framework.

florianv:
Ok, I will review it and make a pull request to the framework in one week or so.

PR #1667, opened 2012-11-08
Adding JRequest Tests

dongilbert:
The `JRequestTest` class only contains tests for methods that are agnostic to the `$_SERVER['REQUEST_METHOD']`.

I broke up tests specifically for `$_POST` and `$_GET` into their own classes, `JRequestPostMethodTest` and `JRequestGetMethodTest` respectively.


tests/suites/legacy/request/JRequestTest.php
```diff
@@ -19,160 +19,191 @@
...
-	 * @todo    Implement testGetMethod().
-	 *
 	 * @return  void
 	 */
 	public function testGetMethod()
 	{
-		// Remove the following lines when you implement this test.
-		$this->markTestIncomplete('This test has not been implemented yet.');
+		$_SERVER['REQUEST_METHOD'] = 'post';
+		$this->assertEquals('POST', JRequest::getMethod());
```
ianmacl: Please break these up into separate tests (i.e. make a method for each test).

tests/suites/legacy/request/JRequestTest.php
```diff
@@ -19,160 +19,191 @@
...
 	 *
-	 * @todo    Implement testGetVar().
-	 *
 	 * @return  void
 	 */
 	public function testGetVar()
 	{
-		// Remove the following lines when you implement this test.
-		$this->markTestIncomplete('This test has not been implemented yet.');
+		$this->assertEquals('from _POST', JRequest::getVar('tag', null, 'post'));
```
ianmacl: These also should be three different test methods.

tests/suites/legacy/request/JRequestTest.php
```diff
@@ -19,160 +19,191 @@
...
 	 *
-	 * @todo    Implement testGetInt().
-	 *
 	 * @return  void
 	 */
 	public function testGetInt()
 	{
-		// Remove the following lines when you implement this test.
-		$this->markTestIncomplete('This test has not been implemented yet.');
+		$_GET['teststr'] = '1';
```
ianmacl: Break these up into two methods.

tests/suites/legacy/request/JRequestTest.php
```diff
@@ -19,160 +19,191 @@
...
 	 *
-	 * @todo    Implement testGetFloat().
-	 *
 	 * @return  void
 	 */
 	public function testGetFloat()
 	{
-		// Remove the following lines when you implement this test.
-		$this->markTestIncomplete('This test has not been implemented yet.');
+		$_POST['teststr'] = '1.320';
```
ianmacl: Ditto here.

tests/suites/legacy/request/JRequestTest.php
```diff
@@ -19,160 +19,191 @@
...
 	 *
-	 * @todo    Implement testGetBool().
-	 *
 	 * @return  void
 	 */
 	public function testGetBool()
 	{
-		// Remove the following lines when you implement this test.
-		$this->markTestIncomplete('This test has not been implemented yet.');
+		$_GET['teststr'] = 'true';
```
ianmacl: Again here.  As an explanation, each test should be testing only one scenario.  More granular tests means that we get better information back from unit test reports (i.e. we can tell that the issue with getBool only happens when it is false).

The other thing to note here is that $_GET, $_POST etc don't get populated with integers, booleans, floats, etc.  If you do a var_dump of a typical request super global you will find that it contains a string.  You should change your tests to match this so that they more closely represent the real world scenario.

dongilbert: Gotcha. I'll get that taken care of and pushed back.

dongilbert:
@ianmacl - I see now why some of the tests were broken up into separate test classes. It makes sense to break them up across test classes, since we need to test each REQUEST_METHOD.

ianmacl:
Looks better.  Fix up the CS issue and then I'll merge it.

dongilbert:
http://developer.joomla.org/pulls/pulls/1667.html

The one CS issue has nothing to do with the pull request or even the test tree.

It's that a blank line needs to be added before the if statement on line 1650 of libraries/joomla/database/query.php Should I fix it in this PR?

dongilbert:
Fixed CS issue in #1674

PR #1671, opened 2012-11-09
Use PHP_EOL instead of \n in tests

okonomiyaki3000:
The actual functions use `PHP_EOL` so the tests should also use them.


LouisLandry:
Thanks!

PR #1672, opened 2012-11-09
Cleaning up structure of JPathway Tests

dongilbert:
Changing JPathwayTest to use a fixture. Each method currently creates a new instance of `JPathwayInspector`, so I abstracted that to the `setUp()` method instead. Also moved inspector class to it's own file.


ianmacl:
I'm going to merge this because it is an improvement.  It should be noted however that for the most part the inspectors are obsolete.  We now have the TestReflection class which has a invoke, getValue and setValue methods that for 99% of cases takes care of everything we were using the inspectors for.

PR #1673, opened 2012-11-09
Object package enhancements

eddieajau:
#### Overview

The `JObject` class has been around for a long time, built in a time before magic methods were available. It is useful to a point, but has some severe limitations particularly when trying to separate class properties that are 'values' held in the object vs properties that can be used for domain logic or other purposes.  This pull request refactors `JObject` to better fit the Value Object pattern in a backward compatible way, embracing magic methods available in PHP 5 and adds some additional functionality.

The `get` and `set` methods are deprecated in favour of using direct property accessors (`$object->key`) that are controlled by magic methods.  The properties are held in a private array within the class, completely separate from the real class properties.  The `setProperties` method is deprecated in favour of a new `bind` method which is used elsewhere in the API (cf. JTable->bind).  A new 'dump' method that allows you to obtain a `stdClass` representation of the object.  This is also used to make the class implement the `JsonSerializable` interface.

A `JObject` can also act as an array iterator allowing you to use it in a `foreach` construct.

Backward compatibility is fully maintained but please refer to the documentation for a subtle difference in behaviour between the old `getProperties` method and the new `dump` method.

A new `JObjectList` class is also provided.  This class provides a convenient way to work with lists of `JObject` objects.  Aside from the regular iterator functionality, the magic get method acts like a "get column" operation on the entire list, and the magic set method acts like a "set column" operation on the entire list.  The magic call method allows you to call a method (if available) on each of the objects in the list.  See the documentation for more information. 

Additional notes and comment can be found in the documentation.
#### Documentation

Documentation is provided with the pull request (in markdown format).

https://github.com/eBaySF/joomla-platform/blob/object/docs/manual/en-US/chapters/classes/jobject.md
https://github.com/eBaySF/joomla-platform/blob/object/docs/manual/en-US/chapters/classes/jobjectlist.md
#### Business Case

As already stated, the current `JObject` class has severe limitations when trying to merge domain logic with the Value Object pattern.  This can be seen in classes like `JTable` and particularly `JUser` where you need to perform some coding gymnastics to avoid column mismatches when saving to the database.  To this end, we needed a better way [for eBay applications] to interact with the value object itself, and also so it could interact with a Data Mapper pattern (a long term alternative/replacement for the JTable-way of doing things) so that the value object could remain independent of the data source (unlike JTable which is a value object that is tightly coupled to a specific datasource).


libraries/joomla/object/list.php
```diff
@@ -0,0 +1,443 @@
+<?php
+/**
+ * @package     Joomla.Platform
+ * @subpackage  Object
+ *
+ * @copyright   Copyright (C) 2005 - 2011 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE
+ */
+
+// @codeCoverageIgnoreStart
```
LouisLandry: I don't think this are particularly necessary, but not a big deal.

libraries/joomla/object/object.php
```diff
@@ -7,111 +7,284 @@
...
- * Joomla Platform Object Class
- *
- * This class allows for simple but smart objects with get and set methods
- * and an internal error handler.
+ * JObject is a 'smart' value object.
  *
  * @package     Joomla.Platform
  * @subpackage  Object
- * @since       11.1
+ * @since       12.3
```
LouisLandry: Why are we changing this?  The class has still existed since 11.1

libraries/joomla/object/object.php
```diff
@@ -7,111 +7,284 @@
...
+	 * @since   12.3
+	 * @throws  InvalidArgumentException
+	 */
+	public function bind($properties, $updateNulls = true)
+	{
+		// Check the properties data type.
+		if (!is_array($properties) && !is_object($properties))
 		{
-			return $this->$property;
+			throw new InvalidArgumentException(JText::sprintf('JOBJECT_INVALID_BIND_DATA', gettype($properties)));
```
LouisLandry: Let's get rid of the translation bit with this exception message.

LouisLandry:
It will come as no surprise that I'm a fan of the direction.  Just a few things to clean up and you get a :+1: from me.

HermanPeeren:
A Data Transfer Object (DTO) is a better name for what you try to do here than a Value Object. The latter name is used quite different nowadays (and causes confusion when used for this JObject). See Fowler's PEAA, page 486-487 (Value Object) and 401-413 (Data Transfer Object).

A downside of overusing this kind of objects is a data-oriented focus. So it should be handled with care if you want OOP. It can be handy, but I wonder if it is a good choice as a kind of basic-object, like JObject was.

eddieajau:
On 9 November 2012 19:37, HermanPeeren notifications@github.com wrote:

> A Data Transfer Object (DTO) is a better name for what you try to do here
> than a Value Object. The latter name is used quite different nowadays (and
> causes confusion when used for this JObject). See Fowler's PEAA, page
> 486-487 (Value Object) and 401-413 (Data Transfer Object).
> 
> I think JObject is fine to consider as a Value Object.  The Data Transfer
> Object is next on my list.
> 
> A downside of overusing this kind of objects is a data-oriented focus. So
> it should be handled with care if you want OOP. It can be handy, but I
> wonder if it is a good choice as a kind of basic-object, like JObject was.
> 
> Yes, this is definitely about changing JObject so we can get to the real
> data, not use some adhoc rules about the class properties (which may or may
> not be data related).  For example, it's a better way to handle things than
> assuming that all the class properties in a JTable class map to database
> table fields.

Regards,
Andrew Eddie

HermanPeeren:
A.E.: "I think JObject is fine to consider as a Value Object". I think it is better to stick to the generally used definitions: that improves communication. What is your definition of a "Value Object" as used here?

Concerning mapping: that is why I use a ORM. No need to reinvent that wheel.

HermanPeeren:
I like the examples; especially of JObjectList. These classes can be nice toys to play with.

But: none of the methods of the current JObject survives platform 13.1 in this proposal. The new JObject serves a different goal (has a different "responsability"). In fact: it is a totally different class, with the same name. Is that strictly necessary? I cannot see that "This pull request refactors JObject (...) in a backward compatible way". 

Does the name JObject tell us something about it's use? You want an object to store data in it. Wouldn't JData (and JDataList) be a better name? Or something else; anything better than "JObject": that suggests a basic object from which you extend other objects. From an OOP point of view, this class is not good for that; too data-centric.

elinw:
Remember that we are not normally extending JObject any more.

HermanPeeren:
elinw: "Remember that we are not normally extending JObject any more".

Exactly: that is why JObject is not such a good name for this new class.

eddieajau:
On 9 November 2012 23:41, HermanPeeren notifications@github.com wrote:

> But: none of the methods of the current JObject survives platform 13.1 in
> this proposal.
> 
> Well, you can propose a different time scale, but honestly who would rather
> use $object->get('property') over $object->property.  I don't understand
> the push back on using standard PHP nomenclature.
> 
> The new JObject serves a different goal (has a different
> "responsability"). In fact: it is a totally different class, with the same
> name. Is that strictly necessary?
> 
> I'm sorry, I don't see how you can say that.  It's performing the same job
> but doing it in a PHP 5 way instead of PHP 4.
> 
> I cannot see that "This pull request refactors JObject (...) in a backward
> compatible way".
> 
> I think that's a pretty simple statement.  It means that JObject will
> continue to work as expected.
> 
> Does the name JObject tell us something about it's use? You want an object
> to store data in it. Wouldn't JData (and JDataList) be a better name?
> 
> Possibly, but the name JObject was chosen a long time ago.
> 
> Or something else; anything better than "JObject": that suggests a basic
> object from which you extend other objects. From an OOP point of view, this
> class is not good for that; too data-centric.
> 
> I really don't get your point.  JObject IS data centric and it's consistent
> with the way I've used it, and what I consider it was supposed to be used
> for, for many, many years.

Regards,
Andrew Eddie

HermanPeeren:
JObject was a base object, to extend most other objects from. Hence it's name. It provided some functionality that was not available in PHP4; that is why it is not necessary anymore in current Joomla: PHP5 provides everything we need from a basic object. We don't need another basic object for general use anymore. 

If you want to provide some specific behaviour in an object, then properly name that object, so it's use is obvious. JData, JStore, JContainer, can't you think of a name describing the kind of object? Or do you want to go back to a general basic object from which most other objects are extended? Thought we had passed that idea long ago.

eddieajau:
On 9 November 2012 20:33, HermanPeeren notifications@github.com wrote:

> A.E.: "I think JObject is fine to consider as a Value Object". I think it
> is better to stick to the generally used definitions: that improves
> communication. What is your definition of a "Value Object" as used here?
> 
> An object that needs to hold properties but does not require persistence.
>  JObject can obviously be extended to go beyond that but that's where it
> starts.  Whatever the case, we should talk about the changes to the code,
> not philosophical differences of opinion.  The main change it leveraging
> the magic methods and it's fully backward compatible.  Do you not approve?
> 
> Concerning mapping: that is why I use a ORM. No need to reinvent that
> wheel.
> 
> I'm just telling you how we use it.  You can use Symfony or whatever takes
> your fancy, just don't force me to use it too.

Regards,
Andrew Eddie

eddieajau:
On 10 November 2012 08:30, HermanPeeren notifications@github.com wrote:

> JObject was a base object, to extend most other objects from. Hence it's
> name. It provided some functionality that was not available in PHP4; that
> is why it is not necessary anymore in current Joomla: PHP5 provides
> everything we need from a basic object. We don't need another basic object
> for general use anymore.
> 
> I understand your opinion, I just don't agree with it.
> 
>  If you want to provide some specific behaviour in an object, then
> properly name that object, so it's use is obvious. JData, JStore,
> JContainer, can't you think of a name describing the kind of object? Or do
> you want to go back to a general basic object from which most other objects
> are extended? Thought we had passed that idea long ago.
> 
> I think JObject is just fine - JData, JStore or JContainer are just
> synonyms with equally obscure means for me.  The fact is, we have a pattern
> for a Value "OBJECT" (<- note "object" is used in the pattern name) so
> JObject fits.  I read Folwer's book last night (the book, not the summary
> on the web site) and it fits the pattern … in my opinion and you have to
> realise that patterns have a subjective element to them.  I'm totally fine
> with you disagreeing, but I don't see any problem with bringing JObject
> into PHP 5 land rather than inventing a new name just for the same of
> inventing a new name.

Regards,
Andrew Eddie

HermanPeeren:
You call it a Value Object and even refer to Fowler, but the JObject you propose here does not have the same function: a small simple object, like money or a date range, whose equality isn't based on identity (and yes I 've reread the whole book often since 2004). A Value Object is used everywhere for instance to contain money (the amount and the currency together). No identity means: it doesn't matter if it is this 5 dollars or that; it's just 5 dollars. 

I won't bother you with my arguments anymore. Good luck with your platform.

eddieajau:
I read and re-read the entire chapter on Value Object and, to me, it's fits
with the description.  Regardless, functionally it's the same as the
current JObject that all Joomla developers are familiar with.  We don't
need to invent something new to do the same job.  I don't mind if you don't
agree with calling it a value object, but I'd don't see how upgrading the
class to the PHP 5 way suddenly changes it (when it's completely backward
compatible).

Regards,
Andrew Eddie

Regards,
Andrew Eddie
http://www.theartofjoomla.com - the art of becoming a Joomla developer

HermanPeeren:
I said I won't bother you with my arguments anymore, because I realised it is useless: I speak another "language". Our basic understanding of Object Oriented Programming is totally different. That happens often and you are in good company. Although useless, I'll just state some things that could once be of help; for you or others.

I often recommend Eric Evans' book on DDD (also from 2004) as a good book to learn to understand Fowler's PEAA. Evans gives a good explaination about what "identity" is, and how the lack of that gives Value Objects. See page 89 - 96 on Entities and 97 - 103 on Value Objects. Please read to understand, instead of misinterpretating to show you are right. Try to understand what "identity" is and what makes up a Value Object. It is not some  vague philosophical discussion. When you talk to people who use notions like Value Object on a daily basis, you'll avoid the trap of redefining it in a different, private and confusing way.

Another book that could help understand the basics of Object Thinking (as opposed to data-centric, procedural thinking) is: Object Thinking by David West. Did I say 2004 was a good year?

I'll state my opinion about JObject one more time: that object was needed to add some basic functionality to objects that was lacking in PHP4. Since we now use PHP5, JObject is not needed anymore: the functionality it provided is now provided by PHP5. Some error-handling functionality is now handled elsewhere. 

So, JObject has become obsolete. After deprecation it can be deleted from the Joomla Platform. 

Reviving it to use it as an "object that needs to hold properties but does not require persistence" is giving another role to that object. Your statement: " JObject IS data centric and it's consistent with the way I've used it, and what I consider it was supposed to be used for, for many, many years" shows the difference in understanding of OOP we have.

eddieajau:
@HermanPeeren on reflection, I'm happy to concede we are not religiously following the Value Object pattern (it really depends on the data you push into the object).  If you like, I can change the docs to say something like "modelled on the Value Object and other patterns".  That said, there is absolutely no intention of deprecating JObject in the short, medium or long term.  What has happened is the we've stopped deriving from JObject just because we can.  I've also always considered JObject as a data object, a replacement for JParameter and other things even if other people have had different ideas about what it's real purpose was or is - so that could also be where we have differing views on what the "right" way is.

That said, this PR doesn't change the API (in a backward incompatible way) so I think it's fine to merge and makes the class more useful - you don't have to use the class if you don't want to ;)

eddieajau:
Code style and version numbers fixed up.  Docs consolidated per the new format.

florianv:
Would it make sense to implement Countable in JObject ?

eddieajau:
Closing this pull request.  See #1703.  I think that forms a reasonable compromise to move forward with.

PR #1675, opened 2012-11-09
Add deprecation tags to legacy classes.

LouisLandry:
Most of these have been marked for removal at 13.3.  This is roughly a year of having them stick around in the legacy tree.  My suspicion is that most -- if not all -- of these classes will end up being picked up and maintained by the CMS project for much longer than Q3 of 2013.


eddieajau:
Looks fine to me.

elinw:
+1

PR #1677, opened 2012-11-10
Update libraries/legacy/updater/adapters/extension.php

Bakual:
The helper function "getClientInfo" takes two arguments, the search "$id" and "$byName" which decides if it should search by id or name.
In our case here we want to search by the name and thus it should be set to "1". Otherwise the function will return null and break all extension updates if the client tag is set instead of the client_id in the updater xml.

### DIFF
--- a/libraries/legacy/updater/adapters/extension.php
+++ b/libraries/legacy/updater/adapters/extension.php
@@ -205,7 +205,7 @@ class JUpdaterExtension extends JUpdateAdapter
 		{
 			if (isset($this->latest->client) && strlen($this->latest->client))
 			{
-				$this->latest->client_id = JApplicationHelper::getClientInfo($this->latest->client)->id;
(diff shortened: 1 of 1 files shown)


pasamio:
This particular line has been in ping pong. Would you mind updating the pull request to do a few things:

1) Check if it's numeric and use the current behaviour. I'd also like to flag this with a deprecated warning with the standard JLog notice that we use elsewhere in the code base. Obviously someone is doing something wrong and I'm sick of this stuff being repeatedly broken.
2) Put in a comment and include a link to here: http://docs.joomla.org/Design_of_JUpdate
3) Keep the change you've got in for if it isn't numeric in an else statement.

Bakual:
Done like Sam suggested

pasamio:
Thanks! Merged :)

PR #1678, opened 2012-11-11
Convert manual to markdown.

ianmacl:
This is the basis of the markdown manual as generated by Louis.  I think we lost some changes so perhaps they'll have to be merged in or something.


LouisLandry:
I can't help but feel like we should find a way to include the changes that have been made.  I'll take a look.

LouisLandry:
OK, there is nothing of importance that is lost.  Just need to convert the outstanding things and fix up a couple of TODO lines.  I'm going to merge this so we have a solid place to work from.  The DocBook is dead.  Long live the Markdown.

PR #1679, opened 2012-11-11
Fix up remaining manual conversion from DocBook to Markdown.

LouisLandry:


dongilbert:
YYYYEEEEESSSSSS!!!!

I'm just a _little_ excited for this. :) Trying to convert the DocBook xml to something I could actually use was getting to be a PITA.

eddieajau:
To be fair, DocBook is far more descriptive than Markdown, and Pandoc converts it readily.  However, I'm also glad we are going to this format :)  It lowers the barrier of entry for people writing docs (or, to put it another way, gives them one less excuse to argue why they shouldn't bother, hehe).

PR #1684, opened 2012-11-12
Refactoring JFormField & Co.

dongilbert:
After several varying discussions on the topic with @elinw and others, and some of my own frustrations, I decided to give a go at refactoing JFormField and it's sister elements. The reason I did this is that a lot of times, the field specific implementation of `getInput()` would process some logic that was already completed by `JFormField`, such as deciding if an element is readonly or disabled.

Doubling up on the workload like that is never good. Now, `JFormField::setup()` does most of the heavy lifting when it comes to determining what settings are active/inactive,  and the child classes simply worry about the html markup that those determinations require. This allows for broader, more expressive rulesets to determine functionality in the parent class, without having to be overly verbose in the child class. For example:

```
/**
 * instead of the child class doing this
 * (which is insufficient, what if the field was disabled="disabled"?)
 */
$disabled = $this->element['disabled'] == true ? 'some html' : '';

// it now does this
$disabled = !empty($this->disabled) ? 'some html' : '';
```

I did this to make it easy to add new features. In my case, I added the `$placeholder` attribute as a protected property to `JFormFieldText` and then extended similar fields that would use said attribute from that parent class.

Another new features is the `$prependToClass` property in `JFormField`. This allows you to set a class name that will always be prepended to the field. For example, look at `JFormFieldEmail`. It extends `JFormFieldText`, but also overrides the `getInput()` method, just to put in a `validate-email` class at the beginning of the input class. This also occurs in  a couple other places, and adding his feature cuts down on that code bloat. Instead of rewriting `getInput()`, do this instead:

```
protected $prependToClass = 'validate-email';
```

Also new to the mix is that each child `JFormField*` class that has it's own specific attributes now has it's own `setup()` method that handles assigning those specific attributes to the field object. See `JFormFieldTextarea` for a good example of that - https://github.com/dongilbert/joomla-platform/blob/abf409ad2feeee98005c0e779dd6c20d4b3eb249/libraries/joomla/form/fields/textarea.php#L69

Last thing - I deprectated the `$forceMultiple` property. It has it's use, but that use can be fulfilled by the `$multiple` property instead. See the `JFormField::setup()` for how I implemented this - https://github.com/dongilbert/joomla-platform/blob/abf409ad2feeee98005c0e779dd6c20d4b3eb249/libraries/joomla/form/field.php#L427

I'm not claiming this is perfect, but it does pass all the tests. And it _shouldn't_ break any backwards compatibility. I just want to start the conversation, since code speaks louder than words. :)

### DIFF
--- a/libraries/joomla/form/fields/password.php
+++ b/libraries/joomla/form/fields/password.php
@@ -9,6 +9,8 @@
 
 defined('JPATH_PLATFORM') or die;
 
+JFormHelper::loadFieldClass('text');
+
 /**
  * Form Field class for the Joomla Platform.
  * Text field for passwords
@@ -19,7 +21,7 @@ defined('JPATH_PLATFORM') or die;
  * @note        Two password fields may be validated as matching using JFormRuleEquals
  * @since       11.1
  */
-class JFormFieldPassword extends JFormField
+class JFormFieldPassword extends JFormFieldText
 {
 	/**
 	 * The form field type.
@@ -28,6 +30,53 @@ class JFormFieldPassword extends JFormField
 	 * @since  11.1
 	 */
 	protected $type = 'Password';
+
+	/**
+	 * Whether or not to use the strength meter
+	 *
+	 * @var    boolean
+	 * @since  12.3
+	 */
+	protected $strengthmeter;
+
+	/**
+	 * Strength threshold for the strength meter
+	 *
+	 * @var    string
+	 * @since  12.3
+	 */
+	protected $threshold;
+
+	/**
+	 * Method to attach a JForm object to the field.
+	 *
+	 * @param   SimpleXMLElement  $element  The SimpleXMLElement object representing the <field /> tag for the form field object.
+	 * @param   mixed             $value    The form field value to validate.
+	 * @param   string            $group    The field name group control value. This acts as as an array container for the field.
+	 *                                      For example if the field has name="foo" and the group value is set to "bar" then the
+	 *                                      full field name would end up being "bar[foo]".
+	 *
+	 * @return  boolean  True on success.
+	 *
+	 * @see     JFormField::setup()
+	 * @since   12.3
+	 */
+	public function setup(SimpleXMLElement $element, $value, $group = null)
+	{
+		parent::setup($element, $value, $group);
+
+		if (!empty($this->element['strengthmeter']))
+		{
+			$this->strengthmeter = (string) $this->element['strengthmeter'];
+		}
+
+		if (!empty($this->element['threshold']))
+		{
+			$this->threshold = (string) $this->element['threshold'];
+		}
+
+		return true;
+	}
 
 	/**
 	 * Method to get the field input markup for password.
@@ -39,18 +88,17 @@ class JFormFieldPassword extends JFormField
 	protected function getInput()
 	{
 		// Initialize some field attributes.
-		$size		= $this->element['size'] ? ' size="' . (int) $this->element['size'] . '"' : '';
-		$maxLength	= $this->element['maxlength'] ? ' maxlength="' . (int) $this->element['maxlength'] . '"' : '';
-		$class		= $this->element['class'] ? ' class="' . (string) $this->element['class'] . '"' : '';
-		$auto		= ((string) $this->element['autocomplete'] == 'off') ? ' autocomplete="off"' : '';
-		$readonly	= ((string) $this->element['readonly'] == 'true') ? ' readonly="readonly"' : '';
-		$disabled	= ((string) $this->element['disabled'] == 'true') ? ' disabled="disabled"' : '';
-		$meter		= ((string) $this->element['strengthmeter'] == 'true');
-		$threshold	= $this->element['threshold'] ? (int) $this->element['threshold'] : 66;
+		$size		= !empty($this->size) ? ' size="' . $this->size . '"' : '';
+		$maxLength	= !empty($this->maxlength) ? ' maxlength="' . $this->maxlength . '"' : '';
+		$class		= !empty($this->class) ? ' class="' . (string) $this->element['class'] . '"' : '';
+		$readonly	= !empty($this->readonly) ? ' readonly="readonly"' : '';
+		$disabled	= !empty($this->disabled) ? ' disabled="disabled"' : '';
+		$auto		= !empty($this->autocomplete) ? ' autocomplete="off"' : '';
+		$threshold	= !empty($this->threshold) ? $this->threshold : 66;
 
 		$script = '';
 
-		if ($meter)
+		if ($this->strengthmeter)
 		{
 			JHtml::_('script', 'system/passwordstrength.js', true, true);
 			$script = '<script type="text/javascript">new Form.PasswordStrength("' . $this->id . '",
--- a/libraries/joomla/form/fields/password.php
+++ b/libraries/joomla/form/fields/password.php
@@ -45,7 +45,7 @@ class JFormFieldPassword extends JFormFieldText
 	 * @var    string
 	 * @since  12.3
 	 */
-	protected $threshold;
+	protected $threshold = 66;
 
 	/**
 	 * Method to attach a JForm object to the field.
@@ -72,7 +72,7 @@ class JFormFieldPassword extends JFormFieldText
 
 		if (!empty($this->element['threshold']))
 		{
-			$this->threshold = (string) $this->element['threshold'];
+			$this->threshold = (int) $this->element['threshold'];
 		}
 
 		return true;
@@ -94,7 +94,6 @@ class JFormFieldPassword extends JFormFieldText
 		$readonly	= !empty($this->readonly) ? ' readonly="readonly"' : '';
 		$disabled	= !empty($this->disabled) ? ' disabled="disabled"' : '';
 		$auto		= !empty($this->autocomplete) ? ' autocomplete="off"' : '';
-		$threshold	= !empty($this->threshold) ? $this->threshold : 66;
 
 		$script = '';
 
@@ -103,7 +102,7 @@ class JFormFieldPassword extends JFormFieldText
 			JHtml::_('script', 'system/passwordstrength.js', true, true);
 			$script = '<script type="text/javascript">new Form.PasswordStrength("' . $this->id . '",
 				{
-					threshold: ' . $threshold . ',
+					threshold: ' . $this->threshold . ',
 					onUpdate: function(element, strength, threshold) {
 						element.set("data-passwordstrength", strength);
 					}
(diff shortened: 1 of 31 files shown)


mbabker:
At a glance, things look pretty good to me.  I'd have to play with this some to make sure nothing breaks, but you've got something good here IMO.

elinw:
Just so people understand, $forceMultiple exists in essence to solve a problem with checkboxes which is that they are required to accept multiples according to the W3C standard (the only field like this as far as I know). That means even if developers don't say multiple-="multiple" multiple still has to be true.

dongilbert:
@elinw - Thanks for the clarification on that. I'm not against `$forceMultiple` per sé - it's just that you can accomplish the same goal by using an existing property that makes just as much, if not more sense. This new implementation reduces 5 lines of code to 1, which is a good thing imho, and it doesn't break b/c (at least until 13.3, if `$forceMultiple` is removed as marked)

elinw:
I haven't played with it but I think it may make more sense to deal with the special needs of checkboxes in checkboxes, but maybe there is a more reason for this that I'm not aware of.  I'm not sure how many people are changing checkboxes to not accept multiple values given that checkboxes itself was problematic unless you extended it until not that long ago

dongilbert:
Correct, but $multiple can be on several fields. Processing that logic in the parent JFormField class allows us to easily add it to any other child class, as needed. 

Makes it super simple for developers too. I've created custom fields that I wanted to be a multiple select list, but I needed a custom getOptions() method. So I tossed in this, set $multiple to true, and voilá, it was set for me, without having to worry about it in the form XML.

libraries/joomla/form/fields/textarea.php
```diff
@@ -29,6 +29,66 @@ class JFormFieldTextarea extends JFormField
...
+		}
+
+		if (!empty($this->element['rows']))
+		{
+			$this->rows = (int) $this->element['rows'];
+		}
+
+		if (!empty($this->element['placeholder']))
+		{
+			$this->placeholder = (int) $this->element['placeholder'];
```
mariopro: Have to correct the cast to string:

 $this->placeholder = (int) $this->element['placeholder'];  to   $this->placeholder = (string) $this->element['placeholder'];

libraries/joomla/form/fields/password.php
```diff
@@ -30,6 +32,53 @@ class JFormFieldPassword extends JFormField
...
+	protected $strengthmeter;
+
+	/**
+	 * Strength threshold for the strength meter
+	 *
+	 * @var    string
+	 * @since  12.3
+	 */
+	protected $threshold;
+
```
mariopro, line 49: This field type could also make use of the placeholder attribute.

dongilbert, line 49: Since this class now extends JFormFieldText, it falls back on the setup method there, so it should have access to the $placeholder property

dongilbert, line 49: We just need to add the placeholder variable in the retuned input string in the getInput method.

mariopro, line 49: Absolutely right. Just adding to the getInput() works fine.

mariopro:
Don, just a though and I would like to know your thoughts about it. In the sql field type, wouldn't make sense now to have a json_decode true/false parameter? I ask because some of the multiselect fields that are populated from the sql, require the data to go through this function in the model.

libraries/joomla/form/fields/textarea.php
dongilbert: Thanks - this has been fixed.

mariopro:
Don, in the radio.php file, imho, seems to make some sense to set the checked radio as disabled. The reason for this comment might seem not so obvious because a checked radio won't change its value or state if the user clicks repeatedly over it, but, if developers want to use the radios to show/hide content, the checked radio not being disabled will respond to the user click and trigger any events that might be associated with it (imho).
In line 56, it could be set as : 
$checked = ((string) $option->value == (string) $this->value) ? ' checked="checked" disabled' : ''; and perhaps change the on click behavior.

dongilbert:
I would say probably not, since this is the general implementation. It
wouldn't be useful in every situation, so I don't think it should be
present in the base class.

mariopro:
You're abs right. Thanks Don.

elinw:
Radio buttons are also a weird special case since if there is only one it can't be unselected and multiple is forced to be false.

dongilbert:
Any further discussion on this topic? Improvements I should make?

dongilbert:
I'm going to add some documentation for JForm - should I attach it to this PR or create a new one? The reason I ask is there currently is none, and writing up the full docs will take some more time than just documenting what's this PR changes.

Maybe I'll document the API as it is now in a separate PR, and then document the changes from this.

elinw:
@dongilbert I was thinking about this project when I was working with a field that won't let me update to blank once there is data. A number of people  have raised this before.  I wonder if we could come up with an attribute that would let us optionally allow this.

mariopro:
As I previously suggested in the platform discussion about this refactoring, imho, the attribute could resemble something as - updateNulls.

dongilbert:
I was thinking something similar. But if that is the case - I think we should also add a default handling for a multiple field while we are at it. Like maybe a serialized array or json string. 

What I've done recently is add a hidden field after a group of checkboxes with a value of 0 and it had the same name as the checkbox set. This seemed to accomplish what I was looking for to do. 

Are there other field types besides checkboxes that don't submit when empty?

dongilbert:
I would agree with that approach Mario, given the recent discussions.

mariopro:
Json serialize would be very useful on multiples (this is how I handle them), but I'm not sure if there is consensus on this. Imho this would be the way to go for the multiples.
I have dealt with the update nulls on multiple with a tweak in the bind array, but in a form with several multiples, this becomes somehow heavy to code.
I don't remember any other checkboxes that don't submit when empty. Probably Elin could be the best person to answer that.

dongilbert:
I would like to propose another addition to the api - the `getAttribute()` method. We would use this method to replace duplicate code within field classes that generate markup for field attributes.

``` php
// We would change this
$disabled = !empty($this->disabled) ? ' disabled="disabled"' : '';

// To this
$disabled = $this->getAttribute('disabled');
```

With this approach, the actual markup generated is handled by the `JFormField` class, and it cleans up the mass of duplicate code while also making it easier to add new attributes if needed without having to copy markup across all `JFormField*` classes.

The method would look something like this: (I know, we'll need to build in some exceptions to the rules here)

``` php
/**
 * Generate markup for specified attribute.
 *
 * @param   string  $attribute  The name of the attribute for which to generate markup
 * @param   string  $default    What to return if $attribute is empty.
 *
 * @return  string  Generated attribute markup if $attribute is found, otherwise it returns default.
 *
 */
protected function getAttribute($attribute, $default = '')
{
    $attr = '';

    if(!empty($this->$attribute))
    {
        $attr = sprintf(' %s="%s"', $attribute, $this->$attribute);
    }
    elseif ($default !== '')
    {
        $attr = sprintf(' %s="%s"', $attribute, $default);
    }

    return $attr;
}
```

mariopro:
Looks very good and an good way to optimize the attributes.

elinw:
Only check boxes as far as I know. 
I agree about multiple storage being standardized though of course options would be nice.

mariopro:
In the checkboxes, wouldn't it be possible to accommodate the bootstrap data-toggle="buttons-checkbox" that, as far as I could find, in Isis is only possible on radios, while bootstrap has this checkbox feature http://twitter.github.com/bootstrap/javascript.html#buttons. I presume that the only impediment now for the attribute to be used is the &lt;li&gt; that is used to enclose the checkboxes fields in JFormFieldCheckboxes class.
Using this attribute instead type="list" for the checkbox could be an option, imho.

elinw:
Defnitely worth looking at.
You may have noticed that the update I did of the email field included multiple to the html5 standard. That calls for a comma separated string of addresses.  I tend to think that as a lowest common denominator that works.

mariopro:
Just a thought - in the JFormFieldFolderList, when the method starts to build the options list from the list of folders, wouldn't it be useful to add a couple of levels of recursiveness in this so that the drop-down that is rendered could accommodate, for instance, three child levels of the filtered folder? Example: 

```
<field name="path_to_images" 
    type="folderList" 
    directory="images" 
    exclude="none" 
    levels="3" 
    label="FOO_LABEL" 
    description="FOOD_DESC"/>
```

The parameter "levels" (or something similar), if defined, could state the max levels of child directories to search and include, hierarchically in the list. This would render, in the drop-down, something similar to what we have in the categories:
Main 
|-First child 
|-- Second child 
|--- Third child

bcordis:
This look very cool. Hope this is don.

On Friday, December 28, 2012, mariopro wrote:

> Just a thought - in the JFormFieldFolderList, when the method starts to
> build the options list from the list of folders, wouldn't it be useful to
> add a couple of levels of recursiveness in this so that the drop-down that
> is rendered could accommodate, for instance, three child levels of the
> filtered folder? Example:
> 
> `<field name="path_to_images" type="folderList" directory="images" exclude="none" levels="3" label="FOO_LABEL" description="FOOD_DESC"/>`
> 
> The parameter "levels" (or something similar), if defined, could state the
> max levels of child directories to search and include, hierarchically in
> the list. This would render, in the drop-down, something similar to what we
> have in the categories:
> Main
> |-First child
> |-- Second child
> |--- Third child
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/joomla/joomla-platform/pull/1684#issuecomment-11740042.

## 

Thanks,
Brent Cordis
2406 Hermitage Park Dr.
Hermitage, TN 37076
615.429.6426

mariopro:
And suddenly I've just stumbled on the JFormFieldFolderList "smarter brother" in the media manager (https://github.com/joomla/joomla-cms/blob/master/administrator/components/com_media/models/manager.php) where the getFolderTree method (class MediaModelManager) does exactly the above... 
Question: couldn't this be transposed or adapted to the JFormFieldFolderList ?

dongilbert:
I'll have a look at that and see about integrating. I think it would be a good addition.

mariopro:
For a possible future review of the the JFormField classes, imho, the onChange parameter could be evaluated for a possibility to change it to "onAction" and allow onClick, OnBlur, onFocus (...) and onChange to be set as the value instead of parameter name (I'm not sure this can be done now on standard fields). An example:

```
<field name="field_name" 
    onAction="onFocus('do_something_with_js_or_jQuery')"
    label="FOO_LABEL" 
    description="FOOD_DESC"/>
```

The method would then add the parameter to the field, depending on what was defined and who knows allow multiple, comma separated, values &lt;field name="foo" onAction="onFocus('do_something_with_js_or_jQuery'), onBlur('do_something_with_js_or_jQuery')" /&gt; 
This would give some space to implement, for instance and among other things, some of the new CSS3 cursor types, or even free some jQuery/js resources that extension's developers might need to implement to act upon a user action on a field. For instance, onClick="jQuery('.class_name').show();"

dongilbert:
I want to recode some of this again, and create a common field interface that can be implemented by those creating custom fields for their extensions. Would that be a welcome change? There's still a lot of work to do to make this a true refactoring worthy of implementing. A lot of what's here already consolidates common calls and cleans things up, which is good, but I want to make it easier to extend and create custom implementations.

Closing for now. :)

PR #1686, opened 2012-11-12
Fix doc block for JRoute::_()

mbabker:
As written, the doc block for `JRoute::_()` does not work as advertised.  The 0 value advertised to use the current scheme never passes the `if ((int) $ssl)` check at line 76 in the unchanged file.  As written, the only solid integer that is expected is a 1 for SSL schemes, so the parameter will accept any other integer for a non-SSL schemes.  We can also look at making the 0 value work for the third param, but as written, the code would never accomodate for that anyways.


eddieajau:
Merged.

PR #1694, opened 2012-11-12
Update email validation to HTML5 standard and to allow longer top level domains

elinw:
The email validation that we are currently using in JFormRule and also in javascript does not match the  standard published by the W3C at http://www.w3.org/TR/html-markup/input.email.html. The validation rule also does not support the multiple attribute.  Finally it does not work for many of the new top level domains that have been added (i.e. those which are more than 4 characters).

This pull request updates the regular expression to the W3C published standard. This is a less strict but more accurate standard than the current one used since it allows for email addresses that do not have top level domains. At the same time, many developers may desire to require  a top level domain so this pull request also includes the option to require that along with a separate regular expression to match that. Unlike the older version this one only requires a minimum of two characters with no maximum (previously 4 was the maximum). The proliferation of top level domains makes it difficult to accurately set a maximum.

The HTML5 standard also includes a multiple attribute which can have a comma separated list of emails. This pull request also implements that.  The multiple attribute may not be used in combination with the unique attribute.


eddieajau:
This branch is not mergable.

elinw:
@eddieajau would you please talk to @louislandry about this ... i believe that if you are just looking at the pull tester it is misunderstanding a squashed update

eddieajau:
Github says it's not.

elinw:
reopen

elinw:
Yup there was a change to one of the files overnight

PR #1697, opened 2012-11-14
Remove evil from joomla platform

oc666:
We should avoid the usage of this function, cause a lot of hosts are disable this function from security reasons. In addition, it's have low performance.
I've replaced each eval usage with the native PHP usage.

### DIFF
--- a/libraries/joomla/document/html/html.php
+++ b/libraries/joomla/document/html/html.php
@@ -505,9 +505,8 @@ class JDocumentHTML extends JDocument
 				: count(JModuleHelper::getModules($name));
 		}
 
-		$str = 'return ' . implode(' ', $words) . ';';
+		return implode(' ', $words);
 
-		return eval($str);
 	}
 
 	/**
--- a/libraries/joomla/updater/adapters/collection.php
+++ b/libraries/joomla/updater/adapters/collection.php
@@ -97,7 +97,7 @@ class JUpdaterCollection extends JUpdateAdapter
 		$tag = $this->_getStackLocation();
 
 		// Reset the data
-		eval('$this->' . $tag . '->_data = "";');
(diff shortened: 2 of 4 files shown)


elinw:
LOL when I read the title .. you probably want to fix that.

oc666:
http://www.blog.highub.com/php/php-core/php-eval-is-evil/

pasamio:
1) this pull request isn't merge able
2) have you actually tested this all still works?

oc666:
Maybe my branch was old.
I make simple tests but not the unit tests.

infograf768:
See CMS tracker #29646

pasamio:
I can tell you the change to count modules will break the samples, the others in not sure about on an iphone but that change to countmodules will break.

oc666:
@pasamio, but it's the same code. can you explain please.

elinw:
See #1307 for count modules

pasamio:
Again I houldnt do this on my phone but its simple:

What happens when $words = array('1', '+', '1'); for the original line and the new line?

oc666:
@pasamio 
Here is an example:;
<?php

function testme1($w)
{
        $output1 = implode(' ', $w);
        return implode(' ', $w);
}

function testme2($w)
{
        $output1 = implode(' ', $w);
        $str = 'return ' . implode(' ', $w) . ';';
        return eval($str);
}

$words = array('1', '+', '1');

print "first test: ";
print_R(testme1($words));
print PHP_EOL;

print "second test: ";
print_R(testme2($words));
print PHP_EOL;

?>

The output of this test script (PHP 5.4.6):
first test: 1 + 1
second test: 2

So you are right, but I think there is a way to do that without eval. 
Let me check it and get back with a proposal.

Thanks

PR #1698, opened 2012-11-16
Update libraries/legacy/model/list.php

artur-stepien:
Count uses less memory and is usually quicker. Tested on mysql, not sure about mssql

### DIFF
--- a/libraries/legacy/model/list.php
+++ b/libraries/legacy/model/list.php
@@ -228,11 +228,11 @@ class JModelList extends JModelLegacy
 		}
 
 		// Load the total.
-		$query = $this->_getListQuery();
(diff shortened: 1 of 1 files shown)


pasamio:
You're assuming the query is an object. What happens if getListQuery returns a string?

I could have sworn that this was changed somewhere already.

artur-stepien:
The first question is good. But about second one: is isn't changed. Not in framework, not in any Joomla CMS version. 
Solution to returning string instead of object is give fall-back compatibility (for bad written components)

pasamio:
I mean I thought I saw someone propose and add this feature where it checked if it was an instance of JDatabaseQuery and then modified the query.

See:
https://github.com/joomla/joomla-platform/blob/master/libraries/legacy/model/legacy.php#L318

PR #1701, opened 2012-11-18
Adding destroy() method to JImage

dongilbert:
Currently, there is no way to release memory associated with a `JImage::$handle` resource. When working with images, it's _normally_ accomplished using `imagedestroy($handle)`, but since the resource is held within the JImage object instance, we don't have access to destroy it and free the memory. This fixes that by allowing you to explicitly destroy the image handle.

### DIFF
--- a/libraries/joomla/image/image.php
+++ b/libraries/joomla/image/image.php
@@ -360,6 +360,9 @@ class JImage
 		// Swap out the current handle for the new image handle.
 		else
 		{
+			// Free the memory from the current handle
+			$this->destroy();
+
 			$this->handle = $handle;
 
 			return $this;
@@ -496,6 +499,9 @@ class JImage
 	 */
 	public function loadFile($path)
 	{
+		// Destroy the current image handle if it exists
+		$this->destroy();
+
 		// Make sure the file exists.
 		if (!file_exists($path))
 		{
@@ -658,6 +664,9 @@ class JImage
 		// Swap out the current handle for the new image handle.
 		else
 		{
+			// Free the memory from the current handle
+			$this->destroy();
+
 			$this->handle = $handle;
 
 			return $this;
@@ -714,6 +723,9 @@ class JImage
 		// Swap out the current handle for the new image handle.
 		else
 		{
+			// Free the memory from the current handle
+			$this->destroy();
+
 			$this->handle = $handle;
 
 			return $this;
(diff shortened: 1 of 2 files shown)


pasamio:
100% more tests than the leading brand pull request? Not sure I agree with your math :p

dongilbert:
:|

Ya, I meant for this method. My bad.

elinw:
I would say it's infinitely more tests.

LouisLandry:
Why wouldn't we just use a destructor for this?

dongilbert:
We could use the destructor to call the destroy method a final time, but a like being able to explicitly run it. I found out it was missing because I have a script that imports a bunch of home listings from an mls XML feed, and downloads all the images and resizes them. In my loop, the JImage object is never unset, so destruct is never called, and memory usage would bloat. So I added this to be able to explicitly destroy the image handle.

pasamio:
So why not handle the clean up in the loadFile method?

dongilbert:
Not sure I follow you Sam. loadFile creates the handle so the other methods can use it. If we handled it there, wouldn't that defeat the purpose?

Unless you are saying, run the code in the top of the loadFile method so that if a handle already exists, it gets destroyed, and then run it again in the  destruct method to clear it out on the last run. 

I think that could work. Then it relieves the developer from having to explicitly call it, and would still solve the problem I was trying to take care of.

Do you think it should still be its own method? Or just integrate/duplicate the code into those two spots mentioned above?

pasamio:
Yes, I was suggesting that if the handles already exists that it should free'd at that point and also the destruct. I think the crop method might also require a destroy as well when it's replacing the handle. I do feel that we should automatically take care of the management of the internal resource as requiring an external destroy method seems a bit of a kludge.

dongilbert:
Updated with the new feature. I left `destroy()` as its own public method, and just call it in the `__destruct()` method as well as at the top of the `loadFile()` method.

dongilbert:
I just checked the automated test log - http://developer.joomla.org/pulls/pulls/1701.html - and it says that this PR can't be cleanly merged. Can someone double check that or let me know what to fix? I'm not too sure what the issue is.

dongilbert:
I think I figured it out - I rebased the changes from upstream/staging

dongilbert:
Not ready to merge - UT Failures

pasamio:
Where you have added the destroy in crop and resize I don't think is correct. It should be in the else clause of the if statement below because if a new JImage is created then the current JImage is still valid and we shouldn't destroy it's internal image data. Only when we replace $this->handle with a new handle should we attempt to free the data for both crop and resize. However the loadFile change I don't have a problem with because it's an intentionally destructive operation so if for some reason it fails midway, I don't expect to be able to re-use what was in the pre-existing handle. Does that make sense?

dongilbert:
I see your point - since you may not be chaining the methods together, you wouldn't want to destroy the handle preemptively. I'll get that fixed.

Quick question while I'm in here. Why do we do this (I see it a lot in the platform):

```
$someVar = SomeClass::getInstance();

return $someVar;

// or specifically in this class

$new = new JImage($handle);

return $new;
```

I know we want to make the code easy to read, but that seems a little ridiculous. It could be easily replaced with the following and be just as readable:

```
return new JImage($handle);
```

pasamio:
Typically that was done when the method is returning an object in PHP4 times where you had to explicitly pass objects by reference. In those cases a return new didn't work and would generate a PHP warning/notice/error. If this is still necessary here is beyond my immediate knowledge but typically when you see that it is because it was necessary with PHP4 or if the function is still marked as returning a reference.

dongilbert:
I'm good with this as is. I was thinking to change `$new = new JImage; return $new;` to `return new JImage;`, but I need to do more research on it prior to committing to that.

PR #1705, opened 2012-11-19
Clean any white spaces and/or tabs

mariopro:
Tabs that may result from manifest source formatting after the translation element
in <submenu><menu>COM_COMPONENT_SUBMENUNAME</menu></submenu> result in invisible bugs as the translation element is saved to the #__menu table with the tabs/white spaces.
Translation will not be done and the alias will be used by the system to build the menu or submenu.
Hard to debug, since white spaces and tabs are invisible in database.
Suggestion is to trim both $menuElement and $child in the _buildAdminMenus method to avoid this issue.

### DIFF
--- a/libraries/legacy/installer/adapters/component.php
+++ b/libraries/legacy/installer/adapters/component.php
@@ -1264,7 +1264,7 @@ class JInstallerComponent extends JAdapterInstance
 			$data = array();
 			$data['menutype'] = 'main';
 			$data['client_id'] = 1;
-			$data['title'] = (string) $menuElement;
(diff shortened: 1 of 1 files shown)


pasamio:
Can you update to trim the alias as well?

mariopro:
The alias is not affected because the building process of the alias, per se, trims the alias before it is saved to the database. On my tests, I didn't find spaces or tabs in the alias db field, so that's the reason it wasn't contemplated here.

pasamio:
Fair enough :)

mariopro:
Thanks

PR #1706, opened 2012-11-20
Payment Library for the Joomla Platform

Skullbock:
During the Joomla World Conference in San Josè, i had a chance to speak to @LuisLandry during his Joomla Platform talk. 
I and some other folks (like Parth from TechJoomla) raised the need for the Joomla Platform to contain a set of libraries that would help managing payment processing; so many different Joomla components and applications are using payment processing and they're reinventing the wheel every time. 

We agreed that some work on this area was needed, so i'm opening this pull request as a centralized place where to discuss ideas and code on how to achieve that. I'm pulling in some code that @LuisLandry had already in one of his branches and that could be use as a great starting point.

Feel free to start commenting on this code and share your ideas on how this coud work. If everyone agrees, we could use my fork at https://github.com/Skullbock/joomla-platform-payment as the main branch for this feature.

### DIFF
--- a/libraries/joomla/commerce/address.php
+++ b/libraries/joomla/commerce/address.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * @package     Joomla.Platform
+ * @subpackage  Commerce
+ *
+ * @copyright   Copyright (C) 2005 - 2011 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE
+ */
+
+defined('JPATH_PLATFORM') or die;
+
+/**
+ * Commerce system address object for the Joomla Platform.
+ *
+ * @package     Joomla.Platform
+ * @subpackage  Commerce
+ * @since       12.1
+ */
+class JCommerceAddress
+{
+	public $title;
+	public $firstName;
+	public $middleName;
+	public $lastName;
+	public $companyName;
+	public $poBox;
+	public $streetNumber;
+	public $streetName;
+	public $streetUnit;
+	public $city;
+	public $region;
+	public $district;
+	public $postalCode;
+	public $country;
+	public $email;
+	public $phone;
+	public $phone2;
+	public $timeZone;
+	public $longitude;
+	public $latitude;
+}
--- a/libraries/joomla/commerce/address.php
+++ b/libraries/joomla/commerce/address.php
@@ -18,24 +18,123 @@ defined('JPATH_PLATFORM') or die;
  */
 class JCommerceAddress
 {
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $title;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $firstName;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $middleName;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $lastName;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $companyName;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $poBox;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $streetNumber;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $streetName;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $streetUnit;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $city;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $region;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $district;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $postalCode;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $country;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $email;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $phone;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $phone2;
+
+	/**
+	 * @var    string
+	 * @since  12.1
+	 */
 	public $timeZone;
+
+	/**
+	 * @var    float
+	 * @since  12.1
+	 */
 	public $longitude;
+
+	/**
+	 * @var    float
+	 * @since  12.1
+	 */
 	public $latitude;
 }
--- a/libraries/joomla/commerce/cart.php
+++ b/libraries/joomla/commerce/cart.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * @package     Joomla.Platform
+ * @subpackage  Commerce
+ *
+ * @copyright   Copyright (C) 2005 - 2011 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE
+ */
+
+defined('JPATH_PLATFORM') or die;
+
+/**
+ * Commerce system cart object for the Joomla Platform.
+ *
+ * @package     Joomla.Platform
+ * @subpackage  Commerce
+ * @since       12.1
+ */
+class JCommerceCart
+{
+	public $items = array();
+	// Order currency code.
+	public $currency = 'USD';
+	// Order subtotal.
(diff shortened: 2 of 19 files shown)


eddieajau:
First off, I'm all for getting a commerce package into the platform.  However, would it be worth taking a few steps back and starting with a wireframe (powerpoint slide, google docs, whatever) to see what all the pieces are and how they wire up rather than diving into code first?  I'm a visual person and I think it's important to establish the flows first.

Just one simple observation though. I don't think an "address" is necessarily tied to "commerce".  I'm pretty sure there are address standards out there so it is probably worth considering making a separate package for that.

elinw:
There are definitely address standards and they are really detailed because I believe they are governed by the international postal union (I think it might have been when I first saw this package). I did some work on that before and it would be really powerful to have that generally available because it is super useful not just for commerce but for CRM and other types of applications. That's the kind of obsessively detail oriented project I don't mind digging into. 

http://www.upu.int/en/activities/addressing/standards.html

Apparently Google uses xNAL

http://microformats.org/wiki/location-formats

http://www.immagic.com/eLibrary/TECH/OASIS/XNAL_V2.PDF

Naouak:
Just skimming through that code I see some problem with the way everything is handled:
- address should not be included in basic order in case of digital store.
- states should not be hardcoded in files because a lot of shops will have other steps (like checking manually the order for example)
- what about recurring payments or payment separated over months
- what to do when the payment method workflow is not like the paypal one
- how does the callback request would be handled ?
- how about analytics(google analytics, piwik) tracking code handling ?

It's just thoughts I had by skimming through this code. It may have already been handled or the code is already able to answer some of these wonderings.

Skullbock:
Hi all, and sorry for not replying sooner. I got back on saturday from the US (jwc12) and i was hit by the jet lag :(

@eddieajau i agree with your points. I pull requested this as per @LouisLandry request so we could have a "code centered" discussion here on github instead of just plain mailing list talk. I prefer something "practical" on which discussion can be started.

@Naouak I totally agree with most of your points. I too do not agree completely on the structure of the order class, nor with the order states. IMHO there is "too much" into this library as a starting point. The library itself should **not** force a particular workflow. It should simply _require what's needed_ for the payment processing. 

In this sense, i like the idea of a placeholder object (like the order and the address) to use as container for the info that we need to pass onto the payment plugins.

Therefore, i would say that the library itself should not handle the workflow itself (like sending the user to a different url, catch the IPN-like calls from the payment processors, etc. This library should just make these things **MUCH** easier, but the workflow should still be implemented by each application, since everyone has a different need.

IMHO, what we should provide with this package is:
- A simple way to interact with payment services (both on site and paypal-like)
- A list of utility classes that will help dealing with the payment data
  - Address
  - Order
  - State
- Some basic plugin widely used (paypal, 2checkout, auth.net) that could serve as an example for the community to develop other plugins

Your thoughts?

Naouak:
IMO : 
- Payment should be a list of simplification API to generate buttons, api calls or handle responses.
- Class utility should be addressed by the CMS(or the application if there is) not the platform. These kind of data object are application specific.
- Address may be done by the platform as it is a more generic type but it should not be part of commerce component. May be more like a location package.

Here is a sample of the workflow I usually do when I have to do stuff with payment method:
When preparing payment:
1. Generate order\* 
2. preparate form  
3. send it to user*
When checking payment:
1. Receive payment\* 
2. check data integrity 
3. check status 
4. do verifications according to status\* 
5. send confirmation to payment system 
6. do stuff with the order*

Actions marked with a \* are for me application specific and should not be implemented in the platform.

I encountered lately a payment system that require to upload some file to a ftp to send a payment notification and then you have to check the FTP the day after to see if payment was accepted. Need to think about a way to implement that too besides other payment methods.

Skullbock:
Generate buttons is something that i would definetly **NOT** have in the platform; the platform could give an array of available payment libraries, but no html should be generated at all. As you said, this is application specific.

I agree that the utility classes are not strictly required, but since each payment method requires a lot of informations on the payment to be executed, like email, country, billing address, VAT ID, taxes, total, etc. it makes sense IMHO to have a class that contains all these informations; we could use an array or a JRegistry, but using dedicated classes makes more sense to me, since they could be extended by the application, and still offer some methods, like validation.

I would leave the CMS out of this talk, since it may take a while to insert something like this into the cms libraries. If the platform would include this, then every cms application / component could use the classes for its own porpouses.

As for the tasks that i would like to see in the platform, i would say that you are including a bit too much into the lower level (the platform). I would leave just these:
- Populate order data
- Sending payment request (no form or html, just the API interactions)
- Validate payment response
- Check data integrity
- Check status
- Add any utility method to interact with the payment system API, like check, void, etc

Does this make sense to you? What i'm trying to achieve here is a system that's not bound to any specific use case, and does not force any particular workflow, but that still takes away a lot of the boring or hard work for the developers, and offer a solid base on top of which we could build our application's payment plugins

AdamDocherty:
Is everyone aware of this: http://www.slideshare.net/coolparth/joomla-common-payments-api

On git: https://github.com/techjoomla/joomla-payments/tree/master/code

I have no real comments on either, just wouldn't want to see duplicate work being done...

parthlawate:
Thanks @Webvida 

Hello All !  
We had presented at the Joomla World Conference on this very subject. We already have quite a bit of code written for this & have it in use in 4 of our own extensions. Also we have 4 other devs who are adopting this over the next couple of months.. We have some 10 odd payment methods ready for it.. 

We have open sourced this code at https://github.com/techjoomla/joomla-payments/tree/master/code

We are already in talks with several developers to adopt this API & have a Skype chat running to discuss this. We are working on getting this ported & written as part of the platform. 

You can also take a look at my presentation on this here for a visual idea http://www.slideshare.net/coolparth/joomla-common-payments-api

Skullbock:
Hi @parthlawate great to have you on board! We already discussed this at jwc12, but as we said, we need something more "abstract" to put into the platform. We could definetly use part of your huge work on this side, but we need to make it more application-independent.

parthlawate:
@Skullbock  Absolutely ! That's where we want to go with it. We are basically already moving parts of this into a library to make it easier to put it into the platform & make it more abstracted. We have a branch of the platform where we are working on this & testing its use in the CMS . 

While we work on that approach, we have basically made it possible to use it today while we all work together to get it into the platform.  Great to see this Pull request ! 

We have been discussing @LouisLandry 's approach to it as well as a few other thoughts on the Skype chat. Oh ya :) Forgot about that .. We have a Payments Skype Chat Setup for this very objective of Getting a Payments API into the platform.  Add me on Skype. My id is tws_parth & send me a request if you want to get on there.. We have some 15 odd devs in there..

Skullbock:
I think i'm already in there as of today ;)

parthlawate:
@Skullbock  You are ? How come i didnt notice ?  Aha ! Just saw you .. Cool ! Welcome aboard !

ssv445:
My 2 cents - We have been discussing a complex thing. 
A lot of complex scenarios are there, so we should propose only a basic processing system.

## A workflow expectation from XYZ component- 

#1 XYZ component concludes to request for a payment,

#2 Collect Payment information from customer via its own means/forms (total-amount, credit-card, user-details, return-url, notfication-url etc)

#3 Ask payment processor to process the amount (with Joomla standards, processor will convert it to gateway standards)
      a) in case of API, payment processor will do it instantly and return the response
      b) in case of redirection - should tell the caller next redirection ?? or might be some other ideas to generalize things

#4 On notification URL, catches notifications, and ask the relavent processore to process the data. 
       a) Payment processor will verify the data and recommend action -> debit amount, credit amount, control-action [payment on hold, order canceled etc.]

#5 payment request and payment notifcation should be stored in DB by XYZ component

## So now we know what we need to build

#1 Base Classes     : Payment, Order, Payment Processor
#2 data structures      : PaymentRequest, PaymentCallback, CreditCard etc
#3 helper classes       : find available plugins, find compatibility, data conversion, verfication and validation etc.
#4 Processor Plugins    : build plugins for gateway-specific-processing, i.e. paypal

Important points : 
#1 : There are multiple variation for payment processing per gateway e.g. PayPal standard, pro, avdnace OR authorize.net SIM, CIM etc.
#2 : We should provide a complete API, for all actions starting from creation to cancellation.
#3 : Plugins should broadcast their support for few features - trial-1-amount, trial-2-amount, trial-1-time, currency,  etc.
#4 : Plugins should provide verify fuctionality for given payment request, so compaonent can tell customer about fault data.

Skullbock:
Just to update this Pull request: as per discussions in the payment api skype group, we changed our approach a bit and we're working on a new version, momentarily placed here:

https://github.com/Skullbock/joomla-platform-payment/tree/jpayment/libraries/joomla/payment

elinw:
Hi just FYI we have a PR now for OpenStreetMap so between that and the JGoogle map I may take a crack  at an address class.

parthlawate:
That's great !

PR #1710, opened 2012-11-22
cache strict standards issue when using system plugin cache in CMS

infograf768:
Solving Strict standards issue in CMS
See tracker http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29665

Strict Standards: Declaration of JCacheControllerPage::store() should be
compatible with JCacheController::store($data, $id, $group = NULL) in
ROOT/libraries/joomla/cache/controller/page.php
on line 196

### DIFF
--- a/libraries/joomla/cache/controller.php
+++ b/libraries/joomla/cache/controller.php
@@ -204,15 +204,14 @@ class JCacheController
 	/**
 	 * Store data to cache by id and group
 	 *
-	 * @param   mixed   $data   The data to store
-	 * @param   string  $id     The cache data id
-	 * @param   string  $group  The cache data group
-	 *
-	 * @return  boolean  True if cache was stored
+	 * @param   mixed    $data        The data to store
+	 * @param   string   $id          The cache data id
+	 * @param   string   $group       The cache data group
+	 * @param   boolean  $wrkarounds  True to use wrkarounds
 	 *
 	 * @since   11.1
 	 */
-	public function store($data, $id, $group = null)
+	public function store($data, $id, $group = null, $wrkarounds = true)
 	{
 		$locktest = new stdClass;
 		$locktest->locked = null;
--- a/libraries/joomla/cache/controller.php
+++ b/libraries/joomla/cache/controller.php
@@ -209,6 +209,8 @@ class JCacheController
 	 * @param   string   $group       The cache data group
 	 * @param   boolean  $wrkarounds  True to use wrkarounds
 	 *
+	 * @return  boolean  True if cache stored
+	 * 
 	 * @since   11.1
 	 */
 	public function store($data, $id, $group = null, $wrkarounds = true)
--- a/libraries/joomla/cache/controller/page.php
+++ b/libraries/joomla/cache/controller/page.php
@@ -117,8 +117,8 @@ class JCacheControllerPage extends JCacheController
 	 * Stop the cache buffer and store the cached data
 	 *
 	 * @param   mixed    $data        The data to store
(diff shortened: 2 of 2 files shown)


libraries/joomla/cache/controller/page.php
```diff
@@ -116,13 +116,16 @@ public function get($id = false, $group = 'page', $wrkarounds = true)
 	/**
 	 * Stop the cache buffer and store the cached data
 	 *
+	 * @param   mixed    $data        The data to store
++	 * @param   string   $id          The cache data id
++	 * @param   string   $group       The cache data group
```
elinw: These look misaligned.

libraries/joomla/cache/controller.php
```diff
@@ -204,15 +204,14 @@ public function get($id, $group = null)
...
-	 * @param   string  $id     The cache data id
-	 * @param   string  $group  The cache data group
-	 *
-	 * @return  boolean  True if cache was stored
+	 * @param   mixed    $data        The data to store
+	 * @param   string   $id          The cache data id
+	 * @param   string   $group       The cache data group
+	 * @param   boolean  $wrkarounds  True to use wrkarounds
 	 *
 	 * @since   11.1
```
elinw, line 214: put back the @return that was there before

elinw:
This is an backward incompatible API change, how much is it going to break?

Chraneco:
Could you please have also a look at this

https://github.com/joomla/joomla-cms/pull/236

while talking about the backward compatibility issue?

infograf768:
It is indeed an API change. Nobody replied to list. Shall we just postpone this to 3.x or even 4.x ?

eddieajau:
That cache controller is only used by the CMS.  It's your call what to do with it and I think you are between a rock and a hard place.

elinw:
Well we have no real idea of what is used by who out there, but it's certainly a reasonable supposition that it is used by few applications other than the CMS. I would say that if we are going to break it, we should fix the various pieces all at once rather than forcing people who are not paying attention to fix in dribs and drabs. I think it's probably preferable to put a log message in now for the next few months and then fix in 3.1 or 3.2.  4 wouldn't make sense since it won't use this cache package.  In fact if this cache moves to legacy before the 3.1 release we would need people to do updates anyway if the CMS chooses to sync with that (which is an open question).

libraries/joomla/cache/controller/page.php
```diff
@@ -116,13 +116,16 @@ public function get($id = false, $group = 'page', $wrkarounds = true)
...
+	 * @param   string   $id          The cache data id
+	 * @param   string   $group       The cache data group
 	 * @param   boolean  $wrkarounds  True to use wrkarounds
 	 *
 	 * @return  boolean  True if cache stored
 	 *
 	 * @since   11.1
 	 */
-	public function store($wrkarounds = true)
+	public function store($data = null, $id = null, $group = null, $wrkarounds = true)
```
eddieajau: I would change this to:

```
public function store($data, $id, $group = null, $wrkarounds = true)
```

just to be completely consistent with the parent.

elinw:
We'd like you to merge this please.  See https://groups.google.com/d/msg/joomla-dev-cms/cTG3bIpxjPs/nPZmgp6qK_gJ

infograf768:
Implemented in CMS master

elinw:
Would someone tag as a b/c break so it's in the change log?

dongilbert:
There you go.

PR #1711, opened 2012-11-22
Fix code style issues in the test tree

elinw:


tests/suites/unit/joomla/database/database/JDatabaseImporterPostgresqlTest.php
```diff
@@ -1,4 +1,4 @@
-<?php
+f<?php
```
LouisLandry: Oops.

tests/suites/unit/joomla/database/database/JDatabaseImporterPostgresqlTest.php
```diff
@@ -406,39 +398,45 @@ public function dataGetAlterTableSQL()
 	}
 
 	/**
-	 * Data for the testGetColumnSQL test.
+	 * Data for the testGetColumnSQL test
 	 *
 	 * @return  array  Each array element must be an array with 3 elements: SimpleXMLElement field, expected result, error message.
-	 *
+	 *
```
LouisLandry: Oops.

tests/suites/unit/joomla/database/database/JDatabaseImporterPostgresqlTest.php
```diff
@@ -605,10 +603,16 @@ public function testGetAddColumnSQL()
 		$instance = new JDatabaseImporterPostgresqlInspector;
 		$instance->setDbo($this->dbo);
 
+
```
LouisLandry: Extra new line isn't necessary.

tests/suites/unit/joomla/database/database/JDatabaseImporterPostgresqlTest.php
```diff
@@ -642,8 +646,9 @@ public function testGetAddColumnSQL()
 	}
 
 	/**
-	 * Tests the getAddSequenceSQL method.
+	 * Tests the getAddSequenceSQL method
 	 *
+	 *
```
LouisLandry: Extra new line isn't necessary.

tests/suites/unit/joomla/database/stubs/nosqldriver.php
```diff
@@ -37,7 +37,7 @@ class JDatabaseDriverNosql extends JDatabaseDriver
 
 	/**
 	 * The null or zero representation of a timestamp for the database driver.  This should be
-	 * defined in child classes to hold the appropriate value for the engine.
+	 * defined in child classes to hold the appropriate value for the engine..
```
LouisLandry: Double period?

tests/suites/unit/joomla/database/database/JDatabaseImporterPostgresqlTest.php
```diff
@@ -20,33 +20,18 @@
 class JDatabaseImporterPostgresqlTest extends PHPUnit_Framework_TestCase
 {
 	/**
-	 * @var    object  The mocked database object for use by test methods.
+	 * @var    object  The mocked database object for use by test methods
```
eddieajau: I'm not a fan of removing all the periods for these types of cases.  For example, say you wanted to added some extra information; you'd need to add a period to separate it.

```
The mocked database object for use by test methods. Say something else about it.
```

elinw: That is our code style in the manual.  A noun+adjectival phrase do not make a sentence. I'm really not going to rewrite the docblocks though if someone wants to do more descriptive work on the test tree I think that would be awesome.

I would personally say if you want to add extra information you should say "The mocked database object for use by test methods which is used to indicate something."   Or you could put the full sentence on a new line to clearly indicate that the first is essentially a label (which it is) and the second is supplemental descriptive text.

tests/suites/unit/joomla/database/database/JDatabaseImporterPostgresqlTest.php
elinw: There is only one line in my repo, not sure what is going on here.

If I look at the full view there isn't a new extra line any more, so I'm not sure why it's not making this a "commented on outdated diff."

elinw: See the other note, this is gone but it's not rendering in the pr for some reason.

tests/suites/unit/joomla/database/database/JDatabaseImporterPostgresqlTest.php
elkuku: May I just throw in here that it is far more easier to write a code sniff that ensures that every comment is terminated with a full stop than to examine if the comment is indeed a complete sentence containing a verb ?

elinw: That's certainly true.  Actually a lot of times the trouble is a missing noun but often in those cases we have imperatives which technically ARE sentences, and those have been bugging me because we really should be using stops and aren't. Of course just as easy to write a sniff highlighting the presence of periods in labels and titles e.g. on the same line as @param and the first line of a method.

A long time ago I proposed to Louis and Andrew that they think of the docblocks as somewhat like a bibliographic reference (because we have title, date,  copyright), but they really didn't like it. Now in many systems of rendering bibliographic listings there are full stops used after a title as well as after date, author etc, basically each element.   Also we are not running all of the elements together, we are putting them on separate lines. It's many many periods to add though and I wonder if it is worth all those extra characters.  Either way it needs to be consistent in the tests and libraries trees. Looking through the codebase a bit it is highly inconsistent.

eddieajau: Bibliographies are a good example of "text" that is terminated by a period yet not forming a proper sentence.  If consistency is the issue, my votes goes to using periods.  That is my practice and it's an easy rule for the developer to remember.  Anything less falls into subjective territory because punctuation rules vary even between English speaking countries.

elinw: So we could say something like: The following tags always take a full stop: the title, @params, @return @note @deprecated  ... but not @since or @link maybe unless you really want full stops all over the place...  I think the rule could be stated as tags with labels, descriptions or comments take a full stop at the end of the label, description or comment. Tags that only have   a value with no additional information do not.

Update: Actually does return always have a description? Definitely not when it is void, so we would need to account for that.

@eddiejau That was my point about bibliographies.

elinw:
Ok since we have an unresolved discussion I returned to the status quo. Eventually we will want to clean up.

tests/suites/unit/joomla/database/database/JDatabaseImporterPostgresqlTest.php
```diff
@@ -13,7 +13,7 @@
  * Test the JDatabaseImporterPostgresql class.
  *
  * @package     Joomla.UnitTest
- * @subpackage  Database
+ * @subpackage  Database
```
eddieajau: There are a few trailing spaces in this file.

elinw:
Okay I think I got them all plus a few more things. I'm not claiming perfect files just steady improvement.

tests/suites/unit/joomla/database/database/JDatabaseImporterPostgresqlTest.php
eddieajau: Still two trailing spaces in this file that I can see.

tests/suites/unit/joomla/database/database/JDatabaseImporterPostgresqlTest.php
```diff
@@ -414,31 +421,37 @@ public function dataGetAlterTableSQL()
 	 */
 	public function dataGetColumnSQL()
 	{
+		$sample = array(
+			'xml-id-field' => '<field Field="id" Type="integer" Null="NO" Default="nextval(\'jos_dbtest_id_seq\'::regclass)" Comments="" />',
+			'xml-title-field' => '<field Field="title" Type="character varying(50)" Null="NO" Default="NULL" Comments="" />',
+			'xml-title-def' => '<field Field="title" Type="character varying(50)" Null="NO" Default="this is a test" Comments="" />',
+			'xml-body-field' => '<field Field="description" Type="text" Null="NO" Default="NULL" Comments="" />',);
```
eddieajau, line 428: Probably should put the closing paren' on a new line.

elinw, line 428: I believe the rule is multiline arrays are on the same line, multi line functions on a new line.  I did run this through phpcs multiple times (obviously since that is how I know were the issues were).

elinw:
Why is phpcs not catching that one? Very odd.

elkuku:
The trailing white space sniff only sniffs code but not comments. `@todo FixMe ;)`

elinw:
@eddieajau I think I got them all now.

ianmacl:
Please rebase.

elinw:
@ianmacl I think it should be okay now.

PR #1712, opened 2012-11-23
New Data package

eddieajau:
#### Overview

This pull request adds a Data package with two new classes: `JData` and `JDataSet`.

While similar to the familiar `JObject` class, the `get` and `set` methods are removed in favour of using direct property accessors (`$object->key`) that are controlled by magic methods.  The properties are held in a private array within the class, completely separate from the real class properties.  The `setProperties` method is replaced with a `bind` method which is used elsewhere in the API (cf. JTable->bind).  A 'dump' allows you to obtain a `stdClass` representation of the data object.  This is also used to make the class implement the `JsonSerializable` interface.

A `JData` can also act as an array iterator allowing you to use it in a `foreach` construct.

The `JDataSet` class provides a convenient way to work with lists of `JData` objects.  Aside from the regular iterator functionality, the magic get method acts like a "get column" operation on the entire list, and the magic set method acts like a "set column" operation on the entire list.  The magic call method allows you to call a method (if available) on each of the objects in the list.  See the documentation for more information. 

Additional notes and comment can be found in the documentation.

This PR supersedes #1673 (and #1703) which looked at making improvements to the existing `JObject` class. After review, it was found reasonable to incorporate the proposed new features and architecture in a new package thereby allowing `JObject`, an artefact of bridging work between PHP 4 and PHP 5, to be deprecated at some point in the future (such work does not form part of this pull request and will be left for others to pursue).
#### Documentation

Documentation is provided with the pull request.

https://github.com/eBaySF/joomla-platform/blob/data-package/docs/manual/en-US/chapters/packages/data.md

Some historic information on `JObject` is also provided:

https://github.com/eBaySF/joomla-platform/blob/data-package/docs/manual/en-US/chapters/packages/object.md
#### Business Case

The `JObject` class has severe limitations when trying to merge domain logic with a value or data object patterns.  This can be seen in classes like `JTable` and particularly `JUser` where you need to perform some coding gymnastics to avoid column mismatches when saving to the database.  To this end, we needed a better way [for eBay applications] to interact with the value object itself, and also so it could interact with a Data Mapper pattern (a long term alternative/replacement for the JTable-way of doing things) so that the value object could remain independent of the data source (unlike JTable which is a value object that is tightly coupled to a specific datasource).

`JData` and `JDataSet` provide a convenient way for working with structured, semi-structured and unstructured data.


eddieajau:
Fixed since #1703:
- PHP strict problem in JDataSet
- Recursive dumping on JData and JDataSet objects changed to JDumpable
- Improved documentation verbiage

libraries/joomla/data/data.php
```diff
@@ -0,0 +1,342 @@
...
+	 * Set a property value if not alreay set.
+	 *
+	 * @param   string  $property  The name of the property to define.
+	 * @param   mixed   $value     The value to give the property.
+	 *
+	 * @return  mixed  If the property is defined, the current value, otherwise, the new value is returned.
+	 *
+	 * @since   11.1
+	 */
+	public function def($property, $value = null)
```
LouisLandry: I really don't think there is a need to have `def()` for this object.

eddieajau:
`def` removed.

florianv:
I think it could be useful to implement `Countable` in `JData` which count only one level of properties (not recursively)

eddieajau:
@florianv that would be fine. Either make a pull request against my branch or you could make a pull request after this is merged.

florianv:
@eddieajau I made a pull request for Countable on your repo some time ago

eddieajau:
Ah ok, sorry, I missed the email for that somehow.  Since it's merged, would you mind making it against the this platform, and also, would you mind fixing up that "instanceOf" in JDataSet that Don found?

Thanks in advance.

florianv:
Done in https://github.com/joomla/joomla-platform/pull/1723

eddieajau:
Thanks.  I'll just wait for the pull tester to roll over.

PR #1713, opened 2012-11-23
Update libraries/joomla/table/table.php

artur-stepien:
JTable:reset() should also reset errors array. If they are not reset JTable::getErrors() will return old errors, from previous record.

### DIFF
--- a/libraries/joomla/table/table.php
+++ b/libraries/joomla/table/table.php
@@ -501,7 +501,7 @@ abstract class JTable extends JObject
 	/**
 	 * Method to reset class properties to the defaults set in the class
 	 * definition. It will ignore the primary key as well as any private class
-	 * properties.
+	 * properties (except $_errors).
 	 *
 	 * @return  void
 	 *
@@ -519,6 +519,8 @@ abstract class JTable extends JObject
 				$this->$k = $v->Default;
 			}
 		}
+		// Reset table errors
+		$this->_errors = array();
 	}
 
 	/**
--- a/libraries/joomla/table/table.php
+++ b/libraries/joomla/table/table.php
@@ -519,6 +519,7 @@ abstract class JTable extends JObject
 				$this->$k = $v->Default;
 			}
 		}
+		
 		// Reset table errors
 		$this->_errors = array();
 	}


LouisLandry:
Completely agree with this.  Will leave it open a little while to see if we have dissenting opinions worth a discussion.

libraries/joomla/table/table.php
```diff
@@ -519,6 +519,8 @@ public function reset()
 				$this->$k = $v->Default;
 			}
 		}
+		// Reset table errors
```
eddieajau, line 523: Need a new line before the comment otherwise the code style checks will complain.

eddieajau:
+1 from me.

libraries/joomla/table/table.php
eddieajau, line 523: @malkowitch if you can make this small fix we can get this merged.  Thanks in advance.

artur-stepien:
I got view more questions:
Why primary key isn't cleared? This function when, I saw it first time looked like it is what I searched for. I wanted to reset all JTable fields so could insert another record without creating another instance.What is the purpose to reset everything without primary keys?

Should this stay like it is now? Should reset() clear all fields? Or there should be another function called for example JTable::clear() that will clear all fields to its default state (with primary keys). 

Also there is performance difference with using getInstance() and reset():

``` code
 0.000 seconds (+0.000); 8.18 MB (+8.179) - before start:
 0.279 seconds (+0.279); 8.28 MB (+0.102) - After creating 10,000 instances:
 0.279 seconds (+0.000); 8.28 MB (-0.003) - before start:
 0.454 seconds (+0.175); 8.28 MB (+0.003) - After resetting table 10,000 times:
```

When it comes to loading record 10,000 times, reset() is not needed. But when user want to insert records using
JTable he will have to recreate instance what leads to performance drop. 

Is this function anywhere used in current version of Joomla! CMS?

I thing that the best solution is to leave reset() like in this commit, and create another function clear() that will clear all fields including primary keys.

libraries/joomla/table/table.php
```diff
@@ -519,6 +519,9 @@ public function reset()
 				$this->$k = $v->Default;
 			}
 		}
+
```
elinw, line 522: jenkins says there is some white space here.

ianmacl:
I know it's been a while, but if you can clean up the one code style violation we'll get this merged in.

eddieajau:
There is 1 code style error in this PR. @malkowitch would you be able to fix this over the weekend so we can merge it? Thanks in advance.

artur-stepien:
I'm really new to github and don't use it often. There is new line before comment. Is it wrong?

dongilbert:
I'll get this wrapped up for you. It's just some indenting space at the end of a line.

PR #1715, opened 2012-11-25
$options['_expire'] in JSession expects minutes not seconds

elinw:
There is a mismatch between the factory (which is converting session expiration time to seconds when it creates a session object) and JSession where the property is described as measured in minutes in several places.

https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/session/session.php#L37

https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/session/session.php#L196

This proposes to use minutes consistently.

### DIFF
--- a/libraries/joomla/factory.php
+++ b/libraries/joomla/factory.php
@@ -573,7 +573,7 @@ abstract class JFactory
 		$handler = $conf->get('session_handler', 'none');
 
 		// Config time is in minutes
-		$options['expire'] = ($conf->get('lifetime')) ? $conf->get('lifetime') * 60 : 900;
+		$options['expire'] = ($conf->get('lifetime')) ? $conf->get('lifetime') : 15;
 
 		$session = JSession::getInstance($handler, $options);
 


elinw:
I should say this issue came up in writing some tests and getting 900 as the result of getExpire while the existing tests expected 20.

eddieajau:
I have no problem with the change, but will this affect downstream behaviour?

elinw:
The downstream behavior in the CMS is just that, the default session is currently being set to 15 hours and the documentation in the form of the tooltip says it is being set to 15 minutes.

realityking:
The issue is even worse than that. We're using this value to set session.gc_maxlifetime (https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/session/session.php#L934) which is measured in seconds (http://php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime).

eddieajau:
I'm not sure how to call this fix? Is the CMS going to be ok if we merge it?

dongilbert:
@elinw @mbabker Should we merge or close this? What is the CMS's view on it?

elinw:
I really am not sure what to do. On the one hand we clearly have a situation where the logic is failing and the unit tests prove it (hence the other issue that Andrew closed was not resolvable). On the other hand, it certainly is an unstable change of behavior.

I think I am going to call it close and fix in Framework when session is reconstructed and JCache is merged.  It's just too risky/too much in the way of potential unknown consequences right now.

dongilbert:
OK - thanks for the update.

Issue #245, opened 2011-08-14
User access levels cache not updated when user added to a new group (Joomla issue tracker id=26420)


jmcameron:
(from http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=26420)
## Test Instructions:
1. Log in back end as admin/super-user 
2. Create a new test group and corresponding access level derived from 'Registered' (do not add Super-user to the new access level). 
3. For debug output, edit libraries/joomla/user/user.php and add the following line of debug code to line 357 (after first 'if' statement in getAuthorizedViewLevels() function): echo "USERLEVELS: ".implode(',', $this->_authLevels)."<br/>"; 
4. Edit an article and notice the lists of access levels permitted. Cancel editing. 
5. Edit the admin/super-user account and make it a member of the new test group. Save. 
6. Edit an article again and notice that you are not in the new access level. 
7. Log out and log back in. Edit an article and notice that NOW the new test access level has been added to the list. 
8. Don't forget to remove the debugging code!
## Summary

user access levels cache not updated when user added to a new group
## Details

When a user is edited and is the groups that they are in are changed so that thier viewing access levels change, the cache in the user object is not invalidated or updated.   Only logging out and logging back in will force that change.   This could be very confusing since you expect to see changes you make to membership, etc, take place immediately.

Note: Elin Waring said that is cache-refresh problem is also an issue when a user changes their timezone.  In order for it to take effect, you must log out and log back in.  I have seen this issue as well.

aaronschmitz:
@elinw, I can work on this one, but I'm going to need some guidance - it's a bit over my head.

elinw:
Ok, check for me tonight or tomorrow on IRC and we can look at it.

jmcameron:
Hi Elin and Aaron,

I'm willing to help out with this issue.  I've been preoccupied on other
things and have neglected it.

I started getting set up with a fresh git checkout of the platform, etc, to
try get this patch fixed.

I'm tied up this evening and tomorrow morning (Pacific time).   Maybe we
can touch base tomorrow afternoon/evening?

-Jonathan

On Fri, Mar 30, 2012 at 4:58 AM, elinw <
reply@reply.github.com

elinw:
ok will look for you

jmcameron:
Ok, sorry I did not connect earlier.  I will not be able to follow up on this any time soon.  So I encourage you both to take it over and follow it up.

[...]

PR #1717, opened 2012-11-25
Allow change of user properties when JUser exists

elinw:
One ongoing issue with the advent of flexible user groups has been that if a JUser object has already been created for a logged in user and the user group assignments are changed elsewhere (either by code such as when a payment is made or manually by another user who has the rights to do so) the JUser is not updated since JUser already exists. The same is also true of user parameters such as time zone. Currently the only solution is to have the user logout which is both awkward and loses potentially useful data.

This pull requests adds a check in JFactory to compare the user groups, authorised view levels, and user parameters that are stored against those in JUser.  If they are different the stored values are then used. 

As a consequence of implementing this I also wrote a number of tests, added a method and modified a number of other tests.

### DIFF
--- a/libraries/joomla/factory.php
+++ b/libraries/joomla/factory.php
@@ -220,8 +220,52 @@ abstract class JFactory
 			{
 				$instance = JUser::getInstance();
 			}
+
+			if ($instance->id == 0)
+			{
+				// Check to see if there is an active session for this user
+				$session = self::getSession();
+
+				if ($session instanceof JSession)
+				{
+					$sessionName = $session->getName();
+				}
+				if (!empty($sessionName))
+				{
+					$sessionInput = new JInputCookie;
+					$sessionId = $sessionInput->get($sessionName);
+
+					if (!empty($sessionId))
+					{
+						// Get the userid associated with this session name if there is one.
+						$db = self::getDbo();
+						$query = $db->getQuery(true);
+						$query->select('userid');
+						$query->from($db->quoteName('#__session'));
+						$query->where($db->quoteName('session_id') . ' = ' . $db->quote($sessionId));
+						$db->setQuery($query);
+						$result = $db->loadResult();
+					}
+
+					// If we found a session and it is not for a guest go ahead and instantiate JUser for it.
+					if (!empty($result) && $result != 0)
+					{
+						$instance = JUser::getInstance($result);
+					}
+					// Otherwise instantiate an empty JUser
+					elseif
+					(!($instance instanceof JUser))
+					{
+						$instance = JUser::getInstance();
+					}
+				}
+			}
+			elseif (!empty($instance->id))
+			{
+				$instance = JUser::getInstance($instance->id);
+			}
 		}
-		elseif (!($instance instanceof JUser) || $instance->id != $id)
+		elseif ($instance->id != $id)
 		{
 			$instance = JUser::getInstance($id);
 		}
@@ -568,7 +612,7 @@ abstract class JFactory
 	 */
 	protected static function createSession(array $options = array())
 	{
-		// Get the editor configuration setting
+		// Get the session handler configuration setting
 		$conf = self::getConfig();
 		$handler = $conf->get('session_handler', 'none');
 
--- a/libraries/joomla/session/session.php
+++ b/libraries/joomla/session/session.php
@@ -1003,7 +1003,7 @@ class JSession implements IteratorAggregate
 			}
 		}
 
-		// Check for clients browser
+		// Check for client's browser
 		if (in_array('fix_browser', $this->_security) && isset($_SERVER['HTTP_USER_AGENT']))
 		{
 			$browser = $this->get('session.client.browser');
--- a/libraries/joomla/user/helper.php
+++ b/libraries/joomla/user/helper.php
@@ -193,7 +193,7 @@ abstract class JUserHelper
 	 *
 	 * @param   integer  $userId  The id of the user.
 	 *
-	 * @return  object
+	 * @return  JObject
 	 *
 	 * @since   11.1
 	 */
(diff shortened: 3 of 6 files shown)


elinw:
I believe this has implications for #245.

pasamio:
Presumably you could shift this up into the is_null($id) block since you're only after the user that is in the session which is only valid when the user ID is null.

elinw:
I'm not sure what you mean by " which is only valid when the user ID is null." I could move it up so it doesn't check new instances.

if ($instance instanceof JUser)
{
    my code

}
elseif (is_null($id))
{
}
else
{
            $instance = JUser::getInstance($id);
}

I'm not sure about the UT error which I'm not getting locally. 
(Okay thought about the UT issue on the way to work, I think I need to make sure a session exists before destroying it.)

pasamio:
Try a rebase on the current staging and see if that fixes the UT issue.

I was thinking that you could just put the code in the `if (is_null($id))` block at the top of the function[1] which is where it by default pulls it out of the session and you'd only need to check there however looking a little closer I notice that if you specify the same ID as the currently logged in user it pulls it from the session directly so you can't really put it in the first block.

That all said, the change somewhat makes useless pulling the user out of the session. The idea there I'm sure was that the session was already loaded and the object would already be available without having to do an extra call to the database to load the user again. With this in place, we're always going to be doing a request to the database to load the user so it rather much defeats the purpose of storing the entire user object in the session in the first place when you're going to be always loading it from the database anyway.

[1] https://github.com/elinw/joomla-platform/blob/1456babc48adcbeb4580bb984418dc38a5f5c554/libraries/joomla/factory.php#L217

pasamio:
While I'm thinking of it, this will add a query to all requests with logged in users even though the reality is the majority of requests (I'd even consider 99% of the requests) aren't actually going to turn up anything different.

elinw:
Yes, I think originally I was going to say this isn't really an ideal way to solve this.  I've been working on a less database intensive approach so I'm going to close this for right now.

pasamio:
One thought I just had was to add some sort of flag in the session table to mark the session as stale. This could easily be set with a simple query when a user record is updated by com_user and then the user object reloaded in its own session and the flag cleared. It would require a database change to implement but would be reasonably clean. There would still be a race condition when a user account is updated quickly in succession but that would be better than no updates at all.

elinw:
That's an interesting idea since the main thing I have been doing is on save in JUser updating the session but that by itself will log the front end user out though they get the same session right back when they log back in (it's not a real logout and the session isn't destroyed). If the application had a nice "you need to re authenticate" UI that wouldn't be too bad but it's a bit of a kludge even when Google makes you d that.

Actually it's reallly not that the user is logged out, they would still show up on a list of logged in users. It's that the factory can't access the session data containing their JUser information at that point.

elinw:
This is an updated different approach to the issue that also addresses some other issues such as forcing blocked users to be logged out.  This has fewer queries and only queries the session table.

pasamio:
This still introduces a race condition as well as a whole heap of code I'd need to look over a lot closer which seems overkill to solve the particular problem.

elinw:
This problem is a lot more complex than it seems on the surface which is
why it has been in various trackers for over two years.  There probably
does need to be a more fundamental look at how JUser and JSession are
interacting.

It's also a lot less new code than there was, so I figured I would push it
up and see what you and others said.

If you are looking at the code :) a couple of things that came up in
working on it were that I was concerned about the issue of what is allowed
to be updated without reauthenticating because of security issues. I talked
to Alan about this, and we concluded that block and password should both
force reauthentication.

It's possible there is a race condition if there are numerous processes
modifying the session and/or the user record in quick sequence. Right now I
guess that exists also but since the session isn't modified it would happen
when some combination of front end user, front end user in a second
browser, back end user, and code all could be modifying at the same time. I
have to say that I have not looked at that.

ghost:
Current practice is a feature, not a bug!  

I am using it to dynamically add groups to a logged in user that should NOT be stored to the database.   For example, on 90% of the sites I've worked on, devs define themselves as super-admins and then have to create a bunch of test accounts for different group levels....plus when they are logged on they are super-admin.  Talk all you want about this being poor practice, it is what it is and will never change.  As such, I simply borrow the solution from the linux world which had to deal with the same issue and lost the fight to tell users not to log out and log back in as root.   Instead use sudo so a regular user can invoke commands temporarily as root.  By the same token, I have a simple sudo component on the front end - authorized users can use it to dynamically add groups to their current session.   So now I log on as me for all my sites and I am just a regular registered user.  If I need to view the site as a paid member, sudo->paid.  If I need to view debug data, sudo->dev.  If I need to make some intensive admin changes, sudo->super.

After 30 minutes of inactivity, the plugin will automatically reset my groups to the ones in the user table.  No forgetting to log out.  No problems with remember me.  30 minutes, goodbye super admin, hello registered user.

By the same token, there already IS an onSave event for users.  If you need to refresh user group membership when it changes, a user plugin onSave to do so is the way to go.  Please don't break functionality...

elinw:
It's definitely a  bug, and you can still write code to allow temporary user groups no problem at all unless you are doing in in some very obscure way. It does sound like a neat development tool that you have, maybe you would want to share it and then I could see what the problem you are anticipating is.

So, if you are using save() it is going to the database. That's a fact.  From what you are saying, you are not using save(). So what impact is this having? 
 All this is changing is the session when the database is changed, it's not changing anything that happens in the database. If you aren't changing the database it has no effect. So in fact I would think this is good for you because you want to be able to change the session if the point is to be able to quickly and temporarily changing groups for a logged in user.  In fact it sounds like what you do is change the session without doing the save.  That would certainly be a nice feature to add and not just for developers, I can see a lot of practical use cases for a temporary add user to group. 

One thing this will let us do in fact is allow for the possibility of opt in user groups that have immediate impact. 

@pasamio do you think #__user should have checkout?

elinw:
I've put the unrelated tests at #1736.

eddieajau:
Doesn't appear to be a clear consensus on this one so I'm going to close it. The CMS can fix it locally if need be.

PR #1718, opened 2012-11-26
Allowing JDatabaseDriver::quote to take an array of strings.

eddieajau:
If supplied with an array of strings, the quote method will return an array of quoted strings.  If neither a string nor array is passed to the quote method, an exception is thrown.
Added documentation for the escape and quote methods.
Updated unit test.

This avoids having to do something like the following every time you need to deal with an array:

``` php
$db = JFactory::getDbo();
$q = $db->getQuery(true);
$closure = function (&$v, $k) use ($q)
{
    $v = $q->q($v);
};
array_walk($strings, $closure);
```

### DIFF
--- a/docs/manual/en-US/chapters/packages/database.md
+++ b/docs/manual/en-US/chapters/packages/database.md
@@ -5,6 +5,50 @@
 The *Database* package is designed to manage the operations of data
 management through the use of a generic database engine.
 
+### Escaping strings
+
+Strings must be escaped before using them in queries (never trust any variable input, even if it comes from a previous database query from your own data source). This can be done using the `escape` and the `quote` method.
+
+The `escape` method will generally backslash unsafe characters (unually quote characters but it depends on the database engine). It also allows for optional escaping of additional characters (such as the underscore or percent when used in conjunction with a `LIKE` clause).
+
+The `quote` method will escape a string and wrap it in quotes, however, the escaping can be turned off which is desirable in some situations. The `quote` method will also accept an array of strings (added in 12.3) and return an array quoted and escaped (unless turned off) string.
+
+```php
+function search($title)
+{
+	// Get the database driver from the factory, or by some other suitable means.
+	$db = JFactory::getDbo();
+
+	// Search for an exact match of the title, correctly sanitising the untrusted input.
+	$sql1 = 'SELECT * FROM #__content WHERE title = ' . $db->quote($title);
+	
+	// Special treatment for a LIKE clause.
+	$search = $db->quote($db->escape($title, true) . '%', false);
+	$sql2 = 'SELECT * FROM #__content WHERE title LIKE ' . $search;
+	
+	// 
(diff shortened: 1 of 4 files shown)


dongilbert:
+1 Very Useful

elinw:
So a null is going to throw an exception?

Actually following up on that, this changes the handling of integers and booleans as well, which would have just been quoted as if they were strings before as far as I can tell (and I've made that mistake plenty of times).   So I'm both tempted to say you should maintain the old api and quote them and even more tempted to  be in favor of  stricter typing  and say why not also test is_bool, is_int, and is_float and not quote in those cases? Maybe only throw an exception if is_object.

eddieajau:
Fair point. Fixed.

elinw:
+1  Thanks

ghost:
I'd rather an option to disable type checking.  PHP is loose about type casting for the convenience.  For example, given the following contrived example:

class MyObject {

static $numObjects = 0

function __construct() {
  self::$numObjects++;
}

function **toString() {
   return __CLASS**.'s';
}
}
$object = new MyObject;
$objectStats = array();
$objectStats['seperator'] = ' ';
$objectStats['number']=MyObject::$numObjects;
$objectStats['object'] = $object;
$objectStats['msg'] = 'have been created. -- Percent declared classes:';
$objectStats['percentage'] = (1/count(get_declared_classes()))*100; 
$objectStats['msg2'] = '-- Objects Exist flag';
$objectStats['exists'] =  ($numObjects) ? true: false;

// store the stats into the database
$checkType = false;
$sqlUpdate = ' set objectStats = CONCAT_WS( implode(',', JDatabaseDriver::quote($updateStrings, $checkType));
$sqlUpdate .= ', objectNumber='.$objectStats['number'];
$sqlUpdate .= ', objectMetrics='.$objectStats['percentage];
$sqlUpdate .= ', objectName='.JDatabaseDriver::quote($objectStats['object]);

So with one array of data it will assign an integer to objectNumber, a float to objectMetrics, a string to objectName[via the magic __toString method and the quote method], and an informational string to objectStats

The first line becomes:
CONCAT_WS(' ', '1', 'MyObjects', 'have been created. -- Percent declared classes:', 'nn.nnn', '-- Objects Exist flag', '1');

Which MySQL then converts to:
'1 MyObjects have been created. -- Percent declared classes: nn.nnn -- Objects Exist flag 1'

Granted, I'm not really thrilled with the readability of the above but I certainly have seen code which uses lots of magic methods and such to create things programatically.   Make the checkType flag default to true and then programmers always have the option of overriding it without writing their own subclass.

eddieajau:
@garyamort, I don't understand what `$updateStrings` is in the example.

> $sqlUpdate = ' set objectStats = CONCAT_WS( implode(',', JDatabaseDriver::quote($updateStrings, $checkType));

elinw:
See #1733 for some new tests.

PR #1719, opened 2012-11-26
Fixing namespace unit test failure

dongilbert:
I found a problem in the JLoader Namespacing feature - if a class is already loaded, the autoloader includes the file again anyways.

I download the latest zip of the platform to make sure I didn't break it locally, and got the same errors. This PR fixes the issue by adding a `class_exists($class)` check to the `loadByNamespace*` methods.

### DIFF
--- a/libraries/loader.php
+++ b/libraries/loader.php
@@ -210,7 +210,7 @@ abstract class JLoader
 		$class = strtolower($class);
 
 		// If the class already exists do nothing.
-		if (class_exists($class))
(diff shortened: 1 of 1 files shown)


LouisLandry:
I feel like you should add a false as the second argument in those `class_exists()` calls.  See: http://php.net/manual/en/function.class-exists.php

Otherwise great catch.

dongilbert:
done

PR #1720, opened 2012-11-27
sprintf complains unless the correct number of arguments are passed

okonomiyaki3000:
An undefined variable was being passed to `JText::sprintf` in two places. This was fixed a little while ago, in one case it was replaced but in another it was omitted. Since `sprintf` throws a warning when the format uses numbered arguments but insufficient arguments are passed to the function... we need to pass something. Luckily, there is already a suitable string: `JLIB_APPLICATION_ERROR_COMPONENT_NOT_FOUND`

### DIFF
--- a/libraries/legacy/component/helper.php
+++ b/libraries/legacy/component/helper.php
@@ -397,7 +397,8 @@ class JComponentHelper
 		if (empty(self::$components[$option]))
 		{
 			// Fatal error.
-			JLog::add(JText::sprintf('JLIB_APPLICATION_ERROR_COMPONENT_NOT_LOADING', $option), JLog::WARNING, 'jerror');
+			$error = JText::_('JLIB_APPLICATION_ERROR_COMPONENT_NOT_FOUND');
+			JLog::add(JText::sprintf('JLIB_APPLICATION_ERROR_COMPONENT_NOT_LOADING', $option, $error), JLog::WARNING, 'jerror');
 
 			return false;
 		}


libraries/legacy/component/helper.php
```diff
@@ -397,7 +397,7 @@ protected static function _load($option)
 		if (empty(self::$components[$option]))
 		{
 			// Fatal error.
-			JLog::add(JText::sprintf('JLIB_APPLICATION_ERROR_COMPONENT_NOT_LOADING', $option), JLog::WARNING, 'jerror');
+			JLog::add(JText::sprintf('JLIB_APPLICATION_ERROR_COMPONENT_NOT_LOADING', $option, JText::_('JLIB_APPLICATION_ERROR_COMPONENT_NOT_FOUND')), JLog::WARNING, 'jerror');
```
eddieajau: I have a feeling the length of this line is just over the limit.

okonomiyaki3000:
Is there a rule about that? phpcs didn't complain.

elkuku:
It should generate a `warning` if the line contains more than 150 chars.

Maybe we should change that to `error` ?

eddieajau:
My personal rule-of-thumb is 120 characters (which gives me a bugger between 120 and 150 if I really need it).  The reason is that 120 chars fits really well in the github code windows.

okonomiyaki3000:
I hope you mean buffer. Anyway, I didn't get a warning but the line was surely over the 150 character limit. I've broken it up into two.

eddieajau:
Oops (blush). Merged. Thanks.

elinw:
I wanted to mention about this that the removal of that $error was also done during thee 12.2 to 12.3 period (#1572), but clearly without being positive what the implications are. And not to be a harpie about testing but ... we should not be doing things like that without serious unit tests in place especially in the legacy folder where applications are particularly counting on stability.

AmyStephen:
What is unfortunate is that everyone missed @ianmacl 's message at the time. Looks like he saw this right away.

But, in all fairness, the application was better off after the patch. It fixed one of two problems. Removing the second undefined variable basically turned in one warning for another. Still, +1 in the overall count. ;-)

PR #1721, opened 2012-11-27
Native Namespace Code Prep

dongilbert:
This PR adds in the root namespace backslash on calls to internal PHP and SPL classes. This will facilitate the platform moving to native PHP namespaces, whenever that may be.

More info can be found here: http://www.php.net/manual/en/language.namespaces.faq.php#language.namespaces.faq.globalclass

Also, I've replaced some explicit calls to class names with `__CLASS__` in callables. For example, in `JLoader` when calling `spl_autoload_register()`. Also, where creating a new instance of the containing class, I've changed `new JClassName` to `new static`. I've not completed all instances of these yet.

This PR does not break B/C at all, as the whole codebase is already operating in the root/global namespace. But this will need to be done when we do move to namespaces, so I thought I could help that along.

No unit tests were harmed in the making of this pull request.

### DIFF
--- a/libraries/joomla/data/data.php
+++ b/libraries/joomla/data/data.php
@@ -17,7 +17,7 @@ defined('JPATH_PLATFORM') or die;
  * @subpackage  Data
  * @since       12.3
  */
-class JData implements JDataDumpable, IteratorAggregate, JsonSerializable, Countable
+class JData implements \JDataDumpable, \IteratorAggregate, \JsonSerializable, \Countable
 {
 	/**
 	 * The data properties.
@@ -130,11 +130,11 @@ class JData implements JDataDumpable, IteratorAggregate, JsonSerializable, Count
 		// Check the properties data type.
 		if (!is_array($properties) && !is_object($properties))
 		{
-			throw new InvalidArgumentException(sprintf('%s(%s)', __METHOD__, gettype($properties)));
+			throw new \InvalidArgumentException(sprintf('%s(%s)', __METHOD__, gettype($properties)));
 		}
 
 		// Check if the object is traversable.
-		if ($properties instanceof Traversable)
+		if ($properties instanceof \Traversable)
 		{
 			// Convert iterator to array.
 			$properties = iterator_to_array($properties);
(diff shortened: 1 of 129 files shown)


eddieajau:
Thanks Don. Would you mind adding a page to the docs, probably namspacing.md under the Code Standards heading, which outlines what we need to be aware of when writing, and obviously reviewing, code?

It makes sense for the platform, but I think this could potentially give the CMS some grief with extension developers that want to retain support for PHP 5.2 on CMS 2.5 yet maintain a dual 2.5+3 version package.  I don't have a problem with the change but it's probably worth giving the CMS dev list a heads up that we are talking about this.

dongilbert:
I was looking at adding to the documentation here - https://github.com/joomla/joomla-platform/blob/staging/docs/manual/en-US/coding-standards/chapters/php.md#class-definitions - but it's probably best to make it it's own page, since I'm sure the section will continue to grow. I'll get that added as well as inform the CMS dev list.

eddieajau:
Thanks. I think it's better for it to be highly visible and I think it's easier to point to a single page on the topic when fielding questions (or criticism {cough}) about it.

dongilbert:
See documentation here - https://github.com/dongilbert/joomla-platform/blob/namespace_prep/docs/manual/en-US/coding-standards/chapters/namespacing.md

It's not completed yet, but it's what I could think of for now.

As for that compatibility layer section. We can take it or leave it, it's just my thoughts of what I would like to see happen, even if I maintain that layer myself for the 2 revisions after going fully namespaced.

eddieajau:
Compat. layer sounds good. I can get behind that if that's something we need (we should do some more clean up and shunting to the /legacy tree before we finalise it).

dongilbert:
I agree - a lot of the compatibility needs could be buffered in the legacy tree, and then the remainder picked up by the proposed compatibility layer.

As you suggested in the mailing list, I think it's a great idea to hold off until 13.1 to merge this. Want to change the milestone?

eddieajau:
Bumped to 13.1.

nicksavov:
Nice job, Don!

dongilbert:
Thanks - we pretty much put all our eggs in the J! basket at work - so if I want to use namespaces and all that has to offer, I need to make it happen.

dongilbert:
Suggested way to handle the compatibility layer.

https://gist.github.com/4162595

dongilbert:
Instead of `__CLASS__`, should we use `$this` when passing to a callable?

libraries/joomla/data/data.php
```diff
@@ -17,7 +17,7 @@
  * @subpackage  Data
  * @since       12.3
  */
-class JData implements JDataDumpable, IteratorAggregate, JsonSerializable, Countable
+class JData implements \JDataDumpable, \IteratorAggregate, \JsonSerializable, \Countable
```
mbabker: JDataDumpable shouldn't have the slash here.

libraries/joomla/database/driver/oracle.php
```diff
@@ -519,7 +519,7 @@ public function unlockTables()
 	 */
 	public static function isSupported()
 	{
-		return class_exists('PDO') && in_array('oci', PDO::getAvailableDrivers());
+		return class_exists('PDO') && in_array('oci', \PDO::getAvailableDrivers());
```
realityking, line 522: Shouldn't this be `class_exists('\PDO')`?

dongilbert, line 522: Should be class_exists('\PDO'); <-- needs two \ since the \ is also the
escape character. You need to escape the escape.

I'm going to merge this in to my PlatformNamespace branch, since I'm going
through every file again as it is, and I know I missed some things before.
So - I should close this request. :)

PR #1722, opened 2012-11-27
Consolidating duplicate code checks into isSupported() method

dongilbert:
8 lines of code were duplicated across every `JImageFilter*` class in the execute method to determine whether or not the `imagefilter` function existed. Seemed a little ridiculous to me, so I consolidated the check into it's own `isSupported()` method, and call that instead.

### DIFF
--- a/libraries/joomla/image/filter.php
+++ b/libraries/joomla/image/filter.php
@@ -34,6 +34,16 @@ abstract class JImageFilter
 	 */
 	public function __construct($handle)
 	{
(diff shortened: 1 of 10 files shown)


eddieajau:
Just a thought.  Why don't you add the code checking in `isSupported` into the constructor of `JImageFilter`?

dongilbert:
That's actually a good idea. Then I can drop the isSupported call in each filter. Why didn't I think of that?

PR #1723, opened 2012-11-28
Implement Countable in JData and various fixes

florianv:
Implement Countable in JData
Remove unused variables
Change instanceOf to instanceof
Defaulting depth = 3 in JDataDumpable as it is in JData and JDataSet


eddieajau:
Merged.  Thanks a bunch.

florianv:
No problem. Thanks

PR #1730, opened 2012-12-03
Fix bug in `JDataSet` when unsetting within a foreach loop.

eddieajau:
The use of `offsetUnset` on the current iterator point within a foreach
loop was causing unpredictable results. If the current pointer is being
unset, the pointer is moved back one unless it's the first element of
the array, in which case the pointer is set to false. The `next` method
then checks for this special case of `false` and rolls to be 'first'
element of the array (which would have been the 'next' one if the offset
had not been unset).

This PR also adds a new `clear` method to reset the contents of the set,
and a `keys` method that works in a similar way to the `array_keys`
function.

Tests and documentation have been updated accordingly. The tests on `JData::count` are also consolidated.

### DIFF
--- a/tests/suites/unit/joomla/data/dataTest.php
+++ b/tests/suites/unit/joomla/data/dataTest.php
@@ -233,6 +233,29 @@ class JDataTest extends TestCase
 	}
 
 	/**
+	 * Tests the count method.
+	 *
+	 * @return  void
+	 *
+	 * @covers  JData::count
+	 * @since   12.3
+	 */
+	public function testCount()
+	{
+		// Tests the current object is empty.
+		$this->assertCount(0, $this->_instance);
+
+		// Set a complex property.
+		$this->_instance->foo = array(1 => array(2));
+		$this->assertCount(1, $this->_instance);
+
+		// Set some more properties.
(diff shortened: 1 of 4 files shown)


tests/suites/unit/joomla/data/dataTest.php
```diff
@@ -233,6 +233,29 @@ public function testBind_exception()
...
 
 	/**
+	 * Tests the count method.
+	 *
+	 * @return  void
+	 *
+	 * @covers  JData::count
+	 * @since   12.3
+	 */
+	public function testCount()
```
elinw, line 243: From the way I've been taught it's always better to have separate tests that way when one fails the later ones still run plus it's much easier  to trace failures.

eddieajau, line 243: I get what you are saying, and that method has been suggested to me as well, and I certainly use it and deliberately ignore to fit the occasion.  The tests here were so trivial it's not worth breaking them up.  Whatever the case, the test case is valid in it's own right anyway (assert an initial condition; do some incremental work; assert; do final work; assert).  If the initial set is not empty, no point in proceeding.  The second assertion is more or less redundant.

PR #1732, opened 2012-12-03
Move config methods to JApplicationBase

mbabker:
To clean up a little bit of duplicated code, this moves the `get`, `set`, `loadConfiguration`, and `fetchConfigurationData` methods from the child classes of `JApplicationBase` into the parent class itself.

### DIFF
--- a/libraries/joomla/application/base.php
+++ b/libraries/joomla/application/base.php
@@ -19,6 +19,12 @@ defined('JPATH_PLATFORM') or die;
 abstract class JApplicationBase
 {
 	/**
+	 * @var    JRegistry  The application configuration object.
+	 * @since  12.3
+	 */
+	protected $config;
+
+	/**
 	 * The application dispatcher object.
 	 *
 	 * @var    JEventDispatcher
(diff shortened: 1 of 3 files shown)


eddieajau:
I think most of it is ok; it's perfectly reasonable to assume an application is going to need configuration out of the gate. The only problem I have is what to do with `fetchConfigurationData`.  I don't think that should be in Base, and looking at it, I'm not sure it technically belongs in Web and Cli.

dongilbert:
What are the arguments against having it in the base class? Most (?) apps would require some sort of configuration data.

eddieajau:
The `loadConfiguration` method is fine.  I'd be happier if `fetchConfigurationData` in Base was abstract.

LouisLandry:
Yep, I agree with @eddieajau on that.  I am fine assuming that all apps will have configuration, but I think I'd rather have the `fetchConfigurationData` method be abstract.  Continuing that it would make sense that it remain abstract in `JApplicationWeb`, `JApplicationCli` and `JApplicationDaemon`.

mbabker:
Question - Were `JApplicationWeb` or `JApplicationCli` intended to be abstract, or is it reasonable to assume they could theoretically work on their own (as the code seems to have indicated for the last year)?  Abstracting the method down to those classes would mean abstracting classes that aren't already, an interesting change in the API.

LouisLandry:
Ya I thought about that.  Realistically you can't build anything useful without extending them, so I'm fine making them abstract.

mbabker:
I'll do some work then to tweak this out and submit a new PR when that's ready.

PR #1733, opened 2012-12-03
Add some new tests for quote.

elinw:
These tests document the quote method's handling of data types that are not strings.

### DIFF
--- a/tests/suites/unit/joomla/database/JDatabaseTest.php
+++ b/tests/suites/unit/joomla/database/JDatabaseTest.php
@@ -430,6 +430,85 @@ class JDatabaseTest extends TestCaseDatabase
 	}
 
 	/**
+	 * Tests the JDatabase::quote method.
+	 *
+	 * @return  void
+	 *
+	 * @since   11.4
+	 */
+	public function testQuoteBooleanTrue()
+	{
+		$this->assertThat(
+			$this->db->quote(true),
+			$this->equalTo("'-1-'"),
+			'Tests handling of boolean true with escaping (default).'
+		);
+	}
+
+	/**
+	 * Tests the JDatabase::quote method.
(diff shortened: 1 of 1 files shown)


elinw:
Hopefully I did this in a way that will work with Andrew's pull request, but this documents the behaviors we were discussing (which were slightly different than I thought) so that we can be sure of having a stable api.

eddieajau:
What are you concerned might happen if we don't have these tests? To me, it's just testing that some particular values are being correctly converted to strings.  We could just as easily do the following to enforce that:

``` php
return '\'' . ($escape ? $this->escape((string) $text) : (string) $text) . '\'';
```

Maybe I'm missing the point?

elinw:
It's not some particular values, it is some particular data types (although it is good to document for example that true goes to '1' not 'true' and that false goes to  null and not '0' or 'false') are being converted to strings. Tests are one of the best ways that we have of documenting behavior especially non obvious behaviors and making sure they don't actually change when we change other things. So as we talked about the other day we could have an api that throws exceptions instead but we don't.   I mean sure you don't really need float and integer but why not be complete?   I was a little bit inspired by the http://php.net/manual/en/function.is-string.php is_string example in its thoroughness.

eddieajau:
I'm not convinced it's necessary, and I would be inclined to roll them into one test with multiple assertions anyway.  If others think it's necessary I'm fine with it.

elinw:
Well yeah a data provider will be the way in the end, but I didn't want to mess up your pr or have yours mess up mine :).

eddieajau:
Ok.  I'd have no issue with a data provide pattern.

elinw:
Okay can I do that after your stuff is merged so it doesn't have to be done twice?

PR #1739, opened 2012-12-04
Change registerPrefix and registerNamespace to LIFO

dongilbert:
When registering a prefix or namespace, the last path registered should be the first path searched. This would make overriding classes MUCH easier.


LouisLandry:
I think this makes a ton of sense, and I would consider it an oversight on my part that it doesn't already behave that way..  :+1: 

That being said, I'd like to leave it open for a day or so to see if we get any strong dissenting opinions.

ianmacl:
+1

elinw:
+1 solves gigantic practical problems.

PR #1741, opened 2012-12-07
JUser bind

fanno:
...w users because the values are not yet set, Fixed by moving the check below setProperties

also see:
https://groups.google.com/forum/?fromgroups=#!topic/joomlabugsquad/ggmaWUzb-lM

This is my first pull request so bare with me

### DIFF
--- a/libraries/joomla/user/user.php
+++ b/libraries/joomla/user/user.php
@@ -546,24 +546,6 @@ class JUser extends JObject
 			// Set the registration timestamp
 
 			$this->set('registerDate', JFactory::getDate()->toSql());
-
-			// Check that username is not greater than 150 characters
-			$username = $this->get('username');
-
-			if (strlen($username) > 150)
-			{
-				$username = substr($username, 0, 150);
-				$this->set('username', $username);
-			}
-
-			// Check that password is not greater than 100 characters
-			$password = $this->get('password');
-
-			if (strlen($password) > 100)
-			{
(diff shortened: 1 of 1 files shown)


libraries/joomla/user/user.php
```diff
@@ -615,6 +597,24 @@ public function bind(&$array)
 			return false;
 		}
 
+        // Check that username is not greater than 150 characters
```
eddieajau, line 600: Seems to be one too many indents here?

eddieajau:
I don't suppose you're able to do the unit test for the bind method while you are at it?

https://github.com/joomla/joomla-platform/blob/b4c34af90a8ea35a510eebfb49fdf7baa8bf990d/tests/suites/unit/joomla/user/JUserTest.php#L394

libraries/joomla/user/user.php
fanno, line 600: This is my first pull request so i do not know what the normal process is ? i do not understand your comment/note ??

fanno:
sure i can, i just don't know how, is there a tutorial for it somewhere ?

fanno:
hmm should the check for to long password not throw an error ? now that i have looked a bit more at it? or else how would you be able to check the login if the salt has changed because it have been cut down in length ? (just thinking out laud)

fanno:
like this ? 
https://github.com/joomla/joomla-platform/pull/1743

PR #1742, opened 2012-12-07
Small array merging optimization

ghost:
Premature and possibly irrelevant optimization of loading language files

### DIFF
--- a/libraries/joomla/language/language.php
+++ b/libraries/joomla/language/language.php
@@ -762,19 +762,19 @@ class JLanguage
 
 		if (file_exists($filename))
 		{
-			$strings = $this->parse($filename);
+			$strings = ksort($this->parse($filename),SORT_STRING);
 		}
 
 		if ($strings)
 		{
 			if (is_array($strings))
 			{
-				$this->strings = array_merge($this->strings, $strings);
+				$this->strings = ksort(array_merge($this->strings, $strings),SORT_STRING);
 			}
 
 			if (is_array($strings) && count($strings))
 			{
-				$this->strings = array_merge($this->strings, $this->override);
+				$this->strings = ksort(array_merge($this->strings, $this->override),SORT_STRING);
 				$result = true;
 			}
 		}


eddieajau:
https://github.com/garyamort/joomla-platform/blob/bc57b93e3f7ae274721df01d2e886111b9752a32/libraries/joomla/language/language.php#L814

Would it be better to put the ksort here?

Really interesting analysis though.  Well done.

ghost:
No idea...honestly when I ran across the repetitive calls to array_merge it struck a memory and I was curious to see if my recollection was correct about there being a noticable difference between merging sorted and unsorted associative arrays.  I hate making tests with manufactured data since it doesn't tell you if it will be effective in the real world - and the Joomla CMS was well positioned to have real data and no effort needed to modify it.  So it let me play with xhprof and pull some before/after data.  Once I finished and noticed the large difference, I went ahead and submitted the change since it seems like a decent gain for no cost.

There is a LOT that could be done to really optimize JLanguage, but that is a rabbit hole of premature optimizing...ie it's fun to do but the benefits really don't justify the time[unless your Facebook, Google, Amazon, Ebay or some other extremely large web business where cutting 50-100 milliseconds off every page run could translates into tens of thousands of dollars in reduced server infrastructure]

So respectfully, I'm resisting strongly the fun factor of playing with more optimization and leaving this pull request as is... ie makes things better with no cost, but leaves room for more improvement.

mbabker:
This is causing unit tests failures on the Jenkins server - http://build.joomla.org/job/platform/1186/testReport/

Debugging a little bit myself, the `$strings` variable previously was an array after running `$this->parse($filename)`.  With the ksort, it is now a boolean.  Don't get why this wasn't picked up by the pull tester, but I'm reproducing it locally.

pasamio:
Perhaps explains how the optimisation works. It is curious that the pull tester didn't pick it up but Jenkins certainly did.

ianmacl:
There are/were some issues with the pull tester that I'll have to keep an eye on.

ghost:
Odd though as I still got translated strings..perhaps an oddity in ksort in different versions.

In any case, I redid the optimization here https://github.com/joomla/joomla-platform/pull/1773 and will look at the xhprof analysis to see if it still does any enhancement or not.

infograf768:
This totally breaks language strings translation in CMS.
Strict standards: Only variables should be passed by reference in ROOT/libraries/joomla/language/language.php on line 764

infograf768:
I see that #1773 has apparently replaced this commit.

PR #1746, opened 2012-12-10
Added JDataMapper class

eddieajau:
This pull request adds a new class to the Data package called `JDataMapper`. The idea is that this sits in between data objects, like JData or even JDataSet, and the data source (which could be a database, could be a caching layer, etc).

The documentation for the class can be found here:

https://github.com/eBaySF/joomla-platform/blob/data-package/docs/manual/en-US/chapters/packages/data.md#jdatamapper

For those familiar with eBay's UCM implementation, the mapper will replace the JDatabaseObject class and the mapper is obviously pivotal in moving forward with UCM (at least for the architecture that eBay has been using).  You can see hints of UCM fundamentals in the JDatabaseTableMapper example in the documentation where it is filtering out unknown table columns in doCreate and doUpdate.

### DIFF
--- a/docs/manual/en-US/chapters/packages/data.md
+++ b/docs/manual/en-US/chapters/packages/data.md
@@ -175,6 +175,156 @@ if (!empty($hurt))
 
 `JDataDumpable` is an interface that defines a `dump` method for dumping the properties of an object as a `stdClass` with or without recursion.
 
+### JDataMapper
+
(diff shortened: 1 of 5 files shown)


pasamio:
Looking at the pull tester, there appear to be some minor check style issues:
http://developer.joomla.org/pulls/pulls/1746.html

Three refer to the empty line before statement stuff that was added, one is to use an incremental operator and one is a minor docblock problem where it looks like there was a longer type that was replaced with a shorter one but the spacing wasn't removed to reduce it back to two spaces.

If there aren't any other objections and once you've fixed up the minor CS stuff, I think it's good to go.

eddieajau:
Thanks Sam.  All fixed.

pasamio:
It's still showing an error on line 120 of mapper.php suggesting to use the incremental operator.

eddieajau:
Oh that's just ridiculous.

elinw:
lol i've had that one before.  Sooooo frustrating.

eddieajau:
Closing. I'll add this to the Framework.

PR #1749, opened 2012-12-11
Fix #29692, fix default value of "client_id"

pilou-:
client_id: use "0" as default value.

### DIFF
--- a/libraries/legacy/table/menu.php
+++ b/libraries/legacy/table/menu.php
@@ -32,6 +32,7 @@ class JTableMenu extends JTableNested
 
 		// Set the default access level.
 		$this->access = (int) JFactory::getConfig()->get('access');
+		$this->client_id = 0;
 	}
 
 	/**


pasamio:
Can you elaborate a little more on this issue?

elkuku:
Some more info: http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29692

pilou-:
Let me know if this is the right way to fix this problem. There are other similar PostgreSQL problem:

http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29685
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29678
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29680

once this one validated I could submit merge requests for others.

pasamio:
Sure, but why does this fix the problem? Is it just that the menu isn't loaded or that something is missing? Is it a check while it's being created that isn't initialised properly? I'd suggest defaulting the property in the table class in general instead of setting it in the constructor as it allows the children to override it easier and it isn't dependent upon any other parameters.

I don't see why this is being quoted as a string when it should be treated as an integer value and that should solve the problem. More importantly I'm worried that something isn't binding data appropriately somewhere.

29685: This is probably a field set incorrectly, I'd suggest that it should update the schema instead. A default value might solve the problem but likely the field should probably be reset to have a default value of an empty string.
29678: I'd suggest much the same as the above.
29680: Sounds like someone is quoting integers again instead of converting them to an integer.

eddieajau:
Doesn't appear to be a resolution here. Closing.

PR #1753, opened 2012-12-12
JImage Docs

dongilbert:
Adding documentation for the `JImage` package. Includes method parameters, code examples, etc...

### DIFF
--- a/docs/manual/en-US/chapters/packages/image.md
+++ b/docs/manual/en-US/chapters/packages/image.md
@@ -6,9 +6,7 @@ Manipulating images in raw PHP using the `GD` image* functions requires a lot of
 
 All classes in this package are supported by the auto-loader so can be invoked at any time.
 
-### JImage
-
-#### Construction
+### Construction
 
 When creating a new `JImage` object, the constructor will check that the `gd` extension is loaded, and throw a `RuntimeException` if it is not.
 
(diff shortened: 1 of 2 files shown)


elinw:
Side issue but have you been wondering like me whether it would make sense to build out the filter class more along the lines of the thumbnails. Right now you need to fill in a lot of details like renaming and writing  etc.  
https://github.com/elinw/jshell/blob/master/createthumbnails.php#L86  
versus 
https://github.com/elinw/jshell/blob/master/filterimage.php#L85

dongilbert:
I haven't, but only because I haven't used the filtering aspect of the package yet. It totally makes sense though.

However, I would like to see the resized image to have the size prepended to the file name, rather than to the end just before the file extension. Currently, I need to break up the file name and then insert the desired size parameter when outputting the image in my code.

``` php
$filename = 'someimage.png';

// Create thumbnails of the image. Then to output it in my component.

$ext = substr($filename, strrpos($filename, '.'));
$name = substr($filename, 0, strrpos($filename, '.');

$image = $name . '_300x15' . $ext;

echo $image;
```

If the size was prepended, we could just do

``` php
$filename = 'someimage.png';

$image = '300x150_' . $filename;

echo $image;
```

elinw:
I like the size postpended because then alphabetical listing still works. I don't find sorting by first digit of the size that helpful.

dongilbert:
I can see your point. The way I implement it though, there aren't ever that many images in a directory, since i give each item it's own upload directory in media/component/images/products/#ID#/

So it would be easier for me to do it that way. It'd be nice to turn the file name creation into a callback or something else that could be overridden.

dongilbert:
@elinw - I created the ability to have a custom filename format and pass it to the `createThumbs` method. See #1754

ianmacl:
Looks good.  I'd like to see this followed up with a pull request to wrap some of the long lines.

dongilbert:
@ianmacl are you referring to the **Accepted Parameters** sections? Those should be in lists. I'll have to check on the formatting.

PR #1754, opened 2012-12-13
Adding ability to custom format filename of generated thumbnails

dongilbert:
This PR adds a new `$format` parameter to the `createThumbs` method that allows you to specify the filename format to use when generating and saving thumbnails. It must contain the following three strings, which will be replaced when saving a thumbnail:

``` php
/**
 * {filename} - replaced with original files name
 * {width} - replaced with width of generated thumbnail
 * {height} - replaced with height of generated thumbnail
 */

// The default format
$format = '{filename}_{width}x{height}';

// Alternative format (that I use)
$format = '{width}x{height}_{filename}';
```

If any one of the three are missing, an InvalidArgumentException is thrown.

### DIFF
--- a/libraries/joomla/image/image.php
+++ b/libraries/joomla/image/image.php
@@ -214,9 +214,13 @@ class JImage
 	 * Method to create thumbnails from the current image and save them to disk. It allows creation by resizing
 	 * or croppping the original image.
 	 *
-	 * @param   mixed    $thumbSizes      string or array of strings. Example: $thumbSizes = array('150x75','250x150');
+	 * @param   mixed    $thumbSizes      String or array of strings. Example: $thumbSizes = array('150x75','250x150');
 	 * @param   integer  $creationMethod  1-3 resize $scaleMethod | 4 create croppping
-	 * @param   string   $thumbsFolder    destination thumbs folder. null generates a thumbs folder in the image folder
+	 * @param   string   $thumbsFolder    Destination thumbs folder. null generates a thumbs folder in the image folder
+	 * @param   string   $format          Filename format for the thumbnails to be saved to. Exclude extension from the format.
+	 *                                    {filename} - This will hold the files original name.
+	 *                                    {width} - This will hold the thumbnail width.
+	 *                                    {height} - This will hold the thumbnail height.
 	 *
 	 * @return array
 	 *
@@ -225,12 +229,18 @@ class JImage
 	 *
 	 * @since 12.2
 	 */
-	public function createThumbs($thumbSizes, $creationMethod = self::SCALE_INSIDE, $thumbsFolder = null)
+	public function createThumbs($thumbSizes, $creationMethod = self::SCALE_INSIDE, $thumbsFolder = null, $format = '{filename}_{width}x{height}')
 	{
 		// Make sure the resource handle is valid.
 		if (!$this->isLoaded())
 		{
 			throw new LogicException('No valid image was loaded.');
+		}
+
+		// Ensure that the filename format includes our required variables.
+		if (strpos($format, '{filename}') === false || strpos($format, '{width}') === false || strpos($format, '{height}') === false)
+		{
+			throw new InvalidArgumentException('Requested filename format is invalid. ' . $format);
 		}
 
 		// No thumbFolder set -> we will create a thumbs folder in the current image folder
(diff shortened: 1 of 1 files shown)


dongilbert:
I'm not convinced this is the best implementation, but the idea is definitely needed. If someone could come up with a better way to format the `$format` or to run a fancy `preg_replace` on it instead of my simple `str_replace`, let me know. I'll integrate it and update the PR.

pasamio:
I disagree with enforcing that you have to include all them. I can think of a few cases where I might not require the other two parameters:
- If I'm dealing with squares and I know I'm always dealing with squares then one of the dimensions is redundant. Digital Blasphemy also only include a single number to identify their various image zips so it isn't an unlikely case.
- To the extreme, the thumbnail directory may be a local specific to an individual image or path. Consider converting an avatar for a user into multiple formats and sticking that in a directory matching the required sizes. This was something that occurred at eBay for the DataHub where the source file was ignored entirely and thrown away once the thumbnails were generated (which were only files in a directory)

I feel it would be reasonable to at least enforce either the width or the height is required. You could scan the array to check that the selected on is unique though that's bordering a bit much. That does rule out a case where the user just wants to quickly resize and output a single thumb as well as the original though I'm not sure if that is a use case we care about - in which case requiring any of them at all is unnecessary. Fundamentally I feel if the user wants to instruct the system to overwrite the same file more than once then this isn't something that we necessarily need to protect against.

dongilbert:
Your comments make sense @pasamio - maybe instead of enforcing the min standard, we just provide clear documentation of what is available to use in the filename, and let them at it.

drmmr763:
I was implementing the generateThumbs method on my own script just yesterday. Take a look at @phproberto's comment at the bottom of this gist (https://gist.github.com/4268599) for how I eventually went about using the method. 

At the beginning I was wondering about the best way to use the generateThumbs method since saving to a file wasn't entirely clear, but after reading Roberto's comment and figuring out the implementation, I actually prefer doing it the way that he does. You can easily come up with a naming scheme that is pretty easy to use and doesn't require modifications to the method. 

So - I'm not sure if the method really requires this change. Its fairly flexible the way it is. Maybe adding a predefined array with some options that are already set would help to prevent writing what I had to do above and make the method a lot easier to use / implement.

dongilbert:
@phproberto's comment there is basically a duplication of the `createThumbs` method. If this PR had already been implemented, you could have saved yourself the time by specifying the format when calling`createThumbs`.

phproberto:
The reason for having a generateThumbs method is to customize the way that thumbs are saved. In fact I originally only created the createThumbs method and someone rightly suggested this (I think it was @realityking ).

I understand what you are trying @dongilbert . I also thought in implementing it. But as @drmmr763 I finally decided that the library should only contain a standard method. 

Any naming method will not probably fit the requirements of two different developers. So if someone needs a different one it can post-process the generateThumbs output or (the best way for me) extend the JImage class and create a customized createThumbs method.

Thanks for your work with this library Donald!

dongilbert:
Some good points there @phproberto.

The `createThumbs` method does a good job in it's own right (great work there) - it does most of what any developer using the library would desire. I actually only recently discovered it, having been implementing my own code to accomplish the same thing in the past. I think the most common use of overriding the method would be to change the format of the filename. While I agree that the platform should contain a standard method, don't you think it would be reasonable for it to support a simple override?

Also, after thinking more on it, I don't think modifying the format via a parameter in the `createThumbs` method is the best way to approach this. Maybe a protected var that can be overwritten by extending classes would be better?

``` php
class JImage
{
    ...
    protected $thumbFilenameFormat = 'STANDARD FORMAT';
}
```

Or, maybe abstract the filename creation to a `createThumbFilename` method?

But at that point, you're needing to extend the class, and you might as well override the `createThumbs` method.

I think I just talked myself out of the PR. :(

drmmr763:
I'd be interested in seeing a method to handle it. It seems like a lot of extra steps to extend the class _just_ so you can specify a custom naming method. If we had a method that did it by just passing a naming scheme, that would be simpler I think.

dongilbert:
So a `setThumbFilenameFormat($format)` method?

drmmr763:
Yes. I'm thinking it could work chained in before using toFile? Something like:

`$thumbs = $imgObj->generateThumbs(array('500','400'));
$format = array('pretext', 'posttext', 'ext');
$setThumbFilenameFormat($format);
foreach($thumbs as $thumb)
{
$thumb->toFile($path);
}`

disclaimer: I put very little thought into the design of this so I may be totally off in what I'm thinking. I'd trust just anyone else's expertise over my own here!

phproberto:
Think in most probable cases:

image.jpg
image_mini.jpg
image_small.jpg
image_normal.jpg
image_big.jpg
image_huge.jpg

or K2 style:

image.jpg
image_XS.jpg
image_S.jpg
image_M.jpg
image_L.jpg
image_XL.jpg
image_XXL.jpg

I think that both deserve to extend the JImage class as you already have to define sizes correspondecies. 

Having in mind the examples if I had to choose a solution I would probably use something simple like:

```
$thumbSizes = array( 
    'XS'  => '100x100',
    'S'   => '150x150',
    'M'   => '200x200',
    'L'   => '250x250',
    'XL'  => '300x300',
    'XXL' => '350x350'
);
$thumbs = createThumbs($thumbSizes);
```

This way the createThumbs method will replace the suffixes with the array KEYS. This method will work also for the mini, smal, etc. example.

eddieajau:
@dongilbert is this worth merging or shall we fix it in the FW?

dongilbert:
I'll put something together for the FW.

PR #1760, opened 2012-12-16
Jloader prefix

ghost:
# Codefix

Fixes loadByPrefix autoloader[_load] so that it doesn't give up on the first match.
# Ex

Prefix: Tree, Path stubs/tree
Prefix: TreeLeaf, Path stubs/tree/leaf
class_exists(TreeLeafLeaf)
# Current behavior

Will try to load either file stubs/tree/leaf/leaf.php or stubs/tree/leaf/leaf/leaf,php - but not both.  Which file is attempted depends on which Prefix was loaded last(?).  Even if the first file does not exist, it will still be the only file attempted.
# New behavior

Check to class_exists is made after loading and only if true will the autoloader return.
# Unit Tests

Implemented testRegisterPrefix in order to add test for this codefix
Included regression test to ensure that for multiple matching prefixes, multiple files are included if required.  This test should fail for older platform code and succeed on this code.

```
    // Regression check that multiple matching prefixes can load multiple files
    $this->assertTrue($fileTreeLeafLeafLoaded, 'Tests that included files not defining class do not abort load.');
```

All other tests should succeed regardless of Platform version.

### DIFF
--- a/libraries/loader.php
+++ b/libraries/loader.php
@@ -561,7 +561,10 @@ abstract class JLoader
 
 			if (strpos($class, $prefix) === 0 && ($chr === strtoupper($chr)))
 			{
-				return self::_load(substr($class, strlen($prefix)), $lookup);
+				if(self::_load(substr($class, strlen($prefix)), $lookup))
+				{
+					return true;
+				}
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


eddieajau:
@garyamort is this something you think CMS 3.1 will need/want?

dongilbert:
I proposed something similar to the PHP-FIG to modify PSR-0 so that it only returns true once the class is loaded. I think this is a good patch, and would help the CMS. If @elinw or @dextercowley could confirm that they would like this in the CMS, then I think it's good to merge.

eddieajau:
Since there has been no response I'm going to close this one.

PR #1761, opened 2012-12-16
Update PHPMailer to 5.2.2

mbabker:


dongilbert:
I'm all for upgrading. +1 on that.

However, I'd much rather see SwiftMailer support instead. It's LGPL - so maybe I should just issue a PR.

elinw:
I think the goal is to replace this with a native package.

dongilbert:
@elinw I see the benefits of creating a monolithic platform to develop on - but why reinvent the wheel when there are great solutions out there like SwiftMailer that can be easily integrated. 

I'd love to see the Platform focus on being the BEST core solution for MVC, Routing, and other base stuff, and then everything else is package / library / component based. We can then focus on each package individually if we feel the need to. Like making the best PHP Mailing package available, and the best Image management package. But - that's probably just the idealist in me talking - we are a long way from that, if ever.

PR #1762, opened 2012-12-17
Only attempt to update a session from JUserHelper if there is a session.

elinw:
If you are creating a user in a cli application updating the current session does not make sense.


elinw:
Right now you can't add a user to groups because that triggers errors caused by going to JFactory.

PR #1764, opened 2012-12-17
Accessibility: add aria-required attribute to JFormField

SniperSister:
The aria-required attribute should be added to any required fields that are generated via JFormField.

### DIFF
--- a/libraries/joomla/form/field.php
+++ b/libraries/joomla/form/field.php
@@ -345,6 +345,7 @@ abstract class JFormField
 		$multiple = (string) $element['multiple'];
 		$name = (string) $element['name'];
 		$required = (string) $element['required'];
+		$ariaRequired = (string) $element['aria-required'];
 		$disabled = (string) $element['disabled'];
 		$readonly = (string) $element['readonly'];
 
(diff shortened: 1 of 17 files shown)


SniperSister:
Okay, I added the attributes to the actual markup output and it seems to be working fine now.

dongilbert:
:+1:

realityking:
Wouldn't it be better to add required="required"? In most browsers (the notable exception being IE8) this is mapped to the corresponding ARIA state. In the CMS we already have a JS based fallback for this.

dongilbert:
He added both in the `required="required"` and `aria-required="true"` in this PR - I think it would be best to add both, since not all browsers support mapping the required attribute to the aria attribute.

ianmacl:
http://developer.joomla.org/pulls/pulls/1764.html

There are issues there that need to be addressed before we look at this.

SniperSister:
I fixed the issues so it should be working now.

dongilbert:
@SniperSister add a comment with the word "reopen" not in quotes, and it will reopen the pull request automagically. Then I can review the changes and it will update the developer.joomla.org pull tester as well (not right away though). :)

SniperSister:
Thanks for the hint ;)

reopen

PR #1769, opened 2012-12-21
Abstract the JApplicationBase children

mbabker:
As discussed in #1732 and https://groups.google.com/d/topic/joomla-dev-platform/XuAhcjebPzs/discussion, this PR abstracts our JApplicationBase instances.  To keep the legacy JApplication working as is, I've only added one abstract method to the root JApplicationBase class, but would like to move others up after it comes out of the Platform's tree.

The get, set, and loadConfiguration methods are moved from the child classes to base since these are duplicated methods across.

The fetchConfigurationData is now abstract in the base class and remains as is in JApplicationCli and JApplicationWeb until 13.3 to give developers time to implement the method in their applications.

The doExecute method is abstract in JApplicationCli and JApplicationWeb and I would like to move its definition to JApplicationBase after the legacy JApplication class comes out of the Platform tree.  As well, I would then abstract the execute method in the JApplicationBase class at that point.

Lastly, this also removes `JApplicationWeb::initialise()` as it was marked deprecated for 13.1.


mbabker:
Applied to the framework repo.

PR #1772, opened 2012-12-23
Move JApplication**::getInstance methods to JApplicationBase

mbabker:
This will probably best work with #1769 if it is merged, FYI.

This pull deprecates the getInstance methods of JApplicationCli and JApplicationWeb in favor of having the getInstance method in JApplicationBase.  Generally, only one application can be running properly within a workspace, so this is IMO a logical move to support that theory.  If accepted, this method could theoretically replace JFactory::getApplication (another idea discussed in #1729).

### DIFF
--- a/libraries/joomla/application/cli.php
+++ b/libraries/joomla/application/cli.php
@@ -25,12 +25,6 @@ class JApplicationCli extends JApplicationBase
 	protected $config;
 
 	/**
-	 * @var    JApplicationCli  The application instance.
-	 * @since  11.1
-	 */
-	protected static $instance;
-
-	/**
 	 * Class constructor.
 	 *
 	 * @param   mixed  $input       An optional argument to provide dependency injection for the application's
@@ -120,23 +114,13 @@ class JApplicationCli extends JApplicationBase
 	 * @return  JApplicationCli
 	 *
 	 * @since   11.1
+	 * @deprecated  14.1  Use JApplicationBase::getInstance() instead.
 	 */
 	public static function getInstance($name = null)
 	{
-		// Only create the object if it doesn't exist.
-		if (empty(self::$instance))
-		{
-			if (class_exists($name) && (is_subclass_of($name, 'JApplicationCli')))
-			{
-				self::$instance = new $name;
-			}
-			else
-			{
-				self::$instance = new JApplicationCli;
-			}
-		}
+		JLog::add(sprintf('%s is deprecated, use JApplicationBase::getInstance() instead.', __METHOD__), JLog::WARNING, 'deprecated');
 
-		return self::$instance;
+		return parent::getInstance($name);
 	}
 
 	/**
--- a/libraries/joomla/application/base.php
+++ b/libraries/joomla/application/base.php
@@ -43,6 +43,12 @@ abstract class JApplicationBase
 	public $input = null;
 
 	/**
+	 * @var    JApplicationBase  The application instance.
+	 * @since  13.1
+	 */
+	protected static $instance;
+
+	/**
 	 * Method to close the application.
 	 *
 	 * @param   integer  $code  The exit code (optional; default is 0).
@@ -67,6 +73,36 @@ abstract class JApplicationBase
 	public function getIdentity()
 	{
 		return $this->identity;
+	}
+
+	/**
+	 * Returns a reference to the global JApplicationBase object, only creating it if it doesn't already exist.
+	 *
+	 * This method must be invoked as: $app = JApplicationBase::getInstance();
+	 *
+	 * @param   string  $name  The name (optional) of the JApplicationBase class to instantiate.
+	 *
+	 * @return  JApplicationBase  Application instance
+	 *
+	 * @since   13.1
+	 * @throws  RuntimeException
+	 */
+	public static function getInstance($name = null)
+	{
+		// Only create the object if it doesn't exist.
+		if (empty(static::$instance))
+		{
+			if (class_exists($name) && (is_subclass_of($name, 'JApplicationBase')))
+			{
+				static::$instance = new $name;
+			}
+			else
+			{
+				throw new RuntimeException('Could not instantiate application: ' . $name);
+			}
+		}
+
+		return static::$instance;
 	}
 
 	/**
--- a/libraries/joomla/application/web.php
+++ b/libraries/joomla/application/web.php
@@ -73,12 +73,6 @@ class JApplicationWeb extends JApplicationBase
 	protected $response;
 
 	/**
-	 * @var    JApplicationWeb  The application instance.
-	 * @since  11.3
-	 */
-	protected static $instance;
-
-	/**
 	 * Class constructor.
 	 *
 	 * @param   mixed  $input   An optional argument to provide dependency injection for the application's
@@ -155,23 +149,13 @@ class JApplicationWeb extends JApplicationBase
 	 * @return  JApplicationWeb
 	 *
 	 * @since   11.3
+	 * @deprecated  14.1  Use JApplicationBase::getInstance() instead.
 	 */
 	public static function getInstance($name = null)
 	{
-		// Only create the object if it doesn't exist.
-		if (empty(self::$instance))
-		{
-			if (class_exists($name) && (is_subclass_of($name, 'JApplicationWeb')))
-			{
-				self::$instance = new $name;
-			}
-			else
-			{
-				self::$instance = new JApplicationWeb;
-			}
-		}
+		JLog::add(sprintf('%s is deprecated, use JApplicationBase::getInstance() instead.', __METHOD__), JLog::WARNING, 'deprecated');
 
-		return self::$instance;
+		return parent::getInstance($name);
 	}
 
 	/**
--- a/tests/suites/unit/joomla/application/JApplicationBaseTest.php
+++ b/tests/suites/unit/joomla/application/JApplicationBaseTest.php
@@ -58,6 +58,41 @@ class JApplicationBaseTest extends TestCase
 	}
 
 	/**
+	 * Tests the JApplicationBase::getInstance method.
+	 *
+	 * @return  void
+	 *
+	 * @since   13.1
+	 */
+	public function testGetInstance()
+	{
+		$this->assertInstanceOf(
+			'JApplicationBaseInspector',
+			JApplicationBase::getInstance('JApplicationBaseInspector'),
+			'Tests that getInstance will instantiate a valid child class of JApplicationBase.'
+		);
+
+		TestReflection::setValue('JApplicationBase', 'instance', 'foo');
+
+		$this->assertEquals('foo', JApplicationBase::getInstance('JApplicationBaseInspector'), 'Tests that singleton value is returned.');
+	}
+
+	/**
+	 * Tests the JApplicationBase::getInstance method for a thrown RuntimeException.
+	 *
+	 * @return  void
+	 *
+	 * @since   13.1
+	 * @expectedException  RuntimeException
+	 */
+	public function testGetInstanceException()
+	{
+		TestReflection::setValue('JApplicationBase', 'instance', null);
+
+		JApplicationBase::getInstance('NonExistingClass');
+	}
+
+	/**
 	 * Tests the JApplicationBase::loadDispatcher method.
 	 *
 	 * @return  void
--- a/tests/suites/unit/joomla/application/JApplicationCliTest.php
+++ b/tests/suites/unit/joomla/application/JApplicationCliTest.php
@@ -284,14 +284,6 @@ class JApplicationCliTest extends TestCase
 		TestReflection::setValue('JApplicationCli', 'instance', 'foo');
 
 		$this->assertEquals('foo', JApplicationCli::getInstance('JApplicationCliInspector'), 'Tests that singleton value is returned.');
-
-		TestReflection::setValue('JApplicationCli', 'instance', null);
-
-		$this->assertInstanceOf(
-			'JApplicationCli',
-			JApplicationCli::getInstance('Foo'),
-			'Tests that getInstance will instantiate a valid child class of JApplicationCli given a non-existent type.'
-		);
 	}
 
 	/**
--- a/tests/suites/unit/joomla/application/JApplicationWebTest.php
+++ b/tests/suites/unit/joomla/application/JApplicationWebTest.php
@@ -949,14 +949,6 @@ class JApplicationWebTest extends TestCase
 			$this->equalTo('foo'),
 			'Tests that singleton value is returned.'
 		);
-
-		TestReflection::setValue('JApplicationWeb', 'instance', null);
-
-		$this->assertInstanceOf(
-			'JApplicationWeb',
-			JApplicationWeb::getInstance('Foo'),
-			'Tests that getInstance will instantiate a valid child class of JApplicationWeb given a non-existent type.'
-		);
 	}
 
 	/**


libraries/joomla/application/cli.php
```diff
@@ -120,23 +114,13 @@ public function get($key, $default = null)
...
-			if (class_exists($name) && (is_subclass_of($name, 'JApplicationCli')))
-			{
-				self::$instance = new $name;
-			}
-			else
-			{
-				self::$instance = new JApplicationCli;
-			}
-		}
+		JLog::add(sprintf('%s is deprecated, use JApplicationBase::getInstance() instead.', __METHOD__), JLog::WARNING, 'deprecated');
```
realityking, line 121: Can we solve this more elegantly with late static binding?

mbabker, line 121: How would you do it?  I still haven't done much with late static bindings, so tips are welcome.

ianmacl:
I would more be in favour of outright getting rid of JApplication(Base)::getInstance for the following reasons:
1. It is less clear than just doing: $app = new MyApplicationWeb or MyApplicationCli
2. There is no reason to make it a static, and having it as a static encourages poor design that is hard to test.
3. While it is usually true that there is only one application object it is not _always_ true.  It is always better to inject your dependencies rather than trying to grab them out of global state variables.

mbabker:
I don't think the application is something we can move completely away from not having static variables for.  Not sure about newer use cases, but there's a high chance that trying to reinstantiate the application or instantiate a second instance will probably cause problems somewhere.  So, IMO, I'd like one global static instance of the app somewhere (be it here or JFactory doesn't matter), but we should keep it.

florianv:
I agree with @ianmacl, applications might be singletons but not always.
Moreover, with the current getInstance method it is impossible to use two different applications in the same script.

Regarding JFactory, its references need to be removed from the Joomla core, and maybe this class will need to be removed later too. It makes the libraries very difficult or impossible to test.

AmyStephen:
I agree, statics are evil. But, I think you are throwing out the baby with the bath water here. 

What is amusing as I watch this discussion is knowing there are OO puritans who believe an approach of using a front controller, which instantiates system objects and then injects child objects with these values, is no better than using singletons, globals or statics. Makes you wonder, is there really "a way?" And if there is, why aren't more people blogging about it instead of what they think is wrong?

There will always be data and functionality with global scope. Session handling, database access, application configuration, user object, and so on. These objects are - by their very nature -- globals. (gasp at the word.)

In an application development framework, like JPlatform, a developer expects a way to configure data and functionality as having a system-wide scope - and expects there is an API available that can be used to access that data and use that functionality throughout the application. 

JFactory performs that function for the Joomla framework. It is a very easy tool for even novice developers to use so that they can determine if SEF is activated, add a CSS lnk to the list that must be loaded in the head, or determine if a registered user is logged on, and so on. If anything, I would like to see JFactory (or it's replacement) be customizable so that developers can define new data and new functionality for global use.

Also, I'm not so sure I agree with some of the conclusions above. For example, why is the Application class relegated to a singleton status in Joomla? I disagree that it is because JFactory uses statics. It's because the index.php file instantiates JFactory, and it instantiates JApplication using a static and calls to JApplication are then resolved by and routed through JFactory. That's where the limit occurs.

Now shift the sequence. Have the index.php instantiate JApplication, and have JApplication use a static (gasp) to instantiates JFactory and now, you can create Application objects until the cows come home. JFactory can continue to instantiate those objects with Application-wide scope (without using statics), and the benefit to easy access to application-wide data and features for developers remains. 

I'd have to understand more about why it's "better design" to force the Application class to instantiate system wide objects and then inject each of it's child objects with the complete list (and those objects have to do the same, and so on). Such an approach is certainly not DRY. It also forces objects to be instantiated in anticipation of use -- if they are shared -- instead of adopting more of a lazy load approach. More importantly, though, in trying to "avoid doing it wrong", capability is simply not provided which starts to diminish the value a developer gets from using even using JPlatform. 

If they have to code it all themselves, why not recommend PHP and a good book on OOP? Or, maybe we are conceding that Rasmus Lerdorf is right. Frameworks are not helpful, they are limiting. http://toys.lerdorf.com/archives/38-The-no-framework-PHP-MVC-framework.html

A good application development framework is going to be opinionated, a great framework will be very opinionated, no framework will stand up to academic debate. Much more important to set clear goals on the purpose of JPlatform, to think very specifically on why developers should choose it over PHP alone, and then measure what you are doing against that standard, keeping in mind development principles, but not limiting yourselves to correctness.

There are many things that were implemented incorrectly in 1.5 and those decisions have created in some places very severe limitations. But, the other side of that is the architecture actually has all the right big concepts to become a very powerful development tool. There is a lot of good thinking to work with in there.

dongilbert:
Statics aren't evil.

mbabker:
IMO, it all depends on how they're used.  Having a static instance of certain higher level classes IMO isn't necessarily a bad thing.  I think there's a reason that `JDatabaseFactory` hasn't seen any adoption in the code over `JFactory::getDbo()`.

Also, if we removed all static references to the application (here and `JFactory`) and every time the application was needed, it was called as `$app = new MyApp;`, how many errors would we immediately have to fix because it is assumed by 99% of our developers that the application is only instantiated once?

AmyStephen:
And @mbabker that's the rub. If one were to create a new instance of the Application, then they would not be able to reference any of the properties or functionality in the existing instance. 

The Application Class really is a Front Controller. http://www.martinfowler.com/eaaCatalog/frontController.html 

It manages all application-wide properties and functionality.

If you instantiate JFactory inside of JApplication and you store the JFactory instance as a static in JApplication, then you can use that static as an entry point to access objects instantiated in JFactory.

Kind of a hack ?, I'll grant that, I suppose, but, a very handy one, IMO.

A must read article by Martin Fowler on Dependency Injection. The "Using a Service Locator" section and "Service Locator vs Dependency Injection" are very good.

http://www.martinfowler.com/articles/injection.html

Related code I shared a couple of weeks ago: https://gist.github.com/4189804

There's always a tradeoff.

mbabker:
If I instantiate a new instance of something, I expect it to be fresh.  I think that's a reasonable assumption to make.  Likewise, if I'm accessing something at a global level, such as how we access the application object now, I should be able to safely assume that certain data will be available after I inject it into that object.

No matter what move we make, it's a sticky situation.  I guess the question is what mid-point can we agree to meet at?

AmyStephen:
I think this is the point that someone comes in and says "This is an Application issue." 

And, I'd almost agree with that -- except first, I wonder if it is it worth considering that JApplication is acting as both a Front Controller and an Application Class? And if it is, does it make sense to split that cleanly so that each is focused on their function and data? 

Then, would it be helpful for applications using JPlatform to have a class like JFactory that is instantiated within the Front Controller that could be used to define global sharing of objects? Having that option would not preclude injection so it would at least provide a choice to developers without issuing a judgement.

Another factor that fits into this is JRegistry which is also "suspect" as it serves as global memory space when coupled with JFactory. What is the intended future for this class and practice?

I know from phrases and half sentences I've read on list or in the issue queues that the platform team is aware of the issues around statics, it's also apparent there is a desire to move away from this practice, but it really would be good to hear what the larger strategy is for access to application-wide data and functionality within JPlatform. 

Bad day to ask for that, given recent resignations.

HermanPeeren:
Service Locator is an  anti-pattern: http://blog.ploeh.dk/2010/02/03/ServiceLocatorIsAnAntiPattern.aspx
Static methods not good for testability:  http://misko.hevery.com/2008/12/15/static-methods-are-death-to-testability/

AmyStephen:
@HermanPeeren 

First, only one static is called for in my suggestion and that is a static defined within the frontcontroller (Application class) for the JFactory (replacement) instance. It's only purpose is to create a nice entry point back into the Frontcontroller/JFactory instance. 

I'm maybe not seeing how the Anti-pattern article fits here. 

The point I see raised is that the dependency to the Service Class is not clear -- the IDE will not provide a hint that an instance is required to a necessary class, for example to the Database class or the Messages class. What the author points out is that the lack of hint will cause developers to  merrily go forward using the resource without first creating an instance to the database or messages class and therefore encounter runtime errors. Am I understanding that correctly?

If so, that's not a problem that is relevant here. There is no instance required in the developer's code to get an instance to the database or message class. The Application's instance of JFactory will hand off the request to a process that will create that instance for the developer and return it for their use. So, no runtime error will result. 

But, maybe I missed something? If so, it would help if you explained a bit more how you see this as fitting.

Related to better communicating breaking changes, though, there's another benefit to a Services Layer approach. If the API is offered within a controlled layer, and direct access to core classes is discouraged, you begin to have a lot more flexibility with changing things without impacting your developer community. In many ways the API doesn't really change - read a file, run a query, get an extension, render this output, and so forth, are stable functions over time. It's how you do it is that evolves. Those kinds of changes can be masked if there is a layer for the API that is separated from the core code.

AmyStephen:
@mbabker If there is interest in putting an optional services layer together, in addition to using dependency injection, I'm willing to help with that and have some starting code we could use. Just let me know.

HermanPeeren:
Although any framework will inevitably be opiniated, I think we should try to avoid the Joomla Platform to prescribe too much just a single way to accomplish things, thereby limiting developers in its use. A guideline for putting something in the Platform could be: is it providing useful possibilities to developers?

If someone thinks a Service Locator is useful in the Joomla Platform, then make a pull request, provide some examples, spread the word about it to some other developers on a mailing list and above all: have fun using it. I personally probably won't use it and maybe even advise against using it, but I won't be against putting it in the JPlatform, as it might be very useful for some developers. You are in good company of frameworks like for instance Symfony2 and Nooku using this "global in disguise with the fancy name" feature too. 

That Mark Seemann blog about Service Locator being an anti-pattern might not be the clearest statement against using a Service Locator. What I liked very much was the first sentence: "...since it was described by Martin Fowler, it must be good, right?". Here is a good posting by Anthony Ferrara, explaining a bit more why globals should be avoided and why (and by which criteria) a Service Locator is just another way to use globals: http://blog.ircmaxell.com/2012/08/object-scoping-triste-against-service.html

Let's not discuss that here, for we have different styles of programming. I'm into Object Thinking (also title of a good book, 2004, by David West). I might even be a bit of an Object Puritan; I wouldn't dislike to be called like that. Others might be more into procedural and data-centric programming, thinking in data and functionality as two separate things. Much of Joomla's way of programming is more in that procedural direction, hence all the static functions (like for instance in JAccess). I believe that in the long term more pure OOP is better, but we can only show it with code. I'm working on some "pure OOP" things in the direction of UCM and hope to show it soon. If some people like it, I can make a pull request and it could eventually be added to the JPlatform. If not, then it's OK too.

Back to Michael's pull request. 

Joomla's Application classes are not pure singletons, with a private constructor and clone-function. You can still choose to use "new" to instantiate another application-class. But if you want to reuse an allready instantiated application, getInstance() is available. So, that matches well the starting point of providing possibilities to developers and let the developers choose what to use. 

There is a small difference between the way the getInstance() works now in JPlatform and how it works in this  pull request: the getInstance()-methods of JApplicationWeb and JApplicationCli check if the class you want to be instantiated with some name is a JApplicationWeb or JApplicationCli respectively.  So it could in theory be used to instantiate a JApplicationCli besides a JApplicationWeb using the same name. Of course you can also instantiate an other JApplication with another name. Multiple JApplications using getInstance() won't work correctly however, because only the last instantiated application will be put in the static property $instance. Although in theory the pull request limits the use of getInstance() a tiny bit, in practice there will not be any difference in behaviour between the current code and this pull request.

In this pull request only 1 JApplicationBase-descendant is possible in the getInstance()-method. Maybe Michael's comment "Generally, only one application can be running properly within a workspace" could be changed to the more factual: "getInstance() can only be used to instantiate and use 1 JApplication". It still leaves the door open to instantiating more applications, if any developer would want to do that, by just using "new"; which can be done in a Service Locator, a DI-container or wherever you want. 

In short: I have nothing against this pull request: it doesn't limit possibilities for developers.

AmyStephen:
I agree with @HermanPeeren - there is no limit imposed here and @mbabker should move forward.

Herman - you really are missing the point. Interestingly, the Symfony2 response to Anthony's post appears to be trying to make the same distinction. Maybe it's just something you gotta play with to see, but suffice it to say you can use DI to create and share "globals" and you can create a service layer that only returns new objects.

libraries/joomla/application/cli.php
realityking, line 121: Sorry, forgot all about this during the holidays. Basically I'd remove the static methods from the subclasses and use this code in JApplicationBase (untested):

``` php

if (empty(self::$instance))
{
    if (class_exists($name) && (is_subclass_of($name, 'JApplicationCli')))
    {
        self::$instance = new $name;
    }
    else
    {
        self::$instance = new JApplicationCli;
    }
}
```

mbabker, line 121: That won't work assuming my pull to make the application classes abstract
is accepted.  Likewise, the check would have to be for an instance of
JApplicationBase; both CLI and Web are having their getInstance methods
consolidated in this pull.

On Monday, January 7, 2013, Rouven Weßling wrote:

> In libraries/joomla/application/cli.php:
> 
> > ```
> >  */
> > public static function getInstance($name = null)
> > {
> > ```
> > -       // Only create the object if it doesn't exist.
> > -       if (empty(self::$instance))
> > -       {
> > -           if (class_exists($name) && (is_subclass_of($name, 'JApplicationCli')))
> > -           {
> > -               self::$instance = new $name;
> > -           }
> > -           else
> > -           {
> > -               self::$instance = new JApplicationCli;
> > -           }
> > -       }
> > -       JLog::add(sprintf('%s is deprecated, use JApplicationBase::getInstance() instead.', **METHOD**), JLog::WARNING, 'deprecated');
> 
> Sorry, forgot all about this during the holidays. Basically I'd remove the
> static methods from the subclasses and use this code in JApplicationBase
> (untested):
> 
> if (empty(self::$instance)){
>     if (class_exists($name) && (is_subclass_of($name, 'JApplicationCli')))
>     {
>         self::$instance = new $name;
>     }
>     else
>     {
>         self::$instance = new JApplicationCli;
>     }}
> 
>  —
> Reply to this email directly or view it on GitHubhttps://github.com/joomla/joomla-platform/pull/1772/files#r2571754.

## 
- Michael

realityking, line 121: As I said untested :P I actually didn't do any modification, I just copy and pasted from above and when I was done fighting with GitHub JS and Safari I forgot to actually change the code..

Lets try again:

``` php

if (empty(static::$instance))
{
    if (class_exists($name) && (is_subclass_of($name, 'JApplicationBase')))
    {
        self::$instance = new $name;
    }
    else
    {
        self::$instance = new static;
    }
}
```

mbabker, line 121: It would work for the time being.  Still leaves unanswered what to do if
the classes become abstract.  Guess we can look at that later on.

If you want to copy my commit over and make the change, go ahead.  The bus
just hit the highway and I don't know when I'll next be online on my laptop.

On Monday, January 7, 2013, Rouven Weßling wrote:

> In libraries/joomla/application/cli.php:
> 
> > ```
> >  */
> > public static function getInstance($name = null)
> > {
> > ```
> > -       // Only create the object if it doesn't exist.
> > -       if (empty(self::$instance))
> > -       {
> > -           if (class_exists($name) && (is_subclass_of($name, 'JApplicationCli')))
> > -           {
> > -               self::$instance = new $name;
> > -           }
> > -           else
> > -           {
> > -               self::$instance = new JApplicationCli;
> > -           }
> > -       }
> > -       JLog::add(sprintf('%s is deprecated, use JApplicationBase::getInstance() instead.', **METHOD**), JLog::WARNING, 'deprecated');
> 
> As I said untested :P I actually didn't do any modification, I just copy
> and pasted from above and when I was done fighting with GitHub JS and
> Safari I forgot to actually change the code..
> 
> Lets try again:
> 
> if (empty(static::$instance)){
>     if (class_exists($name) && (is_subclass_of($name, 'JApplicationBase')))
>     {
>         self::$instance = new $name;
>     }
>     else
>     {
>         self::$instance = new static;
>     }}
> 
>  —
> Reply to this email directly or view it on GitHubhttps://github.com/joomla/joomla-platform/pull/1772/files#r2571823.

## 
- Michael

eddieajau:
If you want to make a singelton, then the constructor should be protected. If not, there's no need for the getInstance. At the moment we are half-and-half and I think we should be one or the other. I'd side with Ian and get rid of the getInstance altogether. If that's satisfactory, let's just do that in the FW and leave this code as is (?).

PR #1773, opened 2012-12-23
Remature optimization

ghost:
ksort redone so as not to replace array with boolean

### DIFF
--- a/libraries/joomla/language/language.php
+++ b/libraries/joomla/language/language.php
@@ -177,6 +177,8 @@ class JLanguage
 		{
 			if (is_array($contents))
 			{
+				// Sort the underlying heap by key values to optimize merging
+				ksort($contents, SORT_STRING);
 				$this->override = $contents;
 			}
 			unset($contents);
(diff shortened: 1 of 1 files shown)


ghost:
Retesting with xhprof shows ksort still provides a performance gain, but the gain is much more minimal for a stock cms install in blog view.
Inclusive Wall Time goes from 28,542 microseconds to 25,646 before and after including this change
Inclusive CPU Time goes from 28,997 microseconds to 24,995  
Or in terms I can understand, it saves 3 milliseconds of page generation/load for a moderate number of language files

Not worth the time to dig into for the speed bonus - but worth implementing once you know about it.  And for me, very worth the time in learning how to use xhprof with simple, concrete, real life examples.  Not worth the time digging further into JLanguage to try to speed it up any further - on the same run JDatabase eats up over 400 milliseconds - just enabling MySQL query cache would provide ten times the benefit this "fix" does. :-)

libraries/joomla/language/language.php
```diff
@@ -775,6 +779,8 @@ protected function loadLanguage($filename, $extension = 'unknown')
 			if (is_array($strings) && count($strings))
 			{
 				$this->strings = array_merge($this->strings, $this->override);
+				//Sort the underlying heap by key values to optimize future merges
```
pasamio: This will probably cause a check style problem just looking at it.

pasamio:
Six commits for six changes in the same file is a bit much, do you mind rebasing?

ghost:
I'll give it a shot later at my coding system. I was having some trouble getting a clean checkout of staging.

ghost:
Cool - had to use --force on the push to get it to apply - looks much better

PR #1779, opened 2013-01-02
Strict error in JDatabaseDriverPostgresql

elinw:


elinw:
Actually I think this is a bug in the cms copy of the database package where the & has been removed. Since it really is a reference in this case can someone please clarify whether whatever happened to remove that was wrong or right? I'm going to look at history now.

elinw:
My fault too :(

PR #1789, opened 2013-01-06
Small refactoring of JRoute::_()

Hackwar:
This refactoring of JRoute::_() simplifies the code a bit, provides a small performance improvement and allows the router to specify if the URL needs to be https and is thus not completely dependent on the layout developer to remember to add the $ssl = true on the JRoute::_() call.

### DIFF
--- a/libraries/joomla/application/route.php
+++ b/libraries/joomla/application/route.php
@@ -65,6 +65,8 @@ class JRoute
 		// Replace spaces.
 		$url = preg_replace('/\s/u', '%20', $url);
 
+		$scheme = array('path', 'query', 'fragment');
+
 		/*
 		 * Get the secure/unsecure URLs.
 		 *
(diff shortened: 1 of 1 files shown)


okonomiyaki3000:
To me this definitely looks better than the current version. I have just a couple of thoughts about it:

1) When you compare `$ssl === 1` to determine what the scheme should be, you should continue the practice of casting `$ssl` as int or else passing `true` as the third argument to the function will not have the desired effect. 

2) It looks like you've added another call to `preg_replace` to replace spaces. Probably this doesn't need to be done twice and, in fact, should not be done with `preg_replace` at all. The preg functions are a little expensive and, whenever possible, alternatives should be used. In this case, `str_replace` can certainly be used instead. I realize the function was already using `preg_replace` but as long as you're changing it, maybe you can make that switch as well. One side point about this, `/\s/u` will match space, tab and probably newline characters. This is probably a flaw in the original code but I'm sure it never became an issue because tabs don't often appear in urls.

eddieajau:

dongilbert:
I agree this should use str_replace instead. I like the looks of it. Also, if a CMS'er could confirm this as good (@elinw) then this can be merged. The framework has dropped this CMS specific router.

PR #1790, opened 2013-01-07
[NEW] JDom Framework : HTML abstraction layer

Cook-Self-Service:


dongilbert:
Looks promising, but will require extensive documentation IMO before I could take a serious look at it. Great work!

AmyStephen:
Along those lines, what would help would be to explain why this approach is better than the existing approach. Does it offer more flexibility in terms of what JS frameworks can be used? How does it (does it?) better separate the presentation layer? Is it more extensible? and so, on. 

Also, unit testing. 

I think one of the problems with getting good patches, and flat-out features, as is the case here, considered and into core is that nothing accompanies the code. The more the developer is able to quantify the benefits and clarify why we can't live without, the better your chances. 

I believe the presentation layer is far too embedded into Joomla core, it seems many agree on that point. So, this is an area where ideas and code might be welcome. But, you'll have to make your case and it's entirely possible other ideas will come out and a working group could form where those interested, code in hand, help build the next generation.

Lastly, kudos for putting this up. Tweeted it an hour after it was shared. It's an area of need and a generous offering. Thanks!

Cook-Self-Service:
Thank you for your comments.

Actually, I am finishing JDom so I cannot present it well for the moment.
JDom is a great framework. The example here is only a Hello World and is showing nothing.

Thanks to JDom :+1: 
- It is possible to run cross compatibilities (1.6 to 3.0).
- It can load easily the requested frameworks (bootstrap, jQuery, jQuery UI + sortables, chosen, icomoon, ...)
- It assembles the PHP, HTML, CSS, images, fonts in the same place for every JDom class.
- The HTML construction is decomposed and inherited, so it is very easy to addCss styles, javascripts commands, for instance.
- The parameters are passed trough an associative array, wich means that it is easy to transfor a list to a combo or to a radio.
- And much much more !...

I will write a documentation and instance it in my Joomla plateform repository to offer some examples.
For the moment it is too early.

I was only testing how works GitHub because I am a newbe here.

Hold on, it will not be long.

Cook-Self-Service:
[quote]I believe the presentation layer is far too embedded into Joomla core, it seems many agree on that point.[/quote]

@AmyStephen 
What do you mean, too much complex, or too much PHP for rendering ?

vecrea:
I use jDom on some projects with Cook. Once you understand the logic behind it, I find it very powerful.
But yes, it lacks a solid documentation, although it has improved over the past months

eddieajau:
I'm going to close this, not because the idea doesn't have merit, but the way it's done just doesn't fit into what the platform should be offering. For example, we don't want to be adding anymore JS to the platform, and we certainly don't want to add CSS files. I think you'd be better trying to offer it directly in the CMS, but I share Amy's concern that you are duplicating some existing functionality.

PR #1791, opened 2013-01-07
[fix] package uninstaller cannot access the manifest

phproberto:
Currently package installer scripts cannot access to the manifest to process it.

For components in the installer script you can get the manifest with something like:

```
function uninstall($parent)
{
    $manifest  = $parent->get('manifest');
}
```

Actually that will not work for packages. Neither:

```
$manifest = $parent->getParent()->manifest;
```


eddieajau:

PR #1793, opened 2013-01-09
Fixing JForm behavior on load without replacing

ghost:
We need to remove the new (but existing) field to avoid duplicates.
unset() seems not to work.

Test instruction: Try to add existing fields without replacing.

### DIFF
--- a/libraries/joomla/form/form.php
+++ b/libraries/joomla/form/form.php
@@ -729,7 +729,8 @@ class JForm
 					}
 					else
 					{
-						unset($field);
+						$dom = dom_import_simplexml($field);
+						$dom->parentNode->removeChild($dom);
 					}
 				}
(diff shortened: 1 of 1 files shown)


elinw:
Do you mean adding new fields with the same name as existing fields but where you don't want to replace the existing fields if they are present? 
$Replace=false doesn't work? I would consider that a bug.

ghost:
Yes, it's a bug. No matter if i want to replace fields or not. JForm has to avoid duplicated fields. But if $replace=false. It doesn't work, because "unset($field)" does not remove the field from the added element.

eddieajau:
I'm going to have to close this one because there is a failing test.
http://developer.joomla.org/pulls/pulls/1793.html

PR #1795, opened 2013-01-10
Fixing extension install from dynamic URLs

Bakual:
Since the JInstallerHelper::downloadPackage function now uses `JHttpFactory::getHttp()` instead of `stream_get_meta_data()`, the "Content-Dispositon" data is now stored directly in the `$response->headers['Content-Disposition']` instead of `$response->headers['wrapper_data']['Content-Disposition']`.

Currently in Joomla 3.0.2, you can't install extensions directly from an URL if the URL points to a dynamic created link (like in Akeeba Release System or CTransifex). This is because it doesn't detect a valid filename.

This commit fixes this.

### DIFF
--- a/libraries/legacy/installer/helper.php
+++ b/libraries/legacy/installer/helper.php
@@ -59,9 +59,9 @@ abstract class JInstallerHelper
 			return false;
 		}
 
-		if (isset($response->headers['wrapper_data']['Content-Disposition']))
+		if (isset($response->headers['Content-Disposition']))
 		{
-			$contentfilename = explode("\"", $response->headers['wrapper_data']['Content-Disposition']);
+			$contentfilename = explode("\"", $response->headers['Content-Disposition']);
 			$target = $contentfilename[1];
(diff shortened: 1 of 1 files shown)


Bakual:
Also created PR in CMS since it's unclear where it belongs to. Currently it's in /libraries/legacy, but the CMS has it in /libraries/joomla.
https://github.com/joomla/joomla-cms/pull/649

Discussion in GoogleGroups: https://groups.google.com/d/topic/joomla-dev-platform/kkBVeh0wygw/discussion

elinw:
Would you please also create an issue for this in the CMS issue tracker and place a link to the two pull requests?

Bakual:
Actually, the trackeritem exists already and is also referenced in the CMS PR, but it sure doesn't hurt to link it here as well :-)
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=29946

PR #1797, opened 2013-01-10
Avoid silent deprecation

florianv:
Avoid silent deprecation by triggering E_USER_DEPRECATED notices.

### DIFF
--- a/libraries/joomla/application/router.php
+++ b/libraries/joomla/application/router.php
@@ -127,7 +127,7 @@ class JRouter
 
 			if (!class_exists($classname))
 			{
-				JLog::add('Non-autoloadable JRouter subclasses are deprecated.', JLog::WARNING, 'deprecated');
+				trigger_error('Non-autoloadable JRouter subclasses are deprecated.', E_USER_DEPRECATED);
 
 				// Load the router object
 				$info = JApplicationHelper::getClientInfo($client, true);
--- a/libraries/joomla/database/database.php
+++ b/libraries/joomla/database/database.php
@@ -31,7 +31,7 @@ abstract class JDatabase
 	 */
 	public function getErrorMsg($escaped = false)
 	{
-		JLog::add('JDatabase::getErrorMsg() is deprecated, use exception handling instead.', JLog::WARNING, 'deprecated');
+		trigger_error('JDatabase::getErrorMsg() is deprecated, use exception handling instead.', E_USER_DEPRECATED);
 
 		if ($escaped)
 		{
@@ -53,7 +53,7 @@ abstract class JDatabase
 	 */
 	public function getErrorNum()
 	{
-		JLog::add('JDatabase::getErrorNum() is deprecated, use exception handling instead.', JLog::WARNING, 'deprecated');
+		trigger_error('JDatabase::getErrorNum() is deprecated, use exception handling instead.', E_USER_DEPRECATED);
 
 		return $this->errorNum;
 	}
@@ -70,7 +70,7 @@ abstract class JDatabase
 	 */
 	public function stderr($showSQL = false)
 	{
-		JLog::add('JDatabase::stderr() is deprecated.', JLog::WARNING, 'deprecated');
+		trigger_error('JDatabase::stderr() is deprecated.', E_USER_DEPRECATED);
 
 		if ($this->errorNum != 0)
 		{
--- a/libraries/joomla/database/driver.php
+++ b/libraries/joomla/database/driver.php
@@ -996,7 +996,8 @@ abstract class JDatabaseDriver extends JDatabase implements JDatabaseInterface
 	 */
 	public function loadNextObject($class = 'stdClass')
 	{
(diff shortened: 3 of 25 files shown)


dongilbert:
This is a good way to handle the deprecations while at the same time reducing package inter-dependence. However, it should be noted, that downstream users will need to set their own error handler so that `trigger_error` doesn't break their sites and show the deprecated notices to non-admins.

florianv:
They can put `error_reporting(E_ALL ^ E_USER_DEPRECATED)` or set any logger as error handler yes. Tests bootstraps are changed too to disregard E_USER_DEPRECATED

florianv:
So if you are a CMS developer and have a few notices `Deprecated`when you test your code, you should remove them.
- If you don't have the time to remove them, you can put this line at the top of your component entry point : 

``` php

<?php 
// All deprecated notices won't be displayed
error_reporting(E_ALL ^ E_USER_DEPRECATED);
```
- If instead you want to log them using `JLog`, you can put this code at the top of your component entry point : 

``` php

<?php 

$deprecatedHandler = function ($errno, $errstr, $errfile, $errline)
{
      $errorString = sprintf('Error : %s, in file %s at line %s', 
          $errstr, $errfile, $errline);

      JLog::add($errorString, JLog::WARNING, 'deprecated');
}

 set_error_handler($deprecatedHandler, E_USER_DEPRECATED);
```

AmyStephen:
Anyway to address it more on a application-wide level, perhaps in the framework file?

https://github.com/joomla/joomla-cms/blob/master/includes/framework.php#L38

(there are also include folders and framework files for the administrator and installer)

florianv:
Yes the handler could be added in the CMS bootstraps or index files so the change is actually invisible for CMS developers. But the idea is that they see the notices in the browser, so they can react before the method/class is removed.

IMO, there should be different configurations : 
- development with max error reporting
- production

and developers could switch between configs but it's an other thing.

eddieajau:
This is a -1 from me but we can see what others say. I appreciate the idea behind this, but it does lead to unintended consequences. In other words, the code will fail because it is deprecated, not because it doesn't work. I don't have any problem with coupling deprecated classes to JLog and, to me, configuring a logger to pick up the notices is easier than writing an error handler.

dongilbert:
does trigger_error cause it to fail even if it's just an E_DEPRECATED?

florianv:
@eddieajau Sometimes it is a method of a "new" class that is deprecated. 

The problem is that there are many devs that don't know if a method has been deprecated unless they follow the mailing list or see the changelog. Many of them are not informed, until their code is broken or there is a flash new on the BBC, they won't know.

So by adding this notice, they can see it in their browser / tests..

@dongilbert no, the code execution is not stopped it is not an exception or a "real" error

florianv:
@eddieajau The code will not fail. It will continue executing while displaying a "notice" that users can simply disable or log like PHP does with E_DEPRECATED. It is not a E_ERROR, it's more like a E_NOTICE. Breaking the code would have been very agressive yes..

eddieajau:
For me, interrupting the output is a major drawback. With JLog, you can add a deprecation logger and do a full run of the application and capture everything in a log file (say, loading the home page of the CMS).

With trigger error, you have to either inspect the output (not always convenient) or you have to override the application's own error handler (PHP only allows you one custom error handler doesn't it?). You also force the developer/tester to add an evil custom error handler where they may not want on(we are in an age of throwing exceptions after all).

So, to me, the end result is more inconvenient.

On the other hand, it's no less inconvenient than testing with full error reporting on to weed out PHP problems, but that's not something we have any control over.

AmyStephen:
I don't know, Andrew. I don't know that there would be an impact of concern, but I agree that it's worth looking at the various logic paths to see for certain if there is an impact. 

@florianv Are you able to draw up a logic matrix with possible options for:
1) JLog setting - limited to a) reports the deprecation error and b) currently does not report the deprecation error, 
2) CMS Configuration setting for error reporting level again - a) display on for this PHP level option or off;
3) PHP Custom Error Handler (no handler defined, one defined that handles this level, one defined but it ignores this level, one defined that handles E_USER_DEPRECATED and passes those on to JLog). 

Of those Andrew mentioned, I don't think there is a difference since the CMS already has that Custom Error code @florianv is proposing (I think @realityking put it in during the JError/JException project.) and the CMS Error Handler uses JLog. So, it should behave as before. 

Now, I agree, that there could be unintended results. But, after playing with this, I think it would be limited to those who: 1) use JPlatform 2) outside of the CMS 3) have full error reporting on 4) have stupidly defined their custom error handler to automatically throw an exception for an E_ALL reporting level. (And, because of this thread, I've now fixed that bug. :-P )

IMO, it's still worth pursuing to see if the dependency can be lifted, but it would be good to take a careful look at the various possibilities to build better assurance of what could happen.

dongilbert:
As Andrew said, it could cause unintended consequences. Deprecated logging has been added to the CMS, and we've switched to PSR-3 logging in the framework. So I'm closing this.

PR #1800, opened 2013-01-12
Move JDocument to legacy tree. Not a Deprecation.

dongilbert:
JDocument is so heavily tied to the CMS that it is difficult to use outside of the CMS in it's current state. With the massive namespace effort, I felt it would be appropriate to move JDocument to legacy, and preserve it there for downstream users. IMO, it should be transferred to the CMS altogether, but I want to provide a viable alternative before that occurs.

For or against, let it be known, or forever hold your peace. :) (only partially kidding)


dongilbert:
So no one is against this strategy? :) I'll post it on the platform list and see if we can get some more action going on here.

AmyStephen:
I had to build my own Theme/Rendering/MVC environment since the JPlatform classes are directly tied to components, modules, templates, and so on. JDocument is all about the CMS.

JPlatform will need a replacement, or at least an interface to Mustache or Twig.

eddieajau:
On 16 January 2013 15:50, AmyStephen notifications@github.com wrote:

> I had to build my own Theme/Rendering/MVC environment since the JPlatform
> classes are directly tied to components, modules, templates, and so on.
> JDocument is all about the CMS.
> 
> JPlatform will need a replacement, or at least an interface to Mustache or
> Twig.
> 
> It doesn't necessarily "need" a replacement. It would be sufficient to
> demonstrate how to replace the functionality using other means, if that is
> an appropriate direction to take (in which case, we could deprecated the
> package with a clear conscience). I'm not quite sure what the end result
> would be, but my guess is it would involve moving on some of the
> functionality of JDocument into the Platform, probably in the form of a
> "Response" package and the application would push whatever "view"
> generation strategy it likes into it.

The other option is we just make a call to deprecate it outright and the
CMS can just copy it into their application stack. I wouldn't be opposed to
that because JDocument is not exactly convenient to use outside of the CMS.

Regards,
Andrew Eddie

AmyStephen:
+1 You just said what I said. @eddieajau

PR #1803, opened 2013-01-14
Updating PHP version requirement in 13.1

dongilbert:
PHP 5.3.10 offers several enhancements to the PHP stack, not the least of which is an arbitrary code execution bug that was fixed in that version.

Improvements to DirectoryIterator as well as a bugfix to BCRYPT (which allows us to properly implement password lib changes that are coming in PHP 5.5 right now) should also influence the decision.

More info on the change set can be found here. http://php.net/ChangeLog-5.php#5.3.10

Look at all the security fixes since 5.3.1 alone!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -18,7 +18,8 @@ Requirements
 ------------
 
 * PHP 5.2.4+ for Platform versions 11.x
-* PHP 5.3.1+ for Platform versions 12.x  
+* PHP 5.3.1+ for Platform versions 12.x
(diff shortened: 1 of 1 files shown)


mahagr:
+1 for the change. The reason is that there are really weird bugs causing random white pages on PHP 5.2 and PHP 5.3 that just go away by using later version of PHP 5.3.

eddieajau:
Let's leave the Platform as is and increase the minimum on the Composer repos.

PR #1811, opened 2013-01-24
Add a date addition/subtraction method to JDatabaseQuery

elinw:
Add a method implementing DATE_ADD  in JDatabaseQuery which will allow dynamic queries on dates.


elinw:
Should be good to go.

elinw:
Can you let me know please if you are not taking this since I want  to be able to run queries for trending tags in 3.1 (as well as for some existing views) and I'll have to do a workaround for jdatabasequery to provide multidb support if you aren't taking it.   This is one space where the different drivers have important differences.

AmyStephen:
It looks good to me, @elinw - saw it when you first posted it, it would be handy, IMO. 

If you are aggregating and trending, it's essential.

elinw:
Thanks Andrew.

PR #1813, opened 2013-01-24
Add JGithubUsers

elkuku:
This will add support for the GitHub users API.

See: http://developer.github.com/v3/users/


ianmacl:
_sigh_.  I just merged this and it breaks a test.  Can you fix it up?

JGithubUsersTest::testGetUsers Expectation failed for method name is equal to string:get when invoked 1 time(s) Parameter 0 for invocation JHttp::get('/users', null, null) does not match expected value. Failed asserting that two strings are equal. --- Expected +++ Actual @@ @@ -'/user' +'/users' .../libraries/joomla/github/users.php:137 .../tests/suites/unit/joomla/github/JGithubUsersTest.php:235

elkuku:
What a dumb mistake... apologies ;)

PR #1815, opened 2013-01-26
Move some repeated code from JGithubObject extending classes to the base class

elkuku:
While playing with JGithub I thought that some repeated code in the "objects" extending JGithubObject could be moved to the base class.

If this is good to go, I would make the changes to the rest of the classes before merging.

I also added type hinting to the new method, which required the unit test to change.
I'm not 100% sure that this is good.


elinw:
It's showing up not merge-able, you may need to rebase.

elkuku:
Should be good now - thanks ;)

PR #1816, opened 2013-01-26
Add Union All to JDatabaseQuery

elinw:
UNION ALL can in the right circumstances have a substantial performance benefit over UNION because it eliminates the sort used to find duplicate rows..


eddieajau:
This branch needs to be updated, then it's fine to merge.

elinw:
Thanks Andrew, I'll try to do that tonight.

elinw:
Should be good now, I fixed a couple of typos in doc blocks while I was at it.

PR #1817, opened 2013-01-28
AUTOMATED: WhiteSpace Removal

WhiteSpace-Strippers:
This is an automated request to remove excessive whitespace from source code. You can find more about this project and opt-out at [whitespacestrippers.com](http://whitespacestrippers.com/)


dongilbert:
The 120 bytes saved in this PR turns into a negative net loss of electricity "savings", considering the time and energy it took to make these automated changes. While noble, your project is misguided. It's like "green" advocates that say "let's tear down this building and put up an energy efficient one!" It would take decades to even see a slight repayment on investment. Sorry to burst your bubble. :-1:

elkuku:
WOW 120 bytes - nice job :P

ianmacl:
So your change results in 11,750 checkstyle errors, 11 unit test failures and 5 unit test errors.  In addition, it touches about 880 files which will more than likely mean that it will cause a merge conflict with at least one other pull request.  Nice attempt I guess, but I think something got missed along the way.

Closing.

PR #1818, opened 2013-01-31
Update libraries/joomla/database/driver/sqlsrv.php

MDunkleSr:
Add ' ORDER BY RowNumber' to limit(). Without ORDER BY, limit function returns results that do not honor the original ORDER BY specified in the query.


dongilbert:
If someone with sqlsrv access can confirm this, I think it's good to merge. @mbabker? (Sorry, always picking on you.)

MDunkleSr:
Not that it changes the need for independent confirmation, but I do have it running w/ desired results in both a test & production environment.  LMK if I need to clarify anything.

Appreciate you taking a look at this & considering it for merge!  

Thanks,

Mike

dongilbert:
That's good to know. I trust testing and production usage more than one off testing, TBH. I just don't have access to sqlsrv (that I know of) to test it myself.

mbabker:
I'll be firing up my Windows & SQL Server environment this weekend for some CMS work, but I'll add this to my to-do list as well.

PR #1820, opened 2013-02-04
Refactor and complete GitHub API implementation

elkuku:
This will refactor and complete the GitHub API implementation for the Joomla! Platform.

This is a "first scratch" - yet a complete one (I hope).
I opened it to obtain some opinions and maybe some assistance - kind of a RFC.
### B/C Issues

Due to the refactoring, some methods have been moved. Since the implementation is still young, I hope this is acceptable.
- [ ] Indentify possible issues.
- [ ] Provide legacy replacements
### @TODO List
- [ ] Doc blocks in unit tests. (HELP pls ! :) )
### Plans ?
- [ ] Build classes that loosely reflect what is available at [developer.github.com/lib/resources.rb](https://github.com/github/developer.github.com/blob/master/lib/resources.rb)
  These classes could be used by the methods to parse the obtained JSON response, pass it to a specific class and return that class to the caller instead of the standard json_decode'd object.
  
  Very useful for auto complete in IDEs. Of course, this should be automated ;)


dongilbert:
Nikolai, there is discussion amongst the maintainers about modernizing the framework, implementing namespaces, utilizing composer and packagist, etc. This would include releasing "Joomla Packages" that can be plugged in to any other framework using composer. It's still in the infant stages, but looks promising.

Where this relates to this PR is that our Github package is arguably the best package that utilizes GitHub API v3, and would be very useful to other devs to consume. What you have here is a complete implementation, and what we have already in the platform is incomplete (judging from the PR's coming in against it). So, the thought is to not introduce BC-incompatible changes to the current platform, and release new code under the new system. More info will be provided in the coming weeks.

elkuku:
Thanks for your comments Don, very much appreciated (please note the letters "RFC" above).

The main question is about the refactoring.
The former approach implemented only a subset of the API, using a custom structure. So it was kind of hard for me (and I guess also for others) to find the corresponding Joomla! methods while reading the GitHub API documentation.
This is why I organized the code following the documentation.

So is it good, bad, ugly ? ...
Does it deserve more work ?

dongilbert:
OH - I thought you were questioning why there was no comment, not about the quality. 

I think it is good, I need to dig through it more thoroughly, but a cursory review gives me good feelings about it. The fact that it is a complete implementation that follows the GitHub API doc flow is a great bonus - very useful.

So, yes, definitely deserves more work.

eddieajau:
@elkuku, can we move this effort to the Framework? Feel free to raise and "Improvement" issue with your TODO list. I'll close this in the mean time.

PR #1826, opened 2013-02-13
Allow Web application router to return controller output.

piotr-cz:
Since [new generic Router](https://github.com/joomla/joomla-platform/pull/1259) has a functionality of a dispatcher (executing controllers), it's handy to pass controller output further to caller (Application), especially when it's JApplicationWeb. It's easy to to override this method in own JApplication instance, but IMHO this gives a hint how to use new MVC.

### DIFF
--- a/libraries/joomla/application/web/router.php
+++ b/libraries/joomla/application/web/router.php
@@ -62,7 +62,7 @@ abstract class JApplicationWebRouter
 	 *
 	 * @param   string  $route  The route string for which to find and execute a controller.
 	 *
-	 * @return  mixed   Controller output
+	 * @return  mixed   The return value of the controller executed
 	 *
 	 * @since   12.2
 	 * @throws  InvalidArgumentException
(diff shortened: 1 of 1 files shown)


piotr-cz:
Maybe it will become clearer 'why' when I attach a chunk of simple application.
If I'm doing it completely wrong, please let me know but at the moment I really think it's quite convenient when `JApplicationWebRouter` returns controller output.

``` PHP

protected function doExecute()
{
    // Initialise the application
    $this->initialise();

    // Dispatch the application
    $this->route();
}

public function route()
{
    // Initialise variables
    $document   = $this->getDocument();
    $router     = $this->getRouter();

    // Execute the task.
    try
    {
        $contents   = $router->execute( $this->get('uri.route') );
    }
    // Mop up any uncaught exceptions.
    catch (RuntimeException $e)
    {
        echo $e->getMessage();
        $this->close($e->getCode());
    }

    $document->setBuffer($contents, 'component');
}
```

dongilbert:
I'm not sure on this one way or the other. Love to have @eddieajau's take on it. I know the new MVC works currently without this, but I could see it as being advantageous, based on how you approach your application architecture.

eddieajau:
I think this is fine if you change the docblock to `@return  mixed  The return value of the controller executed`. I wouldn't want to restrict it to 'output'. That is just one possibly scenario. It could also return another controller in an HMVC setting I would imagine.

PR #1827, opened 2013-02-15
Allow chaining of JDocumentHTML setHtml5 like in other methods

piotr-cz:
Allow chaining of JDocumentHTML setHtml5 like in other methods, ie:

``` php
JFactory::getDocument()
    ->setTitle($this->get('sitename', 'Website'))
    ->setDescription($this->get('MetaDesc'))
    ->setHtml5(true)
;
```

### DIFF
--- a/libraries/joomla/document/html/html.php
+++ b/libraries/joomla/document/html/html.php
@@ -331,7 +331,7 @@ class JDocumentHTML extends JDocument
 	 *
 	 * @param   bool  $state  True when HTML5 should be output
 	 *
-	 * @return  void
+	 * @return  JDocumentHTML instance of $this to allow chaining
 	 *
 	 * @since   12.1
 	 */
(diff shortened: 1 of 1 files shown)


dongilbert:
Looks good and makes sense. If you're able to, could you write a unit test for this change? Currently, JDocumentHTML has 0 tests written, so it would be nice to start here. 

https://github.com/joomla/joomla-platform/blob/staging/tests/suites/unit/joomla/document/html/JDocumentHTMLTest.php

piotr-cz:
I'll try to.

eddieajau:
Would be great if you could add tests, but the actual change is fine.

piotr-cz:
@eddieajau 
I'll try to, but I'm quite busy with work at the moment and I'm completely fresh to unit testing.

PR #1831, opened 2013-02-15
Add JGithubMeta object to interact with Meta API endpoint

mbabker:


ianmacl:
There is a unit test error.  Please fix it up and ask for the request to be reopened.

mbabker:
Any suggestions on how to get past that error?  This method is unique in that it's not strictly returning the response object, but I have it doing post processing on the implied response.  The test failure occurs because the method doesn't have the response data to process.

dongilbert:
Can you use reflection to fake it? I'll have to take a closer look when I get back in front of a computer.

mbabker:
No idea.  Unit testing isn't one of my stronger points, but if so, I'm all for trying.

mbabker:
OK, it was user error. :blush:

dongilbert:
Looks better Michael.

PR #1836, opened 2013-02-19
Memcache fix

ghost:
Updates the ini settings used for memcache so that this will work.


dongilbert:
Gary, your commit says a "potentially multi-dimensional array" - will it always be multi-dimensional?

eddieajau:
I'm going to close this one. If can be fixed against either or both the CMS or the new Framework.

PR #1841, opened 2013-02-27
Joomla CMS [#29922] Detect Chrome for iOS in browser.php

axelsjo:
Fixes tracker item:
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=29922

Simular (CriOS) to a previous fix for Chrome for Android (CrMo):
http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=28039

An alternative approach is to put Chrome|CrMo|CriOS in the rame regexp... I prefer to keep them apart though - some redundant code, but better readability.

### DIFF
--- a/libraries/joomla/environment/browser.php
+++ b/libraries/joomla/environment/browser.php
@@ -256,6 +256,11 @@ class JBrowser
 				$this->setBrowser('chrome');
 				list ($this->majorVersion, $this->minorVersion) = explode('.', $version[1]);
 			}
+			elseif (preg_match('|CriOS[/ ]([0-9.]+)|', $this->agent, $version))
+			{
+				$this->setBrowser('chrome');
+				list ($this->majorVersion, $this->minorVersion) = explode('.', $version[1]);
+			}
 			elseif (strpos($this->lowerAgent, 'elaine/') !== false
 				|| strpos($this->lowerAgent, 'palmsource') !== false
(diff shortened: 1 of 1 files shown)


axelsjo:
Google provides some background on the User-Agent in mobile Chrome versions here:
https://developers.google.com/chrome/mobile/docs/user-agent

`CrMo` is changed to `Chrome/<Chrome Rev> Mobile` in recent versions of the mobile browsers, but the existing CrMo check in browser.php should be kept to catch older versions.

Regarding `CriOS`, we see that the User-Agent string in Chrome for iOS is the same as the Mobile Safari User-Agent, with `CriOS/<ChromeRevision>` instead of `Version/<VersionNum>`.

I haven't tested, but it looks like browser.php today finds Safari but are unable to get the version numbers (in `identifyBrowserVersion()`).

The question is whether Chrome on iOS should be reported as Chrome (as in the pull req) or as Safari. The first is more correct from a "user" perspective, as we report the browser in use with the correct matched version number. The latter is more correct from a "technical" perspective, as Chrome on iOS use the UIWebView rendering engine provided in iOS (but I have a hard time seeing how we can pick a proper version number).

dongilbert:
As it is very similar to an already fixed tracker item, and it's adding no BC problems, I'll merge this.

PR #1842, opened 2013-03-04
[fix] findmanifest should prioritize main folder manifests

phproberto:
Actually the findManifest function looks for manifests in the parent folder and in first level subfolders. 

This example structure:

->fof
--->fof.xml
-libraries
-manifies.xml

Will return the fof.xml manifiest.

This fix ensures that the function prioritises the main folder manifiests. I would like to remove the subfolder check but kept it for B/C reasons.

### DIFF
--- a/libraries/legacy/installer/installer.php
+++ b/libraries/legacy/installer/installer.php
@@ -1836,8 +1836,14 @@ class JInstaller extends JAdapter
 	 */
 	public function findManifest()
 	{
-		// Get an array of all the XML files from the installation directory
-		$xmlfiles = JFolder::files($this->getPath('source'), '.xml$', 1, true);
+		// Main folder manifests (higher priority)
+		$parentXmlfiles = JFolder::files($this->getPath('source'), '.xml$', false, true);
+
+		// Search for children manifests (lower priority)
+		$allXmlFiles    = JFolder::files($this->getPath('source'), '.xml$', 1, true);
+
+		// Create an unique array of files ordered by priority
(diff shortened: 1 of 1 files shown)


dongilbert:
Since this is a CMS specific issue, I would defer the check on this to someone like @dextercowley or @elinw to make sure it is the desired functionality.

phproberto:
I see it as a Platform bug. I think "anything" using JInstaller expects it to load first the main manifest instead of subfolder's. 

I already tested it and sent PR for the 2.5 and the 3.0 CMS branches:

3.0:
https://github.com/joomla/joomla-cms/pull/752

2.5:
https://github.com/joomla/joomla-cms/pull/753

Thanks for reviewing this!

elinw:
Is there any idea that any application that isn't the CMS is using JInstaller?  The code seems so specific that I think we might want to move the whole package from legacy to cms so that it's easier to make it work correctly for the cms.

dongilbert:
I would agree with that Elin, which is the reason for my first comment. It's not that I don't want to fix or I don't think it's a bug, it's just highly CMS specific, IMO.

AmyStephen:
@elinw probably not, the paths are so mixed in, I wasn't able to use it.

dongilbert:
So I agree with Elin, JInstaller should move to the CMS. Any opposition?

AmyStephen:
Nope. Makes sense to me.

dongilbert:
@elinw Do you want to merge this PR before the package moves to CMS?

elinw:
I thought we already merged it?

PR #1843, opened 2013-03-05
Few tests JDocument package

piotr-cz:
JDocumentImage
JDocumentXml
JDocumentHtml (8 methods)

This isn't much, but these are my first unit tests.


dongilbert:
Looks good. All the tests are passing on my end.

piotr-cz:
Pleas take a look at the code quality, as these are the first test I've ever written.

I was going to write one more for JDocumentHtml::[mergeHeadData()](https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/document/html/html.php#L184) but I'm confused, as this method is using internal methods (setMetaData, addStyleDeclaration) whilte setMetadata is setting values directly.

Is this how it's supposed to work?

dongilbert:
OK - I'll take another look when I have time to give it a proper review of quality as well as functionality.

Thanks for doing this.

tests/suites/unit/joomla/document/html/JDocumentHTMLTest.php
```diff
@@ -38,33 +38,136 @@ protected function setUp()
...
-	public function testGetHeadData()
+	protected function tearDown()
 	{
-		// Remove the following lines when you implement this test.
-		$this->markTestIncomplete(
-			'This test has not been implemented yet.'
+	}
+
+	/**
+	 * Terst construct
```
elinw: Test :)

tests/suites/unit/joomla/document/image/JDocumentImageTest.php
```diff
@@ -0,0 +1,157 @@
...
+	 * Test...
+	 *
+	 * @covers  JDocumentImage::render
+	 *
+	 * @return  void
+	 */
+	public function testRender()
+	{
+		JResponse::clearHeaders();
+
```
elinw, line 100: What about doing this as a data provider. That way if one fails the rest of the tests still run.  Also I usually put a message so it's easier to figure out which test is the one that failed.  It isn't super important but it makes php unit easier to use.

piotr-cz, line 100: I'll need your guidance on data provider.

PR #1844, opened 2013-03-06
Update html.php

Paladin:
countMenuChildren can't return anything but zero, because the query is never set. This change sets it.

### DIFF
--- a/libraries/joomla/document/html/html.php
+++ b/libraries/joomla/document/html/html.php
@@ -542,6 +542,7 @@ class JDocumentHTML extends JDocument
 				$query->from('#__menu');
 				$query->where('parent_id = ' . $active->id);
 				$query->where('published = 1');
+				$dbo->setQuery($query);
 				$children = $dbo->loadResult();
 			}
 			else


dongilbert:
Awesome. Great catch.

For everyone else:
THIS IS WHY WE HAVE UNIT TESTS PEOPLE!!!

eddieajau:
I have this funny feeling this fix is going to break a lot of sites that have the directive included.

dongilbert:
It's not used at all in CMS 2.5 or 3 in the core. It's only occurrence is when it is declared.

elinw:
Thousands of developers use the CMS as a platfporm and use parts of packages that the CMS does not use.  These are both CMS developers and develoeprs working on the most widely distributed copy of the platform which is that distributed with the CMS

dongilbert:
What's your point Elin? The code is broke, and needed to be fixed, unless your proposal was to leave it as broken code. I was simply stating an observation with my previous comment, that it's not used anywhere. Obviously Arlen was using it for something, since he discovered the error. I was simply stating that it wasn't use in core.

elinw:
What you learn working with developers is that they have now built out workarounds around that.  Who knows what they are getting but they are NOT at this point getting the documented results unless they are 0.  So I would guess most of them are expecting to not get anything when the result is not 0, but who knows.  So just saying "it is not used in the CMS" is not a valid rationale for making an API change.  IF you want to say it is abug and won't break anyone who is now assuming the current behavior, fine, but saying that ddevelopers (and more importantly  end users who have installed work from developers ) who are not writing for the core cms don't matter is not okay.

That said it is certainly a goofy piece of code and a bug.  But it needs to be properly documented in the API and in where ever it is you are documenting b/c breaks this month.

And the reason we have unit test, one of them, is to avoid undocumented changes in APIs.

AmyStephen:
@elinw - this is a method that _always_ returns 0, isn't executed anywhere in the CMS or the platform, and has been that way for over two years. Not only is this a "valid rationale for making an API change", it's a valid reason to search down and find any developer who complains that it fixed. I want talk to them and fully understand their concern. =)

eddieajau:
> That said it is certainly a goofy piece of code and a bug. But it needs to be properly documented in the API and in where ever it is you are documenting b/c breaks this month.

Issue has been labelled as a "Backward Incompatibilty". If we decide to tag 13.1, it will show in the changelog.

Paladin:
"IF you want to say it's a bug?" IF?????? I reference http://api.joomla.org/Joomla-Platform/Document/JDocumentHTML.html#countMenuChildren:

"return: Number of child menu items"

It didn't do as it was documented to do. It was like checking the time on a stopped clock. What, pray tell, is the definition of a bug if not that?

And the only possible effective workarounds a developer could build for that could not possibly be affected by this change, since the only way to work around it is to write your own routine to do your own counting. Wouldn't be surprised if some did (as the code itself shows, it's a trivial exercise) but so what? This change doesn't affect them.

Let's talk "API change" for a minute, since it's a valid concern. I reject the idea this is an API change. It's a fix. The API was documented to do x and it failed to do x. Dismally. Completely. Making the API work as advertised isn't changing it; making it do something different from what it's documented to do is changing it. If anything, this was restoring the correct API behavior, reverting whatever change made it stop behaving as advertised.

And, like you, Amy, I really don't want to speculate about a developer who gets upset because we actually made some code we're releasing behave as documented. If doing that is a problem, what does that say for the product we've been delivering?

I'm skeptical about "backward compatibility," Andrew. Unless _every_ bug fix is also a backward compatibility issue. Because every bug fix changes current behavior of a documented API call. And if we're going that far, it seems the term is being stretched beyond all utility. But I'm in full on agreement about showing that it's now fixed in the changelog. All changes, including bug fixes, belong there.

As for your earlier remark, Andrew. I think you may be right. Wouldn't surprise me at all if people had cargo culted some template code which had that in it to place a contextual menu on a page, and never saw what the actual appearance of the menu would do to their design, because it was never triggered. But since the template was designed to do that (why else would that code be there? Because someone wanted a long way to write "0"?) and they're using that template, it's hard for me to worry much about it. And I'd tend to worry more about all those potential customers we cost our users because a function they depended on didn't work properly. If someone depends upon buggy, undocumented behavior, they should expect to have to change their code. That's the nature of the industry.

(Sorry about the length, was busy fixing JTable and didn't have time to check in on this thread earlier. BTW, you'll be happy to know the fix doesn't really change JTable, beyond making it come back more often with the right answer. Everything that used to work still works, and now, it even compensates for some sloppy developers, like the ones developing an extension I have to use. But that will be a different thread.)

eddieajau:
It's a backward incompatibility "just in case" someone has this lying around in their template laying domant and suddenly it starts to work as advertised. I'm not saying it's bad, I'm just flagging it because it took 5 seconds to do it and I couldn't be bothered arguing the coin toss. I'm on your side - I merged the PR after all. It's a bug. It's fixed. Done.

AmyStephen:
@Paladin Agree on both points about it not being an API change or a backward compatibility issue.

@eddieajau That's not to say you shouldn't tag it, warn people. 

Not suggesting @elinw shouldn't warn against the impact of change, either. 

This issue gives me a little hope, actually. There has been so much drama about change, sometimes it has felt like there is no way to get a handle on things. Maybe just getting back to those industry basics, seeing the API as it is, a description of how the code should behave, using the API when coding, and so forth. Knowing, as devs, that if the code doesn't behave like the API states, it's on us to report it, etc., maybe, we might stand a chance of reducing angst. To me, that builds a little hope. =)

eddieajau:
I _did_ tag it. I have no idea what all this fuss is about. Can we please move on.

AmyStephen:
No one said you didn't tag it. There is no fuss. There was a discussion of
terms. Yes, moved on.

Issue #1279, opened 2012-06-12
[bug] with JRegistry when the path contains empty fields


palcov:
Hello. JRegistry throws fatal error:

http://forum.joomla.org/viewtopic.php?f=579&t=709422

It can happen if using large and not-so-thoroughly written components.
I not found this issue anywhere else. Thanks.

elinw:
So what you can do is just edit the file and make the changes needed (try to follow joomla coding style). Then make a pull request based on that.

elinw:
@palcov 
If you would browse to the appropriate file here in Github, click edit, and make the changes then save you will be able to send a pull request for this change.

dongilbert:
This is fixed in the Framework.

PR #1849, opened 2013-03-16
Fixing joomla/joomla-platform#1279

dongilbert:
This fixes the fatal error that would occur when passing in a double dotted string to `JRegistry::set($path, $value)` by filtering the nodes, removing empty ones, and they re-keying the array. 

This has already been applied to the Framework.

### DIFF
--- a/libraries/joomla/registry/registry.php
+++ b/libraries/joomla/registry/registry.php
@@ -335,8 +335,14 @@ class JRegistry implements JsonSerializable
 	{
 		$result = null;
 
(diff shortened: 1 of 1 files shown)


dongilbert:
@elinw @dextercowley @mbabker This should probably be merged into CMS asap, including as a bug fix in 2.5.x, since it can result in hard to track fatal errors for end users, and introduces no breaks in backwards compatibility.

elinw:
Ok can you send a pr?

PR #1850, opened 2013-03-16
add attribution and fix several bugs

pwfisher:
When you refactored my code, you introduced a few bugs.

You need to handle the case where the parameter value contains an equals sign.

You're also not handling the case where the next argument is something like -abc (starting with a single hyphen).

Also, don't overwrite a previously set string value with a boolean value.

### DIFF
--- a/libraries/joomla/input/cli.php
+++ b/libraries/joomla/input/cli.php
@@ -113,6 +113,9 @@ class JInputCLI extends JInput
 	 * @return  void
 	 *
 	 * @since   11.1
+	 *
+	 * Based on parseArgs() function by Patrick Fisher <patrick@pwfisher.com>
+	 * @see                 https://github.com/pwfisher/CommandLine.php
 	 */
 	protected function parseArguments()
 	{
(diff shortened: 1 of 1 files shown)


dongilbert:
We didn't use your code as a base for this, as your code was created on github a total of 16 hours ago. Also, you are spamming several projects on github trying to get them to give you attribution for code you didn't create within their project. Please, stop the madness.

pwfisher:
Even if you don't believe I wrote this, why would you ignore the bugfixes?

mbabker:
If you would like the bug fixes to be considered, please submit them
separately from the attribution change.  Internally, we will discuss where
the original writer came up with this code and act accordingly for
attribution purposes.

eddieajau:
> Even if you don't believe I wrote this, why would you ignore the bugfixes?

There is no disputing you probably own the code in the patch, but I'm pretty sure you aren't the original owner of our code (because you said "When you refactored my code").

If you want us to consider this patch, then you need to follow the instructions here:
https://github.com/joomla/joomla-platform/blob/staging/CONTRIBUTING.markdown

Make sure you sign the JCA. Part of the deal with Joomla is that we don't attribute in the code itself. If that's a problem for you, I understand and we'll miss you.

pwfisher:
Reconsidering it, I can't say with certainly that this was derived from my code. I think a drop-in replacement would be the best solution to the logic issues. I'll see what I can put together.

eddieajau:
Make sure you add extra unit test cases to cover the bugs you are fixing. Thanks.

pwfisher:
Please see https://github.com/joomla/joomla-platform/pull/1851 for the new function and unit tests.

PR #1851, opened 2013-03-16
Fix several bugs in parseArguments and add test suite

pwfisher:
The old code will fail these tests:

Expected: $data["funny"] === "spam=eggs"
Old result: isset($data["funny"]) === false

Expected: $data["s"] === "override"
Old result: $data["s"] === true

Expected: $data["key2"] === true && $data["b"] === true
Old result: $data["key2"] === "-b" && isset($data["b"]) === false

### DIFF
--- a/libraries/joomla/input/cli.php
+++ b/libraries/joomla/input/cli.php
@@ -110,6 +110,8 @@ class JInputCLI extends JInput
 	/**
 	 * Initialise the options and arguments
 	 *
+	 * Not supported: -abc c-value
+	 *
 	 * @return  void
 	 *
 	 * @since   11.1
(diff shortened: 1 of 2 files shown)


dongilbert:
Now this is something I could get behind.

eddieajau:
Thanks. We'll take a good look at it. For what it's worth, you do get atttribution in the changelog for any accepted pull requests (just not in the code).

eddieajau:
Patrick, we've tried to do a bit of archeological hunting in the mean time and we can't determine if the code we used is derived from yours one way or the other. It may have been; it may have been inspired by someone else that copied and modified your snippet; it may have been a mashup of many different snippets; it may have been derived from first prinicples independently - we can't say for sure.

pwfisher:
Thanks for looking into it. I reached a similar conclusion. Accepting this pull request would make it a moot point, anyway.

dongilbert:
@eddieajau Should we look into doing this just on the new Framework InputCLI instead of here in the platform?

eddieajau:
I think we should do both. Let's get this one reviewed first. I apologise I've had the week from hell (no pun intended) adn haven't had a spare moment to review this yet.

eddieajau:
Merged. Thanks.

pwfisher:
Thank you.

PR #1853, opened 2013-03-22
Fixed an issue with JApplication redirecting to invalid URLs 

bentasker:
When HTTP headers have already been sent, JApplication::redirect() falls back on a javascript redirect. However it's passing the URL through htmlspecialchars so any URL with and ampersand in the URI will be invalid.


eddieajau:
I'm sorry but we aren't accepting new code into the Joomla Platform anymore. Please consider making a pull request against the joomla-cms or the joomla-framework repo's.

PR #1856, opened 2013-04-01
Check for the onUserAuthenticate function before calling it.

okonomiyaki3000:
onUserAuthenticate is not called in the usual way that events get triggered (by the dispatcher) and, as such, it can cause a crash if you have an authentication plugin that doesn't implement the function. This patch fixes that.

You might think, "Why have an authentication plugin that doesn't handle authentication?" and the answer is that you might have one that only needs to handle authorization. As it is, if you write such a plugin, you would need to throw in an empty 'onUserAuthenticate' function in order to avoid a crash.

### DIFF
--- a/libraries/joomla/user/authentication.php
+++ b/libraries/joomla/user/authentication.php
@@ -271,16 +271,20 @@ class JAuthentication extends JObject
 		{
 			$className = 'plg' . $plugin->type . $plugin->name;
 
-			if (class_exists($className))
-			{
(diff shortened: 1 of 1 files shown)


eddieajau:
I'm sorry but we aren't accepting pull requests for this repository any more. Please see http://developer.joomla.org/cms/report-an-issue.html for how to report and issue for the CMS Platform.

PR #1858, opened 2013-04-05
Fix to use sessions using file (default behaviour)

dmdigital:
Removing the register() function makes file-mode for sessions not working, because the class falls back to JSessionStorage for the register() function, without providing read, write etc. methods.
If you want to use files to save session data, register() needs to exist so that default PHP session-to-file mechanism doesn't stop working.


eddieajau:
I'm sorry but we aren't accepting pull requests for this repository any more. Please see http://developer.joomla.org/cms/report-an-issue.html for how to report and issue for the CMS Platform.

PR #1864, opened 2013-06-20
Form field type checkbox does not work - Proposed Fix

jhattara:
I tried to put checkbox form fields into my custom component using xml forms. In my fieldset there is included: 

&lt;field
    name="named_checkbox"
    type="checkbox"
    label="COM_COMPONENT_CHECKBOX_LABEL"
    description="COM_COMPONENT_CHECKBOX_DESC"
    class="inputbox"
    value="1"
    default="0"  
/&gt;

The checkbox saves fine when it is checked. However when I try to uncheck the checkbox, it does not update the database. After this change any checkbox form element should save value 0 when the checkbox is unchecked.

### DIFF
--- a/libraries/joomla/form/fields/checkbox.php
+++ b/libraries/joomla/form/fields/checkbox.php
@@ -58,7 +58,8 @@ class JFormFieldCheckbox extends JFormField
 		// Initialize JavaScript field attributes.
 		$onclick = $this->element['onclick'] ? ' onclick="' . (string) $this->element['onclick'] . '"' : '';
 
-		return '<input type="checkbox" name="' . $this->name . '" id="' . $this->id . '"' . ' value="'
+		return '<input type="hidden" name="' . $this->name . '" value="0" />'
+			. '<input type="checkbox" name="' . $this->name . '" id="' . $this->id . '"' . ' value="'
 			. htmlspecialchars($value, ENT_COMPAT, 'UTF-8') . '"' . $class . $checked . $disabled . $onclick . $required . ' />';
 	}
 }


wilsonge:
Tracker Item for those interested http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=31192

dongilbert:
This Platform repo is not taking pull requests. Please make this directly against the CMS class you are modifying. Additionally, your code is a bit off, in that you cannot have two HTML elements with the same id.

While not ideal, the current best way to solve this is to add a null check in your the `bind()` method of your table class that corresponds to the form you are using to edit / save the data. Additionally, you should modify any table columns that are associated with a checkbox to be NOT NULL and then if the submitted value is empty, set the field to an empty string.

```
class ComTableFoo
{
    public function __construct(&$db)
    {
        parent::__construct('#__com_foo', 'id', $db);
    }

    public function bind($src, $ignore = array())
    {
        // If the checkbox isn't set, that means it is unchecked. So set the value to empty string.
        if (!isset($src['checkbox_field_name'))
        {
            $src['checkbox_field_name'] = '';
        }

        return parent::bind($src, $ignore);
    }
}
```

jhattara:
The code is perfectly valid html. There is nothing that I know of in the specification that would prevent putting two form elements with the same _name_ in the same form. The latter just overrides the first one. I am well aware that it is suggested that checkbox should be fixed in the bind function, but this should not be necessary.

dongilbert:
I misread the code change. It looked like you had duplicated the _id_ across the hidden and actual field. Sorry for that.

While this should not be necessary (in your opinion) it currently is. If you want to fix the HTML spec so that unchecked checkboxes submit to POST, then by all means contact W3C and have them fix it.

If you want to fix it in Joomla, then you're going to have to submit this PR against the CMS. As stated in the repo description:

```
[READ-ONLY] This repo is not accepting pull requests. Please see joomla/joomla-cms or joomla/joomla-framework.
```

jhattara:
Yeah, I missed that earlier. Sorry.

Issue #1823, opened 2013-02-10
Error handling on database connection problems can be improved


tobixen:
This issue is also reported at http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_id=8103&tracker_item_id=30080

Joomla returns HTTP-code "200 OK" instead of "500 Internal Server Error" when database connection is lost (i.e. mysql dieing due to full disk).

This causes problems for us:

1) Monitoring systems fail to see that there are problems with joomla, hence
we'll have to wait for angry customers to call in before we discover it.

2) Our varnish cache fails to see that there are problems with the content, and
caches it for ten minutes.

3) The varnish backend check fails to see that the backend is broken, so the
"deliver old content"-fallback doesn't work.

Is it possible to fix better error handling?

dongilbert:
Definitely an issue. We'll have to work out the best way to accomplish this, but it surely needs to be done.

AmyStephen:
A while back Fotis provided a patch http://joomlacode.org/gf/project/joomla/tracker/?action=TrackerItemEdit&tracker_item_id=20256 

Is that still working?

dongilbert:
I'm going to push in some code that sets a 500 error, if everyone is ok with that?

I'd like to know what @eddieajau @mbabker @realityking think as well.

dongilbert:
So, after looking at it, this is already taken care of in `JLoader::createDbo()`

https://github.com/joomla/joomla-platform/blob/staging/libraries/joomla/factory.php#L583

khanhlh:
I still get "200" code when connection is lost. Just create a pull request at https://github.com/joomla/joomla-platform/pull/1870 to solve the issue.

PR #1870, opened 2014-09-16
Try to connect to Database when creating DBO

khanhlh:
Try to connect to Database at the time creating DBO to verify database connection. It will fix the issue Joomla still returns "200" code when connection is lost, which reported in #1823 but not solved.


wilsonge:
Hi the Joomla Platform is deprecated as of Spring 2013. The idea of Joomla's standalone platform has been continued in the Joomla Framework (http://framework.joomla.org/) whilst the platform project itself has been merged back into the CMS. So please can you send this code (if relevent) to the correct project please :) This repo exists principally for historical reasons for the CMS to track old conversations. Thankyou!

PR #1871, opened 2014-10-10
Add brackets for database name in sqlsrv driver when selecting 

brittongr:
Database names like 0_MyDatabase needs to be wrapped in [...], otherway the driver will throw an exception.


mbabker:
Thanks for your pull request.  Unfortunately, we aren't accepting pulls on this repo anymore; would you mind sending this to https://github.com/joomla-framework/database please?

